PR #5, opened 2022-12-21
Minor typo, punctuation update

andrewpollack:


src/structure.md
```diff
@@ -15,8 +15,9 @@ days:
 ## Format
 
 The course is interactive and your questions will drive our exploration of Rust!
+You may ask questions are start discussions by filing issues [here](https://github.com/google/comprehensive-rust/issues).
```
qwandor, line 18: s/are/or/

andrewpollack: Good eye -- ty!

src/structure.md
```diff
@@ -15,8 +15,9 @@ days:
 ## Format
 
 The course is interactive and your questions will drive our exploration of Rust!
+You may ask questions or start discussions by filing issues [here](https://github.com/google/comprehensive-rust/issues).
```
mgeisler, line 18: This page (ehem "slide") is shown during the class, so I would prefer to not have this part here. I've added links to the GitHub discussions in the README and on the first welcome page — I think they work better there.

andrewpollack: Apologies -- I missed those inclusions on the first pass. Removed this portion from the PR!

PR #10, opened 2022-12-21
Minor punctuation update

andrewpollack:


mgeisler:
Nice, thanks for keeping things consistent like that!

PR #14, opened 2022-12-21
Cache the artifacts built by the publish workflow

mgeisler:
This should speed up the publication: right now we spend around 4 minutes to build `mdbook` and `mdbook-svgbob`.

With this change, we will only pay that cost once a week (caches created by GitHub Actions [last 7 days][1]).

[1]: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -30,18 +30,26 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+
+      - uses: Swatinem/rust-cache@v2
+
       - name: Install mdbook
         run: cargo install mdbook
+
       - name: Install mdbook-svgbob
         run: cargo install mdbook-svgbob
+
       - name: Build book
         run: mdbook build
+
       - name: Setup Pages
         uses: actions/configure-pages@v2
+
       - name: Upload artifact
         uses: actions/upload-pages-artifact@v1
         with:
           path: 'book'
+
       - name: Deploy to GitHub Pages
         id: deployment
         uses: actions/deploy-pages@v1


.github/workflows/publish.yml
```diff
@@ -30,18 +30,26 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+
+      - uses: actions/checkout@v3
```
qwandor, line 34: Checking out once is enough.

qwandor (commented):
I don't think adding a bunch of blank lines is going to help cache anything.

.github/workflows/publish.yml
mgeisler: Ehem... yeah, good point! I copied the wrong line :-D Should be fixed now.

.github/workflows/publish.yml
```diff
@@ -30,18 +30,26 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+
+      - uses: Swatinem/rust-cache@v2
+
       - name: Install mdbook
         run: cargo install mdbook
```
qwandor, line 37: You should probably pin these to a specific version rather than just whatever happens to be the latest, and add the versions to the cache key.

mgeisler: The `@v2` part pins it to that release. But perhaps you're talking about the `cargo install` commands? Yeah, I agree and I'll move them over to use a version number next.

mgeisler:
> I don't think adding a bunch of blank lines is going to help cache anything.

Agree, I was trying to add the `Swatinem/rust-cache@v2` action, but I copy-pasted the wrong thing :-)

.github/workflows/publish.yml
mgeisler: Yes, let me put up a PR for that next.

PR #15, opened 2022-12-21
`welcome.md`: indicate we skip the subtle details of unsafe Rust

jrvanwhy:
This sentence was missing a negation: it was supposed to say this course does not cover unsafe Rust in depth, but actually said "we go into the subtle details". This PR adds the missing negation.


mgeisler:
Thanks so much!

PR #16, opened 2022-12-21
remove "No memory leaks" from guarantees

sharnoff:
Rust doesn't guarantee this, as evidenced by the safe `Box::leak` function.

### DIFF
--- a/src/why-rust/compile-time.md
+++ b/src/why-rust/compile-time.md
@@ -3,7 +3,6 @@
 Static memory management at compile time:
 
 * No uninitialized variables.
-* No memory leaks.
 * No double-frees.
 * No use-after-free.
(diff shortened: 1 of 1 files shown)


mgeisler:
Hi @sharnoff, thanks for the PR!

> Rust doesn't guarantee this, as evidenced by the safe `Box::leak` function.

Yeah, you're of course technically correct — the best kind of correct :smile:  However, I am thinking of unintended memory leaks here (the kind you get in C if you `malloc` and throw away the pointer).

Could we perhaps add an asterix and a footnote to prompt the presenter to mention `Box::leak` here? I think that could work well: it would let the course participants know that there is more to it than what meets the eye here.

mgeisler:
Let's continue working on this in #28.

PR #21, opened 2022-12-22
"question" typo

Arthur-Milchior:
You actually may want to consider removing this section, or indicating when/where people that are not Googlers may follow this course. 
Plus, this is displayed as a book. You may either want to allow users to get it as Google Slide/pdf, or rename "slide" to "book" here

### DIFF
--- a/src/structure.md
+++ b/src/structure.md
@@ -16,7 +16,7 @@ days:
 
 The course is interactive and your questions will drive our exploration of Rust!
 
-* Please ask quesitons when you get then, don't save them to the end.
+* Please ask questions when you get then, don't save them to the end.
 * Discussions are very much encouraged!
 * We will likely talk about things ahead of the slides.
   * The slides are just a support and we are free to skip them as we like.


mgeisler:
Hi @Arthur-Milchior, thanks for the PR! The typo was fixed in #5 — there were suddenly a lot of PRs after the course was announced.

> You actually may want to consider removing this section, or indicating when/where people that are not Googlers may follow this course.

Yeah, I see what you mean. The [Course Structure](https://google.github.io/comprehensive-rust/structure.html) page is there to start the class — when I teach the course, I use it outline what we'll be doing over the next four days. It's obviously a bit weird to read that when you sit at home and there are no "Day 1" for you :smile: 

> Plus, this is displayed as a book. You may either want to allow users to get it as Google Slide/pdf, or rename "slide" to "book" here

The class is very much meant to be read/used/taken via the website — so that people can run the code samples, modify them and see the results. During an actual class, I execute every single one of them, and I end up modifying most of them based on questions and ideas from the class. So a static PDF doesn't really provide the same experience.

In general, the material I've put up here is not super well-suited for self-study. The material needs someone with experience who can use the code as a starting point for the exploration. I hope to put up videos at some point which can make it a bit better, but still not ideal.

mgeisler:
The typo is fixed, so I think we can close this PR.

PR #22, opened 2022-12-22
Clarify from/into

Arthur-Milchior:
I found the phrasing confusing. Mentionning that a class implements From and then using `into()` seems confusing. It looks like that you meant that it implements `Into`.
So I believe that a little more details may help show why this exercise is valid.

I would have considered clarifying that `into` does not need to specify that we use the implementation for `i16` thanks to type analysis, but I assume this would add too much text, and this can either be ignored by the reader for the time being, or guessed

### DIFF
--- a/src/exercises/day-1/implicit-conversions.md
+++ b/src/exercises/day-1/implicit-conversions.md
@@ -23,8 +23,9 @@ Implementing these traits is how a type expresses that it can be converted into
 another type.
 
 The standard library has an implementation of `From<i8> for i16`, which means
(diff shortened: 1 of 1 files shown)


mgeisler:
> I found the phrasing confusing. Mentionning that a class implements From and then using `into()` seems confusing.

Thanks, I quite agree... this point is always confusing when I get to it in the class :smile: I like your phrasing a lot, so in it goes!

I normally switch between a tab with the course material and the stdlib when teaching the class: so the students get a lot of exposure to how the stdlib looks. For this bit, I use it to find the blanket implementation of `Into<T> for T`.

PR #23, opened 2022-12-22
Limit playground references to tests

Arthur-Milchior:
The system used by https://google.github.io/comprehensive-rust/exercises/day-1/book-library.html allows to execute code directly in the book. So playground references became useless. If I got it wrong, it would be nice to at least explain why it’s not sufficient to edit and run in this book.

The only exception I saw is in the test setting. Playground allows to run tests. In this case, I believe adding a sentence warning that tests are ignored by this book running system means that we really must use the playground and ignore the "run" button provided by the book

### DIFF
--- a/src/exercises/day-1/book-library.md
+++ b/src/exercises/day-1/book-library.md
@@ -14,8 +14,8 @@ fn main() {
 }
 ```
 
-Use this to create a library application. Copy the code below to
-<https://play.rust-lang.org/> and update the types to make it compile:
+Use this to create a library application. Update the types to make the code
+below compile:
 
 ```rust,should_panic
 // TODO: remove this when you're done with your implementation.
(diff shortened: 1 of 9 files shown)


mgeisler:
Hi @Arthur-Milchior,

> If I got it wrong, it would be nice to at least explain why it’s not sufficient to edit and run in this book.

Editing in the inline-playgrounds works — that was how the course worked for the first time I ran it internally. However, there is a crucial problem: the inline-playground loses its state as soon as you navigate away from the page. A single Left-Arrow press with focus outside of the text box is enough to lose everything. So that lead to a lot of sadness in the trial run... tears were shed (almost) :smile:

Copying the text to the "real" stateful playground works better. It also shows `stderr` right away: the inline-playground only shows `stderr` if compilation is unsuccessful. That's something I would like to fix in upstream [mdbook](https://github.com/rust-lang/mdBook/) so that `stderr` is shown always. That would allow me to use the `dbg!` macro throughout and save a lot of typing.

mgeisler:
I could perhaps create a document outside the course which mentions these quirks...

mgeisler:
We can now point this out in the speaker notes: #118. The problem with `stderr` is still there, but we'll probably have to fix this upstream.

PR #24, opened 2022-12-22
Uses into_inter consistently in day 1

Arthur-Milchior:
This page started by using `iter` and then used `into_iter`. No explanation was given. Worse, since the second part is about IntoIterator, intuition would indicate that the word "into" in `IntoIterator` and in `into_iter` refers to the same concept, and this is why `into` was not necessary in the first part of the page.

Since the authors seems not to want to introduce the difference between `into_iter` and `iter` here, I believe it’s more sensical to use `into_iter` everywhere

### DIFF
--- a/src/exercises/day-1/iterators-and-ownership.md
+++ b/src/exercises/day-1/iterators-and-ownership.md
@@ -22,7 +22,7 @@ You use this trait like this:
 ```rust,editable
 fn main() {
     let v: Vec<i8> = vec![10, 20, 30];
-    let mut iter = v.iter();
+    let mut iter = v.into_iter();
 
     println!("v[0]: {:?}", iter.next());
     println!("v[1]: {:?}", iter.next());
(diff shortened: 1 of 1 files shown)


maurer:
My understanding is that this chapter *is* intended to introduce the difference between `.iter()` and `.into_iter()`, thus "What is the type returned by the iterator? Test your answer here:" and "
Why is this type used?", referring to the fact that the returned `Iterator` has `Item=&i8` instead of `Item=i8`, intended as a segue into `IntoIterator`.

If you would prefer to make that more explicit, rather than depending on the speaker as these notes originally were, I think that could be helpful.

mgeisler:
> My understanding is that this chapter _is_ intended to introduce the difference between `.iter()` and `.into_iter()`, thus "What is the type returned by the iterator? Test your answer here:" and " Why is this type used?", referring to the fact that the returned `Iterator` has `Item=&i8` instead of `Item=i8`, intended as a segue into `IntoIterator`.

Yes, that was the idea! :smile: 

We could add a note about this somewhere if you like?

mgeisler:
I'll close this since the idea was to let people play with `iter()` and `into_iter()` on their own to notice the differences.

PR #28, opened 2022-12-22
Add footnote about memory leaks

zdimension:
For https://github.com/google/comprehensive-rust/pull/16


mgeisler:
Nice, thanks a lot!

src/why-rust/compile-time.md
```diff
@@ -3,10 +3,12 @@
...
 * No data races between threads.
 * No iterator invalidation.
+
+[^leaks]: It is technically possible to produce a memory leak in (safe) Rust. The [`Box::leak`](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.leak) method allows getting a raw reference out of a [`Box`](https://doc.rust-lang.org/std/boxed/struct.Box.html) and dropping the [`Box`](https://doc.rust-lang.org/std/boxed/struct.Box.html) afterwards, without running the destructor. A use of this could be to get runtime-initialized and runtime-sized static variables. It is also possible to leak memory by creating a reference cycle, for example by using [`Rc`](https://doc.rust-lang.org/std/rc/struct.Rc.html) and a self-referential type (see [Chapter 15.6 in the Rust Book](https://doc.rust-lang.org/book/ch15-06-reference-cycles.html)). Or simply, the [`std::mem::forget`](https://doc.rust-lang.org/std/mem/fn.forget.html) function, which makes the compiler "forget" about a value meaning the destructor is never run. There are many other ways to create leaks in safe Rust, but for the purpose of this course, "No memory leaks" should be understood as "Pretty much no *accidental* memory leaks".
```
mgeisler, line 14: This is great, though I would prefer to trim it a little (remember this is shown to people on a screen during a class — space is quite limited).

I would leave in the part about `Box::leak` and `mem::forget` since they're both simple to understand (throw away a pointer). The `Rc` part would work much better on the page about [`Rc`](https://google.github.io/comprehensive-rust/std/rc.html).

zdimension: Makes sense, didn't think about screen space when writing it. I removed the part about Rc.

sharnoff:
This looks really good!

I'd still recommend adding a qualification into the bullet point itself - something like "No memory leaks — mostly"? Becuause it _is_ true that Rust's automatic memory management means memory leaks are mostly isolated to certain parts of the language, but "No memory leaks" is a common misconception I think it would be good to avoid.

src/why-rust/compile-time.md
mgeisler: Thanks!

mgeisler:
> "No memory leaks" is a common misconception I think it would be good to avoid.

Yeah, fair enough! I don't want to perpetuate falsehoods :smile:

I would be happy to add the "... mostly" on the line with the bullet point as you suggest. Then there should be no confusion.

mgeisler:
Thank you both very much!

PR #29, opened 2022-12-22
remove `$` to make copy paste easier

anugrahsinghal:
removing the `$` sign user can directly copy and paste the command into terminal

### DIFF
--- a/src/cargo.md
+++ b/src/cargo.md
@@ -8,7 +8,7 @@ and how it fits into this training.
 On Debian/Ubuntu, you can install Cargo and the Rust source with
 
 ```shell
-$ sudo apt install cargo rust-src
+sudo apt install cargo rust-src
 ```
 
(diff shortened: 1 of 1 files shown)


mgeisler:
Hi @anugrahsinghal, thanks for the PR!

> removing the `$` sign user can directly copy and paste the command into terminal

The goal is good, but this has side effects: it will make it harder for us to show command lines together with the output of running the command. Like this:

```shell
$ echo hello
hello
```

I would like to keep it for that purpose. A better solution would be to teach `mdbook` to not select the `$` when you copy the content of a `shell` code block. I've now opened https://github.com/rust-lang/mdBook/issues/1963 to see if we can have `mdbook` support this for us.

anugrahsinghal:
sounds fair @mgeisler , closing this PR

PR #30, opened 2022-12-22
Include version numbers when installing `mdbook`

mgeisler:


.github/workflows/publish.yml
```diff
@@ -34,10 +34,10 @@ jobs:
       - uses: Swatinem/rust-cache@v2
 
       - name: Install mdbook
-        run: cargo install mdbook
+        run: cargo install mdbook --version 0.4.25
 
       - name: Install mdbook-svgbob
-        run: cargo install mdbook-svgbob
+        run: cargo install mdbook-svgbob --version 0.2.1
```
qwandor, line 40: These version numbers should probably be part of your cache key.

mgeisler: Cargo will only be able to reuse the data from the GitHub Action cache if the version number is the same, so this should be fine, no?

PR #34, opened 2022-12-22
Added JNA Documenation for Java Interoperability

Crain-32:
Although the primary focus of the document is on Rust, being able to provide an example of consuming Rust in Java using JNA is helpful for those more familiar with JNA over JNI. The main hiccup I have is the `Android.bp`, as I don't know how to include JNA for the Java side in that build tool.


mgeisler:
Hi @Crain-32, this looks cool, thanks!

I'm glad you could make your way around the `build_all.sh` script... it's a clunky way for me to ensure that the Android part actually builds when you put it into an AOSP checkout. Were you able to make it work?

> Although the primary focus of the document is on Rust, being able to provide an example of consuming Rust in Java using JNA is helpful for those more familiar with JNA over JNI.

> The main hiccup I have is the `Android.bp`, as I don't know how to include JNA for the Java side in that build tool.

Yeah, that won't work directly... Third-party dependencies need to be imported into the `external/` directory. I'm not actually sure how that works for Java code.

Perhaps we should not make this Android-specific? The interoperability is put under Android right now since the course was written for Android engineers :smile: The goal is to show how AOSP does things so that people can get started quickly with Rust.

I would suggest rephrasing the new JNA page in non-AOSP terms for now.

Crain-32:
I have no idea if the `build_all.sh` script works for my example. As my target swapped more to Windows, my example is incomplete on the configuration. As it is still missing a `Cargo.toml` setup for the Rust side to compile correctly. I've decided to ignore that for now, as you mentioned this is targeted at Android.

mgeisler:
Hey agian @Crain-32,

Yeah, I think we'll let this wait a bit for now — the Android part is really about the "Android Open Source Project" right now. We could create a more generic page about Rust and Android, but it should probably be in a new "Other Topics" section for now since it could make the course larger than the four days we have right now.

I'll leave the PR open and then we can see where we can put the material at a later point.

Crain-32:
Sounds good.

PR #36, opened 2022-12-22
Tuples with "..." make no sense, so don't use them

daym:


mgeisler:
Hey @daym, thanks for the PR!

I was actually trying to use `...` as a place holder there, meaning that there could be more `Tn` types following. But I see now that it can be understood as being part of the syntax.

How does it look if we include a few different tuple sizes in the example? Like `(), (T,), (T1, T2)`. I think that could make it clearer without using made-up syntax.

mgeisler:
Hi @daym, I updated the branch with my suggestion above. If you sign the CLA, then we can merge this. Thanks!

daym:
Hi,

> How does it look if we include a few different tuple sizes in the example? Like `(), (T,), (T1, T2)`. I think that could make it clearer without using made-up syntax.

Yes, I took a look and that looks good to me!

Also, I've signed the CLA now.

qwandor (approved):

mgeisler:
> Yes, I took a look and that looks good to me!
> 
> Also, I've signed the CLA now.

Thanks a lot!

PR #41, opened 2022-12-23
Fix C array syntax in memory-management/manual.md

reillyeon:
Heap-allocated arrays in C use `int*` instead of `int[]`.

### DIFF
--- a/src/memory-management/manual.md
+++ b/src/memory-management/manual.md
@@ -8,7 +8,7 @@ You must call `free` on every pointer you allocate with `malloc`:
 
 ```c
 void foo(size_t n) {
-    int[] int_array = (int*)malloc(n * sizeof(int));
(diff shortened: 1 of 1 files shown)


mgeisler:
Hi @reillyeon, thanks for the PR!

> Heap-allocated arrays in C use `int*` instead of `int[]`.

I'm not a regular C programmer, so I thought the two things meant the same? I used `int[]` to name it look like an array for other non-C programmers. I'm happy to merge this, I'm just curious about what the difference is.

reillyeon:
`int[]` isn't a valid type declaration. C puts the brackets on the variable name. When you say `int array[3]` that declares a fixed-size array (either stack-allocated or as a struct field). It's a bit of a special case in the type system though, you can't later say `array = ...`. There's no valid right-hand side to that assignment.

mgeisler:
Okay, thanks for the explanation!

PR #42, opened 2022-12-23
Change "semantics is" to "semantics are"

reillyeon:
Subject-verb agreement.

### DIFF
--- a/src/ownership/copy-clone.md
+++ b/src/ownership/copy-clone.md
@@ -1,6 +1,6 @@
 # Copying and Cloning
 
-While move semantics is the default, certain types are copied by default:
+While move semantics are the default, certain types are copied by default:
 
 ```rust,editable
 fn main() {


mgeisler:
Hi @reillyeon, thanks for catching this!

Non-native speaker here (I'm from Denmark), so these fixes are wonderful :smile:

Now that I read it again, would it not be better to use "While move semantic is the default"? I don't know why I want to add an "s" to "semantic" every time I say it, but I guess it should be singular?

qwandor:
I think you want to say "move semantics are...". The singular sounds wrong.

mgeisler:
> I think you want to say "move semantics are...". The singular sounds wrong.

Okay, excellent! I'll go with that :smile:

reillyeon:
I had to look this up because while I know as a native speaker that "move semantics" is correct and that "semantics" is plural I wasn't sure why. Turns out that "semantic" and "linguistic" are examples where the word that looks like the singular form is an adjective while the word that looks like the plural form is the noun. For example, the study of linguistic rules is called "linguistics". You can't have a singular "linguistic". I guess the same as how you can't have a singular "pant". English is weird.

mgeisler:
> I guess the same as how you can't have a singular "pant". English is weird.

Thanks for the explanation! And yeah, all languages are weird in their own ways 😄

PR #45, opened 2022-12-23
Makes ownership of `self` more accurate

Arthur-Milchior:
It’s entirely possible that a method gives ownership of self to another object, or return `self`.  For example
```
    fn id(self) -> Self {
        return self;
    }
```
or more realistically 
```
    fn or(self, other) -> Self {
      if (self.isValid()) { return self;}
      return other;
    }
```


mgeisler:
Thanks, that's a good point!

PR #48, opened 2022-12-23
Dark Mode SVG Text Fix

nofurtherinformation:
Currently, SVG text in many of the diagrams appears as black on dark blue in dark mode. The SVG `text` elements currently receive the default color, black:

Light mode:
image

Dark mode:
image

This PR suggests a one-liner CSS fix to apply the foreground css variable (`var(--fg)`) to the svg text element to make svg text legible and accessible. This is bundled into the output using the `additional-css` property in mdbook's config `book.toml`. 

This is probably something that could be fixed upstream in `mdbook`/`mdbook-svgbob`  sometime, but this patch would fix it for current usability.

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -14,6 +14,7 @@ class = "bob"
 [output.html]
 curly-quotes = true
 additional-js = ["ga4.js"]
+additional-css = ["svgbob.css"]
 git-repository-url = "https://github.com/google/comprehensive-rust"
 edit-url-template = "https://github.com/google/comprehensive-rust/edit/main/{path}"
 
--- a/svgbob.css
+++ b/svgbob.css
@@ -1,0 +1,4 @@
+/* Ensure text is legible in all themes. */
+svg text {
+  fill: var(--fg);
+}


book.toml
```diff
@@ -14,6 +14,7 @@ class = "bob"
 [output.html]
 curly-quotes = true
 additional-js = ["ga4.js"]
+additional-css = ["ga4.css"]
```
mgeisler, line 17: Please call this something like `svgbob.css` instead — the name `ga4` is short for Google Analytics 4 since that's what the JavaScript file implements :-)

mgeisler:
Hi @nofurtherinformation, thanks a lot for looking into this!

> This PR suggests a one-liner CSS fix to apply the foreground css variable (`var(--fg)`) to the svg text element to make svg text legible and accessible.

Super nice! I definitely want the different themes to be legible so this is important to fix. Could you update the PR description with an image showing how it looks after your fix? (I can check it out myself, but it'll be useful for others to see what the fix does).

I agree that we should file an issue upstream (probably mdbook-svgbob) about this. I'll be happy to do that after we merge this.

book.toml
mgeisler: I renamed the file and pushed to your branch. I'll merge this next.

nofurtherinformation:
Apologies for the slow reply here @mgeisler (I was a bit offline for the holidays) and thanks for merging this in! For posterity, here is a screenshot of the new styling. 
Screen Shot 2023-01-03 at 10 07 02 AM

mgeisler:
No worries at all, I was also out myself :smile: Thanks for the elegant fix, it would have taken me a while to figure this out myself!

mgeisler:
I reported it upstream now, I hope the fix can be included there somehow: https://github.com/boozook/mdbook-svgbob/issues/22.

PR #49, opened 2022-12-23
`anyhow`: Remove `ReadUsernameError::IoError`

### LINKED ISSUE #116: Technical nit: Adding Context to Errors
The `IoError` variant of `ReadUsernameError` enum is not used anymore, after switching to `anyhow::Result` return type. Perhaps this variant should be removed from the example.

ilyagr:
As far as I understand, the example with the `anyhow` example does not use its special `IoError` anymore, unlike the previous example with just `thiserror`.

I think this should be made clear. I considered just adding a `// Now unused` comment, but I think ti's clearer to just remove the unused code.

Fixes #116.

### DIFF
--- a/src/error-handling/error-contexts.md
+++ b/src/error-handling/error-contexts.md
@@ -11,8 +11,6 @@ use anyhow::{Context, Result};
 
 #[derive(Error, Debug)]
 enum ReadUsernameError {
-    #[error("Could not read: {0}")]
-    IoError(#[from] io::Error),
     #[error("Found no username in {0}")]
     EmptyUsername(String),
 }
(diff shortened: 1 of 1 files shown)


mgeisler:
> As far as I understand, the example with the `anyhow` example does not use its special `IoError` anymore, unlike the previous example with just `thiserror`.

Oh, are you saying that calling `.context(...)` converts the type away from `io::Error`? I had not thought of that...

ilyagr:
Exactly.

Tangentially related: https://github.com/dtolnay/thiserror/issues/214

ilyagr:
Actually, not exactly. I think it's still `io:Error`, but wrapped into `anyhow::Error` as opposed to `ReadUsernameError::IoError`.

PR #54, opened 2022-12-25
Typos

Arthur-Milchior:
Not entirely sure of the intended meaning of the second sentence. I assume I’m not too far away


mgeisler:
> I assume I’m not too far away

Yeah, I agree :-) I changed _some_ since we're talking about a specific `T` value found inside `Mutex<T>`.

PR #55, opened 2022-12-25
Typo

Arthur-Milchior:
Not sure of the intended meaning. At least the sentence is true and gramattically correct here

### DIFF
--- a/src/concurrency/send-sync/examples.md
+++ b/src/concurrency/send-sync/examples.md
@@ -37,5 +37,5 @@ These types are not thread-safe and cannot be moved to other threads:
 
 * `Rc<T>`: each `Rc<T>` has a reference to an `RcBox<T>`, which contains a
   non-atomic reference count.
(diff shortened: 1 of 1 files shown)


mgeisler:
Oh my... That looks super weird indeed.

I wanted to write "Rust assumed that...".

QuineDot:
I don't think it's about lifetimes per se, it's more of a general lint.  "If you're doing something on the level of pointers, you have to opt in to being `Send` and/or `Sync`."  (E.g. there's a good chance you're doing something FFI related.)

In fact I think mentioning lifetimes is confusing.  Raw pointers have no lifetime parameter.  Perhaps, "Rust assumes raw pointers may have special concurrency considerations."

mgeisler:
> Rust assumes raw pointers may have special concurrency considerations.

Thanks, I like this phrasing! Let's go with that for now :smile:

PR #57, opened 2022-12-25
Precise what the Android parts are about (fixes #56)

### LINKED ISSUE #56: Note that "Android" means "AOSP" and not "Android app"
I’d strongly suggest, in https://google.github.io/comprehensive-rust/welcome.html , to explicitely note that the Android part considers how to contribute to Android source code, using Android build system.
On first reading, the way I saw it presented on twitter, and in this page, I thought that this was aimed for the Android developers which, for some reason, wanted to write part of their app in Rust. 

I’ll immediately admit that a reader could guess it is not actually for app developer, as the two sentences indicate OASP.
> Building Android components in Rust.
> AIDL servers and clients.

The trouble being that, if you’re not used to working on Android source code, you may not know what AIDL is, and that it’s not a Rust specific term you’ll learn later.

Personally, I admit that, since AnkiDroid, an Android app, is already using Rust for its backend and Kotlin for its front-end, I was wondering whether I could learn things to improve our system. I guess it won’t be the case.

And with this, I believe it’s the last PR/issue you’ll get from me.
Thanks for making this public

zdimension:


mgeisler:
Thanks @zdimension!

I might want to add "AOSP" and "Android Platform" somewhere here (with a link to https://source.android.com/). However, this is much better than no note 🙂

Issue #65, opened 2022-12-26
Technical nits: the try operator


QuineDot:
[You only cover `Result`,](https://google.github.io/comprehensive-rust/error-handling/try-operator.html) but [it works with other types too.](https://doc.rust-lang.org/std/ops/trait.Try.html#implementors)  At least mention `Option`, which is commonly used.

[The desugaring example](https://google.github.io/comprehensive-rust/error-handling/converting-error-types.html) is out of date; [the actual desugaring](https://rust-lang.github.io/rfcs/3058-try-trait-v2.html#desugaring-) is now more complicated.  (And technically it was never just `From::from` directly.)  But I get it, you're not going to help anyone by explaining it on that level.  Probably you can just rephrase this page to "the effective expansion ..." instead of "actual expansion", and "works the same as" instead of "actually becomes".

Consider adding the as-if desugaring for `Option` too.

mgeisler:
> Probably you can just rephrase this page to "the effective expansion ..." instead of "actual expansion", and "works the same as" instead of "actually becomes".

Great idea, thanks!


> [You only cover `Result`,](https://google.github.io/comprehensive-rust/error-handling/try-operator.html) but [it works with other types too.](https://doc.rust-lang.org/std/ops/trait.Try.html#implementors) At least mention `Option`, which is commonly used.

True! I will add this to the speaker notes, or perhaps we can squeeze it into the slide somewhere.

PR #74, opened 2022-12-27
More precise phrasing of what ? expands to

mgeisler:
As per the suggestion from @QuineDot in #65.


maurer (approved):
It might be worth mentioning the `Try` trait in an aside afterwards to avoid the perception that `?` is `Result`-only magic.

mgeisler:
> It might be worth mentioning the `Try` trait in an aside afterwards to avoid the perception that `?` is `Result`-only magic.

Yeah, the instructor should make sure to mention it. It's something which comes up when I teach the course.

Issue #70, opened 2022-12-27
Suggestions: Misc


QuineDot:
[I think you should show the `Option` and `Result` definitions here.](https://google.github.io/comprehensive-rust/std/option-result.html) They'll see the variants everywhere in Rust.

---

[Clarify that a `HashMap` doesn't have to be keyed with a `&str` or `String`](https://google.github.io/comprehensive-rust/std/hashmap.html) (unlike a JSON object, say).

---

[Here,](https://google.github.io/comprehensive-rust/ownership/lifetimes-function-calls.html) I was going to mention that the lifetime can be explicit in some cases... but upon reconsideration that's probably not worth getting into.  However, perhaps you should mention that lifetimes with the same name must indeed be equal.  This will matter a lot should they every run into some `&'a mut Thing<'a>` errors.

But also on that page [and on this one,](https://google.github.io/comprehensive-rust/ownership/lifetimes.html) you say
> Read &'a Point as “a borrowed Point which is valid for at least the lifetime a”. 

That can be misinterpreted as "the borrow is valid for at least `'a`", as opposed to "the `Point` is valid for at least `'a`".  I.e. it might given the mistaken impression that a borrow can be lengthened.

---

[The getters and setters here aren't very idiomatic.](https://google.github.io/comprehensive-rust/exercises/day-2/health-statistics.html)  First of all, nothing here is public -- everything must be in the same module -- so you'd just access the fields directly.  Second of all, instead of a getter an infallible setter, if the struct was public, you'd probably just make those fields public too.  If the fields actually have invariants, the setters shouldn't be infallible, but instead return a `Result<_, _>`.

More generally, it may give the impression that using getters and setters is idiomatic (even in the same module).

---

[Perhaps note this list of receivers is non-exhaustive.](https://google.github.io/comprehensive-rust/methods/receiver.html)  E.g. `Box<Self>`.

---

[I don't know that "omitted" is the right term for `mod in_another_file;`.](https://google.github.io/comprehensive-rust/modules/filesystem.html)  Also perhaps note that (except perhaps for `test` modules) separate files is the norm.

mgeisler:
Hey, just a quick comment for one part above:

[...]

PR #75, opened 2022-12-27
Mention that there are other receiver types

mgeisler:
As per the discussion in #70.

### DIFF
--- a/src/methods/receiver.md
+++ b/src/methods/receiver.md
@@ -14,4 +14,6 @@ are other possible receivers for a method:
 * No receiver: this becomes a static method on the struct. Typically used to
   create constructors which are called `new` by convention.
 
-There are even more types, e.g., `Box<Self>`.
+Beyond variants on `self`, there are also
+[special wrapper types](https://doc.rust-lang.org/reference/special-types-and-traits.html)
+allowed to be receiver types, such as `Box<Self>`.
--- a/src/methods/receiver.md
+++ b/src/methods/receiver.md
@@ -13,3 +13,5 @@ are other possible receivers for a method:
   transmitted.
 * No receiver: this becomes a static method on the struct. Typically used to
   create constructors which are called `new` by convention.
+
+There are even more types, e.g., `Box<Self>`.


src/methods/receiver.md
```diff
@@ -13,3 +13,5 @@ are other possible receivers for a method:
   transmitted.
 * No receiver: this becomes a static method on the struct. Typically used to
   create constructors which are called `new` by convention.
+
+There are even more types, e.g., `Box<Self>`.
```
maurer, line 17: Consider referencing https://doc.rust-lang.org/reference/special-types-and-traits.html so the class can see the enumeration of special receiver types?  (Rc, Arc, Pin are the remaining ones for the moment).

maurer, line 17: Probably rephrase this to make it clear where these types come from, and that they're receiver types, e.g.

"Beyond variants on `self`, there are also [special](https://doc.rust-lang.org/reference/special-types-and-traits.html) wrapper types allowed to be receiver types, such as `Box<Self>`.

maurer, line 17: Should we have a discussion of why you'd want to use alternate receiver types rather than relying on `impl Deref`, or do we want to rely on the instructor for that?

maurer (approved):
LGTM as long as you do something with the sentence to make it clearer that it's part of the receiver discussion - it's clear in the code because it's part or `receiver.md`, but it's not part of the bulleted list on receivers, the sentence doesn't mention receivers, etc.

I've also made some other comments that would improve it, but up to you on those.

src/methods/receiver.md
mgeisler: Great idea to include a link to that page.

mgeisler: I think this is something we'll have to defer to the speaker notes (i.e., to the instructor as you suggest).

As you hint, the pages themselves cannot go into detail on all topics — plus, people are seeing this on Day 2 after having been exposed to Rust for 5-6 hours. I think we should stick to the main point: how does a method look in Rust and related and when do you need to use `&mut self` instead of just `&self`.

mgeisler: Thanks, that's a nice way to put it, I updated the PR to say that.

PR #79, opened 2022-12-27
Comparison: copyedit 'destructors' to 'destructor'

riking:
This is the correct grammatical form for the bullet point, it expands into "Destructors [of objects are] delayed."

### DIFF
--- a/src/memory-management/comparison.md
+++ b/src/memory-management/comparison.md
@@ -25,7 +25,7 @@ Here is a rough comparison of the memory management techniques.
   * Memory leaks.
 * Automatic like Java:
   * Garbage collection pauses.
-  * Destructors delayed.
+  * Destructor delays.
 * Scope-based like C++:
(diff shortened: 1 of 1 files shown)


mgeisler:
Hi @riking,

> This is the correct grammatical form for the bullet point, it expands into "Destructors [of objects are] delayed."

Thanks for catching this. As you've guessed, I'm not a native speaker so I appreciate these grammar fixes a lot!

I think I actually wanted to say "Destructor delays" now that I see it again. How do you find that form?

riking:
Hm, I think that works - one focusing on the fact that work is batched
together, the other focusing on the fact that each individual object is
separately delayed.

mgeisler:
> I think that works - one focusing on the fact that work is batched together, the other focusing on the fact that each individual object is separately delayed.

Okay, great! Will you push the change to your branch or should I do it?

PR #85, opened 2022-12-28
Better Test Case for Day 2 Polygon Exercise

der3318:
`test_shape_circumferences()` is testing the circumference of a `Polygon` with only **2** points in total. However, according to the the general definition:

"A polygon is a closed figure where the sides are all line segments. Each side must intersect **exactly two others sides** but only at their endpoints. The sides must be noncollinear and have a common endpoint."

In short, a valid polygon is supposed to have at least **3** points.

### Proposal

To avoid ambiguity (i.e., users might implement `circumference()` with an early-return condition like `points.len() < 3` and end up with test failure), let's add one more point to the test case.

### DIFF
--- a/src/exercises/day-2/points-polygons.rs
+++ b/src/exercises/day-2/points-polygons.rs
@@ -84,8 +84,8 @@ impl Polygon {
     }
 
     pub fn length(&self) -> f64 {
-        if self.points.is_empty() {
-            return 0.0;
+        if self.points.len() < 3 {
+            return f64::NAN;
         }
 
         let mut result = 0.0;
@@ -207,6 +207,7 @@ mod tests {
     fn test_shape_circumferences() {
         let mut poly = Polygon::new();
         poly.add_point(Point::new(12, 13));
+        poly.add_point(Point::new(17, 11));
         poly.add_point(Point::new(16, 16));
         let shapes = vec![
             Shape::from(poly),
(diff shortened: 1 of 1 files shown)


src/exercises/day-2/points-polygons.rs
```diff
@@ -84,8 +84,8 @@ impl Polygon {
     }
 
     pub fn length(&self) -> f64 {
-        if self.points.is_empty() {
-            return 0.0;
+        if self.points.len() < 3 {
+            return f64::NAN;
```
mgeisler, line 88: I would prefer to not do this change: the goal of the exercise is to "trick" people to write a few lines of Rust. It's not a goal to teach people about the corner cases of IEEE 754 floating point arithmetic :smile:  Basically, keep it simple so that people can focus on writing methods and on exploring the standard library.

mgeisler:
Hey @der3318, thanks for the PR.

> To avoid ambiguity (i.e., users might implement `circumference()` with an early-return condition like `points.len() < 3` and end up with test failure), let's add one more point to the test case.

Thank you, that's a great idea!

src/exercises/day-2/points-polygons.rs
der3318: Make sense to me. I just reverted the part.

mgeisler:
Thanks for the update!

Issue #53, opened 2022-12-24
Add speaker notes


mgeisler:
Now that we've published the course for all to see, we have many people who want to use the course for self-study.

To help with that, I think we should add speaker notes. Perhaps we can use https://github.com/FreeMasen/mdbook-presentation-preprocessor for this.

We should also publish videos, I've created #52 for tracking that.

PR #88, opened 2022-12-30
Add example for shared_state

cuzfrog:
Reading this simple and comprehensive book, while finding the example for shared_shate is not complete. Hope this added simple example could serve.

### DIFF
--- a/src/concurrency/shared_state/example.md
+++ b/src/concurrency/shared_state/example.md
@@ -2,16 +2,18 @@
 
 Let us see `Arc` and `Mutex` in action:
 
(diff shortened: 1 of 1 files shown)


mgeisler:
Hi @cuzfrog,

You have actually solved the exercise which I want to solve with the students in class :smile: 

In other words, the example doesn't compile because I want the students to "help" me fix it during class. That's obviously not very clear right now — I'm working on adding support for speaker notes #53 and then I'll describe the solution in the notes. That should make the course much better for self-study.

I hope that makes sense!

PR #89, opened 2022-12-31
Typo in "Fearless Concurrency" intro

jswalden:
"referred to a _fearless concurrency_" -> "referred to a**s** _fearless concurrency_"


mgeisler:
Thanks for catching this!

PR #93, opened 2023-01-02
Corrected typo on receiver.md

Fosco1:
According to https://dictionary.cambridge.org/it/grammatica/grammatica-britannico/it-s-or-its, "its" should be used instead of "it's"


mgeisler:
Thanks!

PR #96, opened 2023-01-03
Simplify the solution to prefix matching

zhouwfang:


mgeisler:
Hi @zhouwfang, thanks a lot! This looks much better indeed :smile: 

I would be happy to merge this, we just need the CLA to be signed.

zhouwfang:
Hi @mgeisler, CLA has been signed. Thanks.

zhouwfang:
Hi @mgeisler, It should be all set now. Thanks.

mgeisler:
Thanks! Could I also get you to add a few comments to explain the solution? In particular the trick with turning the elements into `Option` so we can detect when the prefix iterator ends.

PR #98, opened 2023-01-03
Simplify the publish workflow

mgeisler:
This removes a bunch of boiler-plate comments which were left from the template I used originally.

It also simplifies the name to just “Publish” since I feel this fits better for what we do here.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,15 +1,11 @@
-# Simple workflow for deploying static content to GitHub Pages
-name: Deploy static content to Pages
+name: Publish
 
 on:
-  # Runs on pushes targeting the default branch
   push:
-    branches: ["main"]
-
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -1,28 +1,25 @@
...
+      - main
   workflow_dispatch:
 
-# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
 permissions:
   contents: read
   pages: write
   id-token: write
 
-# Allow one concurrent deployment
```
qwandor, line 18: This seems like a useful comment, it is not obvious what the config means without it.

qwandor (approved):

.github/workflows/publish.yml
mgeisler: Yeah, good point!

PR #103, opened 2023-01-03
Update match-expressions.md

snoopdougiedougie:
The original code would always print "Unknown program name!".

### DIFF
--- a/src/control-flow/match-expressions.md
+++ b/src/control-flow/match-expressions.md
@@ -5,13 +5,16 @@ that sense, it works like a series of `if let` expressions:
 
 ```rust,editable
 fn main() {
-    match std::env::args().next().as_deref() {
-        Some("cat") => println!("Will do cat things"),
-        Some("ls")  => println!("Will ls some files"),
-        Some("mv")  => println!("Let's move some files"),
-        Some("rm")  => println!("Uh, dangerous!"),
-        None        => println!("Hmm, no program name?"),
-        _           => println!("Unknown program name!"),
+    let args: Vec<String> = std::env::args().collect();
+
+    for s in &args[1..] {
+        match &s[..] {
+            "cat" => println!("Will do cat things"),
+            "ls" => println!("Will ls some files"),
+            "mv" => println!("Let's move some files"),
+            "rm" => println!("Uh, dangerous!"),
+            _ => println!("Unknown program name!"),
+        }
     }
 }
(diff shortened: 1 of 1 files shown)


snoopdougiedougie:
Note that this allows multiple commands, and prints the message for each. So you can run:

```cargo run rm mv```

mgeisler:
Hi @snoopdougiedougie, thanks for spotting this.

> The original code would always print "Unknown program name!".

Yeah, I think I didn't pick a super good example to be honest. Perhaps we can rework it to something which lends itself better to an interactive class.

When teaching the class, I live code using the code on the slides (pages). Here it would be great to have a code snippet which both

1. Is easy to modify in front of a class (`std::env::args()` is not easy to modify in a playground environment 🙂)
2. Shows off the _expression_ part. That is, I would like to assign the value of `match` to something.

Perhaps we can stay in the same theme and hard-code the arguments. We could then use `match` to determine the expected number of arguments... A problem with this is that all of these classical Unix programs take one or more arguments.

Perhaps we can find a better program name to match against — perhaps we say that we're reimplementing `git` instead and want to match its subcommands?

snoopdougiedougie:
Yeah, it's hard to find a good example for a class. You want something "real-world", but not too difficult. The problem with parsing the command line to show how match works isn't how you usually accept command-line options in Rust--you usually use Clap.

mgeisler:
> works isn't how you usually accept command-line options in Rust--you usually use Clap.

True :smile: but I hope the students who take the class will forgive us for this. It would be cool to explain how to actually do it in a speaker note — I've just put up #107 which should help with that.

KunleWire8148:
#107

wire

mgeisler (changes requested):
Please update it to use a fixed vector instead — make _pattern matching_ the focus, not command line shenanigans :smile: 

The call to `as_deref` is also something which raises questions every time I teach the class. We can either leave it and explain what it does in a note, or we can rework the example to make it unnecessary.

mgeisler:
> Please update it to use a fixed vector instead — make _pattern matching_ the focus, not command line shenanigans smile

I'm of course also fine doing this myself, just let me know. We can close this PR and I'll put something else up eventually.

mgeisler:
Let's close this for now — I think the example should be rewritten to not use the arguments at all since that's just extra noise.

PR #112, opened 2023-01-05
Show speaker notes as closed by default

mgeisler:
This makes it clearer that the notes are extra content.

### DIFF
--- a/speaker-notes.js
+++ b/speaker-notes.js
@@ -61,7 +61,7 @@
   // Get the state of the speaker note window: "inline-open", "inline-closed",
   // or "popup".
   function getState() {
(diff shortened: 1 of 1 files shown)


qwandor:
Can something be done with CSS too, so they are hidden by default even before the Javascript loads?

mgeisler:
> Can something be done with CSS too, so they are hidden by default even before the Javascript loads?

Actually, browsers show them as closed by default — that seems to be the built-in behavior of the `details` element I use. So before the JavaScript would open them immediately on load. Now the code will simply end up doing nothing on the initial run.

qwandor (approved):

PR #122, opened 2023-01-06
Create an empty Vec rather than pre-allocating

marshallpierce:
It might give the impression that you can only write to a Vec that has capacity, when in fact Vec's Write impl will grow the storage as needed. While pre-allocating is probably a good efficiency win in many circumstances, I think it's probably worth minimizing the number of concepts in play in this example.


mgeisler:
Okay, fair enough!

mgeisler (approved):

PR #123, opened 2023-01-06
Error handling: clarify printing of Result

ilyagr:
Two examples may print either `Ok(username)` or `Err(error)`.
This commit clarifies this fact.

This is not so much a fix as a way to not confuse users who may not run the code.

### DIFF
--- a/src/error-handling/try-operator.md
+++ b/src/error-handling/try-operator.md
@@ -41,6 +41,6 @@ fn read_username(path: &str) -> Result<String, io::Error> {
 fn main() {
     //fs::write("config.dat", "alice").unwrap();
     let username = read_username("config.dat");
-    println!("username: {username:?}");
+    println!("username or error: {username:?}");
 }
 ```
--- a/src/error-handling/converting-error-types.md
+++ b/src/error-handling/converting-error-types.md
@@ -46,6 +46,6 @@ fn read_username(path: &str) -> Result<String, ReadUsernameError> {
 fn main() {
     //fs::write("config.dat", "").unwrap();
     let username = read_username("config.dat");
-    println!("username: {username:?}");
+    println!("username or error: {username:?}");
 }
 ```


src/error-handling/try-operator.md
```diff
@@ -41,6 +41,6 @@ fn read_username(path: &str) -> Result<String, io::Error> {
 fn main() {
     //fs::write("config.dat", "alice").unwrap();
     let username = read_username("config.dat");
-    println!("username: {username:?}");
+    println!("username or error: {username:?}");
```
mgeisler, line 44: I conceptually think of `username` as "the username", even if it contains an error.

Could I convince you to instead add a speaker note which explains a bit what's going on on this slide? That would be a block like

```
<details>

Key points:

* The `username` variable can be either `Ok(...)` or `Err(...)`.
* Use the `fs::write` call to test out the different scenarios: no file, empty file, file with username.

</details>
```

You can expand this with any other details you think the instructor and the students should know!

mgeisler (changes requested):
Hey @ilyagr, thanks for the PR! I would prefer to keep the text as is so that it remains simple (remember this is shown as a presentation in a classroom).

However, you have a great point and I think we should add a speaker note which explains this instead. I hope I can have you convert the suggestion into such a note?

src/error-handling/try-operator.md
ilyagr: I think putting it in the notes would be OK, but I would definitely prefer to keep it in the code. I personally was quite confused for a bit about where the error went when I looked at it. So, let me try to convince you once more.

How about this: as opposed to what I suggested before, perhaps a better idea would be to keep the `println!` text the same but change the variable name to `username_or_errror`. That would make this code a bit longer, but it would pay off when the variable name changes to just `username` later.

If this doesn't sound good, I'll put it in a note, as you suggested.

mgeisler: > I personally was quite confused for a bit about where the error went when I looked at it. So, let me try to convince you once more.

Okay, we should not leave people confused so thank you for telling me this! It's super hard to recognize what is confusing after having worked with the language for a few years :smile:

I'm okay with changing the `println!` format string like you do in this example.

I would also love for your to add a speaker note explaining what's going on — as a separate change.

> change the variable name to `username_or_errror`. That would make this code a bit longer, but it would pay off when the variable name changes to just `username` later.

I'm not a fan of this naming convention since it makes the variable type (`Result<String, ReadUsernameError>` in this case) part of the variable name. Some languages and frameworks do that a lot (see [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation) for a famous example), but I don't see this as a common or useful thing in Rust.

mgeisler (approved):

mgeisler:
I thought some more about it and you're right: it's useful to show what's going on in the format string.

If you find some time, please add speaker notes to the two pages with more extensive explanation. Explain what the instructor (and thus indirectly the students) should do: test with no file, test with empty file, test with non-empty file.

src/error-handling/try-operator.md
ilyagr: > I'm not a fan of this naming convention

Yeah, on second thought the idea of making it a username wasn't that great.

mgeisler:
Thanks for creating #157!

PR #126, opened 2023-01-06
Use rustls instead of openssl

marshallpierce:
Using the `rustls-tls` feature on reqwest will use rustls and ring, which should compile on any system. (Ring still uses some C and assembly, but doesn't depend on openssl/boringssl AFAIK.)

### DIFF
--- a/src/exercises/day-4/link-checker.md
+++ b/src/exercises/day-4/link-checker.md
@@ -11,7 +11,7 @@ Cargo project and `reqwest` it as a dependency with:
 ```shell
 $ cargo new link-checker
 $ cd link-checker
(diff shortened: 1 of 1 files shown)


src/exercises/day-4/link-checker.md
```diff
@@ -11,7 +11,7 @@ Cargo project and `reqwest` it as a dependency with:
 ```shell
 $ cargo new link-checker
 $ cd link-checker
-$ cargo add --features blocking reqwest
```
marshallpierce, line 14: Oops, I need to add the equivalent change to the Cargo.toml contents below

PR #130, opened 2023-01-08
Add support for translations

### LINKED ISSUE #115: Support translations
It would be great to support translating the course into other languages. I suggest doing this with the infrastructure I wrote here: https://github.com/rust-lang/mdBook/pull/1864.

We don't need to wait for this to merge upstream (infact, we can probably just pull in the code as stand-alone binaries in this repository).

With this system, we can setup something like https://pontoon.rust-lang.org/ to make it easy for people to contribute fixes to the translations.

mgeisler:
This implements a translation pipeline using the industry-standard [Gettext][1] system.

I picked Gettext for the reasons described in https://github.com/rust-lang/mdBook/pull/1864 and https://github.com/rust-lang/mdBook/issues/5#issuecomment-1144887806:

* It’s widely used in open source software. This means that there are graphical editors which will help you in editing the `.po` files.

  There are also many websites which allows you to do translation via an online flow. An example is [Pontoon][4], which is used for the Rust website itself. We can consider setting up such an instance ourselves.

* It is a light-weight yet structured format. This means that nothing changes with regards to how you update the original English text. We can still accept fixes and PRs like normal.

  The structure means that translators can see exactly which part of the course they need to update after a change. This is completely lost if you simply copy over the original text and translate it in-place in the Markdown files.

The code here only adds support for translations. They are not yet published or used for anything. Next steps will be

* Add support for switching languages via a bit of JavaScript on each page.
* Update the speaker notes feature to support translations (right now “Speaker Notes” is hard-coded into the generated HTML). I think we should turn it into a mdbook preprocessor instead.

Fixes #115.

[1]: https://www.gnu.org/software/gettext/manual/html_node/index.html
[4]: https://pontoon.rust-lang.org/


mgeisler:
Please see the new [TRANSLATIONS](https://github.com/google/comprehensive-rust/pull/130/files#diff-9bfdf607e23fb326e78be45b8fa1cb833144528b780ca3d7a17eb5688ae81cb9) file I add in this PR for instructions on how to use the new infrastructure.

@keispace, I hope we can use this to incorporate your Korean translation. We should probably write a small script which finds the quote blocks you've added and extracts the text after them as the translation.

mgeisler:
Hi @keispace,

I've managed to import most of the strings from your translation :tada: Please see the `kr.po` file in this PR. When it's merged, I would need your help to update it with the remaining translations from your fork. Will you be able to assist with that?

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -0,0 +1,120 @@
...
+    let ctx = RenderContext::from_json(&mut io::stdin()).context("Parsing stdin")?;
+
+    fs::create_dir_all(&ctx.destination).with_context(|| {
+        format!(
+            "Could not create destination ({})",
+            ctx.destination.display()
+        )
+    })?;
+    let catalog = create_catalog(&ctx).context("Extracting messages")?;
+    let output_path = ctx.destination.join("messages.pot");
```
qwandor, line 111: Nit: could the filename be passed as a command-line parameter rather than hardcoded?

i18n-helpers/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+use regex::Regex;
+
+static PARAGRAPH_SEPARATOR: Lazy<Regex> = Lazy::new(|| Regex::new(r"\n\n+").unwrap());
+
+/// Extract paragraphs from text.
+///
+/// Paragraphs are separated by at least two newlines. Returns an
+/// iterator over line numbers (starting from 1) and paragraphs.
+pub fn extract_paragraphs(text: &str) -> impl Iterator<Item = (usize, &str)> {
+    // TODO: This could be make more sophisticated by parsing the
```
qwandor, line 25: s/make/made/

src/SUMMARY.md
```diff
@@ -26,19 +26,19 @@
   - [References](basic-syntax/references.md)
     - [Dangling References](basic-syntax/references-dangling.md)
   - [Slices](basic-syntax/slices.md)
-    - [`String` vs `str`](basic-syntax/string-slices.md)
+    - [String vs str](basic-syntax/string-slices.md)
```
qwandor, line 29: Why have you removed formatting from this file? It doesn't seem to be related to adding support for translations.

i18n-helpers/src/bin/mdbook-xgettext.rs
mgeisler: Almost, mdbook will by default invoke the renderer without arguments. I opted to read the path from `ctx.config` instead which means that we need to set it in `book.toml`.

src/SUMMARY.md
mgeisler: Good question! :smile: It turns out that mdbook is removing the formatting from these titles: [summary.rs](https://github.com/rust-lang/mdBook/blob/a8c37ceacee614171a59bf5ca969598b25089d44/src/book/summary.rs#L577).

The result of this stripping ends up as [`Chapter::name`](https://github.com/rust-lang/mdBook/blob/8cdb8d03672f3b5a4a7c6ecf3942fd74c029562b/src/book/book.rs#L157). We use this field when we extract the strings to translate: so instead of saving `"foo _bar_ baz"`, we would end up saving `"foo bar baz"` in `messages.pot`.

I search the `SUMMARY.md` file for the strings we translate so that I can add a proper filename and _line number_ to each translated string. When the string to translate has been altered compared to what is on disk, I cannot find it any longer.

I could potentially redo the parsing of mdbook and strip the formatting while tracking the line numbers. However, since the formatting doesn't show up in the book, it seemed fine to strip it from the source directly.

qwandor: Okay, fair enough. Maybe it would be a good idea to leave a comment in this file saying not to use formatting for chapter names (and explaining why).

keispace:
I got this branch and tested it locally. I think it would be more convenient if we could separate the po file and put it back together after translation, but it's really nice. 
I will re-arrange the contents of the uploaded 'kr.po' until it is merged into the main branch.

+ Is there any way to change the English and translation (Korean) on the web page? Should I just check it on a separate web page? When translating quite a few developer words, the meaning is ambiguous, so it is helpful to refer to English together.

src/SUMMARY.md
mgeisler: I didn't realize this yesterday, but the problem will actually be caught by the test I added: when you run `mdbook-xgettext`, it will error out if it cannot map the chapter names back to lines in `SUMMARY.md`. So we'll immediately know if someone accidentally tries to add such a title to our material.

I've updated the error message a little in that case. If/when we publish `mdbook-xgettext` and `mdbook-gettext`  to crates.io for others to use, we can consider making this less strict.

mgeisler:
Hi @keispace 

> I got this branch and tested it locally. I think it would be more convenient if we could separate the po file and put it back together after translation, but it's really nice.

If you have a copy of `kr.po` locally now, then I will remove it from this PR. I'm happy to add it in later when you've updated it.

> I will re-arrange the contents of the uploaded 'kr.po' until it is merged into the main branch.

Thanks a lot! I tried to describe the workflows in the [TRANSLATIONS](https://github.com/google/comprehensive-rust/pull/130/files#diff-9bfdf607e23fb326e78be45b8fa1cb833144528b780ca3d7a17eb5688ae81cb9) file. Basically, you'll want to run

```shell
```

to create `po/messages.pot`. This is just an intemediary file. You then run

```shell
msgmerge --update po/kr.po po/messages.pot
```

to incorporate all new strings into `po/kr.po`. You can now translate any untranslated entries. When I created the first `kr.po` file, I tried to mark entries with `fuzzy` if the script wasn't sure about them — this happened if the English block of lines had 3 paragraphs, but the Korean block had 4 paragraphs.

At the end of translation, there should be no fuzzy entries left. You can also clean up the obsolete entries if you want (they're at the bottom of your file).

> * Is there any way to change the English and translation (Korean) on the web page? Should I just check it on a separate web page?

Right now the English and Korean output is not linked so you cannot easily switch between them. My plan here is to add a language selector to the page — just a simple drop-down in the corner which sends you to the other book. The output of each language is independent, so all we need is a link between them.

> When translating quite a few developer words, the meaning is ambiguous, so it is helpful to refer to English together.

Yes, I ran into the same problem immediately when starting the Danish translation :smile: Translation is really hard and I guess it's harder than normal for technical texts. For my language, I'll probably include a few English words where I feel they make it easier to understand. Perhaps you can do the same for Korean? In my experience, people in IT often mix in English words when talking together in their native language. I think it's fine if the translation does the same. If people disagree, they can take it up with you afterwards.

keispace:
Thanks @mgeisler!
> If you have a copy of kr.po locally now, then I will remove it from this PR. I'm happy to add it in later when you've updated it.
I received a local copy. Thank you for providing the file. I will work on it so that it can be added later.

mgeisler:
> Thank you for providing the file. I will work on it so that it can be added later.

Thanks! Please let me know if you run into any problems here! There are lots of editors for PO files, such as: https://localise.biz/free/poeditor or https://poedit.net/.

Please note, I named the file `kr.po` by mistake... the actual language code for Korea is **`ko.po`**. Sorry about that!

mgeisler:
Just in case people would like to get started quickly with translating, here is the [messages.zip](https://github.com/google/comprehensive-rust/files/10392886/messages.zip) file from a recent checkout of the `main` branch.

You can download the file, decompress it, and then use any of the many online services to start translating :smile:

mgeisler:
Hi all, I've been looking at this on and off and it seems to work well — I can translate the content and the page titles (in the sidebar) successfully. After this is merged, there are some next steps for us

1. add a language picker (I'm working on this),
2. change the publish action to publish different languages into different sub-directories
3. setup redirects for the `en/` version (mdbook can output redirects)
4. change the JS code to avoid the hard-coded "Speaker Notes" header

Things we can also consider are

- setup translations via an online service, I've looked into https://github.com/mozilla/pontoon/ for this, but there are many others
- publish the `i18n-helpers` onto GitHub
- perhaps we can teach our publish action to only publish the languages that are actually touched (change `po/da.po` -> publish the Danish translation)
- possibly other things I've forgotten :smile:

TRANSLATIONS.md
```diff
@@ -0,0 +1,116 @@
...
+it will redirect the output to `book/xx`.
+
+## Serving a Translation
+
+Like normal, you can use `mdbook serve` to view your translation as you work on
+it. You use the same command as with `mdbook build` above, but additionally
+we'll tell `mdbook` to watch the `po/` directory for changes:
+
+```shell
+$ MDBOOK_BOOK__LANGUAGE=xx \
```
j-vanderstoep, line 112: Should you have a "\" on the next two lines as well?

i18n-helpers/src/bin/mdbook-gettext.rs
```diff
@@ -0,0 +1,230 @@
...
+    let mut target_lineno = 1;
+
+    for (lineno, paragraph) in extract_paragraphs(text) {
+        // Fill in blank lines between paragraphs. This is important
+        // for code blocks where blank lines are significant.
+        while target_lineno < lineno {
+            output.push('\n');
+            target_lineno += 1;
+        }
+        // Subtract 1 because the paragraph is missing a final '\n'
```
j-vanderstoep, line 49: Mostly a nit, but I wonder it would be cleaner to have extract_paragraphs not do this, then to handle it here.

j-vanderstoep (approved):

TRANSLATIONS.md
mgeisler: Thanks, we could do that for consistency and for nicer syntax highlighting — even though there are no special characters on these lines.

mgeisler: By the way, I intend to add some shell scripts to replace these weird incantations... I just wanted us to get a bit of experience with it first.

trdthg:
@mgeisler your i18n-helper works well, but
- would you like to keep the **cli entry** which output markdown? That's what I want.
- glad to see it published

> Btw, I noticed that mdbook-mermaid also provides a cli

mgeisler:
> * would you like to keep the **cli entry** which output markdown? That's what I need.

Are you talking about using the [`markdown` mdbook renderer](https://rust-lang.github.io/mdBook/format/configuration/renderers.html#markdown-renderer)? You can use it to get a full source tree with translated Markdown files.

> * I'm trying to combine [mdbook-i18n](https://github.com/funkill/mdbook-i18n/tree/master/example) with this, which seems to offer a lightweight solution for language switch on the web page

I'm not aiming to use that tool: I found that it doesn't keep all the custom options we set for the main language. So instead I'll simply run `mdbook` once per translation via a script. I'll have this up soon so you can see it in action.

trdthg:
>  I'll have this up soon
Looking forward to it, seems better 😀

mgeisler:
Hi @trdthg,

We've just merged #264 which publishes translations as part of the normal [publish flow](https://github.com/google/comprehensive-rust/blob/main/.github/workflows/publish.yml). We should probably simplify this a bit with a shell script.

I'm simply putting the different languages into subdirectories for now. They're not yet linked — I'll wait with that until the translations are more complete :smile:

trdthg:
> the normal [publish flow](https://github.com/google/comprehensive-rust/blob/main/.github/workflows/publish.yml).

I'll take a look.

And I found a problem here:
```rust
book.for_each_mut(|item| match item {
        BookItem::Chapter(ch) => {
            ch.content = translate(&ch.content, &catalog);
            ch.name = translate(&ch.name, &catalog);
        }
```
did you forget to translate **`ch.sub_items`** ? Because I found some md files in subfolders are not translated

mgeisler:
> And I found a problem here:
> 
> ```rust
> book.for_each_mut(|item| match item {
>         BookItem::Chapter(ch) => {
>             ch.content = translate(&ch.content, &catalog);
>             ch.name = translate(&ch.name, &catalog);
>         }
> ```
> 
> did you forget to handle **`ch.sub_items`** ? Because I found some md files in subfolders are lost

Oh, I hope not... :grimacing: When I look at [`for_each_mut`](https://github.com/rust-lang/mdBook/blob/8cdb8d03672f3b5a4a7c6ecf3942fd74c029562b/src/book/book.rs#L121-L133), I think it does the right thing by recursively going into the `sub_items`.

mgeisler:
@trdthg, can you open a new issue for anything weird you see on your end? That will make it easier to track any problems.

People are starting to use the translation infrastructure internally at Google, so I'm sure we'll soon find out if it doesn't do the right thing.

trdthg:
Sorry, my mistake. 🥲
I was trying to use `?`, so I turn it back to for loop, which made this problem.

Have fun coding.🤗

mgeisler:
Great that you found the problem!

tfpk:
Hi @mgeisler ! I wasn't sure where to post, but just wanted to let you know that [macrokata](github.com/tfpk/macrokata) is now using this translation infrastructure -- it's brilliant, thanks so much for setting it up.

It would be good to have this in a crate somewhere, if I can help on that score please let me know!

tfpk:
Great :)

Adding content on macros sounds very reasonable. Please feel free to take as much from macrokata as makes sense, and reach out if I can help. I've also got the https://tfpk.github.io/lifetimekata project if that's any use to you as well.

mgeisler:
@tfpk and others who might end up here: the translation infrastructure is now published as a separate crate: https://github.com/google/mdbook-i18n-helpers. I hope this makes it easier to reuse in other projects!

PR #131, opened 2023-01-09
Add speaker notes to copy-clone.md

fbornhofen:
I'd like to get some feedback on the level of detail of these notes.
My goal is to eventually not have to have an extra document with notes when teaching the class. Currently, I'm keeping notes on +/- everything I want to show.

### DIFF
--- a/src/ownership/copy-clone.md
+++ b/src/ownership/copy-clone.md
@@ -40,7 +40,7 @@ Copying and cloning are not the same thing:
 
 In the above example, try the following:
 
-* Try printing `p1` twice after the move. What do you expect to happen?
+* Try printing `p1` again after the move. What do you expect to happen?
 * What happens when you add a `String` field to `struct Point`?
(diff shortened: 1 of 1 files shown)


src/ownership/copy-clone.md
```diff
@@ -29,3 +29,20 @@ fn main() {
...
+
+Copying and cloning are not the same thing:
+
+* Copying refers to bitwise copies of memory regions and does not work on arbitrary objects.
+* Cloning is a more general operation and also allows for custom behavior by implementing the `Clone` trait.
+* Copying does not work on types that implement the `Drop` trait.
+
+In the above example, try the following:
+
+* Try printing `p1` twice after the move. What do you expect to happen?
```
qwandor, line 43: Why twice?

fbornhofen: Good catch, was changing things in the playground and didn't notice that the original example is already good enough. Removed that whole bullet point.

qwandor (approved):

PR #133, opened 2023-01-09
Adds speaker notes to Compound Types section

rastringer:
Speaker notes for 6. Basic Syntax, 6.2 Compound Types section. Briefly explains arrays and tuples properties.
Adds option for instructor to check for out of bounds errors using assert!().

### DIFF
--- a/src/basic-syntax/compound-types.md
+++ b/src/basic-syntax/compound-types.md
@@ -35,19 +35,7 @@ Arrays:
 
 *We can use literals to assign values to arrays.
 
-*Demonstrate out of bounds errors by setting n to different values 
-
-```
-a[n+15] = 11 // index out of bounds error since len is 10
-```
-
-*Efficient way to check n is in bounds:
-
-```
-assert!(n+20 < a.len()); // panics
-```
-
-*In the main function, the print statement asks for the debug implementation with the `?` format parameter: `{a}` gives default output, `{a:?}` gives debug output.
+*In the main function, the print statement asks for the debug implementation with the `?` format parameter: `{a}` gives the default output, `{a:?}` gives the debug output.
 
 *Adding `#`, eg `{a:#?}`, invokes a "pretty printing" format, which can be easier to read.
 
--- a/src/basic-syntax/compound-types.md
+++ b/src/basic-syntax/compound-types.md
@@ -31,7 +31,7 @@ Key points:
     
 Arrays:
     
-*Arrays have elements of the same type, T, and length, N, which is fixed. 
+*Arrays have elements of the same type, `T`, and length, `N`, which is a compile-time constant. 
 
 *We can use literals to assign values to arrays.
 
--- a/src/basic-syntax/compound-types.md
+++ b/src/basic-syntax/compound-types.md
@@ -36,14 +36,20 @@ Arrays:
 *We can use literals to assign values to arrays.
 
 *Demonstrate out of bounds errors by setting n to different values 
-    e.g. a[n+15] = 11 // index out of bounds error since len is 10
+
+```
+a[n+15] = 11 // index out of bounds error since len is 10
+```
 
 *Efficient way to check n is in bounds:
-    assert!(n+20 < a.len()); // panics
+
+```
+assert!(n+20 < a.len()); // panics
+```
 
 *In the main function, the print statement contains the debug implementation {a :?}.
 
-*Adding #, eg {a:#?}, invokes a "pretty printing" format, which can be easier to read.
+*Adding `#`, eg `{a:#?}`, invokes a "pretty printing" format, which can be easier to read.
 
 Tuples:
 
--- a/src/basic-syntax/compound-types.md
+++ b/src/basic-syntax/compound-types.md
@@ -47,7 +47,7 @@ a[n+15] = 11 // index out of bounds error since len is 10
 assert!(n+20 < a.len()); // panics
 ```
 
-*In the main function, the print statement contains the debug implementation {a :?}.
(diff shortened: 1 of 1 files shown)


mgeisler (changes requested):
Hi @rastringer, thanks for adding this!

Can I get you to add backticks around the code parts of the text? That way it'll show up as code when the Markdown is processed.

src/basic-syntax/compound-types.md
```diff
@@ -24,3 +24,33 @@ fn main() {
...
+    
+*Arrays have elements of the same type, T, and length, N, which is fixed. 
+
+*We can use literals to assign values to arrays.
+
+*Demonstrate out of bounds errors by setting n to different values 
+    e.g. a[n+15] = 11 // index out of bounds error since len is 10
+
+*Efficient way to check n is in bounds:
+    assert!(n+20 < a.len()); // panics
```
mgeisler, line 42: The newline here and elsewhere is not significant in Markdown.

rastringer: Ack of course. Apologies! Updated PR

src/basic-syntax/compound-types.md
```diff
@@ -24,3 +24,39 @@ fn main() {
...
+Arrays:
+    
+*Arrays have elements of the same type, T, and length, N, which is fixed. 
+
+*We can use literals to assign values to arrays.
+
+*Demonstrate out of bounds errors by setting n to different values 
+
+```
+a[n+15] = 11 // index out of bounds error since len is 10
```
mgeisler, line 41: Hi @rastringer, what is `n` in this line of code?

src/basic-syntax/compound-types.md
```diff
@@ -24,3 +24,39 @@ fn main() {
...
+```
+
+*Efficient way to check n is in bounds:
```
mgeisler, line 44: The `n` should be in backticks here — but a bigger question is why you say this is _efficient_? That is, to check if `n` is within bounds, you do `n < a.len()`.

Oh, are you perhaps thinking of the case where you have code like this:

```rust
fn foo(n: usize, a: &[i32]) -> i32 {
    let mut sum = 0;
    for i in 0..20 {
        sum += a[n + i];
    }
    sum
}
```

Here, the Rust compiler will emit 20 bounds checks — it emits one for each of `n + 0`, `n + 1`, ..., `n + 20`. It does this because it wants to panic with an accurate message if, say, `n + 7` is out of bounds.

In this case, you can add

```rust
assert!(n + 20 < a.len());
```

to the top of the function and you will see that subsequent bounds checks go away. It's a nice trick to know and I would be happy to explain it in the speaker notes.

Hmm... actually, I just tried reproducing this and I don't see the promised change in the assembly output. I made a post about it on the [Internals Forum](https://internals.rust-lang.org/t/indexing-into-a-slice-with-a-known-length/18126.), so hopefully someone there will be able to help clear this up.

src/basic-syntax/compound-types.md
```diff
@@ -24,3 +24,39 @@ fn main() {
...
+
+```
+assert!(n+20 < a.len()); // panics
+```
+
+*In the main function, the print statement contains the debug implementation {a :?}.
```
mgeisler, line 50: Typo: "{a :?}" should be `{a:?}` without a space and with backticks.

mgeisler (changes requested):

src/basic-syntax/compound-types.md
mgeisler:
```suggestion
*In the main function, the print statement asks for the debug implementation with the `?` format parameter: `{a}` gives default output, `{a:?}` gives debug output.
```

src/basic-syntax/compound-types.md
rastringer: Hi Martin, 
I warped the video and slides in my head and missed that the 'n' isn't in the latter. Apologies and good catch. In the video, you show the out of bounds by doing:

let n = 7;
a[n+20] = 30; 
println!("a: {:?}", a);

Perhaps it's best to keep the notes simple for now and remove this part? I could update further if there's some clarity from the Internals Forum later. Nice work checking the Assembly!

mgeisler: > Perhaps it's best to keep the notes simple for now and remove this part?

Yes, I'm a big fan of simplicity, so let's take it out :smile:  That should also solve the compilation error: all code blocks are checked by `mdbook test`, so you either have to make it compile (or add the `compile_fail` attribute if it's broken on purpose).

src/basic-syntax/compound-types.md
```diff
@@ -24,3 +24,27 @@ fn main() {
...
 }
 ```
+
+<details>
+    
+Key points:
+    
+Arrays:
+    
+*Arrays have elements of the same type, T, and length, N, which is fixed.
```
mgeisler, line 34:
```suggestion
*Arrays have elements of the same type, `T`, and length, `N`, which is a compile-time constant. 
```

mgeisler (approved):

src/basic-syntax/compound-types.md
mgeisler: @SkiFire13 [recommended on the Internal Forum](https://internals.rust-lang.org/t/indexing-into-a-slice-with-a-known-length/18126/2?u=mgeisler) using a sub-slice to get great optimization. We could add a note about this "trick" later, it's be very interesting for students who wonder how the performance compares between Rust and C/C++.

src/basic-syntax/compound-types.md
rastringer: Excellent, thanks Martin! That's good to know about the code blocks.

PR #134, opened 2023-01-09
Speaker notes for lifetimes-function-calls.md

fbornhofen:
The existing comment in the code "Put into different scope" might be unclear to somebody going through the course by themselves. Added more detailed instructions in the speaker notes plus an additional thing to try out.

### DIFF
--- a/src/ownership/lifetimes-function-calls.md
+++ b/src/ownership/lifetimes-function-calls.md
@@ -29,7 +29,7 @@ fn main() {
 In the above example, try the following:
 
 * Move the declaration of `p2` and `p3` into a a new scope (`{ ... }`), resulting in the following code:
-  ```rust,editable
+  ```rust,ignore
   #[derive(Debug)]
   struct Point(i32, i32);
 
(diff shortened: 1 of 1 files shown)


mgeisler:
> The existing comment in the code "Put into different scope" might be unclear

Yes, very unclear indeed! It was a left-behind clue to myself so that I can could remember what to do — a prime example of what a speaker note can do much better :smile: Thanks for cleaning this up.

src/ownership/lifetimes-function-calls.md
```diff
@@ -23,3 +23,12 @@ fn main() {
 * Read `&'a Point` as "a borrowed `Point` which is valid for at least the
   lifetime `a`".
   * The _at least_ part is important when parameters are in different scopes.
+
+<details>
+
+In the above example, try the following:
+
+* Move the declaration of `p2` and `p3` into a a new scope (`{ ... }`). You have to declare `p3` as `let p3: &Point;`, in `main`'s scope before that new block, and in the block assign to `p3` without the `let` (just `p3 = left_most(...)`). Note how this does not compile since `p3` outlives `p2`.
```
mgeisler, line 31: Just a note: you could write full code blocks here — it's normal Markdown syntax. I think it would make the explanation both shorter and clearer.

fbornhofen: Makes a lot of sense, thanks for the suggestion.

PR #135, opened 2023-01-09
Add bullet point about mutable references

fbornhofen:
To me, this is an important reference coming from C++


mgeisler:
Thanks @fbornhofen!

PR #137, opened 2023-01-09
Make library mutable in book-library exercise

sahennenkamp:
Otherwise the library cannot have changes (books) added to it when testing via main.

Compile error:
```
error[E0596]: cannot borrow `library` as mutable, as it is not declared as mutable
```

I could definitely be missing something, but I also tried compiling with the given solution (with main uncommented) and received the same error.

### DIFF
--- a/src/exercises/day-1/book-library.rs
+++ b/src/exercises/day-1/book-library.rs
@@ -100,7 +100,7 @@ impl Library {
 // implement the missing methods. You will need to update the
 // method signatures, including the "self" parameter!
 fn main() {
-    let library = Library::new();
+    let mut library = Library::new();
 
     //println!("Our library is empty: {}", library.is_empty());
     //


src/exercises/day-1/book-library.rs
```diff
@@ -100,7 +100,7 @@ impl Library {
 // implement the missing methods. You will need to update the
 // method signatures, including the "self" parameter!
 fn main() {
-    let library = Library::new();
+    let mut library = Library::new();
```
mgeisler, line 103: Hi @sahennenkamp, you're completely correct that you need to add `mut` here.

However... :smile: That is something the students should discover for themselves when solving the exercise.

Can you instead update the comments above to let people know that they (might) have to make things mutable?

mgeisler (changes requested):
It's a good idea to add a note about this — either as a comment or perhaps in a `<details> ... </details>` block (a speaker note).

mgeisler:
Let's open a new PR with the changes requested above (if necessary).

src/exercises/day-1/book-library.rs
sahennenkamp: Hey @mgeisler! Thanks for the input, that makes lots of sense! I'll open a new one that edits the comments. I was thinking about also editing the solution to include the `mut` if that works with you?

sahennenkamp: oh nvm, I see that this is the soln: https://github.com/google/comprehensive-rust/blob/c212a473ba4d060b209b654c0dd403328d6e137f/src/exercises/day-1/solutions-afternoon.md?plain=1#L8
I'll just update the exercise comments

PR #138, opened 2023-01-10
Added some more detail on rustup.

brandonpollack23:
Rusutp is the reccomended way to install rust and maintained by the core team.

I added some organization and suggested that that be the method of installation.

### DIFF
--- a/src/cargo.md
+++ b/src/cargo.md
@@ -34,7 +34,7 @@ $ sudo apt install cargo rust-src
 This will allow [rust-analyzer][1] to jump to the definitions. We suggest using
 [VS Code][2] to edit the code (but any LSP compatible editor works).
 
-Some folks also like to use the [Jetbrains][4] family of IDEs, which do their own analysis but have their own tradeoffs. If you prefer them, you can install the [Rust Plugin][5]. Please take note that as of writing debugging only works on the CLion version of the Jetbrains IDEA suite.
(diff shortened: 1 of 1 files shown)


src/cargo.md
```diff
@@ -5,6 +5,26 @@ used in the Rust ecosystem to build and run Rust applications. Here we want to
...
+## Installation
+
+### Rustup (Recommended)
+
+You can follow the instructions to install cargo and rust compiler, among other standard ecosystem tools with the [rustup][3] tool, which is maintained by the Rust Foundation.
+
+On Unix-like systems you can curl down the install script and pipe it into your shell like so:
+
+```shell
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
```
mgeisler, line 17: I would like us to not add this line of magic... I prefer it if there is one place on the Internet with the recommended command. Put differently: we should absolutely mention and recommend `rustup.rs`, but we should let them explain how to install in details.

mgeisler (changes requested):
Thanks for adding this! It's something which was missing — I only provided the Debian command line since I aimed the course at Googlers with gLinux. So thanks for broadening this to the rest of the world.

src/cargo.md
brandonpollack23: You are absolutely right, this is subject to change, but a link will always be right

brandonpollack23:
Should be good to go

mgeisler (approved):

PR #139, opened 2023-01-10
Add more info and footnotes to rust-ecosystem.md

brandonpollack23:


src/cargo/rust-ecosystem.md
```diff
@@ -44,3 +44,7 @@ Key points:
 [editions]: https://doc.rust-lang.org/edition-guide/
 
 </details>
+
+[^rustc]: Please note, it is quite rare to ever use the compiler directly not through `cargo`.
```
mgeisler, line 48: Please avoid footnotes and instead use the speaker notes as much as possible — it's not immediately clear when you view it on a laptop/desktop, but the material is meant to be shown in a classroom setting. So think of the pages as "slides", not pages :smile: 

The speaker notes can be collapsed or even moved to a popout window. I therefore consider them "free" in terms of space.

src/cargo/rust-ecosystem.md
```diff
@@ -44,3 +44,7 @@ Key points:
...
+
+[^cargo]: Cargo itself is a powerful and comprehensive tool capable of many advanced features including but not limited to: Project structure, [workspaces](https://doc.rust*lang.org/cargo/reference/workspaces.html), dev dependencies, [build scripting](https://doc.rust-lang.org/cargo/reference/build-scripts.html), [binary installation](https://doc.rust-lang.org/cargo/commands/cargo-install.html), etc. It is extensible with sub command plugins as well (such as [cargo clippy](https://github.com/rust-lang/rust-clippy) or [cargo binstall](https://github.com/cargo-bins/cargo-binstall)). Read more from the [official Cargo Book](https://doc.rust-lang.org/cargo/)
```
mgeisler, line 50: I'm trying to move towards a style where I use reference links in Markdown. So something like the below. The advantage: less inline Markdown code for translators to deal with:

```suggestion
[^cargo]: Cargo itself is a powerful and comprehensive tool capable of many advanced features including but not limited to: Project structure, [workspaces], dev dependencies, [build scripting], [binary installation], etc. It is extensible with sub command plugins as well (such as [`cargo clippy`]). Read more from the [official Cargo Book].

[workspaces]: https://doc.rust*lang.org/cargo/reference/workspaces.html
[build scripting]: https://doc.rust-lang.org/cargo/reference/build-scripts.html
[binary installation]: https://doc.rust-lang.org/cargo/commands/cargo-install.html
[`cargo clippy`]: https://github.com/rust-lang/rust-clippy
[official Cargo Book]: https://doc.rust-lang.org/cargo/
```

(I dropped `cargo-binstall` since I haven't heard of it and I don't think we should include more than just the basics here)

mgeisler (changes requested):
Hi @brandonpollack23, thanks a lot for the PR. I would love it if you can refactor it to not use footnotes.

src/cargo/rust-ecosystem.md
brandonpollack23: noted and done

brandonpollack23:
should be good now lmk

PR #140, opened 2023-01-10
Added some speaker notes to variables.md

brandonpollack23:


mgeisler (approved):

mgeisler:
Thanks @brandonpollack23, this is great! I made the language even more direct for the first note: the instructor should definitely talk about type inference at this point.

PR #141, opened 2023-01-10
Added notes, runtime constant eval, and footnoes to static-and-const.md

brandonpollack23:


src/basic-syntax/static-and-const.md
```diff
@@ -36,4 +38,23 @@ fn main() {
...
+
+Sometimes you want a global constant that isn't constructable at compile time, such as a global HashMap or complex structure.  This is done with crates like [OnceCell](https://crates.io/crates/once_cell) or [lazy_static](https://crates.io/crates/lazy_static).
+
+[1]: https://rust-lang.github.io/rfcs/0246-const-vs-static.html
+
+[^lifetime]: We'll discuss lifetimes later today in the section on ownership
+
+<details>
+
+* If the target audience is well versed in C++, it might be helpful to mention that `const` behaves semantically similar to C++'s `constexpr`
```
mgeisler, line 56:
```suggestion
* Mention that `const` behaves semantically similar to C++'s `constexpr`.
```

src/basic-syntax/static-and-const.md
```diff
@@ -36,4 +38,23 @@ fn main() {
...
+* `static`, on the other hand, is much more similar to a `const` or mutable global variable.
```
mgeisler, line 57:
```suggestion
* `static`, on the other hand, is much more similar to a `const` or mutable global variable in C++.
```

src/basic-syntax/static-and-const.md
```diff
@@ -36,4 +38,23 @@ fn main() {
 }
 ```
 
+As noted in the [Rust RFC Book][1], these are not inlined upon use and have an actual associated memory location.  This is useful for unsafe and embedded code, and have a `'static` lifetime[^lifetime].
+
+
 We will look at mutating static data in the [chapter on Unsafe Rust](../unsafe.md).
+
+## Runtime evaluated constants
```
mgeisler, line 46: I think this will make the page too large. We can move it to a separate page or just let the instructor mention it (`->` speaker notes).

mgeisler (changes requested):
It's great to mention these things, but probably in the speaker notes or perhaps even later in the course.

This page is for Day 1 and it's basically keeping us from talking about the highlight of the day: ownership and the borrow checker. With that in mind, we should avoid adding anything which isn't strictly necessary.

src/basic-syntax/static-and-const.md
brandonpollack23: Based on your previous note, totally agreed, I'll actually move it to the thread you opened about helpful libraryes

brandonpollack23:
LMK if it is LGTM

src/basic-syntax/static-and-const.md
```diff
@@ -36,4 +38,15 @@ fn main() {
 }
 ```
 
+As noted in the [Rust RFC Book][1], these are not inlined upon use and have an actual associated memory location.  This is useful for unsafe and embedded code, and have a `'static` lifetime[^lifetime].
```
mgeisler, line 41: Could you move this sentence to the speaker notes as well?

I think the `[^lifetime]` footnote looks like a dangling pointer now? :smile:

src/basic-syntax/static-and-const.md
mgeisler: Thanks a lot!

src/basic-syntax/static-and-const.md
mgeisler: Let me clean that up real quick.

mgeisler (approved):

PR #143, opened 2023-01-10
Rephrase copy/clone speaker notes

fbornhofen:
- Actual instructions instead of questions for consistency with other notes
- Bullet point about copy constructors


mgeisler:
Thanks for making things consistent!

PR #144, opened 2023-01-10
Add speaker notes about RVO in borrowing.md

fbornhofen:


mgeisler:
This is super nice, thank you!

PR #147, opened 2023-01-10
Expand enum sizes example and add speaker notes

fbornhofen:


src/enums/sizes.md
```diff
@@ -19,7 +19,19 @@ enum Foo {
 
 fn main() {
     dbg_size!(Foo);
+    dbg_size!(bool);
```
mgeisler, line 22: I was thinking that the instructor would play with the code during class and put different types here — but we can also put it in directly like you do.

mgeisler:
Thanks @fbornhofen!

PR #149, opened 2023-01-10
Add speaker notes for string.md

fbornhofen:


src/std/string.md
```diff
@@ -12,6 +12,10 @@ fn main() {
     s2.push_str(&s1);
     s2.push('!');
     println!("s2: len = {}, capacity = {}", s2.len(), s2.capacity());
+
+    let s3 = String::from("🇨🇭");
+    println!("s3: len = {}, number of chars = {}", s3.len(),
+             s3.chars().collect::<std::vec::Vec<_>>().len());
```
mgeisler, line 18: This should be the same:

```suggestion
             s3.chars().count());
```

PR #151, opened 2023-01-11
Add speaker notes for structs

cchiw:
Base notes for others to build on

### DIFF
--- a/src/structs.md
+++ b/src/structs.md
@@ -17,3 +17,17 @@ fn main() {
     println!("{} is {} years old", peter.name, peter.age);
 }
 ```
+
+<details>
+
+* The top of the program is how we define a struct, the main function creates a struct, and then at the bottom we access the fields.
+* Structs work similar as they do in other languages, but note there is also no typedef keyword or inheritance between structs. Also note there are no functions. The implementation is separate.
+* We can access the structs fields with the dot notation 
+* Using short-hand syntax, we can initiate the fields with similar-named variables.  [define age variable on it’s own line] 
+* There are multiple types of syntax for structs in Rust
+    * Unit_ like struct, can Define with “struct (name);” and no need for curly brackets and parentheses.
+    * Use: when you need to implement a trait on some type but don’t have any data that you want to store in the type itself. 
+    * next lets see a Tuple struct
+
+
+</details>
--- a/src/structs/field-shorthand.md
+++ b/src/structs/field-shorthand.md
@@ -21,3 +21,14 @@ fn main() {
     println!("{peter:?}");
 }
 ```
+
+<details>
+
+* *“Impl”: This is where we define the new() function and add other boilerplate functions. 
+* Field Shorthand Syntax: Previously we used a field name with a colon, here [demonstrating the fss] when variable name is the same as the field name
+* Function update syntax: 
+  * It’s often useful to create a new instance of a struct that includes most of the values from another instance, but changes some. 
+  * [let dad =  Person {age:57, .. peter};]
+* it’s useful to have output that’s a bit easier to read  [use {:#?}  and re-run]
+
+</details>
--- a/src/structs/tuple-structs.md
+++ b/src/structs/tuple-structs.md
@@ -31,3 +31,21 @@ fn main() {
 }
 
 ```
+
+<details>
+
+* This tuple struct is called a Point, which contains a pair of integers. 
+  * We then have a shorthand for creating it. Note, no curly braces. 
+  * The two fields are called with a dot then 0 and 1.
+* Use case: Create a new type, by wrapping around a single field. 
+  * Here is an example of new type PoundsOfForce and Newtons, which are now different from the f64s type or eachother. 
+  * We can talk about things in terms of these new types. For instance, in what functions should return/accept.
+* Example 
+  * Say you’re a space agency and you want to send a probe up to Mars. As you have your rocket you might want to have a function to compute/set thruster force and you are using f64s but want to consider them different.
+  * [Run example gets an error. Add x= Newtons(2.0); set..(x);]
+* Rust doesn’t do inexplicit things, for instance no using booleans used as integers
+  * Similarly, using the data in a struct requires accessing field appropriately 
+  * [Add float to force force.0+10  in the set function.] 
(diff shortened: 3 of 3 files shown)


src/structs.md
```diff
@@ -17,3 +17,17 @@ fn main() {
     println!("{} is {} years old", peter.name, peter.age);
 }
 ```
+
+<details>
+
+* The top of the program is how we define a struct, the main function creates a struct, and then at the bottom we access the fields.
+* Structs work similar as they do in other languages, but note there is also no typedef keyword or inheritance between structs. Also note there are no functions. The implementation is separate.
+* We can access the structs fields with the dot notation 
+* Using short-hand syntax, we can initiate the fields with similar-named variables.  [define age variable on it’s own line]
```
fbornhofen, line 26: nit: "its own line"

src/structs/tuple-structs.md
```diff
@@ -31,3 +31,21 @@ fn main() {
...
+* Use case: Create a new type, by wrapping around a single field. 
+  * Here is an example of new type PoundsOfForce and Newtons, which are now different from the f64s type or eachother. 
+  * We can talk about things in terms of these new types. For instance, in what functions should return/accept.
+* Example 
+  * Say you’re a space agency and you want to send a probe up to Mars. As you have your rocket you might want to have a function to compute/set thruster force and you are using f64s but want to consider them different.
+  * [Run example gets an error. Add x= Newtons(2.0); set..(x);]
+* Rust doesn’t do inexplicit things, for instance no using booleans used as integers
+  * Similarly, using the data in a struct requires accessing field appropriately 
+  * [Add float to force force.0+10  in the set function.] 
+  * In some cases, like with the + operation, we can overload the operator, we could also use de-refef. We’ll talk about it later
```
qwandor, line 49: What is "de-refef"?

src/structs/field-shorthand.md
```diff
@@ -21,3 +21,14 @@ fn main() {
...
 ```
+
+<details>
+
+* *“Impl”: This is where we define the new() function and add other boilerplate functions. 
+* Field Shorthand Syntax: Previously we used a field name with a colon, here [demonstrating the fss] when variable name is the same as the field name
+* Function update syntax: 
+  * It’s often useful to create a new instance of a struct that includes most of the values from another instance, but changes some. 
+  * [let dad =  Person {age:57, .. peter};]
+* it’s useful to have output that’s a bit easier to read  [use {:#?}  and re-run]
```
qwandor, line 32: Nit: start with a capital letter.

src/structs.md
```diff
@@ -17,3 +17,17 @@ fn main() {
...
+<details>
+
+* The top of the program is how we define a struct, the main function creates a struct, and then at the bottom we access the fields.
+* Structs work similar as they do in other languages, but note there is also no typedef keyword or inheritance between structs. Also note there are no functions. The implementation is separate.
+* We can access the structs fields with the dot notation 
+* Using short-hand syntax, we can initiate the fields with similar-named variables.  [define age variable on it’s own line] 
+* There are multiple types of syntax for structs in Rust
+    * Unit_ like struct, can Define with “struct (name);” and no need for curly brackets and parentheses.
+    * Use: when you need to implement a trait on some type but don’t have any data that you want to store in the type itself. 
+    * next lets see a Tuple struct
```
qwandor, line 30: let's

qwandor (approved):

src/structs.md
```diff
@@ -17,3 +17,17 @@ fn main() {
     println!("{} is {} years old", peter.name, peter.age);
 }
 ```
+
+<details>
+
+* The top of the program is how we define a struct, the main function creates a struct, and then at the bottom we access the fields.
```
mgeisler, line 23: I think we can state this in a shorter way: the notes should not be a "script" for the instructor. The instructor is expected to already know Rust well, so we should instead tell the instructor what to show on each page —we should not teach them Rust.

src/structs.md
mgeisler: Right, and we don't have a convention of putting things into `[ ... ]` brackets.

src/structs.md
```diff
@@ -17,3 +17,17 @@ fn main() {
...
+* Structs work similar as they do in other languages, but note there is also no typedef keyword or inheritance between structs. Also note there are no functions. The implementation is separate.
+* We can access the structs fields with the dot notation 
+* Using short-hand syntax, we can initiate the fields with similar-named variables.  [define age variable on it’s own line] 
+* There are multiple types of syntax for structs in Rust
+    * Unit_ like struct, can Define with “struct (name);” and no need for curly brackets and parentheses.
```
mgeisler, line 28: Please mark all code with backticks. Your curly quotes are nice from a typographical standpoint, but it's the wrong markup here :smile:

src/structs.md
```diff
@@ -17,3 +17,17 @@ fn main() {
...
+    * Use: when you need to implement a trait on some type but don’t have any data that you want to store in the type itself.
```
mgeisler, line 29:
```suggestion
    * Use: when you need to implement a trait on some type but don’t have any data that you want to store in the value itself. 
```

src/structs.md
mgeisler, line 30:
```suggestion
    * next lets see a tuple struct
```

src/structs.md
mgeisler: There are some more typos here: `Unit_ like` should probably be `Unit-like` and the capital `D` in `Define` is out of place.

mgeisler (changes requested):
Please try to tighten the language and make the notes look like the other notes we've added recently.

src/structs.md
```diff
@@ -17,3 +17,17 @@ fn main() {
     println!("{} is {} years old", peter.name, peter.age);
 }
 ```
+
+<details>
+
+* The top of the program is how we define a struct, the main function creates a struct, and then at the bottom we access the fields.
+* Structs work similar as they do in other languages, but note there is also no typedef keyword or inheritance between structs. Also note there are no functions. The implementation is separate.
```
mgeisler, line 24: Sorry, but what does "no functions" mean here?

cchiw:
Thanks for all your feedback! I'm going to open separate PRs for each slide as I think is the way.

PR #159, opened 2023-01-12
Speaker notes: cosmetic fixes for compund-types.md, borrowing.md

fbornhofen:


mgeisler:
Thanks!

I guess it won't have caught this directly, but I plan on introducing an auto-formatter at some point — but in 1-2 weeks when people have submitted most of their PRs.

I've also thought of setting up per-PR deployments: that would have helped here since we could see the rendered output before merging.

PR #161, opened 2023-01-12
Simplify state transitions for speaker notes

mgeisler:
Before, we attempted to change state from “popup” to “inline-open” when the speaker note window was closed.

We did this by listening to “pagehide” and change the state there. The event fires every time you navigate away from the page, so we had a complex setup where we would reset the state to “popup” when the next page was loaded into the speaker note window.

The problem with this is that it’s racy: we could end up in a situation where we set the state to “inline-open” right after the speaker note window was updated. When that happened, we would mark the window as “defunct”, meaning that it was supposed to be closed.

With this change, we no longer try to change the state from the speaker note window. If the window is lost (closed), the user will have to click the “Close speaker notes” button in the top-right to reset the state. This should be much more reliable.

Long-term, a better solution would be to let the speaker notes fetch the current URL using JavaScript instead of doing it via an actual page navigation. That should allow us to react to “pagehide” events again (since they won’t fire on every page transition).

### DIFF
--- a/speaker-notes.css
+++ b/speaker-notes.css
@@ -19,12 +19,13 @@
   width: 40px;
 }
 
-.content details summary a {
-  margin-left: 0.5em;
+.content details summary .pop-out {
+  color: var(--icons);
+  padding: 0 8px;
+  cursor: pointer;
+  transition: color 0.5s;
 }
 
-@media only screen and (max-width: 1080px) {
-  .content details summary a {
-    display: none;
-  }
+.content details summary .pop-out i:hover {
+  color: var(--icons-hover);
 }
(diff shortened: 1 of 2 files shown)


mgeisler:
Hi @fbornhofen, please give this a test! It should fix the brokenness you saw this morning.

fbornhofen:
Thanks for the quick fix!

The one thing that I found is quirky is that if you close the speaker notes window via the button in the main window, you need to reload the speaker ntoes window in order for them to show up. But I'm not planning on doing that. This is now strictly better than before.

mgeisler:
> The one thing that I found is quirky is that if you close the speaker notes window via the button in the main window, you need to reload the speaker ntoes window in order for them to show up.

The whole thing is a little more quirky than I wish... :grimacing: 

One pain point is that we cannot close a popup window if you've navigated to a different page — popups are full of restrictions in JavaScript because of how they were abused over the years. I originally thought I could simply close the window when clicking the button, but that's restricted.

PR #163, opened 2023-01-12
Speaker notes to 6.4 Slices

rastringer:
Brief digest of the slice type, including mentions of borrowing, range syntax and lifetimes.

### DIFF
--- a/src/basic-syntax/slices.md
+++ b/src/basic-syntax/slices.md
@@ -23,7 +23,7 @@ fn main() {
     
 * The same is true for the last index, so `&a[2..a.len()]` and `&a[2..]` are equal.
 
-* We set `s` as a reference of `i32`s. Notice that the type of `s` no longer has an array length. This avoids type errors when performing computations on slices of different sizes.
+* `s` is a reference to a slice of `i32`s. Notice that the type of `s` no longer mentions the array length. This allows us to performing computations on slices of different sizes.
  
 * Slices always borrow from another object. In this example, `a` has to remain 'alive' so we can take a slice from it. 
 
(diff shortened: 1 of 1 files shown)


src/basic-syntax/slices.md
```diff
@@ -14,3 +14,17 @@ fn main() {
...
 * Slices borrow data from the sliced type.
 * Question: What happens if you modify `a[3]`?
+
+<details>
+
+* We create a slice by borrowing `a` and specifying the starting and ending indexes in brackets.
+
+* If the slice starts at index 0, Rust’s range syntax means we can drop the starting index. 
+    
+* The same is true for the last index, so `let s: &[i32] = &a[2..len];` or `let s: &[i32] = &a[2..];` are equal.
```
mgeisler, line 24:
```suggestion
* The same is true for the last index, so `&a[2..a.len()]` and `&a[2..]` are equal.
```

src/basic-syntax/slices.md
```diff
@@ -14,3 +14,17 @@ fn main() {
...
+
+* We set `s` as a reference of `i32`s. Notice that the type of `s` no longer has an array length. This avoids type errors when performing computations on slices of different sizes.
```
mgeisler, line 26:
```suggestion
* `s` is a reference to a slice of `i32`s. Notice that the type of `s` no longer mentions the array length. This allows us to performing computations on slices of different sizes.
```

rastringer: Thanks Martin, we just need 'perform' vs 'performing'...is it easier if you adjust the change?

PR #164, opened 2023-01-12
Add speaker notes for moves-function-calls.md

fbornhofen:


fbornhofen:
Please lmk what you think. The part about cloning is explained in the following slide, and I'm planning to use this example as a segue into an explanation of the details.

src/ownership/lifetimes-data-structures.md
```diff
@@ -20,3 +20,11 @@ fn main() {
 }
 ```
 
+<details>
+
+* In the above example, the annotation on `Highlight` enforces that the data underlying the contained `&str` lives at least as long as any instance of `Highlight` that uses that data.
+* If `text` is consumed before the end of the lifetime of `fox` (or `dog`), the borrow checker throws an error.
+* Data structures with borrowed data force users to hold on to the original data structure. This can be useful for creating lightweight views, but generally make them somewhat harder to use.
```
mgeisler, line 27:
```suggestion
* Types with borrowed data force users to hold on to the original data. This can be useful for creating lightweight views, but it generally makes them somewhat harder to use.
```

src/ownership/lifetimes-data-structures.md
```diff
@@ -20,3 +20,11 @@ fn main() {
...
+* When appropriate, consider creating data structures that own data directly.
```
mgeisler, line 28: I think we can make this recommendation even stronger, echoing advice given to me by @Darksonn a while back:

```suggestion
* When possible, make data structures own their data directly.
```

mgeisler:
> The part about cloning is explained in the following slide, and I'm planning to use this example as a segue into an explanation of the details

Thanks, I think that's a great idea!

PR #167, opened 2023-01-13
Hint towards making library mut

sahennenkamp:
Follow up to https://github.com/google/comprehensive-rust/pull/137 Gives the student a hint that it's okay to edit variables within main.


mgeisler:
Thanks @sahennenkamp for the update, it looks good!

PR #169, opened 2023-01-16
Add some speaker notes for day 3 morning and a page about FromIterator

qwandor:


src/traits/from-iterator.md
```diff
@@ -0,0 +1,23 @@
+# FromIterator
```
mgeisler, line 1: You should also mention this page in `SUMMARY.md` to actually make it appear somewhere :smile:

src/traits/from-iterator.md
```diff
@@ -0,0 +1,23 @@
...
+
+`FromIterator` lets you build a collection from an `Iterator`.
```
mgeisler, line 3: I'm happy to cover this, but we should take into account that we now cover iterators in all their forms in:

* Day 1, [Iterators and Ownership](https://google.github.io/comprehensive-rust/exercises/day-1/iterators-and-ownership.html), which also explains that `IntoIterator` is what makes `for` loops work.
* Day 1, [Type Inference](https://google.github.io/comprehensive-rust/basic-syntax/type-inference.html#speaker-notes) which mentions `FromIterator` and `Iterator::collect`.

This might not be a problem since I keep saying that repetition is important for learning :smile:

mgeisler (changes requested):
Thanks for filling this out!

src/traits/from-iterator.md
qwandor: It looks like you only mention `FromIterator` briefly in the speaker notes in one of those places, without going into much detail of how it works, so I think it's worth talking a bit more about it here in a slide as a common trait that you are likely to use and might want to implement.

mgeisler: Yeah, it's only mentioned briefly in the notes (by @fbornhofen, I believe), but it's something we talk about on Day 1 as a natural part of the exercises. I think it's fine to give is a proper treatment like you do here, I just wanted to point out the redundancy.

mgeisler (approved):

PR #170, opened 2023-01-16
Add more information about error handling

qwandor:
Added some speaker notes, and a slide about `Box<dyn Error>` before we introduce `anyhow`.

### DIFF
--- a/src/error-handling/converting-error-types.md
+++ b/src/error-handling/converting-error-types.md
@@ -57,4 +57,9 @@ Key points:
 * The `username` variable can be either `Ok(string)` or `Err(error)`.
 * Use the `fs::write` call to test out the different scenarios: no file, empty file, file with username.
 
+It is good practice for all error types to implement `std::error::Error`, which requires `Debug` and
+`Display`. It's generally helpful for them to implement `Clone` and `Eq` too where possible, to make
+life easier for tests and consumers of your library. In this case we can't easily do so, because
+`io::Error` doesn't implement them.
+
 </details>
--- a/src/error-handling/converting-error-types.md
+++ b/src/error-handling/converting-error-types.md
@@ -19,13 +19,26 @@ The `From::from` call here means we attempt to convert the error type to the
 type returned by the function:
 
 ```rust,editable
+use std::error::Error;
 use std::{fs, io};
 use std::io::Read;
+use std::fmt::{self, Display, Formatter};
 
 #[derive(Debug)]
 enum ReadUsernameError {
     IoError(io::Error),
     EmptyUsername(String),
+}
+
+impl Error for ReadUsernameError {}
+
+impl Display for ReadUsernameError {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        match self {
+            Self::IoError(e) => write!(f, "IO error: {}", e),
+            Self::EmptyUsername(filename) => write!(f, "Found no username in {}", filename),
+        }
+    }
 }
 
 impl From<io::Error> for ReadUsernameError {
--- a/src/error-handling/deriving-error-enums.md
+++ b/src/error-handling/deriving-error-enums.md
@@ -8,7 +8,7 @@ use std::{fs, io};
 use std::io::Read;
 use thiserror::Error;
 
-#[derive(Error, Debug)]
+#[derive(Debug, Error)]
 enum ReadUsernameError {
     #[error("Could not read: {0}")]
     IoError(#[from] io::Error),
@@ -33,3 +33,13 @@ fn main() {
     }
 }
 ```
+
+<details>
+
+`thiserror`'s derive macro automatically implements `std::error::Error`, and optionally `Display`
+(if the `#[error(...)]` attributes are provided) and `From` (if the `#[from]` attribute is added).
+It also works for structs.
+
+It doesn't affect your public API, which makes it good for libraries.
+
+</details>
--- a/src/error-handling/error-contexts.md
+++ b/src/error-handling/error-contexts.md
@@ -23,7 +23,7 @@ fn read_username(path: &str) -> Result<String> {
         .read_to_string(&mut username)
         .context("Failed to read")?;
     if username.is_empty() {
-        return Err(ReadUsernameError::EmptyUsername(String::from(path)))?;
+        return Err(ReadUsernameError::EmptyUsername(String::from(path)).into());
     }
     Ok(username)
 }
--- a/src/error-handling/error-contexts.md
+++ b/src/error-handling/error-contexts.md
@@ -7,14 +7,7 @@ custom error types:
 ```rust,editable,compile_fail
 use std::{fs, io};
 use std::io::Read;
-use thiserror::Error;
-use anyhow::{Context, Result};
-
-#[derive(Error, Debug)]
-enum ReadUsernameError {
-    #[error("Found no username in {0}")]
-    EmptyUsername(String),
-}
+use anyhow::{Context, Result, bail};
 
 fn read_username(path: &str) -> Result<String> {
     let mut username = String::with_capacity(100);
@@ -23,7 +16,7 @@ fn read_username(path: &str) -> Result<String> {
         .read_to_string(&mut username)
         .context("Failed to read")?;
     if username.is_empty() {
-        return Err(ReadUsernameError::EmptyUsername(String::from(path)).into());
+        bail!("Found no username in {path}");
     }
     Ok(username)
 }
@@ -39,7 +32,8 @@ fn main() {
 
 <details>
 
-* `anyhow::Result<T>` is generic and it can hold any `Error` implementation without changing the type signature.
+* `anyhow::Result<V>` is a type alias for `Result<V, anyhow::Error>`.
+* `anyhow::Error` is generic and it can hold any `Error` implementation without changing the type signature.
 * Actual error type inside of it can be extracted for examination if necessary.
 * Functionality provided by `anyhow::Result<T>` may be familiar to Go developers, as it provides similar usage patterns and ergonomics
   of `(T, error)` from Go. 
--- a/src/error-handling/error-contexts.md
+++ b/src/error-handling/error-contexts.md
@@ -33,9 +33,10 @@ fn main() {
 <details>
 
 * `anyhow::Result<V>` is a type alias for `Result<V, anyhow::Error>`.
-* `anyhow::Error` is generic and it can hold any `Error` implementation without changing the type signature.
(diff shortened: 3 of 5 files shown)


src/error-handling/converting-error-types.md
```diff
@@ -19,15 +19,17 @@ The `From::from` call here means we attempt to convert the error type to the
 type returned by the function:
 
 ```rust,editable
-use std::{fs, io};
+use std::{fs::{self, File}, io};
```
mgeisler, line 22: I know it's a bit of taste, but I really dislike the nested import syntax. It feels very over-engineered to me and I would rather not present it to brand new Rust developers :-)

I'm also not a fan of the `self` — I just like multi-line imports with a nice and consistent hiearchy as you go down the list. Perhaps you restructured this to make the code fit on a slide?

qwandor: Changed to remove nested braces.

src/error-handling/converting-error-types.md
```diff
@@ -19,15 +19,28 @@ The `From::from` call here means we attempt to convert the error type to the
...
+use std::fs::{self, File};
+use std::io::{self, Read};
 
 #[derive(Debug)]
 enum ReadUsernameError {
     IoError(io::Error),
     EmptyUsername(String),
 }
 
+impl Error for ReadUsernameError {}
```
mgeisler, line 33: This code (together with the expanded imports above) make the code too large to comfortably have on a slide in a classroom. Can we take it out again?

I don't think I we show how to `impl Display` anywhere, so that's a great addition — but it's not central to the concept of error handling, so I would suggest we move it elsewhere.

src/error-handling/deriving-error-enums.md
```diff
@@ -4,11 +4,11 @@ The [thiserror](https://docs.rs/thiserror/) crate is a popular way to create an
 error enum like we did on the previous page:
 
 ```rust,editable,compile_fail
-use std::{fs, io};
+use std::{fs::{self, File}, io};
 use std::io::Read;
 use thiserror::Error;
 
-#[derive(Error, Debug)]
+#[derive(Debug, Error)]
```
mgeisler, line 11: Are you sorting them alphabetically? That's not a bad idea, I'm just curious.

Also, I'll eventually do auto-formatting with `rustfmt` via something like https://dprint.dev/. So then we can have `rustfmt` do this kind of changes for us in a consistent way (though I believe sorting derives is unstable).

src/error-handling/error-contexts.md
```diff
@@ -5,25 +5,18 @@ contextual information to your errors and allows you to have fewer
...
 
 fn read_username(path: &str) -> Result<String> {
     let mut username = String::with_capacity(100);
-    fs::File::open(path)
+    File::open(path)
         .context(format!("Failed to open {path}"))?
         .read_to_string(&mut username)
         .context("Failed to read")?;
     if username.is_empty() {
-        return Err(ReadUsernameError::EmptyUsername(String::from(path)))?;
```
mgeisler, line 26: With this change, it's no longer possible to downcast to the original error, right?

I'm not sure why this is a good idea then :slightly_smiling_face:

src/error-handling/converting-error-types.md
qwandor: Is there some way we can hide or collapse the implementation by default in the slide view? I agree that the details of the `Display` implementation are not important, but I want to demonstrate good practices to the students, including implementing `std::error::Error` for all error types.

src/error-handling/deriving-error-enums.md
qwandor: Yes, I just sorted them alphabetically for consistency.

src/error-handling/error-contexts.md
qwandor: True, but if you want to distinguish different types of errors programmatically you probably shouldn't be using `anyhow`. Usually people use `anyhow` or similar dynamic error types because they just want to print an error message or backtrace rather than checking details of the error in code. I think using `anyhow!` or `bail!` is more idiomatic than defining a custom error type only to immediately wrap it in an `anyhow::Error`.

src/error-handling/converting-error-types.md
mgeisler: > Is there some way we can hide or collapse the implementation by default in the slide view?

Partially, you can hide lines like with normal `rustdoc`: https://rust-lang.github.io/mdBook/format/mdbook.html#hiding-code-lines. However, this doesn't combine with the `editable` annotation so you have to pick your poison: an editable example or a correct one :slightly_smiling_face: 

I think it's laudable to want to be correct and show all the good practices. However, you have to keep in mind that these are _slides_. People are probably also struggling to keep everything straights at this point, so you telling them that errors should implement `Display` might make no big difference.

I'm fine either way, but my instinct would be to make the code _as small as possible_ and put all other material into the speaker notes. I'll teach the class next week, so I'll probably have more concrete feedback after that :smile:  (I'll keep the notes open on my laptop and edit them as I go... so that should let me come back with concrete experiences on what fits and what don't.)

src/error-handling/error-contexts.md
```diff
@@ -5,25 +5,18 @@ contextual information to your errors and allows you to have fewer
 custom error types:
 
 ```rust,editable,compile_fail
-use std::{fs, io};
+use std::{fs::{self, File}, io};
```
mgeisler, line 8: I'm still not sure why you change the imports like this :slightly_smiling_face:

The `self` seems unnecessary here — I normally follow the style of trying to minimize imports by importing a _module_, not individual items. That makes it much easier to see where things come from at a glance and it minimizes the necessary changes as you use more/less items from the module.

src/error-handling/error-contexts.md
```diff
@@ -5,25 +5,18 @@ contextual information to your errors and allows you to have fewer
...
 fn read_username(path: &str) -> Result<String> {
     let mut username = String::with_capacity(100);
-    fs::File::open(path)
+    File::open(path)
         .context(format!("Failed to open {path}"))?
         .read_to_string(&mut username)
         .context("Failed to read")?;
     if username.is_empty() {
-        return Err(ReadUsernameError::EmptyUsername(String::from(path)))?;
+        bail!("Found no username in {}", path);
```
mgeisler, line 19: I'm assuming `bail!` supports this the same way as `format!` and friends:

```suggestion
        bail!("Found no username in {path}");
```

qwandor: It does, I just don't like it. Done anyway.

src/error-handling/converting-error-types.md
```diff
@@ -36,7 +49,7 @@ impl From<io::Error> for ReadUsernameError {
 
 fn read_username(path: &str) -> Result<String, ReadUsernameError> {
     let mut username = String::with_capacity(100);
-    fs::File::open(path)?.read_to_string(&mut username)?;
+    File::open(path)?.read_to_string(&mut username)?;
```
mgeisler, line 52: Please don't change the existing code without a good reason. This change just leads to more imports at the top and it's harder to follow for the students who don't know the standard library very well.

src/error-handling/try-operator.md
```diff
@@ -19,11 +19,11 @@ some_expression?
 We can use this to simplify our error handing code:
 
 ```rust,editable
-use std::fs;
+use std::fs::{self, File};
```
mgeisler, line 22: This change seems unnecessary like the other import changes.

mgeisler (changes requested):
Please put the imports back the way they were. I've used that way throughout the course (to the best of my knowledge), so if we want to change how imports look, we need to do it everywhere for consistency.

src/error-handling/error-contexts.md
mgeisler: Yeah, I've been going back and forth on this myself... I was super used to the `"... {} ... {} ...", x, y` style at first: I like how it keeps the format string short and collects everything at the end in a consistent way. Now I'm trying to inline the variables when I can — I find that this typically makes things more readable.

However, it all breaks down and becomes very annoying to look at when you don't have suitably-named variables at hand. When you can either create extra temporaries for the variables (a bit weird) or you can put the expressions at the end of your print call. Mixing the two styles looks really odd to me.

For now, the style used (somewhat consistently, I hope) is to inline the variables.

src/error-handling/converting-error-types.md
qwandor: What I've seen happen before, and I'm pretty sure will happen here, is that people see the example on the slides as an example of a correct implementation, and copy it when they need to write something similar themselves. They might well do so some months later when they get around to writing some Rust in their project, and they've forgotten the details of what you said in the course.  With that in mind, I'd really like as much as possible for the examples we give to be correct and following best practices.

Would you be happier if I split the example out onto its own page, so you don't have to scroll?

src/error-handling/converting-error-types.md
qwandor: The existing code is inconsistent: some examples import `std::fs::File`, some import `fs::File` and use `fs::File` in the code. And other symbols are fully imported with their full paths, like `anyhow::Result` and `thiserror::Error`. The [documentation for `std::fs::File`](https://doc.rust-lang.org/std/fs/struct.File.html) also imports it with the full path.

qwandor:
I've removed the import cleanup from this PR so we can discuss that separately in another PR, and I've rebased this to resolve conflicts. Are you happy to merge this one now, before there are more conflicts?

src/error-handling/converting-error-types.md
mgeisler: Yeah, I guess it has become a bit of a mess... My gut feeling is to use the module name when it's short, from the standard library. Which is not a very clear "rule" :smile: Based on that, I guess it's fine to import the names in the unqualified way you do here.

src/error-handling/converting-error-types.md
mgeisler: > Would you be happier if I split the example out onto its own page, so you don't have to scroll?

Yes, that would be great! The scrolling was something people mentioned to me this week when I taught the course. I'm thinking we need to go through all the pages again and edit them for brevity. It's an annoying battle between being complete and idiomatic while also being useful on a small screen :slightly_smiling_face:

mgeisler: > Yes, that would be great!

Sorry, I was too fast... let's merge this as-is and then refactor this after the training I'm doing on Monday.

mgeisler:
> Are you happy to merge this one now, before there are more conflicts?

Yep, in it goes! It's clearly a much better coverage of this than what we had before :+1:

mgeisler (approved):

PR #177, opened 2023-01-17
Fixes to unsafe chapter

qwandor:
Added speaker notes, safety comments, and a couple more pages.


mgeisler:
> and a couple more pages

Hey Andrew! I like the content a lot that you're adding, but... I would suggest that you try it out first before we commit it as-is. Other instructors, like @maurer and @fbornhofen, might have different experiences, but my experience is that the days are _very full_. That is, the course as it was in ~December was already near the breaking point in terms of content. 

It might not look very full, and we had a comment in #38 where @shinglyu suggested that Day 1 morning could fit into a 1 hour lecture. That's not my experience, at least not when the discussions start flowing.

I guess what I'm saying is that we should know more in a few days when you've had a chance to teach the material here. Is that okay, or do you think it's important to commit and publish it so that your attendees can find the material shown during class at https://google.github.io/comprehensive-rust/?

src/unsafe/calling-unsafe-functions.md
```diff
@@ -1,11 +1,14 @@
...
 
 A function or method can be marked `unsafe` if it has extra preconditions you
-must uphold:
+must uphold to avoid undefined behaviour:
 
 ```rust,editable
 fn main() {
     let emojis = "🗻∈🌏";
+
+    // Safe because the indices are in the correct order, within the bounds of
```
mgeisler, line 10: This duplicates the comment just below. I made the comment short on purpose to make things fit on the screen in the classroom.

src/unsafe/raw-pointers.md
```diff
@@ -9,10 +9,34 @@ fn main() {
     let r1 = &mut num as *mut i32;
     let r2 = &num as *const i32;
 
+    // Safe because r1 and r2 were obtained from references and so are guaranteed to be non-null and
```
mgeisler, line 12: Now that I think about this again: is it not already UB to read from the immutable `r2` when we have the mutable `r1` existing at the same time? That is, I think I've seen that Rust demands that you uphold the normal borrow checker rules, even when you use `unsafe`. I'm not very sure of this, though.

src/unsafe/unsafe-traits.md
```diff
@@ -0,0 +1,32 @@
+# Implementing Unsafe Traits
+
+The `zerocopy` crate has an unsafe trait that looks
```
mgeisler, line 3: You should probably say something like "A trait can be marked as _unsafe_ if it has extra preconditions which the implementer must uphold. Here is an example". You can then explain to the instructor that this comes from zerocopy since this fact doesn't seem central to _what_ an unsafe trait is.

qwandor: Done.

qwandor:
> > and a couple more pages
> 
> Hey Andrew! I like the content a lot that you're adding, but... I would suggest that you try it out first before we commit it as-is. Other instructors, like @maurer and @fbornhofen, might have different experiences, but my experience is that the days are _very full_. That is, the course as it was in ~December was already near the breaking point in terms of content.

Yeah, there is a lot to cover. The course is called Comprehensive Rust though, and right now the introduction to the unsafe section mentions unsafe traits without explaining them, and doesn't talk about defining unsafe functions at all. These are things that often trip people up, and will come up especially with people writing `no_std` code, so I think it's important to at least cover them briefly.

> It might not look very full, and we had a comment in #38 where @shinglyu suggested that Day 1 morning could fit into a 1 hour lecture. That's not my experience, at least not when the discussions start flowing.
> 
> I guess what I'm saying is that we should know more in a few days when you've had a chance to teach the material here. Is that okay, or do you think it's important to commit and publish it so that your attendees can find the material shown during class at https://google.github.io/comprehensive-rust/?

I need to have it somewhere so I can teach it, but I guess I can merge everything into my fork for now.

fbornhofen:
So far, I'd also say that we already have a lot of material. At the same time, I agree with the reasoning that "comprehensive" course shouldn't leave out important information.
As a facilitator, I like to cover all topics in the outline at least superficially, and then go deeper into topics that the participants are interested in. I'm also ok with skipping over some slides when time is running out and the topic seems too advanced.

However, if we end up with slides that start resembling the Rust book, facilitators have to curate topics once again. One easy solution could be extra sections with supplementary materials and exercises for topics that are worth a deep dive, and pointers to those materials in the main slides. WDYT?

src/unsafe/calling-unsafe-functions.md
mgeisler: Yeah, please remove the comment and preserve the vertical height.

src/unsafe/raw-pointers.md
mgeisler: Okay, cool!

src/unsafe/calling-unsafe-functions.md
mgeisler: I think this comment would imply that _the indices are in the correct order_ and _within the bounds of the string slice_:

```suggestion
    // Safe because the indices lie on UTF-8 sequence boundaries.
```

It's just the opposite of the comment below, so it has the same size :-)

src/unsafe/raw-pointers.md
mgeisler: Please consider taking out the comment inside the `unsafe` block — the comment here covers it. Also consider making the comment here shorter. Perhaps just

```suggestion
    // Safe because r1 and r2 are aligned and live pointers.
```

src/unsafe/writing-unsafe-functions.md
```diff
@@ -0,0 +1,38 @@
+# Writing Unsafe Functions
+
+You can mark your own functions as `unsafe` if they require particular conditions to avoid undefined
+behaviour.
+
+```rust,editable
```
mgeisler, line 6: Can you fit 20 lines of code on the screen _without scrolling_ in a conference room?

mgeisler (approved):
I would like to merge this — it's great to cover unsafe Rust in more detail.

I've left some comments saying that the code is probably too big for the screen... we can probably deal with this as we go. I'll teach the course next week and then I'll have a chance to go through every page and edit them a bit to fit :slightly_smiling_face:

mgeisler:
> So far, I'd also say that we already have a lot of material. At the same time, I agree with the reasoning that "comprehensive" course shouldn't leave out important information.

Yep, and I think the content here is a great addition! So I think we should merge it, and then we'll massage it a bit as we go forward.

> facilitators have to curate topics once again

That is a good point: this whole course was in some sense created because sources like [Rust by Example](https://github.com/rust-lang/rust-by-example) are too large to be taught in a course.

> One easy solution could be extra sections with supplementary materials and exercises for topics that are worth a deep dive, and pointers to those materials in the main slides. WDYT?

I think we'll slowly work towards doing precisely that. We'll eventually have a section about async Rust #181 and I hope we can add more specialized topics. When that happens, we can give suggested paths through the material to fit a course in 3, 4, or 5 days.

src/unsafe/writing-unsafe-functions.md
qwandor: Yes.

src/unsafe/calling-unsafe-functions.md
qwandor: That's not enough for the call to be safe though. I really want to drill into people that if you're calling an unsafe function, you need to read all the safety requirements documented for that function and ensure that they are all met, and explain that in your safety comment.

I keep seeing in code reviews of unsafe Rust cases of people not doing that, and writing unsound code as a result, so I think it's important to get people thinking carefully and precisely about unsafe code right from the start, as soon as we introduce it. If we're going to cover unsafe Rust at all, we should make sure people know how to do it properly, otherwise we're going to end up with lots of sloppy unsafe Rust in our codebase and lose a large part of the benefit of using Rust in the first place.

src/unsafe/raw-pointers.md
qwandor: I've removed the comment inside the block. I do think it's important that the safety comment here explains exactly how we know that the requirements are met, rather than just parroting what they are. Writing it out in full forces people to think through exactly what they are doing and why.

qwandor:
I taught this yesterday, and got through it slightly quicker than expected, about 1:40 for day 3 afternoon including a bunch of questions. Error handling took a while with lots of questions, but testing was pretty quick to cover and there weren't any major questions, so we had plenty of time to cover the unsafe chapter.

I was very glad to have the 'Writing Unsafe Functions' example in, because that was one of the first things people asked about. The `zerocopy` example for unsafe traits was also useful, not just as an explanation of unsafe traits, but also because people were asking about safe wrappers of unsafe code and converting between types, and 'zerocopy' is a good example of how to do that safely.

If I were to skip anything from the unsafe chapter it would be the pages on unions and mutable statics, because they are very rarely used in practice.

src/unsafe/writing-unsafe-functions.md
mgeisler: Thanks, I've now added #187 to finally have a way of seeing this at a glance — it confirms that you're right!

PR #180, opened 2023-01-17
Add discussion points for traits

rbehjati:
Suggestion for a few discussion points that I think could be useful :)

### DIFF
--- a/src/traits/drop.md
+++ b/src/traits/drop.md
@@ -33,7 +33,10 @@ fn main() {
 
 Discussion points:
 
-* Why does not `drop` take `self`?
+* Why does not `Drop::drop` take `self`?
+    * Short-answer: If it did, `std::mem::drop` would be called at the end of
+        the block, resulting in another call to `Drop::drop`, and a stack
+        overflow!
 * Try replacing `drop(a)` with `a.drop()`.
 
 </details>
--- a/src/traits/drop.md
+++ b/src/traits/drop.md
@@ -28,3 +28,12 @@ fn main() {
     println!("Exiting main");
 }
 ```
+
+<details>
+
+Discussion points:
+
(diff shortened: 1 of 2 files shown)


mgeisler:
Hi @rbehjati, thanks for the PR!

I think you have some great prompts. Would you be able to also write answers to them? That way the speaker notes become super useful for people who don't have a Rust expert on hand to help them.

rbehjati:
My idea was to pose these questions to people and let them guess, and encourage in a discussion. I am not sure having the answer right there in the slides is a good idea. Even though it is in the speaker notes section, participants often pre-read the contents :)

I could add a hint though. Something like the following. What do you thin about that?

> * Why does not `drop` take `self`?

Hint: Think about how that interacts with `std::mem::drop`, moves its argument into the scope of the function, and is called at the end of every scope. 

> * You could implement `Add` for `&Point`. In which situations is that useful? 

Hint: Think about what happens if `Point` is not copy.

> * Why is `Output` an associated type? Could it be made a type parameter?

Hint: Think about who controls the parameter types, and who controls the associate types.

mgeisler:
> I am not sure having the answer right there in the slides is a good idea. Even though it is in the speaker notes section, participants often pre-read the contents :)

In my opinion, it's great if people read the notes ahead of time! People are here to learn and if they read the notes, I hope they learn something. At Google, we're teaching the class to professional software engineers, so if they show up well-prepared, it should be easy for people to engage in a discussion around the questions.

We can also merge this as it is now and then we'll see how the wider Rust community feels about it. I'm sure we'll get PRs and questions if the notes are not explaining enough.

mgeisler (approved):

rbehjati:
@mgeisler I added **short** answers. Can you take another look? :)

PR #183, opened 2023-01-18
Cache `i18n-helpers/target` explicitly

mgeisler:
Cache the `i18n-helpers` directory explicitly.


mgeisler:
Oh, I see to update the `build.yml` file twice... just a second.

mgeisler:
Nope, this still doesn't work — I think I'll have to create a Cargo workspace instead to get the desired caching effect.

mgeisler:
Fixed via another PR.

PR #184, opened 2023-01-18
Publish translations to `/xx` subdirectories

mgeisler:


mgeisler:
I'll test this out on my own clone first and then put up the PR when I know it works there.

PR #186, opened 2023-01-19
Speaker notes for 6.5 Functions

rastringer:
Adds speaker notes covering functions basics (declaration parameters, return values, the 'unit' type.


mgeisler:
Thanks, looks good!

PR #190, opened 2023-01-19
Fix `Cargo.toml` syntax

andygeorge:
This is very minor fix that uses the correct `Cargo.toml` syntax by quoting a field value.

This matches what is used in [the Rust Book](https://doc.rust-lang.org/book/ch09-01-unrecoverable-errors-with-panic.html).


mgeisler:
Hi @andygeorge, thanks a lot for spotting this!

PR #191, opened 2023-01-20
Added construction chapter

brandonpollack23:


src/methods/receiver.md
```diff
@@ -17,3 +17,11 @@ are other possible receivers for a method:
 Beyond variants on `self`, there are also
 [special wrapper types](https://doc.rust-lang.org/reference/special-types-and-traits.html)
 allowed to be receiver types, such as `Box<Self>`.
+
+<details>
+
+* In reality, all functions in rust are static.  Receivers just add the "dot syntax" syntactic sugar by implicitly passing the struct by move, reference, or mutable reference.
```
qwandor, line 23: That's not really true, method calls on trait objects are dynamic.

mgeisler, line 23:
```suggestion
* In reality, all functions in Rust are static.  Receivers just add the "dot syntax" syntactic sugar by implicitly passing the struct by move, reference, or mutable reference.
```

src/structs/struct-construction.md
```diff
@@ -0,0 +1,43 @@
+# Construction
+
+Unlike in C++ or Java, there is no "constructor" in rust for structs, all
```
mgeisler, line 3:
```suggestion
Unlike in C++ or Java, there is no "constructor" in Rust for structs, all
```

src/structs/struct-construction.md
```diff
@@ -0,0 +1,43 @@
...
+construction is done using struct construction syntax.
+
+It is convention, however, to either implement the Default trait, or create a
```
mgeisler, line 6:
```suggestion
It is convention, however, to either implement the `Default` trait, or create a
```

src/structs/struct-construction.md
```diff
@@ -0,0 +1,43 @@
+# Construction
+
+Unlike in C++ or Java, there is no "constructor" in Rust for structs, all
+construction is done using struct construction syntax.
+
+It is convention, however, to either implement the `Default` trait, or create a
+method called "new" that has no receiver (ie a "static" function).
```
mgeisler, line 7:
```suggestion
method called `new` that has no receiver (i.e., a "static" function).
```

src/structs/struct-construction.md
```diff
@@ -0,0 +1,43 @@
...
+
+    println!("{} is {} years old", peter.name, peter.age);
+}
+```
+
+<details>
+
+* Mention the `Self` static scope accessor, it allows you to access any method of a struct.
+
+* In fact, dot method call syntax is just syntactic sugar, you can even access methods with &self receiver parameters by explicitly passing structs in to the first parameter, eg `Person::display(&peter)` if it had such a method `display(&self)`.
```
mgeisler, line 39:
```suggestion
* In fact, dot method call syntax is just syntactic sugar, you can even access methods with `&self` receiver parameters by explicitly passing structs in to the first parameter, e.g., `Person::display(&peter)` if it had such a method `display(&self)`.
```

mgeisler:
Hey @brandonpollack23 and @qwandor, since I plan on sqush-merging this, I've added a bunch of small formatting fixes.

src/structs/struct-construction.md
```diff
@@ -0,0 +1,43 @@
+# Construction
+
+Unlike in C++ or Java, there is no "constructor" in Rust for structs, all
```
mgeisler, line 3: I would like to add this content — it's great for the course participants if we can cover constructors.

Please move the file to `methods/constructors.md` instead since constructors are not strictly about structs (they apply to enums too).

Then we just need to fix the small problem pointed out by Andrew and we should be good to go!

mgeisler (changes requested):
We just need to move the file and then fix the small part about dynamic calls.

src/methods/receiver.md
mgeisler: Right, we should instead talk about the syntax, which allows you to call methods using the fully qualified syntax.

mgeisler:
Hi @brandonpollack23, can you check if we still need this PR?

src/structs/struct-construction.md
mgeisler: I would still like to see us cover this. Perhaps as part of @djmitche reorg in #1073.

djmitche: I think this will be in the methods slide?

mgeisler: Yeah, I think so too.

mgeisler: Yeah, I see it's mentioned on the slide about methods now.

mgeisler:
This is covered (a tiny bit) by #1073.

PR #192, opened 2023-01-20
Add speaker for Drop trait and lifetime of references.

Marvin182:
Add a little example that demonstrates that the compiler checks lifetimes of references. It's not specific to the Drop trait but the question might come up in the context.

### DIFF
--- a/src/traits/drop.md
+++ b/src/traits/drop.md
@@ -28,3 +28,34 @@ fn main() {
     println!("Exiting main");
 }
 ```
+
+<details>
+
+* References in structs must have a lifetime annotation and the compiler will use it to verify assigments.
+```rust,editable
+struct Droppable<'a> {
+    name: &'a String,
+}
+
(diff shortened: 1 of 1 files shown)


mgeisler:
Hi @Marvin182, thanks for the PR!

> Add a little example that demonstrates that the compiler checks lifetimes of references.

The example is not bad, but I fear that it will be confusing for the course participants to see the instructor copy-paste the example on top of the existing code. The example shows something very similar to what is covered on Day 1: https://google.github.io/comprehensive-rust/basic-syntax/references-dangling.html.

Could you instead add notes which explain what the instructor (in your view) should focus on in this particular slide? For example, the instructor should explain what happens if a value is dropped conditionally (Rust maintains _drop flags_ behind the scenes for this).

mgeisler:
Let's close this PR for now — we can then open another PR for the thing about drop flags later.

PR #195, opened 2023-01-20
Update structs.md

cchiw:


src/structs.md
```diff
@@ -17,3 +17,14 @@ fn main() {
     println!("{} is {} years old", peter.name, peter.age);
 }
 ```
+
+<details>
+Key Points: 
+
+* Structs work similar as they do in other languages, but note there is no `typedef` keyword or inheritance between structs.
```
fbornhofen, line 24: WDYT about splitting this in multiple lines, e.g.:
* Structs work like in C or C++.
  * Like in C++, and unlike in C, no `typedef` is needed to define a type.
  * Unlike in C++, there is no inheritance between structs.

src/structs.md
```diff
@@ -17,3 +17,14 @@ fn main() {
...
 }
 ```
+
+<details>
+Key Points: 
+
+* Structs work similar as they do in other languages, but note there is no `typedef` keyword or inheritance between structs. 
+* Methods are defined in an `impl` block, which we will see in following slides.
+* This may be a good time to let people know there are different types of structs. The next slide will introduce Tuple structs. Here you can define a unit-like struct.
+    * Unit-like structs can be defined with `struct (name);`. There is no need for curly brackets and parentheses.
```
fbornhofen, line 27: The bullet point about different types could also be split up into a list.

I had trouble understanding the line with units and `struct (name)` (first thought about the newtype pattern. I'd suggest either mentioning in a new bullet point that structs can be empty (zero-sized structs) and then either not showing an example, or using a metasyntactic name, e.g., `struct Foo;`.

src/structs.md
```diff
@@ -17,3 +17,14 @@ fn main() {
...
+    * You may use this type of struct when implementing a trait on some type but don’t have any data that you want to store in the value itself.
```
fbornhofen, line 28: One last nitpick: I noticed the indentation is different from what we have in other files, let's try to keep that consistent with 0 spaces for the first and 2 spaces for the second level of bullet points (ex.: https://raw.githubusercontent.com/google/comprehensive-rust/main/src/memory-management/stack-vs-heap.md).

src/structs.md
cchiw: I like it.

cchiw:
ok added an update.

qwandor (approved):

PR #196, opened 2023-01-20
Rename LICENSE.txt to LICENSE

CodeMaster7000:
The file should be named `LICENSE` per https://github.com/google/new-project.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,3 @@
-
                                  Apache License
                            Version 2.0, January 2004
                         http://www.apache.org/licenses/


mgeisler:
Hey @CodeMaster7000,

Thanks, you're right that the file should be named `LICENSE` — I see that https://github.com/google/new-project uses that name too.

Can you update https://github.com/google/comprehensive-rust/blob/main/src/credits.md too?

LICENSE
```diff
@@ -1,4 +1,3 @@
-
```
mgeisler, line 1: Please don't edit the license file — we could keep it exactly like https://github.com/google/new-project/blob/master/LICENSE.

mgeisler (changes requested):
Please update the credits page to link to the new name.

LICENSE
CodeMaster7000: Ok

CodeMaster7000: This has been done in #272.

CodeMaster7000: I think both PRs are ready to be merged now.

CodeMaster7000:
I have done the changes that you requested @mgeisler. Since the head repository has been deleted, I can no longer go back and make changes to this PR. However, #272 has the changes so both PRs should be ready to be merged.

mgeisler:
> Since the head repository has been deleted, I can no longer go back and make changes to this PR. However, #272 has the requested changes so both PRs should be ready to be merged.

Thanks for fixing the credits page.

I believe this PR still removes the line at the top of `LICENSE`? Since the repository was deleted, I suggest we close both PRs and you can then open a new single PR which does both changes in one commit (you'll need to clone it down on your own machine for this).

CodeMaster7000:
@mgeisler Refer to #293 for the updated PR that reflects both of those requirements. I have also closed these 2 PRs. Thank you for the feedback!

PR #198, opened 2023-01-20
Speaker Notes for Field Shorthand

cchiw:
#274


src/structs/field-shorthand.md
```diff
@@ -21,3 +21,13 @@ fn main() {
     println!("{peter:?}");
 }
 ```
+
+<details>
+ Key Points: 
+    
+* This slide illustrates short-hand syntax, by initiating the fields in a struct with similar-named variables.
```
mgeisler, line 28: Can you make this text more direct? Please try to elaborate on the text in the slide instead of reformulating it.

mgeisler (changes requested):
Please make the text more direct: avoid saying "this might", instead tell the instructor what they need to know about the page.

cchiw:
Updated let me know what you think.

cchiw:
Soft ping on review.

src/structs/field-shorthand.md
```diff
@@ -34,4 +33,11 @@ impl Person {
 }
 ```
     
+
+<details>
+
+* Point out the syntax that is used inside the `new` function. To show a comparison you can also rewrite the struct definition using non-shorthand syntax.
```
mgeisler, line 39:
```suggestion
```

src/structs/field-shorthand.md
```diff
@@ -34,4 +33,11 @@ impl Person {
...
+* The `impl` block is where the boilerplate functions are defined. Later slides talk about methods in more detail.
```
mgeisler, line 40:
```suggestion
* Methods are defined in the `impl` block.
```

src/structs/field-shorthand.md
```diff
@@ -34,4 +33,11 @@ impl Person {
...
+* Use function update syntax to define a new structure using `peter`. Note that the variable `peter` will no longer be accessible afterwards.
```
mgeisler, line 41:
```suggestion
* Use struct update syntax to define a new structure using `peter`. Note that the variable `peter` will no longer be accessible afterwards.
```

I'm not sure why we're showing the students that here instead of earlier? This page is really just about the shorthand syntax.

mgeisler (changes requested):

src/structs/field-shorthand.md
mgeisler: It seems to me that this bullet point doesn't say anything new — at this point, students have already seen a method or two.

src/structs/field-shorthand.md
```diff
@@ -22,7 +22,6 @@ fn main() {
 }
 ```
 
-<details>
```
mgeisler, line 25: Hmm, why did we move the `<details>` start tag down? The page becomes very large now.

qwandor (approved):

cchiw:

src/structs/field-shorthand.md
```diff
@@ -33,5 +33,9 @@ impl Person {
     }
 }
 ```
-    
+
+* Methods are defined in the `impl` block.
+* Use struct update syntax to define a new structure using `peter`. Note that the variable `peter` will no longer be accessible afterwards.
+* Note, you can use `{:#?}` while printing structs to change the readability of the output.
```
mgeisler, line 39:
```suggestion
* Use `{:#?}` when printing structs to request the `Debug` representation.
```

mgeisler (approved):

PR #200, opened 2023-01-20
Additional speaker notes for variant payloads

cchiw:
#274


src/enums/variant-payloads.md
```diff
@@ -9,8 +9,15 @@ You can define richer enums where the variants carry data. You can then use the
 
 <details>
 
+Key Points:
+  
+* This might be the first time seeing pattern matching so perhaps introduce the topic, more details are provided in further slides.
+  * Match is very similar to checking against an if expression. The difference is an if expression returns a boolean value, where this match expression can return any type.
```
qwandor, line 15: I'm not sure what you mean by this. The match expression as a whole can return any type, but so can an if expression. The match arms must have patterns matching the type of the expression that is being matched on.

src/enums/variant-payloads.md
```diff
@@ -9,8 +9,15 @@ You can define richer enums where the variants carry data. You can then use the
...
+  * In a match, we check some expressions against some possible options, called match arms. Each option is declared with a pattern, an arrow, some code and then a comma. Sometimes a pair brackets between the arrow and the comma if it's multi-line.
```
qwandor, line 16: Do you mean "a pair of curly braces", i.e. a block?

src/enums/variant-payloads.md
```diff
@@ -9,8 +9,15 @@ You can define richer enums where the variants carry data. You can then use the
...
+  * Starting from the top we look for what pattern matches our use case then implement the code following the arrow. That code is only in-scope for that case. Once we find a match, we stop.
```
qwandor, line 17: s/our use case/the value/
s/implement/run/

I'm not sure what you mean by "in-scope" here.

src/enums/variant-payloads.md
```diff
@@ -9,8 +9,15 @@ You can define richer enums where the variants carry data. You can then use the
 
 <details>
 
+Key Points:
+  
+* This might be the first time seeing pattern matching so perhaps introduce the topic, more details are provided in further slides.
+  * Match is very similar to checking against an if expression. The difference is an if expression returns a boolean value, where this match expression can return any type.
+  * In a match, we check some expressions against some possible options, called match arms. Each option is declared with a pattern, an arrow, some code and then a comma. Sometimes a pair brackets between the arrow and the comma if it's multi-line.
+  * Starting from the top we look for what pattern matches our use case then implement the code following the arrow. That code is only in-scope for that case. Once we find a match, we stop. 
+  * You might demonstrate what happens when the search is inexhaustive. Note the advantage the Rust compiler provides by confirming when all cases are handles.
```
qwandor, line 18: s/handles/handled/

cchiw: updated file, let me know what you think

src/enums/variant-payloads.md
cchiw: I meant compared to if expressions in other languages, but I can see how this is confusing and removed the line.

src/enums/variant-payloads.md
cchiw: yes. :) I rerwote it

src/enums/variant-payloads.md
cchiw: I edited it.

src/enums/variant-payloads.md
```diff
@@ -9,8 +9,15 @@ You can define richer enums where the variants carry data. You can then use the
 
 <details>
 
+Key Points:
+  
+* This might be the first time seeing pattern matching so perhaps introduce the topic, more details are provided in further slides.
```
mgeisler, line 14: Please take this out this line: let the notes _add_ information to the reader (who can be both a student and an instructor). Avoid phrases like "This might" and "so perhaps".

src/enums/variant-payloads.md
```diff
@@ -9,8 +9,15 @@ You can define richer enums where the variants carry data. You can then use the
...
+  * A match expression can return any type.
```
mgeisler, line 15: Using the word "return" here is dangerous: returning has a specific meaning in a programming language and the `match` expression doesn't actually return anything.

Instead, please say something like "The `match` expression has a value. The value is the last expression in the match arm which was executed".

mgeisler (changes requested):

qwandor (approved):

cchiw:

src/enums/variant-payloads.md
```diff
@@ -9,10 +9,14 @@ You can define richer enums where the variants carry data. You can then use the
 
 <details>
 
-* In the above example, accessing the `char` in `KeyPress`, or `x` and `y` in `Click` only works within a `match` or an `if let` statement.
-* `match` and `if let` inspect a hidden discriminant field in the `enum`.
+* In the above example, accessing the `char` in `KeyPress`, or `x` and `y` in `Click` only works within a `match` statement.  
+  * In a match, we check some expressions against some possible options, called match arms. Each option is declared with a pattern, an arrow, some code and then a comma. When it's a multi-line block we use a pair curly braces.
```
mgeisler, line 13: Let's simplify this a bit and explain the lack of fall-through (the syntax is explained by the example itself):

```suggestion
  * The expressions is matched against the patterns from top to bottom. There is no fall-through like in C or C++.
```

src/enums/variant-payloads.md
```diff
@@ -9,10 +9,14 @@ You can define richer enums where the variants carry data. You can then use the
 
 <details>
 
-* In the above example, accessing the `char` in `KeyPress`, or `x` and `y` in `Click` only works within a `match` or an `if let` statement.
-* `match` and `if let` inspect a hidden discriminant field in the `enum`.
+* In the above example, accessing the `char` in `KeyPress`, or `x` and `y` in `Click` only works within a `match` statement.
```
mgeisler, line 12:
```suggestion
* The values in the enum variants can only be accessed after being pattern matched. The pattern binds references to the fields in the "match arm" after the `=>`.
```

mgeisler (approved):

PR #202, opened 2023-01-20
Speaker notes for methods.md

cchiw:


src/methods.md
```diff
@@ -25,3 +25,17 @@ fn main() {
 }
 ```
 
+<details>
+
+Key Points:
+* It can be helpful to introduce methods by comparing them to functions.
+  * Methods are called on an instance of a struct, the first parameter represents the instance as `self`.
```
qwandor, line 32: Not necessarily a struct, they can be implemented for any type (e.g. an enum or a primitive).

cchiw: updated, what do you think

qwandor: I've reworded it in a way that I think is clearer.

cchiw: LGTM

cchiw:
Add Issues #274

PR #259, opened 2023-01-23
Add speaker notes for Android interoperability slide

qwandor:
Also fixed the logging example to be consistent with the output given.

### DIFF
--- a/src/android/logging/src/main.rs
+++ b/src/android/logging/src/main.rs
@@ -15,7 +15,7 @@
 // ANCHOR: main
 //! Rust logging demo.
 
-use log::{debug, error};
+use log::{debug, error, info};
 
 /// Logs a greeting.
 fn main() {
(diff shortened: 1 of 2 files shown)


src/android/logging/src/main.rs
```diff
@@ -25,5 +25,6 @@ fn main() {
             .with_min_level(log::Level::Trace),
     );
     debug!("Starting program.");
+    info!("Things are going fine.");
```
mgeisler, line 28: I believe you need to import `info!` first.

qwandor: Indeed! Done.

mgeisler (approved):

src/android/logging/src/main.rs
mgeisler: Cool :smile: I need to hook up the `build_all.sh` script to CI, but it seems that it'll be a little tricky to do an AOSP checkout in a reasonable time.

PR #264, opened 2023-01-23
Publish translations to `/xx` subdirectories

mgeisler:
This publishes translations (currently only the Danish translation) to subdirectories named after the ISO 639-1 language code: “da/” for Danish, “ko/” for Korean, etc.

The list of translation is an explicit list to make it easy for us to enable/disable translations without being tied to the files in po/. This allows us to experiment with a translation without publishing it immediately.

I propose that we eventually move the English pages to an “en/” directory for symmetry with the other locales. However, for now, the pages remain at the room of our site (which works fine since we don’t have a subdirectory named “en/” in the course).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -18,6 +18,7 @@ concurrency:
 
 env:
   CARGO_TERM_COLOR: always
+  LANGUAGES: da
 
 jobs:
   publish:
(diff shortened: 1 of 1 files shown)


qwandor (approved):

mgeisler:
Thanks @qwandor, the very-incomplete Danish translation is now visible under https://google.github.io/comprehensive-rust/da/. @keispace, we can also publish the Korean translation now.

PR #265, opened 2023-01-24
Speaker notes for destructuring-structs.md

cchiw:
#274

### DIFF
--- a/src/pattern-matching/destructuring-structs.md
+++ b/src/pattern-matching/destructuring-structs.md
@@ -5,3 +5,10 @@ You can also destructure `structs`:
 ```rust,editable
 {{#include ../../third_party/rust-by-example/destructuring-structs.rs}}
 ```
+<details>
+  
+Key Points:
+* Showcase the different ways you can match on a struct by giving the fields literal values, variables, and wildcards.
+    * Illustrate how limitating a literal value can be, if it’s unnecessary.
+    * Point out that you can use a variable instead of a field then use that variable in the following code block.
+</details>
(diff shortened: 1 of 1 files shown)


src/pattern-matching/destructuring-structs.md
```diff
@@ -5,3 +5,10 @@ You can also destructure `structs`:
 ```rust,editable
 {{#include ../../third_party/rust-by-example/destructuring-structs.rs}}
 ```
+<details>
+  
+Key Points:
+* Showcase the different ways you can match on a struct by giving the fields literal values, variables, and wildcards.
+    * Illustrate how limitating a literal value can be, if it’s unnecessary.
```
mgeisler, line 12: Hey @cchiw, I don't understand what you mean by this?

Can you make the steps more direct: instead of "showcase the different ways", try saying what the instructor should do in detail.

cchiw: Ok, modified. Let me know what you think

cchiw:
#274

src/pattern-matching/destructuring-structs.md
```diff
@@ -5,3 +5,8 @@ You can also destructure `structs`:
 ```rust,editable
 {{#include ../../third_party/rust-by-example/destructuring-structs.rs}}
 ```
+<details>
+
+* Change the literal values in `foo` to match with the other patterns.
+* Try adding a new field to the `Foo` struct and make changes to the pattern as needed
```
mgeisler, line 11:
```suggestion
* Add a new field to `Foo` and make changes to the pattern as needed.
```

mgeisler (approved):

PR #266, opened 2023-01-24
Lecture notes for Destructing Arrays

cchiw:
#274


src/pattern-matching/destructuring-arrays.md
```diff
@@ -5,3 +5,10 @@ You can destructure arrays, tuples, and slices by matching on their elements:
 ```rust,editable
 {{#include ../../third_party/rust-by-example/destructuring-arrays.rs}}
 ```
+
+<details>
+  
+Key points:
+* Continuing our journey in pattern matching, this slide just shows another type that can destructured with pattern matching.
```
mgeisler, line 12: Let's take this line out: it's more narrative than a "key point" for the instructor.

src/pattern-matching/destructuring-arrays.md
```diff
@@ -5,3 +5,10 @@ You can destructure arrays, tuples, and slices by matching on their elements:
 ```rust,editable
 {{#include ../../third_party/rust-by-example/destructuring-arrays.rs}}
 ```
+
+<details>
+  
+Key points:
+* Continuing our journey in pattern matching, this slide just shows another type that can destructured with pattern matching. 
+* This is a another great time to do some live coding. You can add/remove values from the array, compare the syntax `..` and `_`, and share any errors you get.
```
mgeisler, line 13: Can you update this with precise instructions on what to do?

Please make sure that the notes _add_ to the slide. They should explain things to people who don't have an instructor around as well as tell the instructor what they should focus on.

mgeisler (changes requested):
Let's make the notes

1. Expand on the content with extra background. Perhaps we can explain if using pattern matching is more/less expensive than using the equivalent `if` expressions?
2. Give precise tips for the instructor on what to do.

cchiw:
updated, let me know what you think

cchiw:

mgeisler (approved):
Sure, let's get this in.

PR #269, opened 2023-01-24
More fixes and additions to Day1 Morning

markozagar:
A few of these improve the language, and a few are extra clarity, arising from questions I was asked when going through the course with a colleague who doesn't know Rust.

There are also a few attempts to make source (md) formatting a bit more consistent, especially around line lengths and such.

### DIFF
--- a/src/basic-syntax/functions-interlude.md
+++ b/src/basic-syntax/functions-interlude.md
@@ -21,3 +21,11 @@ fn main() {
     println!("cash prize: {}", pick_one(500, 1000));
 }
 ```
+
+<details>
+
+* When using generics, the standard library's `Into<T>` can provide a kind of limited
+  polymorphism on argument types. We will see more details in exercise 7.1.
+
+</defails>
+
--- a/src/basic-syntax/functions-interlude.md
+++ b/src/basic-syntax/functions-interlude.md
@@ -25,7 +25,7 @@ fn main() {
 <details>
 
 * When using generics, the standard library's `Into<T>` can provide a kind of limited
-  polymorphism on argument types. We will see more details in exercise 7.1.
+  polymorphism on argument types. We will see more details in a later section.
 
 </defails>
 
--- a/src/basic-syntax/slices.md
+++ b/src/basic-syntax/slices.md
@@ -19,16 +19,18 @@ fn main() {
 
 * We create a slice by borrowing `a` and specifying the starting and ending indexes in brackets.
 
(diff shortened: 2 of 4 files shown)


src/basic-syntax/functions-interlude.md
```diff
@@ -21,3 +21,11 @@ fn main() {
     println!("cash prize: {}", pick_one(500, 1000));
 }
 ```
+
+<details>
+
+* When using generics, the standard library's `Into<T>` can provide a kind of limited
+  polymorphism on argument types. We will see more details in exercise 7.1.
```
mgeisler, line 28: Referencing section numbers is dangerous: they will go out of sync very easily. Please just say "later today" or "tomorrow" or similar.

markozagar: Excellent point. Amended to "later section".

src/basic-syntax/slices.md
```diff
@@ -19,16 +19,18 @@ fn main() {
...
 * We create a slice by borrowing `a` and specifying the starting and ending indexes in brackets.
 
-* If the slice starts at index 0, Rust’s range syntax means we can drop the starting index. 
+* If the slice starts at index 0, Rust’s range syntax allows us to drop the starting index, meaning that `&a[0..a.len()]` and `&a[..a.len()]` are identical.
     
-* The same is true for the last index, so `&a[2..a.len()]` and `&a[2..]` are equal.
+* The same is true for the last index, so `&a[2..a.len()]` and `&a[2..]` are identical.
 
-* `s` is a reference to a slice of `i32`s. Notice that the type of `s` no longer mentions the array length. This allows us to performing computations on slices of different sizes.
+* To easily create a slice of the full array, we can therefore use `&a[..]`.
```
mgeisler, line 26: I've always been wondering when `&a` becomes a slice and when you need to use `&a[..]` (or the more explicit `a.to_slice()`. Is that something we could explain here?

markozagar: I'm not super clear on the distinction myself, but I *think* &a becomes a slice if type elision would have made it a slice anyway (based on the target type), and it defaults to a reference if no other type information is found. I've never gone into the details and always just accepted what the compiler suggested, but it'd be great to have more clarity here from someone more experienced than me, I agree.

mgeisler: Okay, then let's leave it as it is for now. Someone else can come and fill in the details :smile:

PR #270, opened 2023-01-24
Introduce the struct update syntax in the chapter on structs.

markozagar:
I notice we don't mention this at all, but we use it in `structs.md`. I thought it would be a good idea to show it and demonstrate how it is commonly used in various libraries.


qwandor:
This is a nice example, but probably too long to easily fit on screen, and also quite a bit of extra content for a relatively minor detail. Maybe it would be better to just mention it in the speaker notes for the first time it is used in an example? We don't want the course to get too long and unweildy.

markozagar:
I think I agree. I was already a bit on the fence on whether to include it or not, and having a second opinion confirms my thinking. I think I've already added a mention of this in the speaker notes in `structs.md`, so it should be easily covered, and it has a small example there too.

PR #275, opened 2023-01-24
Verbose description of Sync-Send relationship

dguerri:
Add a verbose rationale of the statement:
  `T` is `Sync` if and only if `&T` is `Send`


mgeisler:
Hey @dguerri, thanks for the PR!

I like the text, this is definitely something we should explain! Can you wrap the new text in a `<details> ... </details>` element to turn it into a collapsible speaker note? We have quite limited vertical space on the slides when we use them in a classroom setting.

dguerri:
Good idea :)
I also trimmed the text a bit.
Hope it works now!

qwandor (approved):

PR #277, opened 2023-01-25
Add trailing backslash for multiline command

Mandragorian:
There were some backslashed missing in the command to serve the translated book. Thus the shell would execute the lines one after the other, instead of as one command.

This lead to the environment variables not actually being set, and thus the non-translated version was served anyway.


qwandor (approved):

mgeisler:
Thanks @Mandragorian for catching this! I should copy-paste more carefully! :slightly_smiling_face:

PR #279, opened 2023-01-26
Simply Polygon::length method

upsidedwn:
Propose to simply the length method of Polygon

### DIFF
--- a/src/exercises/day-2/points-polygons.rs
+++ b/src/exercises/day-2/points-polygons.rs
@@ -84,18 +84,12 @@ impl Polygon {
     }
 
     pub fn length(&self) -> f64 {
-        if self.points.is_empty() {
-            return 0.0;
+        let num_points = self.points.len();
+        let mut len: f64 = 0.0;
+        for idx in 0..num_points {
+            len += self.points[idx].dist(self.points[(idx + 1) % num_points]);
         }
-
-        let mut result = 0.0;
-        let mut last_point = self.points[0];
-        for point in &self.points[1..] {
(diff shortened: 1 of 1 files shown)


src/exercises/day-2/points-polygons.rs
```diff
@@ -84,18 +84,12 @@ impl Polygon {
     }
 
     pub fn length(&self) -> f64 {
-        if self.points.is_empty() {
-            return 0.0;
+        let num_points = self.points.len();
+        let mut len: f64 = 0.0;
+        for idx in 0..num_points {
+            len += self.points[idx].dist(self.points[(idx + 1) % num_points]);
```
mgeisler, line 90: I agree that the code is shorter... However, the code here is a little more _dangerous_ since every indexing operation has the potential to fail (of course it won't fail here because you do the wrap-around).

In Rust, you should normally favor using iterators. That way, the _structure_ of your code ensures that you cannot index outside of the collection.

mgeisler:
Hey @upsidedwn, thanks for the PR!

I don't think we should merge this since we want people to prefer iterators over direct indexing in Rust. We could instead add a comment about this to the code.

mgeisler:
I'll close this PR for now, please open another which tells people to use iterators over indexes (where possible).

src/exercises/day-2/points-polygons.rs
upsidedwn: Got it, appreciate you for taking the time to give feedback, sounds logical, agree we should go with your suggestion. Thank  you for pointing this out!

PR #281, opened 2023-01-26
Propose to use the generic term perimeter instead of circumference for shapes

### LINKED ISSUE #280: Rename circumference to perimeter to improve clarity of day 2 morning exercise
I thought that the use of circumference to describe the requirement to calculate the perimeter of a polygon was confusing, propose to rename circumference to use the more generic term perimeter instead, which works for both polygons and circles.

upsidedwn:
Closes #280

### DIFF
--- a/src/exercises/day-2/points-polygons.rs
+++ b/src/exercises/day-2/points-polygons.rs
@@ -204,7 +204,7 @@ mod tests {
     }
 
     #[test]
-    fn test_shape_circumferences() {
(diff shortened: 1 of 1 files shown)


mgeisler:
Someone else brought up the same thing, so thank you for improving my English! :smile:

upsidedwn:
@mgeisler thank you for sharing such a great Rust tutorial, just trying to contribute what little I can to improve the materials :)

PR #288, opened 2023-01-26
Update luhn.rs

mseravalli:
The solution might use more existing rust functionality and constructs.

Additionally in this flow we first parse and validate the input and only in the following phase we run the luhn algorithm.

Probably there are further improvements possible.


mseravalli:
the proposed solution is more inefficient than the original one as it requires more memory and multiple passes on the data.

PR #291, opened 2023-01-27
Add a cloud-translate script

mgeisler:
This little script will simplify starting a new translation by letting you send the untranslated entries through Google Translate.

To use the script, you need a Google Cloud account. Install `gcloud` and [setup application default credentials][1]. You do that by running

    gcloud auth application-default login

When that is done, you should be able to see an access token by running

    gcloud auth application-default print-access-token

You will need to create a new Google Cloud project. The project ID is needed for running `cloud-translate`:

    cargo run -- your-project-id-123456 ../po/xx.po 10000

The `10000` argument denotes the number of Unicode characters you want to translate: Cloud Translate is priced per character you translate. Currently, you get [500k characters for free per month][2].

[1]: https://cloud.google.com/docs/authentication/application-default-credentials#personal
[2]: https://cloud.google.com/translate/pricing


mgeisler:
@Mandragorian, I translated a few messages into Greek using the script found here: [el.zip](https://github.com/google/comprehensive-rust/files/10520523/el.zip). Let me know how hilarious it looks :smile: 

I see a few problems: it translates the code snippets, but without any understanding of what parts is Rust syntax (which should not be translated) and which is comments/strings (which could be translated). That's in some sense a problem in the way we extract the messages in `i18n-helpers`.

Mandragorian:
Mayebe this script can be interactive. I imagine that after translating each message it can ask the user to verify the translation. Then the user could alter the translation if something doesn't make sense, or discard it entirely if, for example, it is a translation of a code snippet.

It would make the translation process more tiring, but I guess we need to verify the translations/skip the code snippets so it might be better this way?

mgeisler:
> I imagine that after translating each message it can ask the user to verify the translation

I would rather not make the script do that: you're "asked" to verify the translation via the `fuzzy` flag which the script adds to every translation it spits out. The flag is the signal to the translator that they need to proof-read the translation.

We don't use fuzzy translations at all when publishing a translation (#212) so it's really important that the translator goes through the messages and removes the flag from each good translation.

mgeisler:
I've moved the code here to its own repository so we can close this PR: https://github.com/mgeisler/gettext-cloud-translate.

PR #295, opened 2023-01-27
Speaker Notes for hashmap.md

cchiw:
#274


src/std/hashmap.md
```diff
@@ -24,3 +24,15 @@ fn main() {
     }
 }
 ```
+
+
+<details>
+
+* The HashMap is not defined in the prelude and needs to be brought into scope.
+* In a hashmap, types with the Copy trait (`i32`) are copied into the hashmap, and owned values (`String`) are moved into the hashmap.
+* Try the following to shortcuts. The first line will see if a book is an the hashmap and if not return an alternative value. The second line will insert the alternative value in the hashmap if the book is not found.
```
qwandor, line 33: I'm not sure what you mean by "Try the following to shortcuts."

"in the hashmap" rather than "an the hashmap"

src/std/hashmap.md
```diff
@@ -24,3 +24,15 @@ fn main() {
     }
 }
 ```
+
+
+<details>
+
+* The HashMap is not defined in the prelude and needs to be brought into scope.
+* In a hashmap, types with the Copy trait (`i32`) are copied into the hashmap, and owned values (`String`) are moved into the hashmap.
```
qwandor, line 32: `Copy`

cchiw:
unsure how to include code in the notes.

src/std/hashmap.md
```diff
@@ -24,3 +24,15 @@ fn main() {
...
+<details>
+
+* The HashMap is not defined in the prelude and needs to be brought into scope.
+* In a hashmap, types with the `Copy` trait (`i32`) are copied into the hashmap, and owned values (`String`) are moved into the hashmap.
+* Try the following lines of code. The first line will see if a book is in the hashmap and if not return an alternative value. The second line will insert the alternative value in the hashmap if the book is not found.
+    ```
+    // let pc1 = page_counts.get("Harry Potter and the Sorcerer's Stone ").unwrap_or(&336);
+    // let pc2 = page_counts.entry("The Hunger Games".to_string()).or_insert(374);
+    ```
+    </details>
```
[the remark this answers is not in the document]
gendx: > unsure how to include code in the notes.

The `ignore` tag should allow to insert code snippets that are not valid on their own.

src/std/hashmap.md
```diff
@@ -24,3 +24,21 @@ fn main() {
     }
 }
 ```
+
+
+<details>
+
+* The HashMap is not defined in the prelude and needs to be brought into scope.
+* In a hashmap, types with the `Copy` trait (`i32`) are copied into the hashmap, and owned values (`String`) are moved into the hashmap.
```
qwandor, line 32: There's not actually any special-casing for `Copy` types is there?

cchiw: I guess not, but I thought this would be a common question. Should I remove it?

qwandor: I'm just not sure what you are trying to say here. Doesn't `HashMap` do exactly the same thing for both `i32` and `String`?

qwandor (approved):

PR #300, opened 2023-01-30
Add Korean translation

jiyongp:
- add ko.po
- translate ko(~23.01.19)
- change speaker-note ko
- change id
- translate ko(~23.01.20)
- ~day3 keynote
- draft done to f3446a91
- add @jiyongp comments of upstream PR #276.
- sync & apply review comments(upstream)
- After 10000 line apply review comments(upstream)
- chgange build.yml
- Fix the inconsistent newline character problem for the KO translation


jiyongp:
This is basically #290 from @keispace. The only change from me is the fix of the inconsistent newline character problem. I've created a new PR because I couldn't amened the original PR.

PR #305, opened 2023-01-30
Fix JetBrains capitalization

friederbluemle:
Minor capitalization fix: `Jetbrains` -> `JetBrains`

### DIFF
--- a/po/da.po
+++ b/po/da.po
@@ -1161,10 +1161,10 @@ msgstr ""
 
 #: src/cargo.md:29
 msgid ""
-"Some folks also like to use the [Jetbrains][4] family of IDEs, which do "
+"Some folks also like to use the [JetBrains][4] family of IDEs, which do "
 "their own analysis but have their own tradeoffs. If you prefer them, you can "
 "install the [Rust Plugin][5]. Please take note that as of January 2023 "
-"debugging only works on the CLion version of the Jetbrains IDEA suite."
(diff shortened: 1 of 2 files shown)


po/da.po
```diff
@@ -1161,10 +1161,10 @@ msgstr ""
 
 #: src/cargo.md:29
 msgid ""
-"Some folks also like to use the [Jetbrains][4] family of IDEs, which do "
+"Some folks also like to use the [JetBrains][4] family of IDEs, which do "
```
mgeisler, line 1164:
```suggestion
"Some folks also like to use the [Jetbrains][4] family of IDEs, which do "
```

po/da.po
```diff
@@ -1161,10 +1161,10 @@ msgstr ""
...
 "their own analysis but have their own tradeoffs. If you prefer them, you can "
 "install the [Rust Plugin][5]. Please take note that as of January 2023 "
-"debugging only works on the CLion version of the Jetbrains IDEA suite."
+"debugging only works on the CLion version of the JetBrains IDEA suite."
```
mgeisler, line 1167:
```suggestion
"debugging only works on the CLion version of the Jetbrains IDEA suite."
```

mgeisler (approved):
Thanks a lot!

I'll revert the changes to the `da.po` file: this file should only be updated by the translators via `msgmerge`. The change in capitalization will flow into the file the next time a translator updates the file.

PR #307, opened 2023-01-31
Simplify error handling pages

mgeisler:
This makes the `use` statements more consistent and shortens some variable names. When teaching the class last week, I noticed that the pages were too large to fit on the screen, so I hope this helps a tiny bit with this.

### DIFF
--- a/src/error-handling/converting-error-types.md
+++ b/src/error-handling/converting-error-types.md
@@ -20,9 +20,8 @@ type returned by the function:
 
 ```rust,editable
 use std::error::Error;
-use std::{fs, io};
+use std::{fs, fmt, io};
 use std::io::Read;
-use std::fmt::{self, Display, Formatter};
 
 #[derive(Debug)]
 enum ReadUsernameError {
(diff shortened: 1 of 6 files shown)


qwandor:
Doesn't using partially qualified names for types make the code samples longer and harder to read?

mgeisler:
> 

To me, it avoids front loading: the imports currently have a lot of _details_ from the code below.

They're complex, with funny dances in the form of `std::fs::{self, ..}` — that statement is terrible in the way that it has two names for the same thing (`std::fs` and the temporary alias `self`). Importing `std::fs` and using `fs` afterwards is simpler: no need to explain the `self` alias.

In addition to making the import itself simpler, I like how related types and functions can be grouped by module.

I know it's not consistent, but some modules seem like they're designed to be imported this way. In particular, `std::fs` and `std::io`. A module that doesn't work that way for me is `std:: collections`.

mgeisler:
Let's drop this, i have other cleanups for the same slides :smile:

PR #308, opened 2023-01-31
Fix raw pointer example which has UB

mgeisler:
The raw pointer example looks safe, but Miri flags it as having undefined behavior. This is a followup to the discussion in #177.

### DIFF
--- a/src/unsafe/raw-pointers.md
+++ b/src/unsafe/raw-pointers.md
@@ -9,14 +9,10 @@ fn main() {
     let r1 = &mut num as *mut i32;
     let r2 = &num as *const i32;
 
-    // Safe because r1 and r2 were obtained from references and so are guaranteed to be non-null and
-    // properly aligned, the objects underlying the references from which they were obtained are
-    // live throughout the whole unsafe block, and they are not accessed either through the
-    // references or concurrently through any other pointers.
     unsafe {
         println!("r1 is: {}", *r1);
         *r1 = 10;
-        println!("r2 is: {}", *r2);
+        println!("r2 is: {}", *r2);  // Undefined behavior!
     }
 }
 ```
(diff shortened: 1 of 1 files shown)


qwandor:
If triggers undefined behaviour then we shouldn't have it as an example, but I'm not clear on why it is UB.

riking:
The two-phase borrows mechanic means that `r1` is considered inactive until it's used to write through, and that write pops `r2` off the borrow stack.

mgeisler:
> The two-phase borrows mechanic means that `r1` is considered inactive until it's used to write through, and that write pops `r2` off the borrow stack.

Thanks @riking!

@qwandor, I would like to merge this so we can warn people against the example. The other examples in this section are similarly full of things not to do, so it fits into the theme.

mgeisler:
Hi @gendx, I think you've played a lot with unsafe... can you review this? Thanks :smile:

gendx (approved):

qwandor (changes requested):
The point of the page was to show how to use raw pointers safely, and how safety comments should look. I think introducing Miri at this point is just going to be confusing. We should write an example that is valid instead.

gendx:
> We should write an example that is valid instead.

How about something like this:

```rust
fn main() {
    let mut array = [5, 6];
    let r1 = &mut array[0] as *mut i32;
    let r2 = &array[1] as *const i32;
    unsafe {
        println!("r1 is: {}", *r1);
        *r1 = 10;
        println!("r1 is: {}", *r1);
        println!("r2 is: {}", *r2);
    }
    println!("array is: {array:?}");
}
```

Miri looks happy about it.

gendx:
Bonus code to modify if you want to get to Miri:

```rust
fn main() {
    let mut array = [5, 6];
    let r1 = &mut array[0] as *mut i32;
    let r2 = &array[1] as *const i32;
    unsafe {
        println!("r1 is: {}", *r1);
        *r1 = 10;
        println!("r1 is: {}", *r1);
        // Undefined behavior because this breaks the aliasing rules, as
        // r1.offset(1) and r2 are aliasing the same memory.
        //*r1.offset(1) = 20;
        println!("r2 is: {}", *r2);
    }
    println!("array is: {array:?}");
}
```

qwandor:
How about #486 instead?

PR #309, opened 2023-01-31
Simplify enum size slide

mgeisler:
The slide is way too large to comfortable use in a classroom. Instead of printing six lines of output, the instructor should carefully walk through the different examples. That way we can take the course participants with us through the explanations instead of simply showing them the end result.

### DIFF
--- a/src/enums/sizes.md
+++ b/src/enums/sizes.md
@@ -17,20 +17,8 @@ enum Foo {
     B,
 }
 
-#[repr(u32)]
-enum Bar {
-    A,  // 0
-    B = 10000,
-    C,  // 10001
-}
-
 fn main() {
     dbg_size!(Foo);
-    dbg_size!(Bar);
-    dbg_size!(bool);
-    dbg_size!(Option<bool>);
-    dbg_size!(&i32);
-    dbg_size!(Option<&i32>);
 }
 ```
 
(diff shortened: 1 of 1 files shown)


src/enums/sizes.md
```diff
@@ -17,20 +17,8 @@ enum Foo {
     B,
 }
 
-#[repr(u32)]
-enum Bar {
-    A,  // 0
-    B = 10000,
-    C,  // 10001
-}
-
```
qwandor, line 26: I think it's useful to explain how enum variant values work, and students coming from C especially are likely to ask about it. If it's too much for this slide maybe move it to another slide or the speaker notes?

src/enums/sizes.md
```diff
@@ -17,20 +17,8 @@ enum Foo {
...
 fn main() {
     dbg_size!(Foo);
-    dbg_size!(Bar);
-    dbg_size!(bool);
-    dbg_size!(Option<bool>);
-    dbg_size!(&i32);
-    dbg_size!(Option<&i32>);
```
qwandor, line 33: These are good examples to consider I think, how about moving them to the speaker notes to remind the instructor to show them in the class?

mgeisler: In my experience, there is no time to go through so many examples. But we can move them to the speaker notes for people to read offline.

src/enums/sizes.md
mgeisler: It's definitely something the instructor can spend a minute on hand-coding — but it's important that the page works as a _slide_. So it has to be short and crisp and to the point. It was very overloaded before.

mgeisler:
I move the examples to the speaker notes. @gendx and @qwandor: I think we'll need to rewrite them a bit since they've grown quite large here — and they show a lot of stuff which cannot be relied upon (as noted in the comments).

qwandor (approved):

PR #313, opened 2023-01-31
Simplify `Vec` slide

mgeisler:
The code is probably meant to illustrate the `vec!` macro, but it feels out of place when teaching: there is already enough material here to get through.


rbehjati (approved):
I agree!

PR #320, opened 2023-02-01
Add optional day about bare-metal Rust

qwandor:
This PR adds the morning sections about no_std in general and microcontrollers; I'll send a separate PR for the afternoon.


src/welcome-bare-metal.md
```diff
@@ -0,0 +1,21 @@
...
+Today we will talk about 'bare-metal' Rust: running Rust code without an OS underneath us. This will
+be divided into several parts:
+
+ * What is `no_std` Rust?
+ * Writing firmware for microcontrollers.
+ * Writing bootloader / kernel code for application processors.
+ * Some useful crates for bare-metal Rust development.
+
+For the microcontroller part of the course we will use the [BBC micro:bit](https://microbit.org/) v2
+as an example. It's a development board based on the Nordic nRF51822 microcontroller.
```
mgeisler, line 12: > Nordic nRF51822 microcontroller

This has no meaning to me :slightly_smiling_face:  Can it be described in more general terms, something like "a popular development board for getting started" or "a board with X, Y, and Z features"?

src/welcome-bare-metal.md
```diff
@@ -0,0 +1,21 @@
...
+
+To get started, install some tools we'll need later:
```
mgeisler, line 14:
```suggestion
To get started, install some tools we'll need later. We will use a Debian machine as our reference:
```

src/bare-metal/alloc-example/src/main.rs
```diff
@@ -0,0 +1,42 @@
...
+
+static mut HEAP: [u8; 65536] = [0; 65536];
+
+pub fn entry() {
+    // Safe because `HEAP` is only used here and `entry` is only called once.
+    unsafe {
+        // Give the allocator some memory to allocate.
+        HEAP_ALLOCATOR
+            .lock()
+            .init(HEAP.as_mut_ptr() as usize, HEAP.len());
```
mgeisler, line 36: As a side question: I wonder why the `init` function doesn't take a slice instead? Is there some reason why we have to break apart the array instead of passing in a slice (and then letting it do this internally)?

src/bare-metal/alloc-example/src/main.rs
```diff
@@ -0,0 +1,42 @@
...
+#[global_allocator]
+static HEAP_ALLOCATOR: LockedHeap<32> = LockedHeap::<32>::new();
+
+static mut HEAP: [u8; 65536] = [0; 65536];
+
+pub fn entry() {
+    // Safe because `HEAP` is only used here and `entry` is only called once.
+    unsafe {
+        // Give the allocator some memory to allocate.
+        HEAP_ALLOCATOR
```
mgeisler, line 34: It looks like you could fit the whole thing on one line:

image

src/bare-metal/alloc-example/src/main.rs
```diff
@@ -0,0 +1,42 @@
...
+
+#[global_allocator]
+static HEAP_ALLOCATOR: LockedHeap<32> = LockedHeap::<32>::new();
+
+static mut HEAP: [u8; 65536] = [0; 65536];
+
+pub fn entry() {
+    // Safe because `HEAP` is only used here and `entry` is only called once.
+    unsafe {
+        // Give the allocator some memory to allocate.
```
mgeisler, line 33: I don't think this comment adds anything which cannot be guessed from the slide already.

```suggestion
```

src/bare-metal/alloc.md
```diff
@@ -0,0 +1,16 @@
+# `alloc`
+
+To use `alloc` you must implement a global (heap) allocator.
```
mgeisler, line 3: I was thinking that the `LockedHeap` type would implement a trait, and it turns out it's called `GlobalAlloc`:

```suggestion
To use `alloc` you must implement a [global (heap) allocator](https://doc.rust-lang.org/nightly/core/alloc/trait.GlobalAlloc.html).
```

src/bare-metal/alloc.md
```diff
@@ -0,0 +1,16 @@
...
+{{#include alloc-example/src/main.rs:Alloc}}
+```
+
+<details>
+
+* `buddy_system_allocator` is a third-party crate implementing a basic buddy system allocator. Other
+  crates are available, or you can write your own or hook into your existing allocator.
+* The const parameter of `LockedHeap` is the max order of the allocator; i.e. in this case it can
+  allocate regions of up to 2**32 bytes.
+
```
mgeisler, line 15: Questions that come to mind:

* What are the differences from this program and a "normal" program without `no_std`?
* Can I use all heap allocating types like normal? 
* Do third-party libraries need to do anything to use the global allocator? (I suppose not since it's, well, global).

src/bare-metal/microcontrollers/mmio.md
```diff
@@ -0,0 +1,5 @@
+# Raw MMIO
```
mgeisler, line 1: Aha, MMIO is short for "memory-mapped I/O" according to https://en.wikipedia.org/wiki/Memory-mapped_I/O_and_port-mapped_I/O. I wanted to suggest adding a speaker note about it, but I cannot add it with a comment since the last like ends with triple-backtick and that breaks the formatting of the suggestion :slightly_smiling_face:

src/bare-metal/microcontrollers/examples/src/bin/mmio.rs
```diff
@@ -0,0 +1,64 @@
...
+#![no_main]
+#![no_std]
+
+extern crate panic_halt as _;
+
+use core::mem::size_of;
+use cortex_m_rt::entry;
+
+/// GPIO port 0 peripheral address
+const GPIO_P0: usize = 0x5000_0000;
```
mgeisler, line 25: I suppose there is a crate somewhere which would define this constant? But perhaps you're doing it by hand to show that there is no magic behind this?

src/bare-metal/microcontrollers/examples/src/bin/mmio.rs
```diff
@@ -0,0 +1,64 @@
...
+
+// GPIO peripheral offsets
+const PIN_CNF: usize = 0x700;
+const OUTSET: usize = 0x508;
+const OUTCLR: usize = 0x50c;
+
+// PIN_CNF fields
+const DIR_OUTPUT: u32 = 0x1;
+const INPUT_DISCONNECT: u32 = 0x1 << 1;
+const PULL_DISABLED: u32 = 0x0 << 2;
```
mgeisler, line 35: Ehm, this is just zero, right? Why would one shift a zero bit like this?

src/bare-metal/microcontrollers/examples/src/bin/mmio.rs
```diff
@@ -0,0 +1,64 @@
...
+            DIR_OUTPUT | INPUT_DISCONNECT | PULL_DISABLED | DRIVE_S0S1 | SENSE_DISABLED,
+        );
+    }
+
+    // Set pin 28 high and pin 21 low to turn the LED on.
+    let gpio0_outset = (GPIO_P0 + OUTSET) as *mut u32;
+    let gpio0_outclr = (GPIO_P0 + OUTCLR) as *mut u32;
+    // Safe because the pointers are to valid peripheral control registers, and no aliases exist.
+    unsafe {
+        gpio0_outset.write_volatile(1 << 28);
```
mgeisler, line 59: When seeing this, I wonder how the equivalent C code would look? Very similar, I assume?

So the advantage of Rust right here is presumably the more expressive type system (thought it's not being used directly here)?

src/bare-metal/microcontrollers/examples/src/bin/mmio.rs
```diff
@@ -0,0 +1,64 @@
...
+        );
+        pin_cnf_28.write_volatile(
+            DIR_OUTPUT | INPUT_DISCONNECT | PULL_DISABLED | DRIVE_S0S1 | SENSE_DISABLED,
+        );
+    }
+
+    // Set pin 28 high and pin 21 low to turn the LED on.
+    let gpio0_outset = (GPIO_P0 + OUTSET) as *mut u32;
+    let gpio0_outclr = (GPIO_P0 + OUTCLR) as *mut u32;
+    // Safe because the pointers are to valid peripheral control registers, and no aliases exist.
```
mgeisler, line 57: The line is clipped here:

image

src/bare-metal/microcontrollers/examples/src/bin/mmio.rs
mgeisler: Oh, the next slide shows this. Nevermind :smile:

src/bare-metal/microcontrollers/examples/src/bin/hal.rs
```diff
@@ -0,0 +1,44 @@
...
+// limitations under the License.
+
+// ANCHOR: Example
+#![no_main]
+#![no_std]
+
+extern crate panic_halt as _;
+
+use cortex_m_rt::entry;
+use nrf52833_hal::{
```
mgeisler, line 22: You could save a bit of vertical space by importing the items over multiple lines like this:

```rust
use nrf52833_hal::gpio::{p0, Level};
use nrf52833_hal::pac::Peripherals;
use nrf52833_hal::prelude::*;
```

src/bare-metal/microcontrollers/examples/src/bin/hal.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    // Configure GPIO0 pins 21 and 28 as push-pull outputs.
+    let mut col1 = gpio0.p0_28.into_push_pull_output(Level::High);
+    let mut row1 = gpio0.p0_21.into_push_pull_output(Level::Low);
+
+    // Set pin 28 high and pin 21 low to turn the LED on.
+    col1.set_low().unwrap();
+    row1.set_high().unwrap();
+
+    loop {}
```
mgeisler, line 43: What is the behavior if we don't loop here? Perhaps a slide on the lifecycle of a program running on a microcontroller would be good?

src/exercises/bare-metal/compass.md
```diff
@@ -0,0 +1,19 @@
+# Compass
+
+We will read the temperature from an I2C compass, and log the readings to a serial port.
+
+Hint: check the documentation for the [`lsm303agr`](https://docs.rs/lsm303agr/latest/lsm303agr/) and
+[`microbit-v2`](https://docs.rs/microbit-v2/latest/microbit/) crates.
+
+If you have time, try displaying it on the LEDs somehow too, or use the buttons somehow.
```
mgeisler, line 8:
```suggestion
If you have time, try making the LED in the direction of North light up.
```

src/bare-metal/aps/mmio.md
```diff
@@ -0,0 +1,17 @@
+# Volatile memory access for MMIO
+
+ * Use `pointer::read_volatile` and `pointer::write_volatile`.
```
mgeisler, line 3: It seems this slide comes a bit too late? There was already talk about MMIO earlier on the "Raw MMIO" slide.

src/bare-metal/aps/examples/src/uart.rs
```diff
@@ -0,0 +1,64 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// ANCHOR: Example
+use core::ptr::write_volatile;
+
+/// Minimal driver for an 8250 UART. This only implements enough to work with the emulated 8250
```
mgeisler, line 18: I suggest wrapping the comments at less than 80 characters to avoid clipping:

image

The `rustfmt.toml` file in the repository has `max_width = 90` but that might even be too generous (and `rustfmt` doesn't wrap comments out of the box).

src/bare-metal/aps/uart.md
```diff
@@ -0,0 +1,5 @@
+# Let's write a UART driver
```
mgeisler, line 1: The flow is a bit abrupt here: I suggest adding a slide first with this title. The slide would explain what UART is and why we should care :slightly_smiling_face:

src/bare-metal/alloc-example/src/main.rs
qwandor: The way we are using a static to reserve space for the heap here is a bit of a hack. Creating a slice to memory which is not allocated would be unsound. A raw slice would be valid I guess, you could send a PR to the crate to take that if you really want.

src/bare-metal/alloc-example/src/main.rs
qwandor: True, but that's not how rustfmt formats it.

src/bare-metal/alloc.md
qwandor: The difference is that you can use the `alloc` crate. The `no_std.md` page has an overview of what that includes. It doesn't include all heap allocating types; `HashMap` is only in `std` for example, but it includes most of them.

src/bare-metal/alloc.md
qwandor: Done.

src/welcome-bare-metal.md
qwandor: Added a note below.

src/welcome-bare-metal.md
qwandor: Added some more details. I do expect people taking this course to know something about embedded development already though.

src/bare-metal/microcontrollers/examples/src/bin/hal.rs
qwandor: The `cortex_m_rt::entry` macro requires that the function have type `fn() -> !`, because returning to the reset handler doesn't make sense.

src/bare-metal/microcontrollers/mmio.md
qwandor: Added a note at the top of the page.

src/bare-metal/microcontrollers/examples/src/bin/mmio.rs
qwandor: Bit 2 specifies whether the pull-up resistor is disabled (0) or enabled (1). It's good practice to be explicit about each setting we are configuring. Usually I'd write out the full set of constants for all the enum values, but to save space here I've only included the ones I actually use.

src/bare-metal/microcontrollers/examples/src/bin/mmio.rs
qwandor: Yes, the typical C code would look pretty similar, maybe with some macros to wrap some things. The point of this series of pages is to show how the progression through levels of abstraction and encapsulation from raw MMIO to a board support crate. Each of the 4 examples is doing the same thing, with less and less code and easier to read because the crates we use provide a higher level of abstraction.

src/bare-metal/microcontrollers/examples/src/bin/hal.rs
qwandor: I guess, but it fits on screen fine as it is and is easier to read.

src/exercises/bare-metal/compass.md
qwandor: I'd like to leave it up to the student to play around and see how they'd like to use the LEDs and buttons here, rather than being too specific, seeing as it's an extension anyway.

src/bare-metal/microcontrollers/examples/src/bin/mmio.rs
qwandor: Wrapped at 80 columns.

qwandor:
I've removed the AP section from this PR for now, and will send a separate PR for that. I've responded to your comments on the microcontroller section, hopefully we can get it merged first.

src/bare-metal/aps/mmio.md
qwandor: The page before was more in the context of building up levels of abstraction for accessing microcontroller peripherals. This one is setting the scene for talking about driver abstractions for more general-purpose systems.

src/bare-metal/aps/examples/src/uart.rs
qwandor: Done (in #482).

src/bare-metal/aps/uart.md
qwandor: We don't particularly care, it's just a simple example to start with.

src/exercises/bare-metal/compass.md
mgeisler: Why make it up to the students when we can give them a concrete task? In my experience, people are not super good at coming up with these ideas by themselves :smile: People will have just seen the board and will only have a tiny bit of context. For that reason, I try to avoid being vague in the suggestions.

src/bare-metal/microcontrollers/examples/src/bin/hal.rs
mgeisler: I have consistently avoided nested imports in the rest of the class and I would like you to follow the same style in the new code. I do that because I personally find such imports very annoying to read (they require me to build kind of mental stack when reading vs having a flat list of imports with helpful repetition).

As for fitting on the screen fine, I see it take up the full vertical height after you scroll it into view. So yeah, it fits, but only barely :smile: 

image

Without scrolling, this is what I see:

image

I would suggest removing things like `// GPIO port 0.` and the blank line above it.

src/bare-metal/alloc-example/src/main.rs
mgeisler: Okay, fair enough! :smile:

src/bare-metal/alloc-example/src/main.rs
mgeisler: Okay, thanks for explaining.

src/bare-metal/alloc.md
mgeisler: Okay, thanks.

src/bare-metal/aps.md
```diff
@@ -0,0 +1,3 @@
+# Application processors
+
+That's enough Cortex-M, let's try a Cortex-A.
```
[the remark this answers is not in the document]
qwandor: Done in #482.

src/bare-metal/microcontrollers/examples/src/bin/hal.rs
qwandor: Do you have a formatter rule for your preferred style? It seems like you're nesting some imports but not others. I prefer to nest as much as possible to save space and make it easier to read, but I'm happy to consistently not nest at all. Going half way and having some things nested but not others seems inconsistent and annoying.

FWIW this page fits on screen fine for me without scrolling, but some of the earlier ones (e.g. the Raw MMIO and PAC examples) don't. I don't think there's any good way around that, some of these examples just are quite long, even being as minimal as possible. I'll see how it goes when I actually teach the course but I don't think having to scroll a bit is going to be a big deal.

mgeisler: > Do you have a formatter rule for your preferred style?

I would like to think that I use [`import_granularity: "Module"`](https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=modul#Module%5C%3A) for the code snippets. Within that style, I prefer to import modules over individual items when the module name is short. I know it's not a hard rule :smile: 

> FWIW this page fits on screen fine for me without scrolling

So, I suppose the aspect ratio helper is affected by zoom levels...? I guess I should have know that :smile: 

> I'll see how it goes when I actually teach the course but I don't think having to scroll a bit is going to be a big deal.

Yeah, it's near impossible to keep everything short when you want to show real world examples. Scrolling being a problem was given to me as direct feedback in Munich and that's why I'm trying to become better at avoiding it.

src/exercises/bare-metal/compass.md
mgeisler: Okay, the first part is concrete for sure. But everything we write down becomes instructions for someone — either in your class or on the Internet. So that's why I'm suggesting to avoid vague language where possible.

It's not a big deal, I just thought it was easy to make the extra task concrete as well. My thinking is that the focus of the class is on using the microcontroller — finding things to do with the microcontroller should be our task and we should not leave it to the reader.

src/bare-metal/alloc-example/src/main.rs
qwandor: I'd rather be overly clear than not clear enough, and there's enough space on the slide.

src/welcome-bare-metal.md
```diff
@@ -0,0 +1,32 @@
...
+For the microcontroller part of the course we will use the [BBC micro:bit](https://microbit.org/) v2
+as an example. It's a [development board](https://tech.microbit.org/hardware/) based on the Nordic
+nRF51822 microcontroller with some LEDs and buttons, an I2C-connected accelerometer and compass, and
+an on-board SWD debugger.
+
+To get started, install some tools we'll need later:
+
+```bash
+sudo apt install gdb-multiarch picocom
+rustup update
```
mgeisler, line 24: I think this line should go: it's not our job to tell people to update `rustup` (just like we don't have an `apt update` above:

```suggestion
```

src/welcome-bare-metal.md
```diff
@@ -0,0 +1,32 @@
...
+
+For the microcontroller part of the course we will use the [BBC micro:bit](https://microbit.org/) v2
+as an example. It's a [development board](https://tech.microbit.org/hardware/) based on the Nordic
+nRF51822 microcontroller with some LEDs and buttons, an I2C-connected accelerometer and compass, and
+an on-board SWD debugger.
+
+To get started, install some tools we'll need later:
+
+```bash
+sudo apt install gdb-multiarch picocom
```
mgeisler, line 23: I needed `libudev-dev` as well:

```suggestion
sudo apt install gdb-multiarch picocom libudev-dev
```

src/welcome-bare-metal.md
```diff
@@ -0,0 +1,32 @@
...
+rustup update
+rustup target add thumbv7em-none-eabihf
+cargo install cargo-binutils cargo-embed
+echo 'SUBSYSTEM=="usb", ATTR{idVendor}=="0d28", MODE="0664", GROUP="plugdev"' | sudo tee /etc/udev/rules.d/50-microbit.rules
```
mgeisler, line 27: I know the line is not meant to be read carefully, but it's too long to see without scrolling.

Many people will probably have questions about what this does, especially since you're asking them to execute it as root.

src/bare-metal/no_std.md
```diff
@@ -0,0 +1,65 @@
+# `no_std`
+
+<table>
```
mgeisler, line 3: Have you considered ditching the table and instead use three slides, one for each crate?

I used to have a table myself for some other part of the course, but a linear list with headings proved to be much simpler to read at a glance. The advantage of such a list is that you can slowly scroll up and down:

image

src/welcome-bare-metal.md
qwandor: That's weird, what happens if you don't install it?

src/bare-metal/no_std.md
```diff
@@ -0,0 +1,65 @@
...
+* `println!`, `Read`, `Write`, `Stdin`, `Stdout` and the rest of `io`
+* `Path`, `OsString`
+* `net`
+* `Command`, `Child`, `ExitCode`
+* `spawn`, `sleep` and the rest of `thread`
+* `SystemTime`, `Instant`
+
+</td>
+</tr>
+</table>
```
mgeisler, line 58: The hand-written HTML does not look very nice, it bumps into the speaker notes below:

image

I instead format it as a simple list, possibly split across three pages like I suggested above:

```suggestion

## `core`

* `&[T]`, `&str`, `CStr`
* `NonZeroU8`...
* `Option`, `Result`
* `Display`, `Debug`, `write!`...
* `Iterator`
* `panic!`, `assert_eq!`...
* `NonNull` and pointer-related functions
* `Future` and `async`/`await`
* `fence`, `AtomicBool`, `AtomicPtr`, `AtomicU32`...
* `Duration`

## `alloc`

* `Box`, `Cow`, `Arc`, `Rc`
* `Vec`, `BinaryHeap`, `BtreeMap`, `LinkedList`, `VecDeque`
* `String`, `CString`, `format!`

## `std`

* `Error`
* `HashMap`
* `Mutex`, `Condvar`, `Barrier`, `Once`, `RwLock`, `mpsc`
* `File` and the rest of `fs`
* `println!`, `Read`, `Write`, `Stdin`, `Stdout` and the rest of `io`
* `Path`, `OsString`
* `net`
* `Command`, `Child`, `ExitCode`
* `spawn`, `sleep` and the rest of `thread`
* `SystemTime`, `Instant`
```

src/welcome-bare-metal.md
qwandor: Updating the toolchain is the important point here. Some of the examples depend on a recent rustc version, and I'd rather not have to go around helping people debug confusing error messages due to trying to build with an out of date compiler.

src/bare-metal/no_std.md
qwandor: I had a list before, but I decided to go with a table so that I can have everything on one screen and discuss the distinctions without having to scroll or switch pages back and forth.

src/welcome-bare-metal.md
qwandor: Well okay, they are welcome to ask questions.

src/bare-metal/alloc-example/src/main.rs
```diff
@@ -0,0 +1,42 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// ANCHOR: Alloc
+#![no_main]
+#![no_std]
+
+extern crate alloc;
+extern crate panic_halt as _;
```
mgeisler, line 20: Same question, any reason do this import?

src/bare-metal/alloc-example/src/main.rs
mgeisler: This what I see:

image

How does it look on your side?

src/bare-metal/alloc-example/src/main.rs
qwandor: It's to provide a panic handler, rather than writing one ourselves and making the example longer.

src/bare-metal/alloc-example/src/main.rs
qwandor: image

src/bare-metal/microcontrollers/debugging.md
```diff
@@ -0,0 +1,38 @@
...
+
+In GDB, try running:
+
+```gdb
+b src/bin/board_support.rs:28
+b src/bin/board_support.rs:29
+b src/bin/board_support.rs:31
+c
+c
+c
```
mgeisler, line 35: Could you describe what the expected output is? I tried it and got

```
% gdb-multiarch target/thumbv7em-none-eabihf/debug/board_support --eval-command="target remote :1337"

GNU gdb (Debian 12.1-4) 12.1
Copyright (C) 2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
target/thumbv7em-none-eabihf/debug/board_support: No such file or directory.
Remote debugging using :1337
warning: No executable has been specified and target does not support
determining executable automatically.  Try using the "file" command.
0x00000100 in ?? ()
(gdb) b src/bin/board_support.rs:28
No symbol table is loaded.  Use the "file" command.
Make breakpoint pending on future shared library load? (y or [n]) y
Breakpoint 1 (src/bin/board_support.rs:28) pending.
(gdb) b src/bin/board_support.rs:29
No symbol table is loaded.  Use the "file" command.
Make breakpoint pending on future shared library load? (y or [n]) y
Breakpoint 2 (src/bin/board_support.rs:29) pending.
(gdb) b src/bin/board_support.rs:31
No symbol table is loaded.  Use the "file" command.
Make breakpoint pending on future shared library load? (y or [n]) y
Breakpoint 3 (src/bin/board_support.rs:31) pending.
(gdb) c
Continuing.
c
c
c


(gdb) 
```

The light turned on, but that's it :smile: The many warnings about the lack of a symbol table were a bit confusing.

src/SUMMARY.md
```diff
@@ -233,6 +233,40 @@
...
+  - [Board support crates](bare-metal/microcontrollers/board-support.md)
+  - [The type state pattern](bare-metal/microcontrollers/type-state.md)
+  - [embedded-hal](bare-metal/microcontrollers/embedded-hal.md)
+  - [probe-rs, cargo-embed](bare-metal/microcontrollers/probe-rs.md)
+    - [Debugging](bare-metal/microcontrollers/debugging.md)
+  - [Other projects](bare-metal/microcontrollers/other-projects.md)
+- [Exercises](exercises/bare-metal/morning.md)
+  - [Compass](exercises/bare-metal/compass.md)
+
+# Day 5A: Afternoon
```
mgeisler, line 255:
```suggestion
# Bare Metal Rust: Afternoon
```

src/SUMMARY.md
```diff
@@ -245,3 +279,4 @@
   - [Day 3 Morning](exercises/day-3/solutions-morning.md)
   - [Day 3 Afternoon](exercises/day-3/solutions-afternoon.md)
   - [Day 4 Morning](exercises/day-4/solutions-morning.md)
+  - [Day 5A Morning](exercises/bare-metal/solutions-morning.md)
```
mgeisler, line 282:
```suggestion
  - [Bare Metal Rust Morning](exercises/bare-metal/solutions-morning.md)
```

src/exercises/bare-metal/solutions-morning.md
```diff
@@ -0,0 +1,9 @@
+# Day 5A Morning Exercise
```
mgeisler, line 1:
```suggestion
# Bare Metal Rust Morning Exercise
```

.github/workflows/build.yml
```diff
@@ -25,6 +25,16 @@ jobs:
       - name: Test code snippets
         run: mdbook test
 
+  format:
```
mgeisler, line 28: We should take this out into it's own PR, then we could merge it much sooner.

mgeisler: And we would avoid the unrelated changes in this PR :slightly_smiling_face:

src/SUMMARY.md
```diff
@@ -233,6 +233,40 @@
 
 ----
 
+# Day 5A: Morning
```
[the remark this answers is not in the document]
qwandor: Done.

src/SUMMARY.md
qwandor: Done.

src/SUMMARY.md
qwandor: Done.

src/exercises/bare-metal/solutions-morning.md
qwandor: Done.

src/exercises/bare-metal/compass/src/main.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    // Configure serial port.
+    let mut serial = Uarte::new(
+        board.UARTE0,
+        board.uart.into(),
+        Parity::EXCLUDED,
+        Baudrate::BAUD115200,
+    );
+
+    // Set up the I2C controller and IMU.
```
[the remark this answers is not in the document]
qwandor: Done.

src/exercises/bare-metal/compass.md
```diff
@@ -0,0 +1,27 @@
+# Compass
+
+We will read the temperature from an I2C compass, and log the readings to a serial port.
+
+Hint: check the documentation for the [`lsm303agr`](https://docs.rs/lsm303agr/latest/lsm303agr/) and
```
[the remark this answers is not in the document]
qwandor: Thanks for the feedback! I've added some more hints to the exercise. Part of the issue is that I2C is also called TWI (for historical trademark-related reasons I think), which is not obvious if you haven't come across it before, so I've added a note mentioning that.

COMP is a comparator, it's not relevant here. You can read more about the peripherals in the [nRF 52833 datasheet](https://infocenter.nordicsemi.com/pdf/nRF52833_PS_v1.5.pdf), but it shouldn't be necessary for the exercise.

I'd like people to get practice looking through the documentation so I don't want to give them the answer too directly. There are two ways I'd expect someone to figure out what they need to do:
1. Find `Board::i2c_internal` (as you did), see that that can be converted to either `twim::Pins` or `twis::Pins` (i.e. a TWI master or slave, now that they have the hint that I2C and TWI are the same thing), look up one level to the module to find `twim::Twim`, and see that that can be constructed by `Twim::new` with things that they have.
2. See that `Lsm303agr::new_with_i2c` needs something implementing `embedded_hal::blocking::i2c::WriteRead`, look in https://docs.rs/microbit-v2/latest/microbit/hal/prelude/trait._embedded_hal_blocking_i2c_WriteRead.html#implementors to see that `Twim` does so. See what `Twim::new` needs and figure out how to get them.

src/welcome-bare-metal.md
mgeisler: I cannot install `cargo-embed` without it. It fails with

```
error: failed to run custom build command for `hidapi v2.2.0`

Caused by:
  process didn't exit successfully: `/tmp/cargo-installegDUk8/release/build/hidapi-1b03c572ab5df1f8/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-env-changed=LIBUDEV_NO_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG
  cargo:rerun-if-env-changed=LIBUDEV_STATIC
  cargo:rerun-if-env-changed=LIBUDEV_DYNAMIC
  cargo:rerun-if-env-changed=PKG_CONFIG_ALL_STATIC
  cargo:rerun-if-env-changed=PKG_CONFIG_ALL_DYNAMIC
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG_PATH
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH
  cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG_LIBDIR
  cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR
  cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG_SYSROOT_DIR
  cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR

  --- stderr
  thread 'main' panicked at 'Unable to find libudev: `"pkg-config" "--libs" "--cflags" "libudev"` did not exit successfully: exit status: 1
  error: could not find system library 'libudev' required by the 'hidapi' crate

  --- stderr
  Package libudev was not found in the pkg-config search path.
  Perhaps you should add the directory containing `libudev.pc'
  to the PKG_CONFIG_PATH environment variable
  No package 'libudev' found
  ', /usr/local/google/home/mgeisler/.cargo/registry/src/github.com-1ecc6299db9ec823/hidapi-2.2.0/build.rs:57:54
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: failed to compile `cargo-embed v0.16.0`, intermediate artifacts can be found at `/tmp/cargo-installegDUk8`
```

src/welcome-bare-metal.md
mgeisler: Yes, I would like to move this command to it's own code block, with some text in front of it which explains why we're running commands as root. You'll want to explain to people what the command does and why it's needed -- that should be helpful for people who don't use `udev` or who are on a system where the config files are somewhere else.

The command should then be nicely formatted so that they're readable without overflow.

src/welcome-bare-metal.md
qwandor: Aha! That makes sense then, added.

src/welcome-bare-metal.md
qwandor: Done.

src/bare-metal/alloc-example/src/main.rs
mgeisler: > It's to provide a panic handler, rather than writing one ourselves and making the example longer.

Right, and what you're implicitly saying is that even a "no-op import" like this has an effect? I'm used to only using imports to bring symbols into scope — but here the import also does something more behind the scenes?

src/bare-metal/no_std.md
qwandor: This is my attempt at a top-level summary. The module names aren't necessarily entirely obvious, and often there are modules with the same name split across `core`/`alloc`/`std` with different contents, so I think it's more useful to give examples of commonly used items from the modules than the modules themselves. My plan is to spend a while on this page talking through the distinctions and the reasons for them, and having to keep switching back and forth would be distracting. I don't expect students to read everything, but being able to spot a few things they have used before and see where they fall is useful.

src/bare-metal/microcontrollers/debugging.md
mgeisler: Okay, it works better when I run it in the same directory:

```
Reading symbols from target/thumbv7em-none-eabihf/debug/board_support...
Remote debugging using :1337
microbit_common::display::nonblocking::control::{impl#0}::display_row_leds (
    self=<error reading variable: Cannot access memory at address 0x2002006c>, 
    row=<error reading variable: Cannot access memory at address 0x20020070>, 
    cols=<error reading variable: Cannot access memory at address 0x20020074>)
    at src/display/nonblocking/control.rs:134
134	            p0.outset.write(|w| w.bits(rows_to_set | cols_to_set));
(gdb) b src/bin/board_support.rs:28
Breakpoint 1 at 0x18e: file src/bin/board_support.rs, line 29.
Note: automatically using hardware breakpoints for read-only addresses.
(gdb) b src/bin/board_support.rs:29
Note: breakpoint 1 also set at pc 0x18e.
Breakpoint 2 at 0x18e: file src/bin/board_support.rs, line 29.
(gdb) b src/bin/board_support.rs:31
Breakpoint 3 at 0x1b8: file src/bin/board_support.rs, line 32.
(gdb) c
Continuing.

Breakpoint 1, board_support::__cortex_m_rt_main () at src/bin/board_support.rs:29
29	    board.display_pins.col1.set_low().unwrap();
(gdb) c
Continuing.

Program received signal SIGINT, Interrupt.
nrf_hal_common::gpio::p0::{impl#144}::set_low<nrf_hal_common::gpio::PushPull> (self=0x2001fff4) at /usr/local/google/home/mgeisler/.cargo/registry/src/github.com-1ecc6299db9ec823/nrf-hal-common-0.14.1/src/gpio.rs:581
581	                    fn set_low(&mut self) -> Result<(), Self::Error> {
(gdb) c
Continuing.

Breakpoint 3, board_support::__cortex_m_rt_main () at src/bin/board_support.rs:32
32	    loop {}
(gdb) 
```

However, the first two breakpoints seem to be duplicated? The `Program received signal SIGINT, Interrupt.` also looks scary?

src/bare-metal/no_std.md
qwandor: To give some examples of why module names aren't enough: `collections::BinaryHeap` is in `alloc`, but `collections::HashMap` is only in `std`. `fmt::write!` is in `core`, but `fmt::format!` is in `alloc`.

mgeisler: > My plan is to spend a while on this page talking through the distinctions and the reasons for them

Okay, I would love some details about this in the speaker notes so that someone (like me!) can say something smart about the page. Otherwise I would probably go over it in much too little time. I'll of course also take notes when we do a trial run of this new material.

src/exercises/bare-metal/compass.md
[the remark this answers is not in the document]
mgeisler: Okay, thanks for the hints. It'll be very interesting to see how the audience does in the trial run! :smile:

src/bare-metal/alloc-example/src/main.rs
qwandor: Ah, yes, the `extern crate` is necessary for the crate to be linked in, as we are not using any items from it in the code. If we leave this line out then it will fail to compile due to a missing panic handler. I've added a mention of this to the speaker notes.

src/welcome-bare-metal.md
mgeisler: I see, in that case, I would suggest writing

```suggestion
rustup default 1.65.0 // or later
```

to ensure they use a known version. When I first ran these commands, I was on Rust 1.59.0 since I was testing something with that version — all the `rustup update` commands in the world wouldn't have helped me :smile:

mgeisler (approved):
Thanks for all the updates, let's get this in for the trial run on Friday! :tada:

src/bare-metal/microcontrollers/debugging.md
qwandor: Hmm, looks like the line numbers are wrong, I'll send a PR to fix that. I'm not sure what the message about `SIGINT` is from, there's no such thing as `SIGINT` here. I have also seen it halt at lines other than the breakpoints I set, and I'm not sure why. I think it's still worth demonstrating debugging though, even if it's a bit buggy.

PR #321, opened 2023-02-01
Add a cargo workspace containing both crates

qwandor:
This lets everything be tested by running `cargo test` in the root directory, which simplifies CI slightly. I've also added a CI step to run `cargo build`, as some of the exercises weren't building, and then fixed them to build properly.


mgeisler:
This is a nice idea! We should also adjust the Rust caching GitHub action — I believe the `workspaces` key can go now.

qwandor:
Done.

mgeisler (approved):
Very cool!

PR #332, opened 2023-02-03
Remove dangling links in cargo/rust-ecosystem.md.

gendx:
I'm not sure whether they should point to any relevant content, but these footnotes are currently undefined.


mgeisler:
Thanks! The content was moved to the speaker notes in https://github.com/google/comprehensive-rust/commit/c2c3d4892009bf4b0e011fa346a7345f519be93d.

PR #336, opened 2023-02-03
Add a small illustration of bounds checking.

gendx:
Not strictly necessary at this stage of the course, but this page didn't have much content and I found it interesting that the compiler even checks the bounds at compile time when the array length is already known :)

Let me know what you think!

### DIFF
--- a/src/why-rust/runtime.md
+++ b/src/why-rust/runtime.md
@@ -5,6 +5,26 @@ No undefined behavior at runtime:
 * Array access is bounds checked.
 * Integer overflow is defined.
 
+```rust
+fn main() {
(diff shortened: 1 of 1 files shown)


mgeisler:
Hey @gendx, I think I'm not a fan of adding more code right here — we will see the exact same thing in a few slides when we talk about arrays. Keeping the slide simple and focused is more important to me than showing all the things as early as possible.

mgeisler:
I think we can close this — I would really like an actual diagram on the relevant page where we talk about arrays and stack/heap etc, but I think we should keep this slide focused.

PR #338, opened 2023-02-03
Add a solution for day 1's bonus question with generics.

gendx:
The current solution mentions that it doesn't work so well. It's not quite true as one can use more advanced concepts such as the [AsRef](https://doc.rust-lang.org/std/convert/trait.AsRef.html) trait to generalize over an owned vs. borrowed slice.

It's probably too early to introduce all of these concepts in the class, but I think the solution should mention that it's possible to do. In the end, showing this advanced solution or not is a matter of balancing between giving the following impressions to the students:

1. "Rust is complicated", by showing advanced syntax with generics and trait bounds before they are explained in the course,
2. "Rust is not powerful enough", by (incorrectly) stating that it's not possible to generalize between `&[&[i32]]` and `Vec<Vec<i32>>`.

IMO, generics are a cornerstone of the Rust language, so we shouldn't leave a false impression about 2. And given that we are in the "bonus" section, it's fine to mention some advanced concepts that will be covered later in the course.

### DIFF
--- a/src/exercises/day-1/solutions-morning.md
+++ b/src/exercises/day-1/solutions-morning.md
@@ -9,8 +9,39 @@
 ```
 ### Bonus question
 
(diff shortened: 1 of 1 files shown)


mgeisler:
Wow, that's cool! I did not think of this :smile:

PR #341, opened 2023-02-03
Add highly-unsafe speaker notes to inspect the memory layout of string.

gendx:
The slide shows the assumed memory layout of a String, so we can use some `unsafe` code to check that (and see that the capacity is actually laid out first - although Rust provides no guarantees about that).

### DIFF
--- a/src/memory-management/stack.md
+++ b/src/memory-management/stack.md
@@ -29,6 +29,23 @@ fn main() {
 
 * If students ask about it, you can mention that the underlying memory is heap allocated using the [System Allocator] and custom allocators can be implemented using the [Allocator API]
 
+* We can inspect the memory layout with `unsafe` code. However, you should point out that this is rightfully unsafe!
+
+```rust,editable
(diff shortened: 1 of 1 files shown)


mgeisler (approved):
It's a bit of a wild code sample to show here (as you point out) 😅

Since it's only going into the speaker notes, I guess er can add it.

gendx:
I've showed it in the class this week and the audience liked it (especially the "transmute" name :p)!

We ended up following up on what the layout of `(i8, i64, i8)` might look like in practice (with and without `#[repr(c)]`) - how the compiler may optimize the layout, how the unused bytes can contain other things, how you shouldn't do it at home, etc.

Overall, I think these kinds of sneak peeks can keep the classroom engaged and spawn other questions :)

mgeisler:
Nice! It's also been my impression that people appreciate these detours into more advanced topics. It shows them that the class is there to answer their questions, whatever they might be 🙂

PR #344, opened 2023-02-03
Make sure exercise 1.afternoon can be (almost) dropped in as is in the Playground

gendx:
One caveat: the exercise required changing `len(self)` to `len(&self)` and I'm not sure how best to encode that with the anchors.


mgeisler:
Thanks @gendx! The exercise got reworked after your PR so I'll close this. #1073 is further reworking the exercises of the course.

PR #346, opened 2023-02-03
Add speaker notes for pattern-matching/destructuring-arrays.md

gendx:
I found it interesting to see that array patterns also work for slices of unknown length.


mgeisler (approved):
Nice, those are great notes and something which I've found myself hand-coding in class!

qwandor (approved):

PR #347, opened 2023-02-03
Remove unused main() function in day-2.morning exercise.

gendx:
This exercise is about making the unit tests pass, so we can remove the `main()` function, which is empty and unused. Otherwise the Rust playground complains about dead code.

### DIFF
--- a/src/exercises/day-2/points-polygons.md
+++ b/src/exercises/day-2/points-polygons.md
@@ -35,9 +35,6 @@ tests pass:
 {{#include points-polygons.rs:Shape}}
 
 {{#include points-polygons.rs:unit-tests}}
-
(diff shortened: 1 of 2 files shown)


gendx:
Looks like the `cargo build` check is not happy about that.

mgeisler:
> Looks like the `cargo build` check is not happy about that.

Yeah, it's probably listed as a binary in the `Cargo.toml` file and so the build breaks without a `main`.

mgeisler:
We could unlist it there if it's covered equivalently by `mdbook test`.

mgeisler:
This looks a bit stale — we can revisit it after #1073 is merged.

PR #348, opened 2023-02-03
Add a slide with more advanced visibility rules.

gendx:
I think it's useful to introduce `pub(crate)`. The others are probably not super (pun intended) relevant.

I've shifted the "Paths" slides before "Visibility", given that showcasing the advanced visibility rules depends on paths.

Let me know what you think!


mgeisler:
> I think it's useful to introduce `pub(crate)`.

Yeah, for sure. I normally mention it and sometimes hand-code it for the class.

> I've shifted the "Paths" slides before "Visibility", given that showcasing the advanced visibility rules depends on paths.

Sure, I think it makes sense to introduce paths earlier :+1: 

I'm not sure we should have a new page for the advanced visibility keywords. The course doesn't focus much on modules since most things are based on the inline playgrounds — so we can get away with not talking much about visibility and such. I normally just explain that Rust has a module system and that it works similarly to other languages. People seem happy to believe that :smile: 

However, having an exercise or two which lets the students work with multiple files (modules) would be great. The focus on inline playgrounds is great to get started, but by Day 2 or 3, we should strongly nudge people towards using Cargo on the command line for their exercises.

Could I get you to to move the discussion of `pub(crate)` to the speaker notes instead of adding a new page?

mgeisler (changes requested):
I think we can make do with an update to the speaker notes. Test it out when you teach the course and let me know :smile:

gendx:
Migrated the changes to the speaker notes as requested.

mgeisler (approved):
Thanks for moving the details.

PR #352, opened 2023-02-06
Setup redirects after #120 and #177

mgeisler:
This makes `mdbook` output a simple redirect at the location of the old pages. I’ll try to add such pages when we shuffle around our pages to make sure external links stay valid.

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -42,3 +42,7 @@ level = 0
 
 [output.html.playground]
 editable = true
+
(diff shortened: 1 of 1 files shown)


mgeisler:
I tested this on a local checkout — the redirects are a single page which will auto-load the new page. The new page path is relative to the generated page.

qwandor (approved):

Issue #317, opened 2023-02-01
Brazilian Portuguese translation


mgeisler:
I've created Google Cloud Translate file to get started: [pt_BR.zip](https://github.com/google/comprehensive-rust/files/10555855/pt_BR.zip).

We might have to update our publishing machinery to handle the `xx_YY` form of the locale, but we'll deal with this when we get there :-)

mgeisler:
The Rust book is translated here, I hope it can be a good source of inspiration: https://github.com/rust-br/rust-book-pt-br.

rastringer:
Brilliant, thanks for this it's just what I needed to check some of the vocab. About 15% done.

mgeisler:
#340 translated the first part of the course — I'll reopen this issue so we can track the translation of the full course. After the first pass, there will be some ongoing work to keep the translation up to date. I don't think we need an issue to track that, we just need PRs once in a while

mgeisler:
The translation is now published! Thanks a lot for the hard work here!

mgeisler:
Hi @rastringer, @hugojacob, @joaovicmendes, @henrif75!

It would be great if you could refresh the translation with msgmerge after installing the new version of mdbook-i18n-helpers. After the normalization in #1104, you need a refresh to get new strings incorporated and to mark old strings as obsolete.

rastringer:
Thanks @mgeisler, PR submitted

mgeisler:
> Thanks @mgeisler, PR submitted

Thanks, looks good! You can decide if you want to take out unnecessary fuzzy entries in the PR directly or in a follow-up.

henrif75:
I'm working to clean up the fuzzy entries.

rastringer:
Thanks Henri! Would it help to merge this initial PR or would you like to
do it all at once?

henrif75:
I'll do it in another PR, thanks!

mgeisler:
Hi all! I wanted to let you know about #1460. This is an update to mdbook-i18n-helpers which will remove ~6k lines from your PO file. It changes the way code blocks are processed: we now extract strings and comments directly.

You need to run `mdbook-i18n-normalize` on your PO file to preserve the translations found here: https://google.github.io/comprehensive-rust/pt-BR/hello-world/small-example.html (and any other places where you have translated comments).

You can do this before/after refreshing the translation. See #1461 for an example PR which shows the state after normalization.

henrif75:
[...]

PR #355, opened 2023-02-06
Additional work on Brazilian Portuguese translation

rastringer:
Code snippets days 1-4 are fixed, with comments in Br. Portuguese where appropriate. There will be the odd dangling error / mistranslation, which will be sorted as they are found. 
Generally all four days of the course are readable now to a decent standard. We will keep improving the grammar / vocab etc. 
Part of continuing work on issue #317.


rastringer:
The tests don't like the the `C` code in section 33.5.1.1 Calling C with Bindgen. Is there a way around this please?
Run msgfmt --statistics -o /dev/null po/pt_BR.po
po/pt_BR.po:17662: keyword "typedef" unknown
po/pt_BR.po:17662:8: syntax error
po/pt_BR.po:17663: end-of-line within string
msgfmt: found 3 fatal errors
Error: Process completed with exit code 1.

rastringer:
Please disregard comment re `C` above, realized there was a quotation marks missing! Fixed and tests passing.

rastringer:
Excellent, thank you I'll download that one for the next endeavour.

mgeisler:
> Excellent, thank you I'll download that one for the next endeavour.

Thanks, I see you got the strings normalized now, looks good :+1: 

You just let me know when you think we should merge this into the `main` branch.

> Generally all four days of the course are readable now to a decent standard.

It sounds like the translation is in an okay condition now? You and Hugo can still look it over after we merge it. I'll like to publish it next week if you two are comfortable with that?

rastringer:
Yes definitely, I think we should merge then will do another review / check with Hugo before publication. Thanks for all your comms on this.

mgeisler (approved):
Thanks Robin!

PR #361, opened 2023-02-07
Update points-polygons.md

cchiw:
#274


qwandor (approved):

src/exercises/day-2/points-polygons.md
```diff
@@ -43,11 +43,11 @@ fn main() {}
 <details>
 
 Since the method signatures are missing from the problem statements, the key part
-of the exercise is to specify those correctly.
+of the exercise is to specify those correctly. Do not modify any of the tests.
```
mgeisler, line 46: People are actually free to modify things... we should write something like "You don't have to modify the tests" instead.

PR #364, opened 2023-02-07
Update option-result.md

cchiw:
#274

### DIFF
--- a/src/std/option-result.md
+++ b/src/std/option-result.md
@@ -21,5 +21,5 @@ fn main() {
 * `binary_search` returns `Result<usize, usize>`.
   * If found, `Result::Ok` holds the index where the element is found.
   * Otherwise, `Result::Err` contains the index where such an element should be inserted.
-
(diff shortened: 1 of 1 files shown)


src/std/option-result.md
```diff
@@ -21,5 +21,5 @@ fn main() {
 * `binary_search` returns `Result<usize, usize>`.
   * If found, `Result::Ok` holds the index where the element is found.
   * Otherwise, `Result::Err` contains the index where such an element should be inserted.
-
+* `Option` and `Result` are `enums`, which means we can use enum functions on them. Ex. `first.unwrap_or(0)`.
```
qwandor, line 24: What is an "enum function"? `unwrap_or` is just a method defined on `Result` and `Option`.

cchiw: OK,  Let me know if this makes sense.

qwandor: I'm still not sure what you mean by an "enum method". What do these methods have to do with the fact that `Option` and `Result` are enums?

PR #367, opened 2023-02-07
Update match-expressions.md

cchiw:
#274

### DIFF
--- a/src/control-flow/match-expressions.md
+++ b/src/control-flow/match-expressions.md
@@ -21,3 +21,12 @@ expression of the block, if any. In the example above, the type is `()`.
 
 See [pattern matching](../pattern-matching.md) for more details on patterns in
 Rust.
+
+<details>
+    
+* Save the match expression to a variable and print it out.
+* Remove `.as_deref()` and explain the error.
+    * `std::env::Agrs().next` returns an option borrowed string. 
+    * `as_deref()` takes an `option` type and calls the deref on the inside value in the case of a `Some` variant.
+    * By using `as_deref()`, we are able to match against a String literal.
+</details>
--- a/src/control-flow/match-expressions.md
+++ b/src/control-flow/match-expressions.md
@@ -26,7 +26,7 @@ Rust.
     
 * Save the match expression to a variable and print it out.
 * Remove `.as_deref()` and explain the error.
-    * `std::env::Agrs().next` returns an option borrowed string. 
-    * `as_deref()` takes an `option` type and calls the deref on the inside value in the case of a `Some` variant.
-    * By using `as_deref()`, we are able to match against a String literal.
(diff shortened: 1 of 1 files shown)


src/control-flow/match-expressions.md
```diff
@@ -21,3 +21,12 @@ expression of the block, if any. In the example above, the type is `()`.
 
 See [pattern matching](../pattern-matching.md) for more details on patterns in
 Rust.
+
+<details>
+    
+* Save the match expression to a variable and print it out.
+* Remove `.as_deref()` and explain the error.
+    * `std::env::Agrs().next` returns an option borrowed string.
```
mgeisler, line 29:
```suggestion
    * `std::env::Agrs().next` returns an `Option<&String>`, but we cannot match against `String`.
```

src/control-flow/match-expressions.md
```diff
@@ -21,3 +21,12 @@ expression of the block, if any. In the example above, the type is `()`.
 
 See [pattern matching](../pattern-matching.md) for more details on patterns in
 Rust.
+
+<details>
+    
+* Save the match expression to a variable and print it out.
+* Remove `.as_deref()` and explain the error.
+    * `std::env::Agrs().next` returns an option borrowed string. 
+    * `as_deref()` takes an `option` type and calls the deref on the inside value in the case of a `Some` variant.
```
mgeisler, line 30:
```suggestion
    * `as_deref()` transforms an `Option<T>` to `Option<T::Target>`. In our case, this turns `Option<&String>` into `Option<&str>`.
```

src/control-flow/match-expressions.md
```diff
@@ -21,3 +21,12 @@ expression of the block, if any. In the example above, the type is `()`.
...
+    * By using `as_deref()`, we are able to match against a String literal.
```
mgeisler, line 31:
```suggestion
    * We can now use pattern matching to match against the `&str` inside `Option`.
```

mgeisler (approved):

src/control-flow/match-expressions.md
```diff
@@ -21,3 +21,12 @@ expression of the block, if any. In the example above, the type is `()`.
 
 See [pattern matching](../pattern-matching.md) for more details on patterns in
 Rust.
+
+<details>
+    
+* Save the match expression to a variable and print it out.
+* Remove `.as_deref()` and explain the error.
+    * `std::env::Agrs().next` returns an `Option<&String>`, but we cannot match against `String`.
```
qwandor, line 29: I guess you mean `Args()` and `next()`.

qwandor: Or rather, `args().next()`, case is important.

cchiw: Follow-up in #400

PR #368, opened 2023-02-07
Add Speaker notes string.md

cchiw:
#274


src/std/string.md
```diff
@@ -27,8 +27,14 @@ fn main() {
 
 <details>
 
+* `new` heap allocated buffer. `StringL::with capacity` is used when you know how much you want to push on
 * `len` returns the size of the `String` in bytes, not its length in characters.
 * `chars` returns an iterator over the actual characters.
-* `String` implements `Deref<Target = str>` which transparently gives it access to `str`'s methods.
+*  When people refer to strings they could either be talking about &str or String.
```
mgeisler, line 33:
```suggestion
*  When people refer to strings they could either be talking about `&str` or `String`. 
```

src/std/string.md
```diff
@@ -27,8 +27,14 @@ fn main() {
...
+* `new` heap allocated buffer. `StringL::with capacity` is used when you know how much you want to push on
 * `len` returns the size of the `String` in bytes, not its length in characters.
 * `chars` returns an iterator over the actual characters.
-* `String` implements `Deref<Target = str>` which transparently gives it access to `str`'s methods.
+*  When people refer to strings they could either be talking about &str or String. 
+* Implementing Deref (trait), gives the compiler ability to take a value of any type, call the deref method, and know how to dereference it  
+    * `String` implements `Deref<Target = str>` which transparently gives it access to `str`'s methods.
+    * Write and compare `let s3 = s1.deref();` and  `let s3 = &*s1`;.
+* String isimplemented as a wrapper around a vector of bytes, many of the operations you see supported on vectors are also supported on String, but with some extra guarantees.
+* Show the danger of indexing Strings by (1) adding a unicode character to an above string (Ex. `Hβello`) and slicing it at various locations and (2) unwraping the characters `.chars.nth(i).unwrap()` where `i` is in-bound and out-of-bounds given the number of chars.
```
mgeisler, line 38: I think this is outdated now: we already have a Unicode flag (🇨🇭) above.

src/std/string.md
```diff
@@ -27,8 +27,14 @@ fn main() {
...
 
+* `new` heap allocated buffer. `StringL::with capacity` is used when you know how much you want to push on
 * `len` returns the size of the `String` in bytes, not its length in characters.
 * `chars` returns an iterator over the actual characters.
-* `String` implements `Deref<Target = str>` which transparently gives it access to `str`'s methods.
+*  When people refer to strings they could either be talking about &str or String. 
+* Implementing Deref (trait), gives the compiler ability to take a value of any type, call the deref method, and know how to dereference it  
+    * `String` implements `Deref<Target = str>` which transparently gives it access to `str`'s methods.
+    * Write and compare `let s3 = s1.deref();` and  `let s3 = &*s1`;.
+* String isimplemented as a wrapper around a vector of bytes, many of the operations you see supported on vectors are also supported on String, but with some extra guarantees.
```
mgeisler, line 37:
```suggestion
* `String` is implemented as a wrapper around a vector of bytes, many of the operations you see supported on vectors are also supported on `String`, but with some extra guarantees.
```

src/std/string.md
cchiw: I updated the text. Let me know what you think.

src/std/string.md
```diff
@@ -27,7 +27,14 @@ fn main() {
 
 <details>
 
+* `new()` heap allocated buffer. `StringL::with capacity` is used when you know how much you want to push on
```
mgeisler, line 30:
```suggestion
* `String::new` returns a new empty string, use `String::with capacity` when you know how much data you want to push to the string.
```

src/std/string.md
```diff
@@ -27,7 +27,14 @@ fn main() {
...
 * `len()` returns the size of the `String` in bytes, not its length in characters.
```
mgeisler, line 31:
```suggestion
* `String::len` returns the size of the `String` in bytes (which can be different from its length in characters).
```

src/std/string.md
```diff
@@ -27,7 +27,14 @@ fn main() {
...
 * `chars()` returns an iterator over the actual characters.
+*  When people refer to strings they could either be talking about `&str` or `String`. 
+* Implementing Deref (trait), gives the compiler ability to take a value of any type, call the deref method, and know how to dereference it
```
mgeisler, line 34:
```suggestion
* When a type implements `Deref<Target = T>`, the compiler will let you transparently call methods from `T`.
```

src/std/string.md
```diff
@@ -13,7 +13,7 @@ fn main() {
     s2.push('!');
     println!("s2: len = {}, capacity = {}", s2.len(), s2.capacity());
 
-    let s3 = String::from("🇨🇭");
+    let s3 = String::from("A Flag:🇨🇭");
```
mgeisler, line 16: We need to be able to reason about the string length at a glance in both bytes and characters. So we should not add extra text here:

```suggestion
    let s3 = String::from("🇨🇭");
```

src/std/string.md
```diff
@@ -27,7 +27,14 @@ fn main() {
 
 <details>
 
+* `new()` heap allocated buffer. `StringL::with capacity` is used when you know how much you want to push on
 * `len()` returns the size of the `String` in bytes, not its length in characters.
 * `chars()` returns an iterator over the actual characters.
```
mgeisler, line 32:
```suggestion
* `String::chars` returns an iterator over the actual characters. Note that a `char` can be different from what a human will consider a "character" due to [grapheme clusters](https://docs.rs/unicode-segmentation/latest/unicode_segmentation/struct.Graphemes.html).
```

mgeisler (approved):

PR #370, opened 2023-02-08
Add examples of using the default trait

cchiw:
We had some questions on how to create default values for structs so I added a slide. It uses the continuous example on `Person`. Let me know what you think and if there are other files to edit to add a slide.
#365. #475


mgeisler (changes requested):
Hey @cchiw, we already cover the `Default` trait here: https://google.github.io/comprehensive-rust/traits/default.html. That was added by @igor-petruk a while ago.

I think you have a good point that the instructor should mention `Default` somewhere when talking about structs. So I suggest we add a note about it to the speaker notes instead.

src/structs/default.md
```diff
@@ -0,0 +1,34 @@
...
+}
+
+fn main() {
+    let peter = Person::new(String::from("Peter"), 27);
+    println!("{peter:?}");
+    
+    let tmp = Person {..Default::default()};
+    println!("{tmp:?}");
+    
+    let tmp = Person {age:12, .. Default::default()};
```
mgeisler, line 27: Please note that we try to format everything the way `rustfmt` would do it:

```suggestion
    let tmp = Person { age: 12, ..Default::default() };
```

cchiw: I removed this file and moved content to Speaker notes.

src/structs/field-shorthand.md
```diff
@@ -32,6 +32,20 @@ impl Person {
         Self { name, age }
     }
 }
+```    
+* Implement a `default` setting for the struct.
```
mgeisler, line 36: I don't think there is a concept of a "default setting" in Rust — we should just talk about the `Default` trait instead:

```suggestion
* Implement the `Default` trait for the struct.
```

src/structs/field-shorthand.md
```diff
@@ -32,6 +32,20 @@ impl Person {
...
+```rust,ignore
```
mgeisler, line 37: Why do you have the `ignore` attribute on the code here? It will make `mdbook test` [ignore](https://rust-lang.github.io/mdBook/format/mdbook.html) the code — but we actually want to compile it so that we know it works.

src/structs/field-shorthand.md
```diff
@@ -32,6 +32,20 @@ impl Person {
...
+impl Default for Person {
+  fn default() -> Person { Person { name:"Bot".to_string(), age:0 } }
```
mgeisler, line 39: Spacing is wrong here. Please follow `rustfmt` in the formatting.

src/structs/field-shorthand.md
```diff
@@ -32,6 +32,20 @@ impl Person {
...
+}
+```
+* Create a new struct by using the default values.
+```rust,ignore
+let tmp = Person { ..Default::default() };
+```
+* Define some fields and use the default values for the other fields.
+```rust,ignore
```
mgeisler, line 47: The code blocks are not indented correctly so they don't become part of the list items:

image

You can actually see this as well in the rich diff shown in GitHub:

image

cchiw: Noticed this wasn't done consistently and opened issue #475

mgeisler (approved):
Thanks!

PR #374, opened 2023-02-09
Clarify the Box layout diagrams.

gendx:
At first it wasn't clear to me what "tag", "0" and "1" were supposed to represent. I initially read the diagram as "the tag is two bits, 01".

It's clearer with a header for each column.


mgeisler:
> It's clearer with a header for each column.

Yes, kind of... except that this is how the diagrams looked at first until I re-drew them to remove clutter the other diagrams :smile: 

I would love it if we could find an even better representation... perhaps we could drop the idea of the field names and just show the payloads instead. Something like this (made with https://asciiflow.com):

```
+----+----+--+  +----+----+----+  +----+---------+
|Cons| 1  |  +->|Cons| 2  |    |->|Nil |/////////|
+----+----+--+  +----+----+----+  +----+---------+
```

That is a nice symbolic representation and it should be possible to change this into a drawing where the pointer shares the space with the `Cons`/`Nil` tags.

gendx:
Applied the suggestion.

src/std/box-niche.md
```diff
@@ -18,14 +18,15 @@ allows the compiler to optimize the memory layout:
...
-:   | "1/Tag"| o-----+----+-----+-'  | "1/Tag"|  o-----+-'  | "1/Tag"| null |   :
-:   +--------+-------+    :     :    +--------+--------+    +--------+------+   :
-:                         :     :                                               :
-:                         :     :                                               :
-`- - - - - - - - - - - - -'     '- - - - - - - - - - - - - - - - - - - - - - - -'
+.- - - - - - - - - - - -.     .- - - - - - - - - - - - - - - - - - - - - - -.
+:                       :     :                                             :
+:    list               :     :                                             :
+:   +----+----------+   :     :    +----+----------+    +----+----------+   :
+:   | .0 | ".1/Tag" |   :     :    | .0 | ".1/Tag" |    | .0 | ".1/Tag" |   :
```
mgeisler, line 25:
```suggestion
```

src/std/box-niche.md
```diff
@@ -18,14 +18,15 @@ allows the compiler to optimize the memory layout:
...
-:   | 0      | 1     |    :     : .->| 0      |  2     | .->| ////// | //// |   :
-:   | "1/Tag"| o-----+----+-----+-'  | "1/Tag"|  o-----+-'  | "1/Tag"| null |   :
-:   +--------+-------+    :     :    +--------+--------+    +--------+------+   :
-:                         :     :                                               :
-:                         :     :                                               :
-`- - - - - - - - - - - - -'     '- - - - - - - - - - - - - - - - - - - - - - - -'
+.- - - - - - - - - - - -.     .- - - - - - - - - - - - - - - - - - - - - - -.
+:                       :     :                                             :
+:    list               :     :                                             :
+:   +----+----------+   :     :    +----+----------+    +----+----------+   :
```
mgeisler, line 24:
```suggestion
```

src/std/box-recursive.md
```diff
@@ -20,24 +20,24 @@ fn main() {
 .- - - - - - - - - - - - -.     .- - - - - - - - - - - - - - - - - - - - - - - -.
 :                         :     :                                               :
 :    list                 :     :                                               :
-:   +--------+-------+    :     :    +--------+--------+    +--------+------+   :
-:   | Tag    | Cons  |    :     : .->| Tag    | Cons   | .->| Tag    | Nil  |   :
-:   | 0      | 1     |    :     : |  | 0      | 2      | |  | ////// | //// |   :
-:   | 1      | o-----+----+-----+-'  | 1      | o------+-'  | ////// | //// |   :
-:   +--------+-------+    :     :    +--------+--------+    +--------+------+   :
+:   +------+----+----+    :     :    +------+----+----+    +------+----+----+   :
```
mgeisler, line 23:
```suggestion
```

src/std/box-recursive.md
```diff
@@ -20,24 +20,24 @@ fn main() {
...
+:   | Tag  | .0 | .1 |    :     :    | Tag  | .0 | .1 |    | Tag  | .0 | .1 |   :
```
mgeisler, line 24:
```suggestion
```

mgeisler (approved):
Thank you! I'll go a step further and completely remove the confusing `.1/Tag` part.

src/std/box-niche.md
mgeisler: Ah, I had hoped I could sneak this in :-) Let me do it in a new PR.

PR #381, opened 2023-02-10
Publish Brazilian Portuguese translation

mgeisler:
This will trigger publication of the translation on https://google.github.io/comprehensive-rust/pt_BR/.

@rastringer, can you review this when you feel the translation is good enough?

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -18,7 +18,7 @@ concurrency:
 
 env:
   CARGO_TERM_COLOR: always
-  LANGUAGES: da
+  LANGUAGES: da pt_BR
 
 jobs:
   publish:


mgeisler:
> translation on https://google.github.io/comprehensive-rust/pt_BR/.

Does the `pt_BR` part look "normal" for Brazilian Portuguese sites? I'm wondering if we would want to use `pt-br` instead?

hugojacob:
> > translation on https://google.github.io/comprehensive-rust/pt_BR/.
> 
> Does the `pt_BR` part look "normal" for Brazilian Portuguese sites? I'm wondering if we would want to use `pt-br` instead?

I've seen various ways: `pt_BR`, `pt-BR` or `pt-br`. The [IANA registry](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry) says `pt-BR` and I think that's also the same used by [ISO 639-3:2007]. [GNU gettext](https://www.gnu.org/software/gettext/manual/html_node/Locale-Names.html) uses `pt_BR`.

Having said that, no strong feelings.

rastringer (approved):
Happy with the content, we can keep improving things here and there once it's in the wild.

mgeisler:
> Great question on the language abbreviation, no strong feeling either. I think pt-br is the nicest looking (I think that's what they go with in the Rust Book) but happy with whichever.

Thanks! I looked a bit more and found that `pt-BR` seems to be the correct name. I've updated the PR description with links about it :smile:

PR #382, opened 2023-02-10
CODEOWNERS: initial check-in

hugojacob:
That's the initial check-in of the CODEOWNERS file.

By reading through the [CODEOWNERS documentation](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners), it's not clear to me whether a default assignee (e.g `* <someone>`) is needed, though.


mgeisler (approved):
Thanks, I think that looks good.

Issue #71, opened 2022-12-27
Don't silence warnings in exercises


QuineDot:
All the exercises start with something like
```
// TODO: remove this when you're done with your implementation.
#![allow(unused_variables, dead_code)]
```
You're training people that it's normal and a good idea to ignore warnings during development.  Instead, [continue to emphasize what you mention in passing elsewhere](https://google.github.io/comprehensive-rust/why-rust/modern.html) -- Rust has great compiler errors!

Of course it's going to have a lot of warnings at the start since you've given them incomplete code.  You can
* Just let them deal with it
* Or mention `RUSTFLAGS=-A... cargo ...` perhaps

But you shouldn't imply ignoring warnings until you're "done with your implementation" is a good idea, or IMO that ignoring warnings in your source code (as opposed to a command line flag) is a good idea.

mgeisler:
> But you shouldn't imply ignoring warnings until you're "done with your implementation" is a good idea, or IMO that ignoring warnings in your source code (as opposed to a command line flag) is a good idea.

I generally agree... except that most students use the playground for the exercises. At least for the first 1-2 days. I did it like that to minimize the work I (and other instructors) would have to get things started: I used Linux and know little about how to debug Windows firewall settings which prevents people from installing Cargo and Rust on their laptops.

I added the warnings after the first few classes internally: I noticed that people would get a ton of warnings on the playground and that they had a hard time scrolling around in all the output.

Perhaps there is a way to re-design the exercises to not trigger warnings? That would be ideal. One way would be to comment out the unused code — though that also leads to less coverage by `mdbook test`.

I would be happy to hear ideas for improving this.

mgeisler:
Hey @cchiw, I see you've assigned this issue to yourself. How do you plan on tackling this?

cchiw:
Hola,
 
After trying it out,  I thought we could remove the line (to silence warnings) by either:

1. Just commenting out the minimal amount of code necessary [Example PR](https://github.com/google/comprehensive-rust/pull/389/files). Note might be a lot for some exercises.
2.  Using the unused code. That can include all the fields in a struct [Example PR](https://github.com/google/comprehensive-rust/pull/388) or parameters to a function [Example PR](https://github.com/google/comprehensive-rust/pull/391) and [Example](https://github.com/google/comprehensive-rust/pull/392). 

What do you think?

mgeisler:
[...]

PR #388, opened 2023-02-10
Remove silenced warnings in book-library exercise

cchiw:
In book-library.rs we create instances of the structs (library books) so that the struct fields are not unused. These changes allow us to remove the line to silence warnings for unused variables https://github.com/google/comprehensive-rust/issues/71.

### DIFF
--- a/src/exercises/day-1/book-library.rs
+++ b/src/exercises/day-1/book-library.rs
@@ -118,8 +118,8 @@ fn main() {
     //}
     //
     //println!("Our library has {} books", library.len());
-    for b in library.books {
-        println!("{b}");
+    for book in library.books {
+        println!("{book}");
     }
 }
 // ANCHOR_END: main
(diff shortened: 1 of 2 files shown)


src/exercises/day-1/book-library.rs
```diff
@@ -116,6 +118,9 @@ fn main() {
     //}
     //
     //println!("Our library has {} books", library.len());
+    for b in library.books {
```
mgeisler, line 121:
```suggestion
    for book in library.books {
```

mgeisler (approved):
Thanks!

src/exercises/day-1/book-library.rs
```diff
@@ -116,6 +118,9 @@ fn main() {
     //}
     //
     //println!("Our library has {} books", library.len());
+    for book in library.books {
+        println!("{book}");
+    }
```
mgeisler, line 123: I just realized that this inadvertently shows the students what they are supposed to write in the `print_books` method. I'll fix this for the class I'm teaching next week.

PR #391, opened 2023-02-10
Remove silence warnings in exercises

cchiw:
This code had variables in the function arguments that were not being used (yet).
To address that, in this PR I use the function arguments in a print statement. This changes allow us to remove the line to silence warnings for unused variables https://github.com/google/comprehensive-rust/issues/71.

Alternatively, we could comment out the functions. As seen in #389.

### DIFF
--- a/po/da.po
+++ b/po/da.po
@@ -7339,7 +7339,7 @@ msgstr ""
 #: src/exercises/day-3/simple-gui.md:18
 #: src/exercises/day-3/safe-ffi-wrapper.md:25
 msgid ""
-"```rust,should_panic\n"
+"```rust,should_panic"
 msgstr ""
 
 #: src/exercises/day-3/simple-gui.md:22
--- a/po/da.po
+++ b/po/da.po
@@ -2659,8 +2659,6 @@ msgstr ""
 #: src/exercises/day-2/health-statistics.md:13
 msgid ""
 "```rust,should_panic\n"
-"// TODO: remove this when you're done with your implementation.\n"
-"#![allow(unused_variables, dead_code)]"
 msgstr ""
 
 #: src/exercises/day-1/for-loops.md:56
@@ -5125,8 +5123,6 @@ msgstr ""
 #: src/exercises/day-2/strings-iterators.md:12
 msgid ""
 "```rust\n"
-"// TODO: remove this when you're done with your implementation.\n"
-"#![allow(unused_variables, dead_code)]"
 msgstr ""
 
 #: src/exercises/day-2/points-polygons.md:11
@@ -7344,8 +7340,6 @@ msgstr ""
 #: src/exercises/day-3/safe-ffi-wrapper.md:25
 msgid ""
 "```rust,should_panic\n"
-"// TODO: remove this when you're done with your implementation.\n"
-"#![allow(unused_imports, unused_variables, dead_code)]"
 msgstr ""
 
 #: src/exercises/day-3/simple-gui.md:22
--- a/po/da.po
+++ b/po/da.po
@@ -7340,8 +7340,6 @@ msgstr ""
 #: src/exercises/day-3/safe-ffi-wrapper.md:25
 msgid ""
 "```rust,should_panic\n"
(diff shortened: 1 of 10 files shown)


src/exercises/day-2/luhn.md
```diff
@@ -21,10 +21,8 @@ function:
 
 
 ```rust
-
-{{#include luhn.rs:luhn}}
+pub fn luhn(cc_number: &str) -> bool {
+    println!("{cc_number}");
```
mgeisler, line 25: Please just add this line below the include we had before. That way we keep the connection between the exercise and the solution.

mgeisler (changes requested):

src/exercises/day-2/luhn.md
cchiw: Sure, does it make sense to add a new function to the original luhn.rs file and refer to it in this file or just add the line back in as is? The commit I just made does the latter.

mgeisler: Hehe, this comment seems very similar to what we have in the other PR.

This reminds me of something I've been thinking about: can you try to avoid doing similar changes in different PRs? That is, please put up one PR for each conceptual change. Fewer and bigger PRs would make life easier for me as a reviewer.

Based purely on the `cchiw-patch-11` branch name, I assume you're doing these changes in the GitHub client? If so, that probably means that you haven't seen them live using `mdbook serve`?

cchiw: Yeah, I tried two _different_ approaches in the PRs because I wanted the reviewer to see the options. I think it totally make sense to just work on one PR at a time if it helps with the load. No problem.

I'll just work on this one, which has the line added back in as you requested, but now has a build error.

mgeisler: > * Ah, yeah I wanted to show the _different_ approaches. Moving forward I'll just do one CL at a time to lower the load. Thanks for bringing this up.

Thanks! What you cannot know is that I review many of these CLs while commuting to work... so I'm often doing this on my phone. This means that I have little context and it's hard for me to compare between different PRs. So when I saw that you had several different PRs with different approaches, I put them back on the "for later" queue.

cchiw: Got it. I consolidated the open PRs. Last night found more places to remove the silence warnings and included them here. It's ready for review, but no rush.

po/da.po
```diff
@@ -2659,8 +2659,6 @@ msgstr ""
 #: src/exercises/day-2/health-statistics.md:13
 msgid ""
 "```rust,should_panic\n"
-"// TODO: remove this when you're done with your implementation.\n"
```
mgeisler, line 2662: Please don't modify the `.po` files — only the translators can update those.

third_party/rust-on-exercism/health-statistics.rs
```diff
@@ -6,26 +6,32 @@ struct User {
 
 impl User {
     pub fn new(name: String, age: u32, weight: f32) -> Self {
+        println!("Use name {:?} age {:?} and weight{:?}", name, age, weight);
```
mgeisler, line 9: I've tried to inline the format strings everywhere else in the course:

```suggestion
        println!("Use name {name:?} age {age:?} and weight {weight:?}");
```

I also fixed a space, which makes me think that you've not looked at this in the browser after making the change?

third_party/rust-on-exercism/health-statistics.rs
```diff
@@ -6,26 +6,32 @@ struct User {
...
         unimplemented!()
     }
 
     pub fn name(&self) -> &str {
+        println!("Use name {:?}", self.name);
```
mgeisler, line 14: I don't know if we've agreed that this is a good way to "fix" the warnings? I would have added such print statements already if this is a nice way to solve it :slightly_smiling_face:

mgeisler:
Hi @cchiw,

I don't think adding a lot of `print("Use {foo}")` lines is a good way to silence the warnings. Why not? Because it not how actual code looks like. Let's discuss on #71.

PR #392, opened 2023-02-10
Remove silenced warnings in string-iterators exercise

cchiw:
This code had variables in the function arguments that were not being used (yet). 
To address that, in this PR I use the function arguments in a print statement. This changes allow us to remove the line to silence warnings for unused variables #71.

Alternatively, we could comment out the functions. As seen in #389.

### DIFF
--- a/src/exercises/day-2/strings-iterators.md
+++ b/src/exercises/day-2/strings-iterators.md
@@ -10,10 +10,8 @@ pass. Try avoiding allocating a `Vec` for your intermediate results:
 
 
 ```rust
-// TODO: remove this when you're done with your implementation.
-#![allow(unused_variables, dead_code)]
-
-{{#include strings-iterators.rs:prefix_matches}}
+pub fn prefix_matches(prefix: &str, request_path: &str) -> bool {
+    println!("Use parameters {prefix} and {request_path}");
     unimplemented!()
 }
 
--- a/src/exercises/day-2/strings-iterators.md
+++ b/src/exercises/day-2/strings-iterators.md
@@ -10,7 +10,10 @@ pass. Try avoiding allocating a `Vec` for your intermediate results:
 
 
 ```rust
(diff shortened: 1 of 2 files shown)


mgeisler:
> Remove flag to silence warnings #71

Please move this to the PR title. The title shows up prominently in the history afterwards so it's helpful when it says more than just "Update foo.md" 🙂

Also, please explain _why_ you make the change. What did you see before and after this PR? That gives me a lot more context as a reviewer.

src/exercises/day-2/strings-iterators.md
```diff
@@ -10,10 +10,8 @@ pass. Try avoiding allocating a `Vec` for your intermediate results:
 
 
 ```rust
-
-{{#include strings-iterators.rs:prefix_matches}}
+pub fn prefix_matches(prefix: &str, request_path: &str) -> bool {
```
mgeisler, line 13: Doing this makes us lose the coupling with the function signature in `strings-iterators.rs`. Could you put the include back so that we can keep the connection?

mgeisler (changes requested):
Doing this makes us lose the coupling with the solution in `strings-iterators.rs`. That's not a

src/exercises/day-2/strings-iterators.md
cchiw: Sure, does it make sense to add a new function to the original string-iterators.rs file and refer to it in this file or just add the line back in as is? The commit I just made does the former.

mgeisler: The latter is better: that way we avoid duplication in the solution file and we avoid showing `prefix_matches_unimplemented` to the students (there is no reason for them to see such a function name).

My goal is to reuse as much as possible from the solution file. We compile the code there in CI  so we know it works — when we reuse it, we also have some confidence that the unsolved version is aligned with the solution.

PR #393, opened 2023-02-11
Simplify JNI function names via the `jni_fn` crate

haraldreingruber:
Just a small idea, to improve the readability and simplicity of the JNI function names.

The `jni_fn` crate (https://docs.rs/jni_fn/) provides a macro, which allows generating the mangled version JNI is expecting (including package and class name).

Please let me know what you think. Further improvement ideas welcome :)

### DIFF
--- a/src/android/interoperability/java.md
+++ b/src/android/interoperability/java.md
@@ -4,6 +4,8 @@ Java can load shared objects via [Java Native Interface
 (JNI)](https://en.wikipedia.org/wiki/Java_Native_Interface). The [`jni`
 crate](https://docs.rs/jni/) allows you to create a compatible library.
 
+The [`jni_fn` crate](https://docs.rs/jni_fn/) provides a macro for easier JNI function name generation. 
+
 First, we create a Rust function to export to Java:
 
 _interoperability/java/src/lib.rs_:
--- a/src/android/interoperability/java/src/lib.rs
+++ b/src/android/interoperability/java/src/lib.rs
@@ -18,10 +18,11 @@
 use jni::objects::{JClass, JString};
 use jni::sys::jstring;
 use jni::JNIEnv;
+use jni_fn::jni_fn;
 
 /// HelloWorld::hello method implementation.
-#[no_mangle]
-pub extern "system" fn Java_HelloWorld_hello(
+#[jni_fn("HelloWorld")]
+pub fn hello(
     env: JNIEnv,
     _class: JClass,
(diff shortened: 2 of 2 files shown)


qwandor:
Thanks for the suggestion!

However, we don't have the `jni_fn` crate [in AOSP](https://cs.android.com/android/platform/superproject/+/master:external/rust/crates/), so there's not much point recommending it here (in the Android section) unless we decide to import it into AOSP.

haraldreingruber:
> However, we don't have the `jni_fn` crate [in AOSP](https://cs.android.com/android/platform/superproject/+/master:external/rust/crates/), so there's not much point recommending it here (in the Android section) unless we decide to import it into AOSP.

That makes a lot of sense.
Do you think it would be a good idea to suggest it there?

mgeisler:
Hi @haraldreingruber, thanks for the PR!

> Do you think it would be a good idea to suggest it there?

You mean suggesting that it's imported in AOSP? In general, we try to vendor as little as possible since every vendored crate incurs a cost in terms of maintenance. So a small helper crate like this will probably not be imported (sorry).

haraldreingruber:
No worries, I just wanted to check if it could be valuable to you/the AOSP team. 😊

mgeisler:
> No worries, I just wanted to check if it could be valuable to you/the AOSP team. blush

Actually, let me ask people internally what they think of it... they might not have heard of the crate. I see that we have [~10 Rust files which use JNI](https://cs.android.com/search?q=%22use%20jni%22%20f:%5C.rs$%20-f:external%2Frust%2Fcrates) so the people who wrote those would probably like to simplify things.

haraldreingruber:
Thanks @mgeisler for checking with the team(s)! 

Maybe worth mentioning is that the crate is super simple and excluding test code it's just 150 lines of code.

But the initial response makes totally sense.

mgeisler:
> Thanks @mgeisler for checking with the team(s)!

No problem, thanks for suggesting it in the first place :smile: 

> Maybe worth mentioning is that the crate is super simple and excluding test code it's just 150 lines of code.

In this context, that kind of counts against the crate: we only want to vendor things in Android if they come with must-have functionality. The problem is that someone will have to review the code, and also review updates to the code. I guess there won't be many updates to a small crate like this.

I think adding the "waiting" label was a bad idea since the PR keeps showing up (of course). So I'll instead close it. People can then find it later and resurrect as needed.

PR #395, opened 2023-02-13
Translate "Running the course" section to German

ronaldfw:
I started using poedit to work on the translation, which has reformatted large parts of the file, so the change is a lot larger than just the newly translated parts.

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -1,16 +1,16 @@
-
 msgid ""
 msgstr ""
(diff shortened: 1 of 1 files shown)


mgeisler:
> I started using poedit to work on the translation, which has reformatted large parts of the file, so the change is a lot larger than just the newly translated parts.

Thanks and that's all good — the sooner we get the file normalized the better :slightly_smiling_face:

mgeisler (approved):

PR #401, opened 2023-02-13
Let user know they can modify the tests, but do not need to.

cchiw:
Follow up on comment in merged PR #361, working on issue #274


mgeisler (approved):
Thank you very much, I like this phrasing!

PR #405, opened 2023-02-14
Remove explicit typing in Mutex example

smides-nest:
Example contained unnecessary explicit type info for the vector in Mutex v. Rust will magically do the needful conversions for us. Code looks cleaner/simpler without the explicit typing.


mgeisler (approved):
Thanks a lot!

PR #408, opened 2023-02-14
Day2: Afternoon - speaker note details and minor cosmetic changes.

markozagar:
I added some extra examples and speaker notes especially for HashMap and Modules, since it was useful in my own projects in the past.

### DIFF
--- a/src/std/string.md
+++ b/src/std/string.md
@@ -27,8 +27,7 @@ fn main() {
 
 <details>
 
-* `len` returns the size of the `String` in bytes, not its length in characters.
-* `chars` returns an iterator over the actual characters.
-* `String` implements `Deref<Target = str>` which transparently gives it access to `str`'s methods.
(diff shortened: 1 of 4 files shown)


src/std/string.md
```diff
@@ -27,8 +27,7 @@ fn main() {
 
 <details>
 
-* `len` returns the size of the `String` in bytes, not its length in characters.
-* `chars` returns an iterator over the actual characters.
-* `String` implements `Deref<Target = str>` which transparently gives it access to `str`'s methods.
```
qwandor, line 32: Why remove this line?

markozagar: A very similar statement with the same information is already written on the slide itself, it seemed redundant to repeat it again in speaker notes. The one in the slide also links to the appropriate place in the STD reference.

qwandor (approved):

PR #409, opened 2023-02-14
Further simplify `Box` diagrams

mgeisler:
I think we can improve the drawing by simplifying them and making them more symbolic. Followup to #374.


qwandor (approved):

src/std/box-niche.md
```diff
@@ -18,15 +18,13 @@ allows the compiler to optimize the memory layout:
...
-:   | 1  |    o-----+---+-----+--->| 1  |    o-----+--->| 2  |  null    |   :
-:   +----+----------+   :     :    +----+----------+    +----+----------+   :
-:                       :     :                                             :
-:                       :     :                                             :
-`- - - - - - - - - - - -'     '- - - - - - - - - - - - - - - - - - - - - - -'
+.- - - - - - - - - - - - -.     .- - - - - - - - - - - - - - - - - - - - - - -.
+:                         :     :                                             :
+:    list                 :     :                                             :
+:   +----+----+           :     :    +----+----+    +----+----+               :
+:   | 1  | o--+-----------+-----+--->| 2  | o--+--->| // | // |               :
```
gendx, line 25: I'd still leave a `null` in the second half of the last box, to make it clear that a null pointer is how the compiler notices that the variant is `Nil`.

mgeisler: Thanks, that's a nice idea

```suggestion
:   | 1  | o--+-----------+-----+--->| 2  | o--+--->| // | null |             :
```

src/std/box-niche.md
```diff
@@ -18,15 +18,13 @@ allows the compiler to optimize the memory layout:
...
+:   +----+----+           :     :    +----+----+    +----+----+               :
```
mgeisler, line 26:
```suggestion
:   +----+----+           :     :    +----+----+    +----+------+             :
```

PR #411, opened 2023-02-14
Add language picker menu

mgeisler:
The picker is a drop-down menu using the same design as the theme picker in the top-left.

There doesn’t seem to be an easy way to pass in a list of languages and descriptions, so for now we’ll have to expand the menu by hand as we add new languages. A comment has been added to `publish.yml` to remind us of this.

Part of #414.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -18,6 +18,7 @@ concurrency:
 
 env:
   CARGO_TERM_COLOR: always
+  # Update the language picker in index.hbs to link new languages.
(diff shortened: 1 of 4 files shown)


mgeisler:
To test this, you could download one or more of the translations from [the test run](https://github.com/google/comprehensive-rust/actions/runs/4178679467) and then look at them in a browser :slightly_smiling_face:

rastringer (approved):

rastringer:
Amazing, thank you

PR #421, opened 2023-02-16
Run `links` preprocessor before `gettext` preprocessor

### LINKED ISSUE #419: Including .md file doesn't work with translations
I noticed that gettext works as expected with the inclusion. But when rendered with translation, those included texts are not translated even when {lang}.po file contains the translations.

For now, there's only a single instance of it.
https://github.com/google/comprehensive-rust/blame/main/src/exercises/day-2/health-statistics.md#L3

image

image

jooyunghan:
fixes #419

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -13,6 +13,7 @@ extra-watch-dirs = ["po"]
 
 [preprocessor.links]
 renderers = ["html"]
+before = ["gettext"]
 
 [preprocessor.index]
 renderers = ["html"]


mgeisler:
Hi @jooyunghan, thanks a lot for diagnosing this!

Please also update the documentation in `TRANSLATIONS.md`.

I didn't know what we could use the `before` and `after` keys with unknown preprocessors. I tried with

```toml
[preprocessor.svgbob]
renderers = ["html"]
class = "bob"
after = ["foobar"]
```

and it works fine — though we do get a warning when running `mdbook`:

```
2023-02-16 11:04:31 [INFO] (mdbook::cmd::serve): Building book...
2023-02-16 11:04:31 [WARN] (mdbook::book): preprocessor.svgbob.before contains "foobar", which was not found
2023-02-16 11:04:31 [INFO] (mdbook::book): Book building has started
```

Still, we should be able to simplify our instructions a little based on this.

jooyunghan:
With this PR I don't think we need to update TRANSLATIONS.

mgeisler (approved):

Issue #415, opened 2023-02-15
Translation error in Brazilian translation for pattern-matching.html


gendx:
See:
- English page: https://google.github.io/comprehensive-rust/pattern-matching.html
- Brazilian page: https://google.github.io/comprehensive-rust/pt-BR/pattern-matching.html.

It looks like the pattern for the character `a` got translated into two letters `um` in the code, which breaks compilation.

Maybe the GitHub Actions checking the translations should also run `mdbook test` on the translated contents to check that the code snippets still compile?

---

![image](https://user-images.githubusercontent.com/4346404/219077042-fc444820-fa94-4d6f-ad8b-bf81de2d272f.png)

mgeisler:
> Maybe the GitHub Actions checking the translations should also run `mdbook test` on the translated contents to check that the code snippets still compile?

Yes, we actually do exactly that — but it doesn't work as intended until https://github.com/rust-lang/mdBook/pull/1986 is merged.

mgeisler:
This was fixed by #416 and will be long-term fixed by rust-lang/mdBook#1986.

PR #425, opened 2023-02-16
More polishing on the pt-BR translation

hugojacob:


rastringer (approved):
Thanks Hugo! These PRs are just what we need.

po/pt-BR.po
```diff
@@ -1929,17 +1882,17 @@ msgid ""
...
-"fn main() { // Point de entrada do programa\n"
-"    let mut x: i32 = 6; // Vinculação de variável mutável\n"
-"    print!(\"{x}\"); // Macro para impressão, como printf\n"
-"    while x != 1 { // Sem parênteses em torno da expressão\n"
-"        if x % 2 == 0 { // Matemática como em outras linguagens\n"
-"            x = x/2;\n"
+"```rust,editable\n"
+"fn main() {              // Ponto de entrada do programa\n"
+"    let mut x: i32 = 6;  // Atribuição de uma variável mutável\n"
+"    print!(“{x}”);       // Macro para escrever na tela, como printf\n"
```
mgeisler, line 1888: These beautiful "typographic" quotes won't actually work. I'm waiting on #415 to be fixed upstream so that we could catch this.

hugojacob: Done.

mgeisler (approved):

PR #428, opened 2023-02-17
Publish Korean translation

jiyongp:


theme/index.hbs
```diff
@@ -160,6 +160,9 @@
                           <li role="none"><button role="menuitem" class="theme">
                               <a id="pt-BR">Brazilian Portuguese</a>
                           </button></li>
+                          <li role="none"><button role="menuitem" class="theme">
+                              <a id="ko">한국어</a>
```
mgeisler, line 164: Right now, all of the interface parts are in English — because we don't have access to the strings via the preprocessor/renderer setup. I would suggest writing this as either

* Korean (한국어)
* 한국어 (Korean)

I think I've seen other pages do the same for this kind of situation. @rastringer, it would be great if you could update the Brazilian Portuguese line the same way.

mgeisler:
Yay! Please update the new `translations.md` file with your names as well!

theme/index.hbs
rastringer: Will do! I'm just going to be out of action for a day or two, will be back soon.

jiyongp:
> Yay! Please update the new `translations.md` file with your names as well!

Done!

mgeisler (approved):
Awesome!

PR #429, opened 2023-02-17
`svgbob` should run after `gettext`

jooyunghan:
svgbob translates `_text_` into `*text*`, which makes gettext fail to translate paragraphs with `_text_`.

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -20,6 +20,7 @@ renderers = ["html"]
 
 [preprocessor.svgbob]
 renderers = ["html"]
+after = ["gettext"]
 class = "bob"
 
 # Enable this preprocessor to overlay a large red rectangle on the


mgeisler:
> svgbob translates `_text_` into `*text*`, which makes gettext fail to translate paragraphs with `_text_`.

Oh, interesting! I did not know about that problem yet...

Our instructions [already ask](
https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md#building-a-translation) people to run `svgbob` after `gettext` — but that was simply done so that translators have the possibility to translate the diagrams.

Now that we know that we only get a warning due to the undefined preprocessor, I'm fine with adding the `before`/`after` key to the `book.toml`.

Going further, we could consider adding the `gettext` preprocessor to the normal flow — people who want to test the book locally already have to install both `mdbook` and `mdbook-svgbob`, so asking them to install `i18n-helpers` too isn't much of a change. Doing so should simplify the `publish.yml` script a bit.

jooyunghan:
This was a regression from #421. Adding `before` to `links` caused `svgbob` running *before* `gettext`.

I agree that adding gettext to fix the warning (undefined preprocessor). We can make it a no-op when po-file is not specified.

jooyunghan:
Can we merge this?

PR #447, opened 2023-02-22
Make some more progress on the German translation

ronaldfw:


ronaldfw:
Hey @fechu, thanks for taking a look at this. 

Two relevant pieces of information: 
1. We have agreed to use informal manner of addressing people in our translations (that is, "Du" instead of "Sie" in German). 
2. For vocabulary, I'm aiming to use the same translations as in https://rust-lang-de.github.io/rustbook-de/

po/de.po
```diff
@@ -1098,7 +1098,6 @@ msgstr ""
 "Auch deine Teilnehmer sind herzlich eingeladen, [uns Rückmeldung zu senden][4]!"
 
 #: src/running-the-course.md:55
-#, fuzzy
```
fechu, line 1101: What is this `fuzzy` line that you removed in multiple places?

po/de.po
```diff
@@ -1178,72 +1168,63 @@ msgid ""
...
 "give a brief overview of what Cargo is and how it fits into the wider ecosystem\n"
 "and how it fits into this training."
 msgstr ""
-"Wenn Sie anfangen, über Rust zu lesen, werden Sie bald [Cargo](https://doc.rust-lang.org/cargo/), "
-"das Standardwerkzeug, kennenlernen\n"
-"Wird im Rust-Ökosystem verwendet, um Rust-Anwendungen zu erstellen und auszuführen. Hier wollen "
-"wir\n"
-"geben einen kurzen Überblick darüber, was Cargo ist und wie es in das breitere Ökosystem passt\n"
-"und wie es in dieses Training passt."
+"Wenn Du anfängst über Rust zu lesen, wirst du sehr bald [Cargo](https://doc.rust-lang.org/"
```
fechu, line 1186: Incosistent use of "Du" vs "du".

po/de.po
mgeisler: The translations that are marked "fuzzy" are not yet used when building the HTML output. They are the either the result of a script like [cloud-translate](https://github.com/google/comprehensive-rust/pull/291) or they come from `msgmerge`, which is used to import new English strings into a translation.

By removing the fuzzy marker, the translation becomes active and will show up in the output.

po/de.po
mgeisler: I recently discovered the "suggestion" button in GitHub (the left-most button in the toolbar):

```suggestion
"Wenn du anfängst über Rust zu lesen, wirst du sehr bald [Cargo](https://doc.rust-lang.org/"
```

You can even commit the suggestions yourself afterwards and thus cut down on review round-trips when you simply want to fixup a small detail.

ronaldfw:
Thanks @fechu for the review and thanks @mgeisler for the helpful comments. Sorry for the long delay on this, the PR should be in better shape now. PTAL.

po/de.po
fechu: 🤯  This is amazing! Thanks for mentioning it!

fechu (approved):

PR #448, opened 2023-02-22
Clarify that `Box` cannot be `null`

DarthGandalf:


mgeisler (approved):
Nice, thanks for adding this detail!

PR #449, opened 2023-02-22
Parse Markdown in mdbook-xgettext

### LINKED ISSUE #318: Extract text more carefully in `mdbook-xgettext`
Right now, we simply split the text on `\n\n+`, but this leads to a number of problems:

* We split code blocks into different messages when there are one or more blank lines in the middle of the block.
* We extract bullet point lists as a single message.

In general, it would be awesome if we could

* Make the extracted messages independent of the precise formatting of the Markdown text. In particular, a hard-wrapped paragraph should be extracted without the paragraph breaks.
* Remove formatting such as `#` from headers and `*` from bullet points.
* Extract code blocks as a single message.

So Markdown like

~~~markdown
# This is a heading

A _little_
paragraph.

```rust,editable
fn main() {
    println!("Hello world!");
}
```

* First
* Second
~~~

should result in these messages

* `This is a heading` (heading type is stripped)
* `A _little_ paragraph.` (softwrapped lines are unfolded)
* `fn main() {\n    println!("Hello world!");\n}` (info string is stripped)
* `First` (bullet point extracted individually)
* `Second`

You could imagine done something nice with links too: `foo [bar](https://example.net) baz` could be stored as `foo [bar] baz`. This might be a poor idea, though: it means that the translator cannot change the destination URL.

djmitche:
This uses a full Markdown parser to parse the book contents into messages for translation.

The existing functionality broke messages on double-newlines (`\n\n`), otherwise returning the text exactly as found in the original file.  The updated approach still takes string slices from the original file, but uses the markdown parser to better identify the breaks between messages.

*Status*: work in progress

Fixes #318.


i18n-helpers/src/lib.rs
```diff
@@ -54,33 +56,215 @@ pub fn extract_paragraphs(text: &str) -> impl Iterator<Item = (usize, &str)> {
...
+                self.push_message(Message::Code(offset));
+                self.in_codeblock = true;
+            }
+            Event::Start(_) => self.finish_message(),
+
+            Event::End(Tag::Emphasis | Tag::Strong | Tag::Strikethrough) => {
+                self.push_message(Message::Text(offset))
+            }
+            Event::End(Tag::CodeBlock(_)) => {
+                self.in_codeblock = false; // (assumes codeblocks aren't nested!)
```
mgeisler, line 174: I believe this assumption is true according to the specification.

i18n-helpers/src/lib.rs
```diff
@@ -54,33 +56,215 @@ pub fn extract_paragraphs(text: &str) -> impl Iterator<Item = (usize, &str)> {
...
-        assert_iter_eq!(extract_paragraphs(""), vec![]);
+    fn offset_to_line_multiline() {
+        let input = "abc\ndef\nghi";
+        let acc = MsgAccumulator::new(input);
+        let line_nums: Vec<_> = (0..=input.len()).map(|o| acc.line_number(o)).collect();
+
+        assert_eq!(
+            line_nums,
+            vec![
+                1, // a
```
mgeisler, line 235: I think you could rewrite the test to something like

```
let char_line_nums = input.chars().enumerate().map(|(ch, idx)| (ch, acc.line_number(idx)).collect::<Vec<_>>();
```

Then your `assert_eq!` would not need the comments since the character is visible in the output.

i18n-helpers/src/lib.rs
```diff
@@ -120,4 +304,57 @@ mod tests {
...
+                (2, "```rust\nfn hello() {\n  todo!()\n}\n```"),
+                (7, "Postamble")
+            ],
+        );
+    }
+
+    #[test]
+    fn extract_msgs_details() {
+        assert_eq!(
+            // This isn't great, because the parser treats any data following a tag as also HTML,
```
mgeisler, line 331: Yeah, I was worried about how the `details` elements would influence the Markdown parsing.

I had a plan to rework the speaker notes into a mdbook preprocessor instead. The idea would be to use Markdown syntax like

```markdown
# Speaker Notes {#speaker-notes}

Blah blah.
````

The `mdbook-xgettext` would extract this like normal and the translator would translate it (keeping the `speaker-notes` `id` attribute intact).

The new preprocessor could then run on the page and find the speaker notes and do the wrapping into `<details> ... </details>` as necessary.

I think that would simultaniously simplify the job of `mdbook-xgettext` and allow for a straight forward way to translate the heading. It would also mean less JavaScript code running on page load.

For now, I'm happy to rely on the pages having blank lines around the `details` tags.

File: i18n-helpers/src/lib.rs [hunk 4]
```diff
@@ -54,33 +56,215 @@ pub fn extract_paragraphs(text: &str) -> impl Iterator<Item = (usize, &str)> {
...
+            Event::Html(content) => self.push_message(Message::Html(offset)),
+            // TODO: list items
+            // TODO: headings
+            _ => {}
+        }
+    }
+
+    /// Get the resulting list of messages.
+    fn into_msgs(mut self) -> Vec<(usize, &'a str)> {
+        let parser = Parser::new(self.text);
```
mgeisler, line 192: You will eventually want to use [`new_cmark_parser`](https://docs.rs/mdbook/latest/mdbook/utils/fn.new_cmark_parser.html) here to have the same Markdown extensions enabled as in mdbook.

i18n-helpers/src/lib.rs
mgeisler: Oh, this plan would rely on [`ENABLE_HEADING_ATTRIBUTES`](https://docs.rs/pulldown-cmark/latest/pulldown_cmark/struct.Options.html#associatedconstant.ENABLE_HEADING_ATTRIBUTES) being set. It's not enabled as the moment by [`new_cmark_parser`](https://docs.rs/mdbook/latest/mdbook/utils/fn.new_cmark_parser.html) but I don't see why it shouldn't be enabled.

mgeisler:
This looks like a great start, thanks for working on it!

i18n-helpers/src/lib.rs
djmitche: I think that ENABLE_HEADING_ATTRIBUTES wouldn't matter for `mdboox-gettext`

mgeisler: > I think that ENABLE_HEADING_ATTRIBUTES wouldn't matter for `mdboox-gettext`

Right, I agree — it would only matter for a new preprocessor which would use the ID to find the right tag to wrap.

djmitche:
A bit of experimentation with the existing translations shows no difference for some of them, especially de and da.  I suspect that these are just out of date?  I do see differences for ko and pt-BR.

My thinking is that I will run a `msgmerge` and then update the `.po` file by hand to where the resulting output is the same as it currently is on main.  But, I suspect that this is going to drop a lot of translations (probably almost everything for de and da) that have not yet been updated due to other changes in the text that have landed on `main`.

I'm not sure what the other options are, though.  What do you think?

mgeisler:
> A bit of experimentation with the existing translations shows no difference for some of them, especially de and da. I suspect that these are just out of date? I do see differences for ko and pt-BR.

Yes, the German and Danish translations are very incomplete right now. The only full translations are Korean and Portuguese.

> My thinking is that I will run a `msgmerge` and then update the `.po` file by hand to where the resulting output is the same as it currently is on main. But, I suspect that this is going to drop a lot of translations (probably almost everything for de and da) that have not yet been updated due to other changes in the text that have landed on `main`.

Yeah, if the translations are out of date, there's not much we can do about it.

However, I was not thinking that you would `msgmerge` anything into the existing translations. Instead, I was thinking that you would do something like this:

1. Use the `msgid` fields to reassemble the original English input to the `.po` file. This should be possible since the current `.po` files are complete: they only leave out blank lines.
2. Do the same with the `msgstr` fields to get a (partially) translated file.
3. Run the new `extract_msgs` function on both files. This ought to give you pairs of messages which you can use as `msgid` and `msgstr` for a `.po` file.

The effect of this would be a `.po` file where the code blocks are no longer split. It should also give you a `.po` file where headings can be saved without `#`, and where list items can be saved item-by-item.

However, this only works when the translated paragraphs have the same Markdown structure as the original text. @jooyunghan indicated that this is not always the case for Korean. @jiyongp, @rastringer, @hugojacob, and @ronaldfw what do you think about this?

See also the discussion in #318.

djmitche:
There are a _lot_ of complications and edge cases in trying to follow this clever approach:
 * the .po file contains a concatenation of all messages, with duplicates removed, so only _sort of_ in order
 * concatenating the msgids leads Markdown to sometimes combine lists from multiple files (e.g., where a header is duplicated and thus omitted)
 * there are a lot of occurrences of `<details>` and `</details>` in the .po file that cause spurious differences; fixing these causes lots more cases of merging lists
 * there are places where lists are separated by `\n  \n` and not `\n\n`, which causes different parses

I've pushed a commit with my helper in it, after having manually edited `po/ko.po` enough that it ran successfully.  You can replicate what I'm seeing with
```
cargo run -p i18n-helpers --bin helper
bash compare.sh
```

Here's what's happening:
 * The `helper` binary takes in `po/ko.po`, reconstructs a large markdown file from the `msgids` it finds there, including only messages which actually have a translation.  It then re-parses that markdown with the new parser and uses the offsets of the old and new parses to generate a new set of messages by concatenting msgstr's from the old catalog.  The edits to `po/ko.po` in this commit are there to line up the boundaries properly and avoid the `bail!` case.  It then writes the result out to `po/ko-new.po`.
 * The `compare.sh` script generates a full markdown output of both the original (from `origin/main`) translation and the new (`po/ko-new.po`) translation, and diffs the two.

The result is that lots of things aren't translated anymore, for various fiddly reasons suggested above.  Surprisingly, some things are _newly_ translated, for example
```diff
18,21c15
< * Experience with Java, Go, Python, JavaScript...: You get the same memory safety
<   as in those languages, plus a similar high-level language feeling. In addition
<   you get fast and predictable performance like C and C++ (no garbage collector)
<   as well as access to low-level hardware (should you need it)
---
> * Java, Go, Python, JaveScript: 이 언어들과 동일한 메모리 안정성과 함께, '하이레벨'언어의 느낌을 느낄 수 있습니다. 거기에 더해, 가비지 컬렉터가 없는 C/C++와 유사한 수준의 빠르고 예측 가능한 성능을 기대할 수 있습니다. 그리고 필요한 경우 저수준 하드웨어를 다루는 코드로 작성할 수 있습니다.
```

I feel like this is a never-ending struggle, and maybe what I've got is good enough?  What do you think?

djmitche:
(and, I see there are conflicts in `po/ko.po`, suggesting my work is bitrotting, too!)

mgeisler:
> * the .po file contains a concatenation of all messages, with duplicates removed, so only _sort of_ in order
> * concatenating the msgids leads Markdown to sometimes combine lists from multiple files (e.g., where a header is duplicated and thus omitted)

Right! I was conceptually thinking that you would do this on a per-file basis. So for a given `foo.md` file, you can find all `msgid` entries that are relevant for this file. That should let you reconstruct the English version of `foo.md` at the time the `xx.po` file was last touched. From that you can get a `foo-xx.md` file by doing the same with the `msgstr` entries.

> I feel like this is a never-ending struggle, and maybe what I've got is good enough? What do you think?

Yes, that's a good point... this should not turn into a marathon project where everything must be perfect. When I look at the differences between the `messages.pot` file before and after, most changes look exactly like what I would expect:

image

Here we have a code block which was split before and is now joined into a single message.

The only other big change is the bullet points in the lists:

image

Here individual bullet points have been turned into one big message. If they would be split into individual messages, then the output would be just like before (and thus safe for the existing translations).

I would suggest implementing this: splitting lists into individual bullet points. That transformation alone sounds like something that can be safely executed on all `msgid` and `msgstr` fields today in a lossless fashion. You could even detect when a `msgid` is results in _n_ messages and the corresponding `msgstr` becomes _m != n_ messages (then just mark the _m_ new messages fuzzy so that the translators can quickly proofread them).

With that, we should have a drop-in replacement: the code blocks will lose their translation, but we don't have a lot of those strings (I looked for relevant strings in `msggrep -K -e ';' po/messages.pot | msggrep -K -e '//' | msggrep -K -v -e '// ANCHOR'` and found ~50 translatable comments in total).

Does that approach sound doable?

djmitche:
OK, I've finished all four languages.  There are still some discrepancies, but only a handful.

i18n-helpers/src/lib.rs
```diff
@@ -12,112 +12,487 @@
...
+}
+
+impl Message {
+    fn new(line: usize, span: Range<usize>) -> Self {
+        Self { line, span }
+    }
+
+    /// Get the text of this message, as a slice of the document from which it was generated.
+    pub fn text<'doc>(&self, document: &'doc str) -> &'doc str {
+        &document[self.span.start..self.span.end]
```
mgeisler, line 36: I suppose this could be simplified to just

```suggestion
        &document[self.span]
```

i18n-helpers/src/lib.rs
```diff
@@ -12,112 +12,487 @@
...
+    }
+
+    /// Extend this message to the given offset.
+    fn extend(&mut self, to_end: usize) {
+        self.span.end = to_end;
+    }
+
+    /// Trim trailing newlines from this message.
+    fn trim_right(&mut self, document: &str) {
+        // Note that we cannot index `document[self.span.end]` as that may not
```
mgeisler, line 56: This seems odd to me: could we not require that all `to_end` arguments to `extend` above are valid character boundaries?

i18n-helpers/src/lib.rs
```diff
@@ -12,112 +12,487 @@
...
+        // be a valid UTF-8 character.
+        while document[self.span.clone()].ends_with("\n") {
```
mgeisler, line 58: This ought to be the same as

```suggestion
        self.span.end = self.span.start + document[self.span.clone()].trim_end('\n').len();
```

Or similar? :slightly_smiling_face:

i18n-helpers/src/lib.rs
```diff
@@ -12,112 +12,487 @@
...
+        Self { line, span }
+    }
+
+    /// Get the text of this message, as a slice of the document from which it was generated.
+    pub fn text<'doc>(&self, document: &'doc str) -> &'doc str {
+        &document[self.span.start..self.span.end]
+    }
+
+    /// Get the line number at which this message begins.
+    pub fn lineno(&self) -> usize {
```
mgeisler, line 40: I see you have a method called `line_number` below — I suggest renaming this method to `line_number` as well for consistency.

i18n-helpers/src/lib.rs
```diff
@@ -12,112 +12,487 @@
...
+            Event::End(
+                Tag::Emphasis | Tag::Strong | Tag::Strikethrough | Tag::Link(..),
+            ) => self.push_message(span),
+
+            // We want to translate everything: text, code (from backticks, `..`), or HTML.
+            Event::Text(_) | Event::Code(_) | Event::Html(_) => self.push_message(span),
+
+            // For many event types we just take the entire text from Start to End, which is
+            // already encompassed in the event span.
+            Event::Start(Tag::CodeBlock(_))
```
mgeisler, line 161: You could consider moving all the tags into the `Event::Start` pattern (I don't know it looks nicer): `Event::Start(Tag::CodeBlock(_) | Tag::Heading(..) | ...)`.

po/da.po
```diff
@@ -2631,484 +2633,6 @@ msgid ""
 "columns):"
 msgstr ""
 
-#: src/exercises/day-1/for-loops.md:41
```
mgeisler, line 2634: I'm surprised to see all the messages disappear. Is it because the file is very outdated?

mgeisler (approved):
Wow, I think this looks great!

Please take out the commit with the helper — unless it's somehow useful in the future when we do further transformations?

I agree with you that we should get this merged so that we can start streamlining the PO files to the new parsing.

i18n-helpers/src/lib.rs
djmitche: That comment should be `[self.span.end-1]`, sorry.  And it's the `1` there that is the problem.

po/da.po
djmitche: None of them were translated, but were causing alignment issues, so I just dropped them.

Although now that I think of it, I searched for `^msgstr ""` which may not have been an appropriate search pattern.  I'll take another look.

djmitche: fixed :)

i18n-helpers/src/lib.rs
djmitche: With a `.clone()`, yes.

PR #451, opened 2023-02-23
Add fechu as owner to German translation

fechu:


mgeisler:
Thanks Sandro, I've invited you to the repository and this should fix the error.

mgeisler (approved):

PR #456, opened 2023-02-26
Fix typo in string.md

adamac:


src/std/string.md
```diff
@@ -35,6 +35,6 @@ fn main() {
     * `String` implements `Deref<Target = str>` which transparently gives it access to `str`'s methods.
     * Write and compare `let s3 = s1.deref();` and  `let s3 = &*s1`;.
 * `String` is implemented as a wrapper around a vector of bytes, many of the operations you see supported on vectors are also supported on `String`, but with some extra guarantees.
-* Compare the different ways to inde a Strings by using `s3[i]` and `s3.chars.nth(i).unwrap()` where `i` is in-bound, out-of-bounds, and "on" the flag unicode character.
+* Compare the different ways to index a String by using `s3[i]` and `s3.chars.nth(i).unwrap()` where `i` is in-bound, out-of-bounds, and "on" the flag unicode character.
```
jooyunghan, line 38: Thanks for the catch! 
Btw, I think we can fix more in this sentence. For example, s/String/`String` and s/`chars`/`chars()`

Besides, I'm curious what's the intent of `s3[i]` here. It's not available for `String`, right? (unless `i` is a range) Was it a typo for something else like `s3.as_bytes()[i]`?

mgeisler: The text is from #368, so we can ask @cchiw what she meant here?

mgeisler: > Btw, I think we can fix more in this sentence. For example, s/String/`String` and s/`chars`/`chars()`

It's often easier to do this in follow-up commits. Alternatively, we can use the suggestion feature in GitHub to quickly add more commits:

```suggestion
* Compare the different ways to index a `String` by using `s3[i]` and `s3.chars().nth(i).unwrap()` where `i` is in-bound, out-of-bounds, and "on" the flag Unicode character.
```

mgeisler: The nice part about that is that I can commit directly to the branch (thanks @adamac for marking the branch as accepting changes from maintainers) and thus we can get fixes in super quick :smile:

mgeisler (approved):

Issue #434, opened 2023-02-17
Make `i18n-helpers` a requirement for running `mdbook`


mgeisler:
We can simplify the setup a bit by making the `i18n-helpers` a mandatory requirement for running `mdbook`.

I was originally hesitant to complicate the setup instructions, but since people already have to install `mdbook-svgbob`, I now think it will be fine to also ask them to install `i18n-helpers`.

We should also think about publishing `i18n-helpers` to crates.io so that people can install it easily for other projects. That's more of a long-term plan, though.

mgeisler:
Doing this would remove some warnings triggered by #429 and #421.

jooyunghan:
Fixed with #461 

Publishing it could be another story.

PR #461, opened 2023-02-28
Make `i18n-helpers` a requirement

jooyunghan:
plus `gettext` now finds a PO file based on `book.language` for simpler setup.


jooyunghan:
This should fix #434

i18n-helpers/src/bin/mdbook-gettext.rs
```diff
@@ -66,19 +68,30 @@ fn translate(text: &str, catalog: &Catalog) -> String {
...
-        .ok_or_else(|| anyhow!("Missing preprocessor.gettext.po-file config value"))?
-        .as_str()
-        .ok_or_else(|| anyhow!("Expected a string for preprocessor.gettext.po-file"))?;
-    let catalog = po_file::parse(Path::new(path))
-        .map_err(|err| anyhow!("{err}"))
-        .with_context(|| format!("Could not parse {path} as PO file"))?;
+    let po_dir = cfg.get("po-dir").and_then(Value::as_str).unwrap_or("po");
+    let path = ctx.root.join(po_dir).join(format!("{language}.po"));
 
+    // no-op when PO file is missing
```
mgeisler, line 87: Would it not be clearer to error out if the file doesn't exist? I'm thinking that the normal case is to have `book.language` unset, so _if_ it is set, it means that the user has actively asked for a non-standard language. This in turn makes me think that the user would like to be told about a missing PO file.

book.toml
```diff
@@ -13,14 +13,14 @@ extra-watch-dirs = ["po"]
...
-before = ["gettext"]
+
+[preprocessor.gettext]
 
 [preprocessor.index]
 renderers = ["html"]
 
 [preprocessor.svgbob]
 renderers = ["html"]
-after = ["gettext"]
```
mgeisler, line 23: Ah, are you able to remove this because the preprocessor order is determined by the order in `book.toml`? That's nice!

mgeisler (approved):

book.toml
jooyunghan: Yup, you're right.

i18n-helpers/src/bin/mdbook-gettext.rs
jooyunghan: It's for the English (the original book), which requires no translation but sets `book.language = en`. 

I was a bit hesitant as well for this _fallback_. Probably we can add `skip-for-languages = ["en"]` kind of setting.

jooyunghan: And I was thinking additional "language specific override" like following:
```
[preprocessor.gettext]
en.skip = true
ko.po-file = "path-to-ko.po"
```

mgeisler: > It's for the English (the original book), which requires no translation but sets `book.language = en`.

Ah... I had forgotten that we set the language even in that case.

> And I was thinking additional "language specific override" like following:
> 
> ```
> [preprocessor.gettext]
> en.skip = true
> ko.po-file = "path-to-ko.po"
> ```

Okay, do you have a use case for this right now?

PR #468, opened 2023-03-01
Remove Last-* properties in po/ko.po

jooyunghan:


mgeisler:
Looks good — but I would like to apply it uniformly across all our translations. Could you update the other `.po` files too?

mgeisler:
I just checked, and `msgmerge` will not update the date fields. This is probably because we don't populate the `POT-Creation-Date` field at all. So we should be good on that front. Other translators might have editors which touch the fields and that's okay for me.

jooyunghan:
Cleaned up other po files as well. I hope this doesn't bother anyone ;-)

mgeisler (approved):
Thanks!

PR #480, opened 2023-03-06
Update mdbook to 0.4.28

mgeisler:
This version contains https://github.com/rust-lang/mdBook/pull/1986 which will allow us to test the code in each translation.


jooyunghan:
Just curious, why did you need to update ko.po?

However, the formatting for the changed text in English looks like:
image

And in Korean:
image

I think the original text should have used ``` instead of \`.

The Korean translation seems wrong ;-) I'll fix it in a next round.

mgeisler:
> I think the original text should have used ``` instead of `.

Yes, I completely agree: the formatting in English is weird and should be fixed there. I'm sorry I didn't spot that in #169 (@qwandor).

The golden rule of working with i18n is that you cannot "fix" anything in the translations, apart from typos in the source text. You have to follow the original in terms of formatting, otherwise the two versions will start to drift apart and that makes your life very hard when it comes to tracking changes.

PR #483, opened 2023-03-07
Reuse Rust cache across different jobs

mgeisler:
By default, the rust-cache action will derive a cache key from the GitHub action job. That is, our `translations` job uses a different cache from our `i18n-helpers` job.

This is not useful in our case since all our jobs install roughly the same set of binaries: `mdbook`, `mdbook-svgbob`, `i18n-helpers`. We should therefore use a single cache for all of these jobs.

If we’re wrong, we pay with an increase in compilation time: regardless of what we cache, Cargo will install the right thing when asked.

### DIFF
--- a/.github/workflows/setup-rust-cache/action.yml
+++ b/.github/workflows/setup-rust-cache/action.yml
@@ -7,3 +7,6 @@ runs:
   steps:
     - name: Setup Rust cache
       uses: Swatinem/rust-cache@v2
+      with:
+        prefix-key: "v0"
+        shared-key: "rust"


mgeisler:
Hmm, this doesn't work very well: a version of the cache was created without `~/.cargo/bin/mdbook`, and now the job that creates this file thinks the cache is up-to-date.

mgeisler:
I believe this doesn't work because `rust-cache` uses filenames for caching instead of looking at the mtime or using a checksum: https://github.com/Swatinem/rust-cache/issues/59.

Luckily, the caches fix themselves after some time as the old ones expire. So I think we can close this for now.

PR #486, opened 2023-03-08
Make Miri happy

qwandor:
I think this is a better solution than #308. It fixes the Miri error, and mostly keeps the original intent of demonstrating how to use raw pointers.


mgeisler (approved):
Thanks!

PR #487, opened 2023-03-08
Generalize the day-4 afternoon

djmitche:
This is in preparation for adding more options for this portion of the course, and reflects an existing practice of substituting other materials for this last half-day.

See #181 for the first of those new topics.

PR #492, opened 2023-03-09
Introduce an alternative day-4-afternoon: async

### LINKED ISSUE #181: Write chapter about async Rust
We should write a chapter about async Rust. This would be a good alternative to the Android part on Day 4. A possible agenda could look like this:

* Introductory Example
* Terminology
* Futures
* `await`
* Runtimes
* Spawning
* IO-Compatible Cunctions

djmitche:
Fixes #181.

### DIFF
--- a/src/async.md
+++ b/src/async.md
@@ -1,0 +1,17 @@
+# Async Rust
+
+"Async" is a concurrency model where multiple tasks are executed concurrently by
+executing each task until it would block, then switching to another task that is
+ready to make progress. The model scales to higher concurrency than threads
+because the per-task overhead is typically very low and operating systems
+provide means of efficiently selecting tasks that can make progress.
+
+## Comparisons
+
+ * Python has a similar model in its `asyncio`. However, its `Future` type is
+   callback-based, and not polled. Async Python programs require a "loop",
+   similar to an executor in Rust.
+
+ * JavaScript's `Promise` is similar, but again callback-based. The language
+   runtime implements the event loop, so many of the details of Promise
+   resolution are hidden.
--- a/src/async.md
+++ b/src/async.md
@@ -2,15 +2,23 @@
 
 "Async" is a concurrency model where multiple tasks are executed concurrently by
 executing each task until it would block, then switching to another task that is
-ready to make progress. The model scales to higher concurrency than threads
-because the per-task overhead is typically very low and operating systems
-provide means of efficiently selecting tasks that can make progress.
+ready to make progress. The model allows running a larger number of tasks on a
+limited number of threads. This is because the per-task overhead is typically
+very low and operating systems provide primitives for efficiently identifying
+I/O that is able to proceed.
+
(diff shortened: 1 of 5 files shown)


djmitche:
Note that this depends on #487.  Once that lands, I'll rebase and the diff will contain only the async bits.

src/async.md
```diff
@@ -0,0 +1,17 @@
+# Async Rust
+
+"Async" is a concurrency model where multiple tasks are executed concurrently by
+executing each task until it would block, then switching to another task that is
+ready to make progress. The model scales to higher concurrency than threads
```
rbehjati, line 5: This sentence is a bit vague in my opinion. I prefer something like: 
```suggestion
ready to make progress. The model allows running a larger number of concurrent tasks on a limited number of threads...
```

src/async.md
```diff
@@ -0,0 +1,17 @@
...
+because the per-task overhead is typically very low and operating systems
+provide means of efficiently selecting tasks that can make progress.
```
rbehjati, line 7: nit: It sounds like there is inherent OS support for async/await. Perhaps reword to something like "the async/await implementation relies on OS primitive to efficiently select the next tasks to run".

src/async.md
```diff
@@ -0,0 +1,17 @@
...
+
```
rbehjati, line 8: Can you add a few words here about the concept of a "Future", since it is mentioned below?

src/async/async-await.md
```diff
@@ -0,0 +1,52 @@
+# `async`/`await`
+
+At a high level, async Rust code looks very much like "normal" sequential code:
+
+```rust,editable
+use tokio::time;
```
rbehjati, line 6: Is it possible to skip using `tokio` in this first example, and use futures and `block_on`?

rbehjati:
> rebase'd
> 
> @rbehjati do you want to write some of these chapters? I think you can push to this PR, or I could rebase this PR to point to an upstream branch and we could make a few PRs against that branch.

Sure. I can help. Most likely will have time to work on it in late March.

src/async/async-await.md
djmitche: `futures` does not provide any time-based waking functionality. The only functionality in `futures` that would be non-trivial would be to send data over a channel, but I want to keep this example simple and parallel to the threads example.

Using `futures` for one of the later examples is a good idea, though.

src/async.md
djmitche: I think I addressed this, but I'm happy for more editing :)

djmitche:
> Also, I think it should be possible to add the content gradually (in separate PRs) and only add a link to it in src/SUMMARY.md when we are happy with the content.

I think this won't work, or at least would be confusing; this PR itself is getting errors because SUMMARY is incorrect.

Instead, I'll push to an `issue181` branch on the upstream repo, and we can both make PRs against that.

PR #493, opened 2023-03-10
Update Korean Translation (20230310)

jooyunghan:


jooyunghan:
I found that lots of fuzzy entries were due to new xgettext(?) around `<details>` and code blocks.

PR #496, opened 2023-03-10
Add an "async" session

djmitche:
This is a continuation of #492, but against a branch in the upstream repo so that @rbehjati and I can both make PRs against it.  Once the session is complete we can merge this PR to put it into `main`.


src/async/pitfalls.md
```diff
@@ -0,0 +1,9 @@
+# Pitfalls of async/await
+
+Async / await provides convenient and efficient abstraction for concurrent asynchronous programming. However, the async/await model in Rust also comes with its share of pitfalls and footguns. We illustrate some of them in this chapter:
+
+---
```
rbehjati, line 5:
```suggestion
```

nit: This formatting is not consistent with the rest of the course.

src/async/pitfalls/blocking-executor.md
```diff
@@ -0,0 +1,60 @@
...
+
+Most async runtimes only allow IO tasks to run concurrently.
+This means that CPU blocking tasks will block the executor and prevent other tasks from being executed.
+An easy workaround is to use async equivalent methods where possible.
+
+```rust,editable,compile_fail
+use futures::future::join_all;
+use std::time::Instant;
+
+// Uncomment to try with `spawn_blocking` around `std::thread::sleep`.
```
rbehjati, line 11: There is nothing to uncomment.

src/async/pitfalls/blocking-executor.md
```diff
@@ -0,0 +1,60 @@
...
+    } else {
+        std::thread::sleep(std::time::Duration::from_millis(duration_ms));
+    }
+}
+
+async fn tokio_sleep_ms(duration_ms: u64) {
+    tokio::time::sleep(tokio::time::Duration::from_millis(duration_ms)).await;
+}
+
+// Single threaded executor for better reproducibility in runtime.
```
rbehjati, line 30: I don't get this comment. Did you mean to make is single threaded?

src/async/pitfalls/blocking-executor.md
```diff
@@ -0,0 +1,60 @@
+# Blocking the executor
+
+Most async runtimes only allow IO tasks to run concurrently.
+This means that CPU blocking tasks will block the executor and prevent other tasks from being executed.
+An easy workaround is to use async equivalent methods where possible.
+
+```rust,editable,compile_fail
```
rbehjati, line 7: This example looks too complicated to me. I suggest starting with a simple version, then adding steps in the speaker notes for building on top of it. You already have some steps in the speaker notes, perhaps move some of the code blocks here for reference to the speaker notes.

src/async/pitfalls/blocking-executor.md
```diff
@@ -0,0 +1,60 @@
...
+
+```
+
+<details>
+
+- Using `std::thread::sleep` blocks the thread, so it prevents the executor from running. It means that while all futures are spawned at the same time, they all run one after the other. The runtime is the sum of all the `sleep` times. Try changing the runtime to `multi_thread` in a multi core environment to see how it impacts the run time.
+- A simple fix is to use `tokio::time::sleep`. Now, the `sleep` calls are `async` and they are properly scheduled by the executor.
+- Another fix would be to `tokio::task::spawn_blocking` which spawns an actual thread and transforms its handle into a future without blocking the executor. This thread is also scheduled as part of the executor's threadpool to grant better performance.
+
+- You should not think of tasks as OS threads. They do not map 1 to 1 and most executors will allow many tasks to run on a single OS thread. This creates multiple gotchas:
```
rbehjati, line 54: Does this note belong to the page on tasks?

src/async/pitfalls/async-traits.md
```diff
@@ -0,0 +1,50 @@
...
+    let sleepers: Vec<Box<dyn Sleeper>> = vec![
+        Box::new(FixedSleeper { sleep_ms: 50 }),
+        Box::new(FixedSleeper { sleep_ms: 100 }),
+    ];
+    run_all_sleepers_multiple_times(sleepers, 5).await;
+}
+```
+
+<details>  
+
```
rbehjati, line 46: Perhaps add a note on why this is not supported (i.e., the resulting type not having a known type at compile time), and the overhead of using async_trait (because it does heap allocation).

rbehjati (commented):
Thanks for putting this together. 

My general impression is that this is more like a chapter on `tokio` than `async`. Is that the intention? In particular, I think most of the basic concepts can be demonstrated using the futures crate, and with less boilerplate code. 

Also, I remember there were some discussions about whether the content is for a full-day or a half-day session. Looks like we are aiming for the latter (i.e., day 4 afternoon after a morning on concurrency). This means we have only 2 hours for teaching this content. I feel like the content is too heavy for only 2 hours.

rbehjati:
> As for whether it's a `tokio` tutorial or not: the options for async things that will run in the playground are pretty limited, and in particular `tokio` appears to be the only available runtime. Is that what you're referring to, or are you concerned with the more particular bits about channels and async mutexes and so on?

I was thinking mostly about the first few slides. I think it is too much to talk about `tokio` and runtimes in the first or second slides. I can try and rewrite src/async/async-await.md without using `tokio`, or at least without using `time` and `spawn`. I think in that first slide, we just want to talk about the basic syntax. I feel like there are currently too many new things in that slide.

djmitche:
That's fair - let's see what you can come up with. I felt pretty constrained by having to write something that ran in the playground, but maybe I missed something, or maybe that constraint can be relaxed.

src/async/control-flow/select.md
```diff
@@ -0,0 +1,63 @@
...
+
+    assert_eq!(winner, Animal::Dog {name: String::from("Rex")});
+}
+```
+
+<details>
+
+* In this example, we have a race between a cat and a dog. `first_animal_to_finish_race` listens to both channels and will pick whichever arrives first. Since the dog takes 5 seconds, it wins against the cat that take 10 seconds.
+* You can use `oneshot` channels in this example as the channels are supposed to receive only one `send`.
+* You can try adding more contestants to the race and return a leaderboard. Also, you can add a deadline after which contestants get eliminated.
```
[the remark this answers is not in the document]
djmitche: I think the deadline part of this "Try .." feels less contrived, so I'll stick with that.

src/async/pitfalls/async-traits.md
djmitche: By "supported", do you mean why this isn't stabilized?

src/async/pitfalls/blocking-executor.md
djmitche: I think it makes sense here, as this is the incorrect intuition that might lead someone to block in a task.

src/SUMMARY.md
```diff
@@ -225,6 +225,27 @@
...
 
+# Day 4: Afternoon (Async)
+
+----
+
+- [Async](async.md)
+  - [async/await](async/async-await.md)
+  - [Async Blocks](async/async-blocks.md)
+  - [Futures](async/futures.md)
+  - [Runtimes](async/runtimes.md)
```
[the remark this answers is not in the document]
djmitche: I'm not sure how or where.. they are simple to understand but a little more complicated to demonstrate.

djmitche (commented):
OK! I've made a PR addressing most of these comments: #556.

src/async/pitfalls/async-traits.md
rbehjati: Yes :)

rbehjati: Sure. I can add something once you've merged #556. I think I learnt about it in an episode of Crust of Rust. We could also leave it out, as it might be too much detail. 

I think the instructors can choose to go into more depth as they see fit when teaching the course. We don't need to cover everything in the notes, but since the speaker notes for this section are are already so comprehensive, I was tempted to suggest adding this too :)

djmitche:
I want to use some new stuff that's in `main` for the exercises, so I'll do that in a separate PR.  This is ready for review.

src/async/pitfalls/async-traits.md
sakex: If think I know why but I would have a hard time explaining, so let me provide a pointer and you can explain better than me:

If you had a trait with `fn fetch_data_count(&self) -> usize`, then the compiler knows that it needs to push 64 bits to the stack for the return value, whatever the type of the caller.

Now if we had `async fn fetch_data_count(&self) -> usize`, since we actually return a future, the stack of the future needs to also be allocated, so the size of the return value depends on the actual method.

Consider both implementations:
```
async fn fetch_data_count(&self) -> usize {
    let data: [i32; 1000] = fetch().await;
    let filter: i32 = fetch_filter().await;
    data.iter().filter(|&item| item != filter).count()
}
```

```
async fn fetch_data_count(&self) -> usize {
    let data: Vec<i32> = fetch().await;
    let filter: i32 = fetch_filter().await;
    data.iter().filter(|&item| item != filter).count()
}
```

Example one, with the data on the stack probably has a higher size than the second one with the Vec. Using dynamic dispatch, it would be impossible for the compiler to know ahead of time the size.

sakex (approved):

src/async/pitfalls/async-traits.md
sakex: Heap allocating, fixes the size issue. Note that, Box does not always allocate and the compiler is often able to omit allocations if it can figure out the size

rbehjati: Yes. Exactly. And I just noticed that I have a typo in my original comment; it should be "the resulting type not having a known **size** at compile time". 

We cannot add all the details in the notes, so I was thinking of just a two sentence summary. I'll create a PR.

rbehjati (approved):
Looks great. Thanks.

src/async/pitfalls/pin.md
```diff
@@ -0,0 +1,110 @@
...
+                _ = &mut timeout_fut => { println!(..); },
+            }
+        }
+        ```
+
+    * This compiles, but once the timeout expires it is `Poll::Ready` on every
+      iteration (a fused future would help with this). Update to reset
+      `timeout_fut` every time it expires.
+
+* Box allocates on the heap. In some cases, `tokio::pin!` is also an option, but
```
mauvealerts, line 106: Maybe prefer or mention `std::pin::pin!`?

We're at an awkward point where the standard library has a pin macro (as of 1.68, stable on 2023-03-09) but
* It might not be supported if they haven't updated, or used Rust from a distro
* Folks are likely to run into the Tokio macro in existing code

src/async/tasks.md
```diff
@@ -0,0 +1,64 @@
...
+async fn main() -> io::Result<()> {
+    let listener = TcpListener::bind("127.0.0.1:6142").await?;
+	println!("listening on port 6142");
+
+    loop {
+        let (mut socket, addr) = listener.accept().await?;
+
+        println!("connection from {addr:?}");
+
+        tokio::spawn(async move {
```
mauvealerts, line 25: This is the first time async blocks appear (also alluded to in async-await and pin sections). They probably merit a sentence somewhere.

src/async/tasks.md
```diff
@@ -0,0 +1,64 @@
+# Tasks
+
+Runtimes have the concept of a "Task", similar to a thread but much
+less resource-intensive.
+
+A Task has a single top-level Future which the executor polls to make progress.
```
mgeisler, line 6:
```suggestion
A task has a single top-level future which the executor polls to make progress.
```

Alternatively, do you want to mark this as code:

```suggestion
A `Task` has a single top-level `Future` which the executor polls to make progress.
```

src/running-the-course/day-4.md
```diff
@@ -23,5 +23,18 @@ Ensure that `adb sync` works with your emulator or real device and pre-build
 all Android examples using `src/android/build_all.sh`. Read the script to see
 the commands it runs and make sure they work when you run them by hand.
 
+## Async
+
+If you chose Async for Day 4 afternoon, you will need a fresh crate set up and
+the dependencies downloaded and ready to go. You can then copy/paste the
+examples into `src/main.rs` to experiment with them.
```
mgeisler, line 30: I would suggest adding the large async examples to the repository as full Cargo projects (similar to how the Android chapter handles things).

That makes it trivial for an instructor to find known-good versions of the code, and we also get coverage from `cargo build` when you add those projects to the workspace for the repository.

mgeisler (approved):
Thanks @djmitche, @rbehjati, and @sakex for all your work on this!

I really like the notes, they helped me a lot in understanding the slides.

I've left a few small comments which can be fixed after merging the PR — I think it'll be great to get this published so that we can get some more real-world feedback on it.

src/exercises/day-4/async.md
```diff
@@ -0,0 +1,3 @@
+# Exercises
+
+TBD
```
mgeisler, line 3: I think you actually decided on this in #557?

djmitche: I did, and was planning to land this PR before that one was done. I suppose I can just put it in here now :)

djmitche:
Thanks! I'll make up some new PRs and/or issues to address these items.

src/async/tasks.md
djmitche: That's mentioned in the speaker notes. In general, it's a non-goal to put all explanation on the slide itself.

mauvealerts: Ah, this was a failure in my search. I agree the note in speaker notes is good and sufficient.

For context, I noticed `async move` later and was trying to find where async blocks was explained/mentioned. Both "async" and "block" appear a lot, so I searched for `async block`, `async {`, or `async move`. (I don't think there's anything actionable here.)

src/running-the-course/day-4.md
mgeisler: Thanks!

PR #497, opened 2023-03-10
Add futures page

djmitche:
This adds a page about futures, showing how `async` and `await` interact with them.

### DIFF
--- a/src/async/futures.md
+++ b/src/async/futures.md
@@ -41,7 +41,7 @@ pub enum Poll<T> {
 
 An async function returns an `impl Future`, and an async block evaluates to an
 `impl Future`. It's also possible (but uncommon) to implement `Future` for your
-own types. For example, the `JoinHandle` returns from `tokio::spawn` implements
+own types. For example, the `JoinHandle` returned from `tokio::spawn` implements
 `Future` to allow joining to it.
 
 The `.await` keyword, applied to a Future, causes the current async function or
--- a/src/async/futures.md
+++ b/src/async/futures.md
@@ -2,7 +2,7 @@
 
 What is the type of an async operation?
 
-```rust, editable
+```rust,editable,compile_fail
 use tokio::time;
 
 async fn count_to(count: i32) -> i32 {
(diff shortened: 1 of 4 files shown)


src/async/futures.md
```diff
@@ -0,0 +1,51 @@
...
+
+pub enum Poll<T> {
+    Ready(T),
+    Pending,
+}
+```
+
+An async function returns an `impl Future`, and an async block evaluates to an
+`impl Future`. It's also possible (but uncommon) to implement `Future` for your
+own types. For example, the `JoinHandle` returns from `tokio::spawn` implements
```
rbehjati, line 22:
```suggestion
own types. For example, the `JoinHandle` returned from `tokio::spawn` implements
```

I think!

src/async/futures.md
```diff
@@ -0,0 +1,51 @@
...
+
+The `.await` keyword, applied to a Future, causes the current async function or
+block to pause until that Future is ready, and then evaluates to its output.
+
+<details>
+
+* These types are conceptually quite simple, and implemented as such in
+  `std::task`.
+
+* We will get to `Pin` and `Context` shortly, so don't get into any depth on
```
rbehjati, line 44: Cool. So you actually want to talk about `Pin` too.

rbehjati (approved):
Thanks :)

src/async/futures.md
djmitche: Based on discussion in the issue, I'm not sure it's a good idea, actually.  The ecosystem is _almost_ to the point where understanding Pin is only required for people working on low-level plumbing, and the rest of us can happily type `.await` and `async` and assume things just work.

Maybe I'll refocus this to a "how futures work, simplified" page and not revisit it in more detail.

rbehjati: I like that. But maybe @mgeisler has other ideas :)

mgeisler: I'm happy to wait a bit and see where the three of you take this — you know more about async than me!

PR #498, opened 2023-03-10
The code block in 27.1 probably should not be runnable

dgorelik:
Trying to run it causes an error since there is no `main`.


mgeisler:
Yes, good point... Thanks for fixing it!

mgeisler (approved):

PR #499, opened 2023-03-11
Remove unnecessary `Mutex::lock` from shared state example

tekumara:
It's not necessary to explicitly take the lock to print the Vec.

I'm using rustc 1.66.1


mgeisler (approved):
Thanks, that looks better indeed!

PR #506, opened 2023-03-14
Don't allocate in success case

qwandor:
`format!` was being called (and allocating a string) even in the success case.


mgeisler (approved):
I know Clippy suggests this, but I hate to think that it wouldn't optimize to the same code.

There must be an if statement somewhere that guards the _use_ of the string. Now we instead guard the _construction_ of the string. What I'm hoping for is that the compiler will already recorder things so that the string construction is also guarded by the if statement.

I should of course look at the assembly to check this 😄

mgeisler:
Can you find a way for us to run Clippy on all our cod?

PR #507, opened 2023-03-14
On mac use Cmd-Enter to run the code blocks

rbehjati:


mgeisler:
Can you check if you can simply add another binding instead of changing the current one? It would be nice to be consistent with other `mdbook` projects and so I would prefer to keep the current binding.

mgeisler (changes requested):
Please also update the page about keyboard shortcuts.

rbehjati:
> Can you check if you can simply add another binding instead of changing the current one? It would be nice to be consistent with other `mdbook` projects and so I would prefer to keep the current binding.

I agree. Multiple bindings would have been nice, but I could not find a way to do this. 
Also, it looks like most people have re-mappings that makes `Ctrl-Enter` work on mac. So I am going to close this :)

mgeisler:
> I agree. Multiple bindings would have been nice, but I could not find a way to do this.
> Also, it looks like most people have re-mappings that makes `Ctrl-Enter` work on mac. So I am going to close this :)

Okay! Thanks for looking!

Issue #491, opened 2023-03-09
Include doc comments


djmitche:
Somewhere in the syntax section, we should address doc comments and how those appear in rustdocs.

mgeisler:
We cover this indirectly here: https://google.github.io/comprehensive-rust/testing/doc-tests.html. I normally use that slide as a starting point for talking about how real code is documented.

PR #509, opened 2023-03-15
Add a brief mention of doc comments

djmitche:
This addresses both inner and outer doc comments, and that documenting things is idiomatic Rust.

The first commit here removes a redundant slide about methods. I did this partly to "make space" for the docstrings during the first day, and partly because it's probably a little early to talk about methods before talking about types.


src/basic-syntax/methods.md
qwandor, line 1: The trouble with removing this page is that the day 1 afternoon exercises use methods. I agree that the duplication is bad but I'm not sure this is an improvement.

src/modules/filesystem.md
```diff
@@ -21,6 +21,15 @@ The `crate` root is in:
 * `src/lib.rs` (for a library crate)
 * `src/main.rs` (for a binary crate)
 
+Modules defined in files can be documented, too, using "inner doc comments".
+These document the item that contains them -- in this case, a module.
+
+```rust,editable,compile_fail
+//! This module implements the garden, including a highly performant germination
+//! implementation.
+pub mod vegetables;
```
qwandor, line 30: I'm not sure why the submodule here is necessary, but if you want to have it then I suggest at least adding a blank line before it, to make clear that the Rustdoc comment is not associated with it.

src/basic-syntax/doc-comments.md
```diff
@@ -0,0 +1,27 @@
+# Documentation Comments
```
qwandor, line 1: Rustdoc is the name I've usually heard used for this, where are they called Documentation Comments?

It would be good to link to https://doc.rust-lang.org/rustdoc/what-is-rustdoc.html too.

src/basic-syntax/doc-comments.md
```diff
@@ -0,0 +1,27 @@
...
+/// If the second argument is zero, the result is false.
+fn is_divisible_by(lhs: u32, rhs: u32) -> bool {
+    if rhs == 0 {
+        return false;  // Corner case, early return
+    }
+    lhs % rhs == 0     // The last expression in a block is the return value
+}
+```
+
+The contents are treated as Markdown. All published Rust code is automatically
```
qwandor, line 17: More specifically, all Rust library crates (not binary crates) published to crates.io.

qwandor (commented):
Thanks!

src/basic-syntax/methods.md
djmitche: Hm, I'll leave it in and file an issue.

djmitche: #510

mgeisler:
Hi @djmitche, I don't think we should include this here — in my experience, the first day is already crammed full with other useful things.

I love that you wanted to remove the page on methods. As @qwandor points out, it was added to make it easier for people to work with them in [the exercise](https://google.github.io/comprehensive-rust/exercises/day-1/book-library.html). Though, in my experience, people grasp the basics of methods very quickly.

Longer term, we need to rework that exercise so that we don't talk about methods yet (but that's for a later refactoring).

As for this PR, I would like to move the content to a later point. Like I wrote on #491, we already talk about doc comments when we talk about testing. (This reminds me, if you mention the issue number in your PR description, things will be linked nicely).

qwandor:
I disagree with @mgeisler on this. I've found that the first day is the shortest, we finish well before the allotted time for both morning and afternoon sessions even with a bunch of questions, whereas the other days are longer. In the current course I'm teaching I got several questions about Rustdoc on day 2 or 3, especially when doctests were mentioned on day 3, so I think it's worth mentioning Rustdoc earlier when talking about basic syntax to get people used to it.

qwandor (approved):

djmitche:
> This reminds me, if you mention the issue number in your PR description, things will be linked nicely

This isn't required -- you can also link it under "Development" on the right, which I've done :)

src/basic-syntax/doc-comments.md
djmitche: BTW these are referred to as doc comments in the language reference - https://doc.rust-lang.org/reference/comments.html

djmitche:
I want to push back a bit, too. Idiomatic rust has its public items documented, and I think people should hear that early, at the same time as other basic syntax items. Implying the existence doc comments in the section on testing, right up against other "advanced" concepts like unsafe, seems to indicate they are not very important.

It feels like day 1 is a little bit of "let's get enough Rust by noon to do some exercises".  I notice that `match` and `if` and `for` are already in there, too, all of which are actually defined on day 2.  So perhaps it's too early to include information that's not _required_ to be successful with the exercises.

Still, I'd like to address it explicitly somewhere. The day-1 afternoon seems to focus on ownership and isn't a good fit for doc comments. What if we moved this to the "Structs" section on the second day?

mgeisler (approved):
Okay, sure, let's take this in for now :smile:  I'll be reworking things over the next few weeks but it's good content to have in general.

djmitche:
On the first day, or moved to Structs?

PR #512, opened 2023-03-16
Run `cargo fmt` in presubmit

qwandor:
I ran `cargo fmt` once to format the existing code, and added a presubmit workflow to check it automatically in future.


qwandor:
This was merged as part of #320.

PR #514, opened 2023-03-19
fix: fix typo at ko.po

eunsukimme:


jiyongp (approved):
Thanks for contributing!

PR #522, opened 2023-03-22
Add Runtimes & Tasks

djmitche:
These concepts are closely related, and there's not much else to know about runtimes other than "they exist".

This removes the bit about futures being "inert" because it doesn't really lead anywhere. Maybe it should be in the speaker notes?

### DIFF
--- a/src/async/runtimes.md
+++ b/src/async/runtimes.md
@@ -15,67 +15,15 @@ Several larger applications have their own runtimes. For example,
 [Fuchsia](https://fuchsia.googlesource.com/fuchsia/+/refs/heads/main/src/lib/fuchsia-async/src/lib.rs)
 already has one.
 
-## Tasks
-
-All of these runtimes have the concept of a "Task", similar to a thread but much
-less resource-intensive.
-
-A Task has a single top-level Future which the executor polls to make progress.
-That future may have one or more nested futures that its `poll` method polls,
-corresponding loosely to a call stack. Concurrency is possible within a task by
-polling multiple child futures, such as racing a timer and an I/O operation.
-
-```rust,editable,compile_fail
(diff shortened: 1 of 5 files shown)


src/async/runtimes.md
```diff
@@ -0,0 +1,81 @@
...
+   [Tonic](https://github.com/hyperium/tonic) for gRPC.
+ * [async-std](https://async.rs/) - aims to be a "std for async", and includes a
+   basic runtime in `async::task`.
+ * [smol](https://docs.rs/smol/latest/smol/) - simple and lightweight
+
+Several larger applications have their own runtimes. For example,
+[Fuchsia](https://fuchsia.googlesource.com/fuchsia/+/refs/heads/main/src/lib/fuchsia-async/src/lib.rs)
+already has one.
+
+## Tasks
```
rbehjati, line 18: I think it is still better to split it into two pages. This is too much content in one page when teaching.

src/async/runtimes.md
```diff
@@ -0,0 +1,81 @@
...
+                println!("socket error: {e:?}");
+            }
+        });
+    }
+}
+```
+
+<details>
+
+Note that of the listed runtimes, only Tokio is supported in the Rust
```
rbehjati, line 70: Nice speaker notes :)

Perhaps change to a bullet pointed list, to be more consistent with the rest of the slides.

rbehjati (approved):

src/async/runtimes.md
djmitche: Done

src/async/runtimes.md
djmitche: Done

rbehjati (approved):
Thanks :)

PR #524, opened 2023-03-23
Async chapter

sakex:
Async control flow + Async channels


src/async/control-flow.md
```diff
@@ -0,0 +1,200 @@
...
+
+async fn race_finish_line(
+    mut cat_rcv: Receiver<String>,
+    mut dog_rcv: Receiver<String>,
+    timeout: Duration,
+) -> Option<Vec<AnimalRacePerformance>> {
+    let mut performances: Vec<AnimalRacePerformance> = Vec::new();
+    let timeout_sleep = sleep(timeout);
+    // See the chapter about pin.
+    tokio::pin!(timeout_sleep);
```
sakex, line 129: Alternatively, that can be done with a oneshot channel, so we don't need to talk about pin

src/async/control-flow.md
```diff
@@ -0,0 +1,200 @@
...
+    println!("The winner of the race is {:?}", winner);
+
+    assert_eq!(
+        winner,
+        Animal::Dog {
+            name: String::from("Rex")
+        }
+    );
+}
+```
```
djmitche, line 106: This example seems a bit lengthy. Could this start with a shorter, simpler example with ideas in the speaker notes to improve it?

src/async/control-flow.md
djmitche: Let's move this to the "pitfalls" section. AIUI the issue here is that you have a single Future that you'd like to `await` until it is Ready. This is different from the futures returned from `.recv()`, which can be created fresh every time around the loop. Maybe the example (in the pitfalls section) can start with
```rust
loop {
  select! {
    ..
    _ = sleep(timeout) => break,
  }
}
```
and then understand that this starts the timeout over for every iteration of the loop, and work out how to solve it from there.

I have to be honest, I don't think I could explain why `tokio::pin` is necessary here without a good bit of refreshing myself!

src/async/pitfalls.md
```diff
@@ -0,0 +1,79 @@
+# Pitfalls of async/await
```
djmitche, line 1: Let's make each pitfall its own sub-chapter.

djmitche (commented):
Looking good! I think we want to keep each chapter pretty short, and at least initially the code examples should be short as well.

src/async/pitfalls.md
```diff
@@ -0,0 +1,79 @@
...
+
+    let now = Instant::now();
+    join_all(tokio_sleep_futures).await;
+    let runtime = now.elapsed().as_millis();
+    assert!(runtime < 150 && runtime >= 100);
+}
+```
+
+Using `std::sync::mutex` in an `async` runtime is also very dangerous. When you lock the mutex in a thread then yield the executor using `.await` the thread might try to lock the mutex once more in a different task.
+Hence, prefer `async` alternatives like `tokio::sync::mutex`.
```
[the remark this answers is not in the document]
sakex: It's actually more of a `footgun`. I had a bad experience with this in the past. I will address your comments tomorrow, thank you :)

sakex:
@djmitche I addressed your comments and added a section about `daemon` tasks. Please have a look. I will start working on my exercise later this week!

src/async/pitfalls/reentering-threads.md
```diff
@@ -0,0 +1,51 @@
+# Reentering threads
```
djmitche, line 1: I'm not sure how this is "reentering" - perhaps there's a better title?

djmitche (approved):
This looks good! I'll get it merged and we can continue to adjust in the issue181 branch.

PR #525, opened 2023-03-24
Use `panic-write` to get panics on the serial port

mgeisler:
This is helpful for debugging the compass exercise code.


qwandor:
This looks like a handy crate, but I'm wary of adding extra magic to the exercise without explanation. I think if we want to recommend using `panic-write` then we should add a page about it first.

mgeisler:
> This looks like a handy crate, but I'm wary of adding extra magic to the exercise without explanation.

I felt this was on the same level as the `panic-halt` crate which is use already in a few places: the name tells me what it does and I can imagine that it emits a panic handler which writes into the value I register.

mgeisler:
Thanks for looking at the code, I created https://github.com/icewind1991/panic-write/issues/1 based on what you wrote above.

PR #526, opened 2023-03-24
Format title consistently

mgeisler:


qwandor (approved):
I think I had a reason for this, but fair enough.

Issue #283, opened 2023-01-26
Greek translation


mgeisler:
For reference: https://github.com/TChatzigiannakis/rust-book-greek.

PR #534, opened 2023-03-29
Add initial and incomplete Greek translation

Mandragorian:
Part of #283.


mgeisler (approved):
Yep, that's all Greet to me! :smile: Thanks!

PR #538, opened 2023-03-30
Add tool to build exercise templates

### LINKED ISSUE #529: Build zip files for bare metal exercises
The bare metal exercises currently require copying and pasting a lot of files to get started, which is annoying and error-prone. Some students check out the repository instead, but then they get model solutions rather than just the starting template. It would be better to have a script and GitHub action which automatically builds the template, using the same marker comments as the mdbook includes use, and then makes it available to download somewhere (maybe as a release or artifact?) linked from the mdbook.

qwandor:
This will parse the Markdown in `src/exercises/bare-metal/compass.md` and `src/exercises/bare-metal/rtc.md` and resolve the `#import` directives to build starter templates for the exercises, and publish it as a zip file to Pages along with the book.

For now this is only used for the bare metal exercises, as they have quite a few files.

Fixes #529.

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -50,4 +50,4 @@ editable = true
 "generics/trait-bounds.html" = "traits/trait-bounds.html"
 
 [output.exerciser]
-output-directory = "exercise-templates"
+output-directory = "comprehensive-rust-exercises"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -133,7 +133,7 @@ jobs:
         run: mdbook build
 
       - name: Zip exercise templates for ${{ matrix.language }} translation
-        run: cd book/exerciser && zip --recurse-paths ../html/exercises.zip exercise-templates/
+        run: cd book/exerciser && zip --recurse-paths ../html/comprehensive-rust-exercises.zip comprehensive-rust-exercises/
 
       - name: Upload ${{ matrix.language }} translation
         uses: actions/upload-artifact@v3
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -41,7 +41,7 @@ jobs:
         run: mdbook build -d book
 
       - name: Zip exercise templates
-        run: cd book/exerciser && zip --recurse-paths ../html/exercises.zip exercise-templates/
+        run: cd book/exerciser && zip --recurse-paths ../html/comprehensive-rust-exercises.zip comprehensive-rust-exercises/
 
       - name: Build all translations
         run: |
@@ -50,7 +50,7 @@ jobs:
               MDBOOK_BOOK__LANGUAGE=$po_lang \
               MDBOOK_OUTPUT__HTML__SITE_URL=/comprehensive-rust/$po_lang/ \
               mdbook build -d book/$po_lang
-              (cd book/$po_lang/exerciser && zip --recurse-paths ../html/exercises.zip exercise-templates/)
+              (cd book/$po_lang/exerciser && zip --recurse-paths ../html/comprehensive-rust-exercises.zip comprehensive-rust-exercises/)
               mv book/$po_lang/html book/html/$po_lang
               echo "::endgroup::"
           done
--- a/mdbook-exerciser/README.md
+++ b/mdbook-exerciser/README.md
@@ -15,5 +15,12 @@ fn some_more_code() {
 ```
 ````
 
-It will create a file `book/exerciser/exercise-templates/example/src/main.rs` with the appropriate
-contents.
+and mdbook configuration in `book.toml` like:
+
+```toml
+[output.exerciser]
+output-directory = "comprehensive-rust-exercises"
+```
+
+It will create a file `book/exerciser/comprehensive-rust-exercises/example/src/main.rs` with the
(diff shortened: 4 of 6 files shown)


exerciser/src/main.rs
```diff
@@ -0,0 +1,155 @@
...
+    let input_directory = input_filename
+        .parent()
+        .with_context(|| "Input file has no parent directory.")?;
+    let input_contents = read_to_string(input_filename)
+        .with_context(|| format!("Failed to open {:?}", input_filename))?;
+    let parser = Parser::new(&input_contents);
+
+    let mut next_filename: Option<String> = None;
+    let mut current_file: Option<File> = None;
+    for event in parser {
```
mgeisler, line 59: Parsing Markdown like this makes me a little nervous: the coupling between the source and this script is far from obvious. Even reading this now, it's not obvious to me what the Markdown looks like which will trigger some useful action in this code :slightly_smiling_face: 

One thing I've noticed is that your pages don't follow the same style as the rest of the pages. Compare

image

with the existing style:

image

The way I would like to handle this is with a new preprocessor, similar to the `links` preprocessor which is responsible for resolving the `{{#include` syntax. The new preprocessor, call it `file` would take

~~~markdown
{{#file some/path/foo.rs:Foo}}
~~~

and expand it to

~~~markdown
_some/path/foo.rs_:

```rust
{{#include some/path/foo.rs:Foo}}
```
~~~

(It can guess the syntax based on the file extension.)

That would give us a consistent way to show files in the course.

Ideally, the preprocessor would also be able to emit something which we can recognize clearly in the rendered book (probably in a book rendered as Markdown). I'm thinking it could emit `<!-- Start include -->` or similar and we would then find that in a [`Event::Html`](https://docs.rs/pulldown-cmark/latest/pulldown_cmark/enum.Event.html#variant.Html), if I'm guessing correctly.

I would feel better about that than assigning as special meaning to a line of code followed by a code block.

mgeisler (changes requested):

exerciser/src/main.rs
```diff
@@ -0,0 +1,155 @@
...
+/// Writes the given output file based on the given code text from the Markdown input, processing
+/// include directives as necessary.
+fn write_output(
+    text: CowStr,
+    input_directory: &Path,
+    output_file: &mut File,
+) -> anyhow::Result<()> {
+    for line in text.lines() {
+        info!("Line: {:?}", line);
+        if let (Some(start), Some(end)) =
```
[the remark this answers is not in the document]
qwandor: Unfortunately it looks like it's not possible to run an mdbook preprocessor by itself without processing the whole book. It would be possible to do that, but it would make the process more complicated, and I'd rather keep this as a fairly simple lightweight tool if possible.

qwandor:
I've fixed the bug with translations, PTAL.

mgeisler (approved):
Thanks!

.gitignore
```diff
@@ -1,5 +1,6 @@
 # Build artifacts
 /book/
+/exercise-templates/
```
mgeisler, line 3: Is this actually necessary? When I ran the code on the branch, it created `book/exerciser/` as expected.

book.toml
```diff
@@ -48,3 +48,6 @@ editable = true
 "generics/closures.html" = "traits/closures.html"
 "generics/impl-trait.html" = "traits/impl-trait.html"
 "generics/trait-bounds.html" = "traits/trait-bounds.html"
+
+[output.exerciser]
+output-directory = "exercise-templates"
```
mgeisler, line 53: Could we use a name which is more distinct for both the top-level folder inside the zip file _and_ the zip file itself?

I would suggest `comprehensive-rust-exercises` for both.

exerciser/src/main.rs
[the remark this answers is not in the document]
mgeisler: > Unfortunately it looks like it's not possible to run an mdbook preprocessor by itself without processing the whole book.

Right, I was simply thinking you would call `mdbook build` with the right option to make it use the `markdown` renderer, then inspect the generated Markdown files in your temporary build output directory.

That would be an external tool which could be called as part of the publishing process -- instead of a tool which is called all the time like the renderer we have now.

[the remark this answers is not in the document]
mgeisler: > Unfortunately it looks like it's not possible to run an mdbook preprocessor by itself without processing the whole book.

Right, I was simply thinking you would call `mdbook build` with the right option to make it use the `markdown` renderer, then inspect the generated Markdown files in your temporary build output directory.

That would be an external tool which could be called as part of the publishing process -- instead of a tool which is called all the time like the renderer we have now.

exerciser/src/main.rs
mgeisler: > because it needs to handle including multiple sections interspersed with inline code

Right... I had not thought of that. The special comments are also nice and very flexible :+1:

.gitignore
qwandor: Not anymore, this was leftover from a previous iteration. Removed.

book.toml
qwandor: Done.

PR #539, opened 2023-03-31
Start Spanish translation

mgeisler:
This is a fixed up version of the branch in #495 by @wfranck. I applied the suggestions by @kevinislas2, except a few where there were some discussions that I don't know the answer to :smile: I'll let people who actually speak Spanish correct this in the future.


mgeisler:
I decided to merge this since the change was already approved in #495.

PR #540, opened 2023-03-31
Update if-let-expressions.md

### LINKED ISSUE #536: Missing let-else
I am not certain if this should be included or not, but in a number of cases students could encounter this construct in a language, so perhaps it might make sense to mention it -- the [let-else](https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html) construct.

nyurik:
add a note about let-else expressions.  Closes #536


src/control-flow/if-let-expressions.md
```diff
@@ -22,5 +22,20 @@ Rust.
     * For the similar use case consider demonstrating a newly stabilized [`let else`](https://github.com/rust-lang/rust/pull/93628) feature.
```
mgeisler, line 22:
```suggestion
```

mgeisler (approved):
Thank you! I saw that the feature was already mentioned in passing, so I removed that at the same time. I also indented the code block to make it nest correctly under the bullet point.

PR #551, opened 2023-04-05
Inline format vars to make the format str simpler

nyurik:
minor cleanup of a number of samples to inline format args


mgeisler (approved):
Thanks, this follows the style we've been using throughout most of the course (which Clippy now also suggests by default).

i18n-helpers/src/bin/mdbook-gettext.rs
```diff
@@ -90,7 +90,7 @@ fn translate_book(ctx: &PreprocessorContext, mut book: Book) -> anyhow::Result<B
 
     let catalog = po_file::parse(&path)
         .map_err(|err| anyhow!("{err}"))
-        .with_context(|| format!("Could not parse {:?} as PO file", path))?;
+        .with_context(|| format!("Could not parse {path:?} as PO file"))?;
```
mgeisler, line 93: Can I get you to submit this to https://github.com/google/mdbook-i18n-helpers/ instead? I moved the code to a stand-alone repository yesterday.

mgeisler: I'm removing the code you see here in #552 which should be merged later today, hopefully :smile:

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -31,7 +31,7 @@ use std::{fs, io};
 
 fn add_message(catalog: &mut Catalog, msgid: &str, source: &str) {
     let sources = match catalog.find_message(None, msgid, None) {
-        Some(msg) => format!("{}\n{}", msg.source(), source),
+        Some(msg) => format!("{}\n{source}", msg.source()),
```
qwandor, line 34: In cases like this where not everything can go inline, I think it's probably more readable to keep all format args consistently out of line. Having them out of order is confusing.

i18n-helpers/src/bin/mdbook-xgettext.rs
```diff
@@ -77,7 +77,7 @@ fn create_catalog(ctx: &RenderContext) -> anyhow::Result<Catalog> {
         })?;
         last_idx += idx;
         let lineno = summary[..last_idx].lines().count();
-        let source = format!("{}:{}", summary_path.display(), lineno);
+        let source = format!("{}:{lineno}", summary_path.display());
```
qwandor, line 80: Ditto.

mgeisler, line 80: I'm personally a bit in doubt about the mixed inlining and I think that's why I didn't inline it already. One approach I've seen people use is to do

```rust
format!("{foo}:{bar}", foo = complex + expressions)
```

I don't see it often, though, so I don't know if we want to bother with it. In any case, I'm fine with the partial inlining as well as a good first step.

mgeisler: @qwandor what do you think of the style above? I'm not 100% sure I like it mself :smile:

i18n-helpers/src/bin/mdbook-gettext.rs
mgeisler: @nyurik the PR was merged and this caused a few conflicts here... sorry about that. I hope you can revert the changes in your PR (and probably take out the controversial mixed cases as well just to make things simple).

i18n-helpers/src/bin/mdbook-xgettext.rs
qwandor: Personally I prefer the original style where nothing is inline.

nyurik: Thx for all the feedback.  I found the inline **tends to be** more readable if most/all of the values are inlined, and also it kinda forces me to move complex computations to a separate line, which helps readability as well.  I guess this specific one (tiny format string just to join two long values with a colon) is indeed more readable without inlining.

nyurik:
I merged this with the main branch, i hope most suggestions are now less controversial

mgeisler (approved):
Nice, this looks great now! Thanks a lot!

i18n-helpers/src/bin/mdbook-xgettext.rs
mgeisler: Yeah, I agree that it looks better in _some_ but not all cases :-) Having to create a local variable instead of using `summary_path.display()` directly is annoying for example. I often wish we could have just a _bit_ more complex expressions here — like in Python or JavaScript. However, that's a discussion for others to have :smile:

mgeisler: This got reverted.

i18n-helpers/src/bin/mdbook-xgettext.rs
mgeisler: This got reverted.

i18n-helpers/src/bin/mdbook-gettext.rs
mgeisler: Thanks for updating the PR!

PR #552, opened 2023-04-05
Use mdbook-i18n-helpers crate

mgeisler:
The i18n-helpers are now available as a stand-alone crate: https://crates.io/crates/mdbook-i18n-helpers.


mgeisler:
I downloaded the Korean translation from the build output and opened it locally — it looks fine.

PR #556, opened 2023-04-05
Modifications to the async section

djmitche:
This is a pretty heavy edit to `issue181` branch, in response to comments on #496.

* Remove the "Daemon" slide, as it largely duplicates the "Tasks" slide. The introduction to the "Control Flow" section mentions tasks as a kind of control flow.

* Reorganize the structure in SUMMARY.md to correspond to the directory structure.

* Simplify the "Pin" and "Blocking the Executor" slides with steps in the speaker notes to demonstrate / fix the issues.

* Rename "join_all" to "Join".

* Simplify some code samples to shorten them, and to print output rather than asserting.

* Clarify speaker notes and include more "Try.." suggestions.

* Be consistent about where `async` blocks are introduced (in the "Tasks" slide).

* Explain `join` and `select` in prose.

* Fix formatting of section-header slides.


src/async/control-flow/daemon.md
```diff
@@ -1,39 +1 @@
 # Daemon
```
sakex, line 1: File needs to be deleted

src/async/control-flow/join_all.md
```diff
@@ -1,48 +1 @@
...
-
-* `join_all` should soon be stabilized as part of the standard library in `std::future`.
-* For multiple futures of disjoint types, you can use `join!` but you must know how many futures you will have at compile time.
-* You can also combine `join_all` with `join!` for instance to join all requests to an http service as well as a database query.
-* The risk of `join` is that one of the future could never resolve, this would cause your program to stall. 
-* Try adding a timeout to the future.
-
-</details>
-
+# Join
```
sakex, line 1: :eyes:

sakex (approved):

sakex:
Thanks for simplifying the examples, it looks much better

src/async/pitfalls/blocking-executor.md
```diff
@@ -8,53 +8,41 @@ An easy workaround is to use async equivalent methods where possible.
...
+* Another fix would be to `tokio::task::spawn_blocking` which spawns an actual
+  thread and transforms its handle into a future without blocking the executor.
+  This is useful when calling external libraries that block the thread.
 
-- You should not think of tasks as OS threads. They do not map 1 to 1 and most executors will allow many tasks to run on a single OS thread. This creates multiple gotchas:
-  - For instance, using `std::sync::mutex` in an `async` runtime is very dangerous. When you lock the mutex in a thread then yield the executor using `.await` the thread might try to lock the mutex once more in a different task. Hence, prefer `async` alternatives like `tokio::sync::mutex`.
-  - Thread-local storage should also be used with care in async contexts as it doesn't map to specific tasks.
-  - Device drivers sometimes map to specific OS threads (for instance CUDA.) Prefer `tokio::task::spawn_blocking` when dealing with those.
-  - Some C libraries rely on thread local storage as well.
+* You should not think of tasks as OS threads. They do not map 1 to 1 and most executors will allow many tasks to run on a single OS thread. This creates multiple gotchas:
```
rbehjati, line 42: Perhaps format this part too :)

src/async/pitfalls/blocking-executor.md
```diff
@@ -8,53 +8,41 @@ An easy workaround is to use async equivalent methods where possible.
...
+  * For instance, using `std::sync::mutex` in an `async` runtime is very dangerous. When you lock the mutex in a thread then yield the executor using `.await` the thread might try to lock the mutex once more in a different task. Hence, prefer `async` alternatives like `tokio::sync::mutex`.
```
rbehjati, line 43: I think this particular topic is debatable. A more practical advice is to use `std::sync::mutex` for short critical sections, as async mutexes have performance overhead.

src/async/pitfalls/pin.md
```diff
@@ -1,108 +1,108 @@
 # Pin
 
-When you await a future, you effectively move the whole stack frame from which you called `.await` to an internal data structure of your executor. If your future has pointers to data on the stack, the addresses might get invalidated. This is extremely unsafe. Therefore, you want to guarantee that the addresses your future point to don't change. That is why we need to `pin` futures. In most cases, you won't have to think about it when using futures from common libraries unless you use `select` in a loop (which is a pretty common use case). If, you implement your own future, you will likely run into this issue.
+When you await a future, you effectively move the whole stack frame from which
```
rbehjati, line 3:
```suggestion
When you await a future, you effectively move the whole stack frame from where
```

I guess?

src/async/pitfalls/pin.md
```diff
@@ -1,108 +1,108 @@
...
-    println!("{:?}", finish_line_future.await.expect("Failed to collect finish line"));
-    // [Runner { name: "Milo" }, Runner { name: "Oliver" }]
+    let (tx, rx) = mpsc::channel(10);
+    spawn(worker(rx));
+    let resp = do_work(&tx, Duration::from_millis(10)).await;
+    println!("work result: {resp}");
 }
 ```
 
+Try making the worker print the number of work items handled every 100ms.
```
rbehjati, line 57: Should this go to the details part?

src/async/pitfalls/pin.md
```diff
@@ -1,108 +1,108 @@
...
 
 <details>
 
-* `tokio::pin!` only works on futures that implement `Unpin`. Other futures need to use `box::pin`.
-* Another alternative is to not use `tokio::pin!` at all but spawn another task that will send to a `oneshot` channel after the end of the `sleep` call. 
+* Students may recognize this as an example of the actor pattern. Actors
```
rbehjati, line 61:
```suggestion
* You may recognize this as an example of the actor pattern. Actors
```

I feel like the general theme is that the audience of these notes are anyone reading them, not just the instructors.

src/async/pitfalls/pin.md
```diff
@@ -1,108 +1,108 @@
...
+  typically call `select!` in a loop.
 
-```rust,editable,compile_fail
-use tokio::sync::mpsc::{self, Receiver};
-use tokio::time::{sleep, Duration};
-use tokio::sync::oneshot;
+* This serves as a summation of a few of the previous lessons, so take it
```
rbehjati, line 64:
```suggestion
* This serves as a summation of a few of the previous lessons, so take your time with it:
```

nit :)

rbehjati (approved):
This is very nice. Thanks. I've left a number of, mostly minor, and copy-edit comments.

src/async/pitfalls/blocking-executor.md
sakex: If you lock a sync Mutex and then use `.await` while the Mutex is locked, you're almost guaranteed to deadlock

rbehjati: Yes. By a short critical section I meant one that does not include any `.await` or heavy computation.

sakex: Yes, it's true. I'm maybe projecting my own bias but I would prefer people to default to async mutexes and use std Mutex only if there is a strong reason

src/async/control-flow/join_all.md
djmitche: Oh, I think that's the vim plugin doing that. It's annoying. I'll fix!

src/async/pitfalls/blocking-executor.md
djmitche: This is probably too detailed a point to get into in the course. I think understanding _why_ it might be a bad idea to use sync primitives in an async context is good enough. A good engineer can then figure out whether it makes sense to use `std::sync::mutex` in some specific situation. I'll try to scope this to explanation without making a recommendation.

src/async/pitfalls/pin.md
djmitche: This was a bit of a weird case because the example as-written does not illustrate the issue. But maybe I could save some time by finishing the first few steps in the speaker notes. I'll adjust.

src/async/pitfalls/blocking-executor.md
rbehjati: I agree. For me "very dangerous" sounded too strong. Perhaps just add that one should also consider the overhead of `tokio::sync::mutex`.

src/async/pitfalls/pin.md
djmitche: I think "from which" is more correct, but I'll rephrase.

PR #561, opened 2023-04-12
Extend bare metal afternoon exercise

qwandor:
Add a second step to use the match register to busy-wait for a period of time.


mgeisler (approved):
Very nice extension! I think I would have been able to do this after some reading of the documentation.

Issue #285, opened 2023-01-26
German translation


mgeisler:

mgeisler:
> I'll aim to be consistent with the translation there, in particular when it comes to technical terms.

That's a great idea! We should try to apply that principle to other translations as well :smile:

mgeisler:
Here is a machine-generated file to get started: [de.zip](https://github.com/google/comprehensive-rust/files/10578020/de.zip).

Throvn:
Just dropping by in case you need some help.

mgeisler:
Hi @Throvn, help would be very much appreciated!

There has been some PRs to the German translation, but there are still many strings to translate. It would be helpful if you would follow the instructions in `TRANSLATIONS.md` to sync the `de.po` file with the latest strings added (without translating them). Then it would be easier to do followup PRs with translations.

Throvn:
I will take on the task thanks for letting me know.

mgeisler:
Hi @ronaldfw and @Throvn, how complete is the German translation now?

Throvn:
I'm sorry @mgeisler but from our conversation above it was not clear to me that I should help with the german translation. If this is still the case, I will start working on it and open a WIP PR. So far though, there is 0 progress from my side.

PR #568, opened 2023-04-18
Translated table of contents to German

Throvn:
I've reviewed the **table of contents** of the existing translations, and added missing ones for the German language according to #285 .

I've tried to make the commit history as easy to review as possible, however, if there are some further improvements I could do, please tell me.

Otherwise, I would continue by translating the book page by page.


po/de.po
```diff
@@ -748,201 +743,250 @@ msgid "With Java"
...
-#: src/SUMMARY.md:245
+#: src/SUMMARY.md:265
 msgid "PACs"
 msgstr ""
 
-#: src/SUMMARY.md:246
+#: src/SUMMARY.md:266
 msgid "HAL Crates"
-msgstr ""
+msgstr "HAL Kisten"
```
mgeisler, line 847: You should decide if you want to translate "crate" to "Kiste" or not? I would personally not translate it, but it's mostly up to you (and other German speakers in the community).

mgeisler:
> I've tried to make the commit history as easy to review as possible, however, if there are some further improvements I could do, please tell me.

Sieht gut aus, danke dir!

We squash-merge the PRs in this repository: this means that your commits don't appear as individual commits in the final history. It is thus not super important how the individual commits look: only the PR title and the actual changes are important.

mgeisler (approved):

po/de.po
Throvn: When looking at the german translation of the Rust Book, they used "Kisten (crates)" as a translation. However, I think it makes sense to just stick with "crates" too, since this is used commonly. So from here on, I would just use "crates" and when it makes sense "crates (Kisten)".

PR #569, opened 2023-04-20
Main - Fix bug in exercise solution

jaewan-github:
To fix bug in the day 3's morning session's solution

### DIFF
--- a/src/exercises/day-3/simple-gui.rs
+++ b/src/exercises/day-3/simple-gui.rs
@@ -70,6 +70,13 @@ impl Window {
     fn add_widget(&mut self, widget: Box<dyn Widget>) {
         self.widgets.push(widget);
     }
+
+    fn inner_width(&self) -> usize {
+        std::cmp::max(
+            self.title.chars().count(),
+            self.widgets.iter().map(|w| w.width()).max().unwrap_or(0),
+        )
+    }
 }
 
 // ANCHOR_END: setup
@@ -78,10 +85,7 @@ impl Window {
 impl Widget for Window {
     fn width(&self) -> usize {
         // ANCHOR_END: Window-width
-        std::cmp::max(
-            self.title.chars().count(),
-            self.widgets.iter().map(|w| w.width()).max().unwrap_or(0),
-        )
+        self.inner_width() + 4
     }
 
     // ANCHOR: Window-draw_into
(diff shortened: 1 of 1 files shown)


src/exercises/day-3/simple-gui.rs
```diff
@@ -78,10 +85,7 @@ impl Window {
 impl Widget for Window {
     fn width(&self) -> usize {
         // ANCHOR_END: Window-width
-        std::cmp::max(
-            self.title.chars().count(),
-            self.widgets.iter().map(|w| w.width()).max().unwrap_or(0),
-        )
+        self.inner_width() + 4
```
mgeisler, line 88:
```suggestion
         // The `| ... |` formatting adds an additional 4 characters to the width.
        self.inner_width() + 4
```

src/exercises/day-3/simple-gui.rs
```diff
@@ -155,6 +159,17 @@ fn main() {
         "Click me!",
         Box::new(|| println!("You clicked the button!")),
     )));
+    let mut inner_window = Box::new(Window::new("Sub-window"));
```
mgeisler, line 162: This now shows up in the exercise text, right? The exercise shows the expected output and this is now out of sync. I would like to remove the code from here to keep things simple in the exercise.

Instead,  we could move the code to a unit test — it would be nice to have tests for this code in general. Would you be up for writing a few simple unit tests below?

src/exercises/day-3/simple-gui.rs
mgeisler: It took me a while to figure out which bug you're fixing... until it finally clicked :smile:

mgeisler:
> To fix bug in the day 3's morning session's solution

Hi @jaewan-github, thanks for the PR! Well spottet, I had not thought about adding a sub-window like this. I would like to keep the exercise code simple and then move the new extended example to a unit test. That way people can see it when they look at the solution.

jaewan-github:
I tried github's cloud code editor for the first time, and not certain whether I uploaded new revision correctly or not.

Please tell me if I made any mistake.

mgeisler:
> Then I'll remove extra code for adding nested window, and only fix the issue.

Great! :+1: 

> I tried github's cloud code editor for the first time, and not certain whether I uploaded new revision correctly or not.

It's perfect: the editor only allows you to create new commits (not amend existing commits). We squash-merge all PRs here anyway, so the precise history of your branch doesn't matter.

mgeisler (approved):

PR #571, opened 2023-04-20
Fix some broken fragment ID encodings in rustdoc links

eepeep:
I'm not exactly sure why they aren't decoded and considered equivalent, but the links were not working as-is for me in Chrome.

### DIFF
--- a/src/std/hashmap.md
+++ b/src/std/hashmap.md
@@ -61,6 +61,6 @@ fn main() {
   but it can lead into complications with the borrow checker.
   * Try removing `to_string()` from the example above and see if it still compiles. Where do you think we might run into issues?
 
-[1]: https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#impl-From%3C%5B(K%2C%20V)%3B%20N%5D%3E-for-HashMap%3CK%2C%20V%2C%20RandomState%3E
+[1]: https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#impl-From%3C%5B(K,+V);+N%5D%3E-for-HashMap%3CK,+V,+RandomState%3E
 
 </details>
--- a/src/std/vec.md
+++ b/src/std/vec.md
@@ -30,7 +30,7 @@ fn main() {
 methods on a `Vec`.
 
 [1]: https://doc.rust-lang.org/std/vec/struct.Vec.html
-[2]: https://doc.rust-lang.org/std/vec/struct.Vec.html#deref-methods-[T]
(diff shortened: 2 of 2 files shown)


mgeisler:
Thanks for the PR!

I feel I've fixed this before... or at least changed the percent encoding here before. I'll have to try and figure out what's going on and why some people see broken links here.

mgeisler:
Hi @djmitche, this is similar to your #917 — did you find out why the links changed in your PR?

mgeisler (approved):
I just checked all URLs by coping them to my address bar — and indeed, the new ones work and the old ones don't :smile: 

I was hoping that the CommonMark spec would have something to say about this, but it only briefly mentions "URL encoding" in the [About this document](https://spec.commonmark.org/0.30/#about-this-document) and when talking about an [absolute URI](https://spec.commonmark.org/0.30/#absolute-uri):

> An [absolute URI](https://spec.commonmark.org/0.30/#absolute-uri), for these purposes, consists of a [scheme](https://spec.commonmark.org/0.30/#scheme) followed by a colon (:) followed by zero or more characters other [ASCII control characters](https://spec.commonmark.org/0.30/#ascii-control-character), [space](https://spec.commonmark.org/0.30/#space), <, and >. If the URI includes these characters, they must be percent-encoded (e.g. %20 for a space).

So it seems we're at the mercy of what [mdbook](https://github.com/rust-lang/mdBook/) does, which is probably what [pulldown-cmark](https://github.com/raphlinus/pulldown-cmark) does.

For now, let's merge this. Perhaps someone can setup a link checker at some point to check the links in CI.

src/exercises/day-1/iterators-and-ownership.md
```diff
@@ -104,7 +104,7 @@ What is the type of `word` in each loop?
 
 Experiment with the code above and then consult the documentation for [`impl
 IntoIterator for
-&Vec<T>`](https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-IntoIterator-for-%26%27a%20Vec%3CT%2C%20A%3E)
+&Vec<T>`](https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-IntoIterator-for-%26'a+Vec%3CT,+A%3E)
```
mgeisler, line 107: Sorry, just a spurious space to kick off a new set of builds:

```suggestion
&Vec<T>`](https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-IntoIterator-for-%26'a+Vec%3CT,+A%3E) 
```

mgeisler:
I rebased the commit and this should merge after the tests are done.

PR #578, opened 2023-04-21
Direct users to "smccc" crate rather than "psci"

johnscheible:
psci's docs.rs page redirects users to `smccc` instead as the crate has been renamed.


qwandor (approved):
Thanks! We should update the examples too.

qwandor:
Ah, can you sign the CLA at https://cla.developers.google.com/ please, so we can accept this PR?

johnscheible:
Apologies, I'm a googler. I believe I've taken care of the internal process and issue should be resolved now. Let me know if I've missed something.

PR #583, opened 2023-04-26
Use smccc crate rather than psci in examples and exercise.

qwandor:
It was renamed.

This is a follow-up to #578.


mgeisler (approved):
Okay, sure!

PR #585, opened 2023-04-27
Test not implementing `Send`

mgeisler:


mgeisler:
Yep, it fails as it should! :smile:

PR #587, opened 2023-04-27
Cleanup control flow slides

mgeisler:
This avoids calling all the looping constructs “expressions” since they all (except for `loop`) return trivial values.

### DIFF
--- a/src/control-flow/loop-expressions.md
+++ b/src/control-flow/loop-expressions.md
@@ -1,7 +1,9 @@
 # `loop` expressions
 
-Finally, there is a `loop` keyword which creates an endless loop. Here you must
-either `break` or `return` to stop the loop:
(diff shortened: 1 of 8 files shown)


src/control-flow/loop-expressions.md
```diff
@@ -23,5 +25,8 @@ fn main() {
 <details>
     
 * Break the `loop` with a value (e.g. `break 8`) and print it out.
+* Note that `loop` is the only looping construct which returns a non-trivial
+  value. This is because it's guaranteed to be entered at least once (unline
```
djmitche, line 29:
```suggestion
  value. This is because it's guaranteed to be entered at least once (unlike
```

djmitche (approved):

PR #593, opened 2023-04-27
Remove exotic string types from scalar type slide

mgeisler:
When teaching the class, I’ve noticed that raw strings and byte strings tend to cause confusion. The slide is meant to be a friendly introduction and show how Rust is similar to other languages like C and C++. Instead, the string types cause a ton of questions which are unnecessary at this early point.

The information is still there, but now in the form of a speaker note.

### DIFF
--- a/src/basic-syntax/scalar-types.md
+++ b/src/basic-syntax/scalar-types.md
@@ -5,9 +5,8 @@
 | Signed integers        | `i8`, `i16`, `i32`, `i64`, `i128`, `isize` | `-10`, `0`, `1_000`, `123i64` |
 | Unsigned integers      | `u8`, `u16`, `u32`, `u64`, `u128`, `usize` | `0`, `123`, `10u16`           |
 | Floating point numbers | `f32`, `f64`                               | `3.14`, `-10.0e20`, `2f32`    |
-| Strings                | `&str`                                     | `"foo"`, `r#"\\"#`            |
+| Strings                | `&str`                                     | `"foo"`, `"two\nlines"`       |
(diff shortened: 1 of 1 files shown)


src/basic-syntax/scalar-types.md
```diff
@@ -16,3 +15,29 @@ The types have widths as follows:
 * `isize` and `usize` are the width of a pointer,
 * `char` is 32 bit wide,
 * `bool` is 8 bit wide.
+
+<details>
+
+There are a few syntaxes which are not shown above:
+
+- Raw strings allow you to create a `&str` value with escapes disabled: `r"\n"
+  == "\\\\n"`. You can embed double-quotes by using an equal amount of `#` on
```
rbehjati, line 24:
```suggestion
  == "\\n"`. You can embed double-quotes by using an equal amount of `#` on
```

I think.

rbehjati (approved):
Nice. Thanks.

src/basic-syntax/scalar-types.md
mgeisler: You are actually right according to the Commonmark spec: https://spec.commonmark.org/0.30/#example-338. However, I looked at the page and this is what I see with the weird number of backslashes:

image

I think there is a bug somewhere. I briefly tested https://github.com/raphlinus/pulldown-cmark and it handles the backticks correctly from what I can tell. So perhaps they get mangled further down the line.

mgeisler: I tried disabling our preprocessors and the result is the same. I filed https://github.com/rust-lang/mdBook/issues/2079 to track this.

Issue #64, opened 2022-12-26
Technical nits: Slices and other DSTs


QuineDot:
I don't think you ever cover what a non-`Sized` type, aka dynamically sized type (DST), is.

- [x] [Here](https://google.github.io/comprehensive-rust/basic-syntax/scalar-types.html) you call `&str` and `&[u8]` "scalar types", and... while the *reference* is, I suppose... the underlying (DST) slice contains multiple values, so I found their inclusion surprising.

- [x] [Then later on](https://google.github.io/comprehensive-rust/basic-syntax/slices.html) you refer to `&[T]` as a slice.  This is common nomenclature, but I feel it's worth pointing out that a `[T]` is more properly the slice, it is dynamically sized, and this is why you typically see it behind a reference (or other pointer like in a `Box`).  And [then on this page](https://google.github.io/comprehensive-rust/basic-syntax/string-slices.html) you could say a `str` is basically a `[u8]` with invariants (valid UTF8 encoding).  

Incidentally, that page says there are two Rust string types, but there are many others, such as `Arc<str>` or `Cow<'_, str>`.

---

- [ ] [Then later on](https://google.github.io/comprehensive-rust/std/box-recursive.html) you say
> Recursive data types or data types with dynamic sizes need to use a Box

But that's incorrect:
* There are no dynamic sizes here in Rust terminology; `List<T>` has a static size
* If you attempt to write `List` without the `Box`, it has *infinite* size -- but still not dynamic
* Slices have dynamic sizes but don't need a `Box`

---

- [x] [When you get around to trait objects,](https://google.github.io/comprehensive-rust/generics/trait-objects.html) you can mention that they're DSTs too, and that's why you typically see them in a `Box` (but may also seem them behind a reference, in an `Arc`, ...).

fw-immunant:
- `&str` and `&[u8]` are no longer described as scalar types.
- `&[T]` is more commonly known as a slice than the DST, just because DST usage is uncommon in general. I think it's best to avoid talking about DSTs early on as it requires invoking several concepts at once to be able to work with them.
- Filed PR #2283.
- We now mention the required pointer indirection for trait objects [here](https://google.github.io/comprehensive-rust/generics/dyn-trait.html).

PR #600, opened 2023-04-27
Talk about options in `?` chapter

sakex:
Fixes https://github.com/google/comprehensive-rust/issues/65


sakex:
Fixes #64

src/error-handling/try-operator.md
```diff
@@ -4,42 +4,47 @@ The try-operator `?` is used to return errors to the caller. It lets you turn
 the common
 
 ```rust,ignore
-match some_expression {
+let some_variable = match some_expression {
```
mgeisler, line 7: This seems unnecessary for the point we're trying to make: that `foo?` expands to something else. It's not `let x = foo?` that expands, it's just the `foo?` part. As you know, you can use `?` in deeply nested expressions which don't match the pattern your showing here.

(Please take this change out)

src/error-handling/try-operator.md
```diff
@@ -4,42 +4,47 @@ The try-operator `?` is used to return errors to the caller. It lets you turn
...
-        Ok(_) => Ok(username),
-        Err(err) => Err(err),
+    let mut email = String::new();
+    if let Err(err) = email_file.read_to_string(&mut email) {
+        return Err(err);
+    }
+    if let Some((username, hostname)) = email.split_once('@') {
+        Ok((String::from(username), String::from(hostname)))
+    } else {
+        Err(Error::new(ErrorKind::InvalidData, "Invalid email address"))
```
mgeisler, line 40: I think this adds too much extra logic to the example. The point here is to quickly (in 30 seconds or so) transform the code into using `?`. I would like to just keep the speaker notes here.

src/error-handling/try-operator.md
```diff
@@ -49,5 +54,14 @@ Key points:
 
 * The `username` variable can be either `Ok(string)` or `Err(error)`.
 * Use the `fs::write` call to test out the different scenarios: no file, empty file, file with username.
+* The return type of the function has to be compatible with the nested functions it calls. For instance,
+a function returning a `Result<T, Err>` can only apply the `?` operator on a function returning a 
+`Result<AnyT, Err>`. It cannot apply the `?` operator on a function returning a `Result<T, OtherErr>` 
+or an `Option<AnyT>`. Reciprocally, a function returning an `Option<T>` can only apply the `?` operator 
+on a function returning an `Option<AnyT>`.
+    * You can coerce incompatible types into one another with the different `Option` and `Result` methods
```
mgeisler, line 62:
```suggestion
    * You can convert incompatible types into one another with the different `Option` and `Result` methods 
```

src/error-handling/try-operator.md
```diff
@@ -49,5 +54,14 @@ Key points:
...
 * The `username` variable can be either `Ok(string)` or `Err(error)`.
 * Use the `fs::write` call to test out the different scenarios: no file, empty file, file with username.
+* The return type of the function has to be compatible with the nested functions it calls. For instance,
+a function returning a `Result<T, Err>` can only apply the `?` operator on a function returning a 
+`Result<AnyT, Err>`. It cannot apply the `?` operator on a function returning a `Result<T, OtherErr>` 
+or an `Option<AnyT>`. Reciprocally, a function returning an `Option<T>` can only apply the `?` operator 
+on a function returning an `Option<AnyT>`.
+    * You can coerce incompatible types into one another with the different `Option` and `Result` methods 
+    such as `Option::ok_or`, `Result::ok`, `Result::err`.
+    * Error types can be implicitely converted into the return type (let's call it`TopLevelErrorT`) with the `?` operator if they implement the `Into<TopLevelErrorT>` trait. This can be a good exercise to
```
mgeisler, line 64: This is explained on the following slides, no?

mgeisler (changes requested):
I would like to keep the slides very short and focused: they should only show the bare essentials to bring across the point the `?` expands into a `match` statement.

mgeisler:
Hey @sakex, can you try simplifying this a bit so we can merge it?

src/error-handling/try-operator.md
```diff
@@ -7,39 +7,44 @@ the common
 match some_expression {
     Ok(value) => value,
     Err(err) => return Err(err),
-}
+};
```
mgeisler, line 10:
```suggestion
}
```

src/error-handling/try-operator.md
sakex: Shortening this now

src/error-handling/try-operator.md
```diff
@@ -13,33 +13,38 @@ match some_expression {
 into the much simpler
 
 ```rust,ignore
-some_expression?
+let some_variable = some_expression?;
```
mgeisler, line 16: We should focus on just the expansion of `expr?`:

```suggestion
some_expression?
```

src/error-handling/try-operator.md
sakex: What do you think about the change?

src/error-handling/try-operator.md
```diff
@@ -20,26 +20,31 @@ We can use this to simplify our error handing code:
...
 
-    let mut username = String::new();
-    match username_file.read_to_string(&mut username) {
-        Ok(_) => Ok(username),
-        Err(err) => Err(err),
+    let mut email = String::new();
+    if let Err(err) = email_file.read_to_string(&mut email) {
+        return Err(err);
+    }
+    if let Some((username, hostname)) = email.split_once('@') {
```
mgeisler, line 32: The extra complexity of splitting and returning a `(String, String)` tuple seems unnecessary to me.

The example is supposed to focus laser-sharp on transforming `match foo { Ok(foo) => foo, Err(err) => return Err(err) }` to `foo?`. There should be very very little business logic apart from that.

sakex: I need to illustrate how to transform `Option` to `Result`. I thought `split` was a good example in that case because missing the `@` in the string would mean invalid email address.  Is it fine to drop the `tuple` and only return the username?

mgeisler: Oh... :sweat_smile:  See, I had not even noticed that this was the point of the change.

We've started putting larger and larger examples in the speaker notes over the last few months. Basically, when teaching the class, it's very poor style when we have to scroll up and down to explain an example. This is what the [aspect ratio helper](https://github.com/google/comprehensive-rust/blob/main/STYLE.md#vertical-space) is for and I suggest you check it out.

This is all to say that we should not add more vertical space to the slide. We should instead mention these point (important and great points!) in the speaker notes.

> Is it fine to drop the `tuple` and only return the username?

If this can be done without adding more vertical space, then sure! Alternatively, we could add a new page for to explicitly mention the most common conversion methods. That would help students by showing the explicit nature of Rust.

src/error-handling/try-operator.md
```diff
@@ -19,27 +19,27 @@ some_expression?
 We can use this to simplify our error handling code:
 
 ```rust,editable
-use std::{fs, io};
-use std::io::Read;
+use std::fs;
+use std::io::{self, Error, ErrorKind};
```
mgeisler, line 23: I would prefer not having to talk about this here — the slide is not about the standard library error types, it's about the `?` operator :smile:

sakex: Yes, I have diverted from my original goal at the time I realise.

I wanted to illustrate that a function that returns `Option` can also leverage the `?` operator. But it is not a good example here. I will roll back the change and should I add a new slide instead to show the `Option` case ?

mgeisler: I would start by putting it in the speaker notes like you do in this PR. You should try teaching the class, then you will see how much time there is for all the details (not a lot, in my experience :smile: )

mgeisler (approved):
Thanks for rolling back the code changes — it's much less complicated to add to the notes as you do here!

src/error-handling/try-operator.md
sakex: I am scheduled in 2 weeks :)

mgeisler: Oh right, sorry that I missed that! I'll look forward to hear what you think of it!

Issue #66, opened 2022-12-27
Suggestion: Globals and state


QuineDot:
[You say:](https://google.github.io/comprehensive-rust/basic-syntax/static-and-const.html)
> Global state is managed with static and constant variables.

But `const`s can't manage a state.  Statics can... but preferably only with interior mutability of some sort.  Your examples have no actual managing of state.

You do say "We will look at mutating static data in the chapter on Unsafe Rust", and indeed [have a page on `static mut`.](https://google.github.io/comprehensive-rust/unsafe/mutable-static-variables.html)

That slide should just be deleted.  It's so nigh-on impossible to get right even for experts that [there's a push for its deprecation.](https://github.com/rust-lang/rust/issues/53639)  Definitely no introduction to Rust should convey that it might be reasonable thing to reach for.

---

What should take its place?  A more accurate portrayal of how global state is actually done in *safe* idiomatic Rust.

First, it usually isn't!  Pass your state around, use an `Arc<Mutex<_>>` or the like.

But when it is, it's done with interior mutability.  `static` atomics, thread local `RefCell`, [`once_cell`](https://crates.io/crates/once_cell) ([eventually to be part of `std`](https://github.com/rust-lang/rust/issues/74465), [`lazy_static`](https://crates.io/crates/lazy_static).  

---

If you *must* leave in `static mut`, at least cover [`Once`](https://doc.rust-lang.org/std/sync/struct.Once.html#method.call_once).  But really that's just a tedious and error prone way of building up things like `once_cell` manually.  In practice people just use `once_cell` and similar instead of rolling their own.

Reaching for `unsafe` to do global state yourself is just a bad idea.

mgeisler:
Hi @QuineDot, thanks a lot! 

> You do say "We will look at mutating static data in the chapter on Unsafe Rust", and indeed [have a page on `static mut`.](https://google.github.io/comprehensive-rust/unsafe/mutable-static-variables.html). That slide should just be deleted.

I have not actually used this particular feature myself, so this is valuable information. To be honest, the slide is only there because I wanted to at least mention all the "super powers" you get when you use the `unsafe`  keyword. I wanted to de-mystify what it does — when teaching the class, I make a point out of telling people that `unsafe` doesn't "disable the borrow checker" like you sometimes see people imply.

[...]

PR #591, opened 2023-04-27
Add a style guide

mgeisler:
This is an attempt at documenting some things to do and don’t do when updating the course text or a translation.

Fixes #560.

PR #601, opened 2023-04-27
Rephrase static-and-const

sakex:
Fixes https://github.com/google/comprehensive-rust/issues/66


mgeisler (dismissed):
Hey Alex, please start by reverting all changes in `po/`. The files there are only touched by the translators, never by anybody else

po/pt-BR.po
```diff
@@ -3448,15 +3448,15 @@ msgid "# Static and Constant Variables"
...
+msgid "Static and constants are two kind of variables with special constraints."
 msgstr "O estado global é gerenciado com variáveis estáticas e constantes."
 
 #: src/basic-syntax/static-and-const.md:5
 msgid "## `const`"
 msgstr "## `const`"
 
 #: src/basic-syntax/static-and-const.md:7
-msgid "You can declare compile-time constants:"
+msgid "Constants are values created at compile time. They are inherently more efficient because they don't need to be computed at runtime and the compiler can inline them and apply special optimizations that would not be possible with a larger set of values:"
```
mgeisler, line 3459: Ah, this illustrates the problem well: by changing the `msgid` field here, you're claiming that

> Constants are values created at compile time. They are inherently more efficient because they [...]

should be translated into Portuguese as

> Você pode declarar constantes de tempo de compilação:

I don't speak Portuguese, but that's clearly wrong :smile:

src/basic-syntax/static-and-const.md
```diff
@@ -1,10 +1,10 @@
 # Static and Constant Variables
 
-Global state is managed with static and constant variables.
+Static and constants are two kind of variables with special constraints.
 
 ## `const`
 
-You can declare compile-time constants:
+Constants are values created at compile time. They are inherently more efficient because they don't need to be computed at runtime and the compiler can inline them and apply special optimizations that would not be possible with a larger set of values:
```
mgeisler, line 7: You're turning single-line sentences (like the ones found on a slide deck) into multi-line paragraphs. That's an anti-pattern and something I'm putting together a style guide (#591) to try and avoid.

Basically, please try looking at the slide with the "aspect ratio helper" enabled in your `book.toml` file. I'm pretty sure it overflow the height by a lot now.

mgeisler:
Hey Alex, thanks for taking up that old issue! We should find a nice way to outline how to handle global state since it comes up immediately when people start migrating to Rust.

About mutable statics: @Manishearth told me recently that it's effectively impossible to use a mutable static correctly. I don't know the details, but you end up triggering UB in almost all cases (if I understood correctly).

sakex:
Ok, good to know 😅 I didn't know if I had to edit those things or not. I will revert.On Apr 27, 2023 23:33, Martin Geisler ***@***.***> wrote:
@mgeisler commented on this pull request.



In po/pt-BR.po:
>  msgstr "O estado global é gerenciado com variáveis estáticas e constantes."
 
 #: src/basic-syntax/static-and-const.md:5
 msgid "## `const`"
 msgstr "## `const`"
 
 #: src/basic-syntax/static-and-const.md:7
-msgid "You can declare compile-time constants:"
+msgid "Constants are values created at compile time. They are inherently more efficient because they don't need to be computed at runtime and the compiler can inline them and apply special optimizations that would not be possible with a larger set of values:"

Ah, this illustrates the problem well: by changing the msgid field here, you're claiming that

Constants are values created at compile time. They are inherently more efficient because they [...]

should be translated into Portuguese as

Você pode declarar constantes de tempo de compilação:

I don't speak Portuguese, but that's clearly wrong 😄

src/basic-syntax/static-and-const.md
```diff
@@ -1,10 +1,11 @@
 # Static and Constant Variables
 
-Global state is managed with static and constant variables.
+Static and constants are two kind of variables with special constraints.
 
 ## `const`
 
-You can declare compile-time constants:
+Constants are values created at compile time. They are inherently more efficient because they don't need to be computed at runtime
```
mgeisler, line 7: Can you show me how this looks with the aspect ratio helper?

I feel you're turning a slide (with short bullet points) into a long-form text suitable for reading offline.

src/basic-syntax/static-and-const.md
```diff
@@ -1,10 +1,11 @@
 # Static and Constant Variables
 
-Global state is managed with static and constant variables.
+Static and constants are two kind of variables with special constraints.
```
mgeisler, line 3: This changes the sentence from something concrete:

* Global state is managed by
  * static and
  * constant variable

to something more loose:

* _two kind_ — are there more kinds?
* _special constraints_ — which constraints? How are the constraints special compared to other variables?

mgeisler:
Hi @sakex, let's get this merged or closed. I saw that we now also have #1019, which updates the same slide.

Can you trim down your edits so the slide stays very concrete and then we can merge this?

sakex:
Oh yes, I completely forgot about this. Fixing before lunch

mgeisler:
> Oh yes, I completely forgot about this.

Me too :smile: I was just going over the old PRs to see which we could merge.

mgeisler:
Thanks for the update @sakex!

@qwandor and @domenukk, I think you two know much more about `static` and `const` than me... can I get you to review this PR?

src/basic-syntax/static-and-const.md
mgeisler: Let's resolve this, we probably have worse offenders in our slides :smile:

src/basic-syntax/static-and-const.md
```diff
@@ -38,18 +42,31 @@ fn main() {
...
 
-As noted in the [Rust RFC Book][1], these are not inlined upon use and have an actual associated memory location.  This is useful for unsafe and embedded code, and the variable lives through the entirety of the program execution.
+As noted in the [Rust RFC Book][1], these are not inlined upon use and have an actual associated memory location.  This is useful for unsafe and 
+embedded code, and the variable lives through the entirety of the program execution.
 When a globally-scoped value does not have a reason to need object identity, `const` is generally preferred.
 
-Because `static` variables are accessible from any thread, they need to be guarded, for example by a [`Mutex`](https://doc.rust-lang.org/std/sync/struct.Mutex.html), or accessed using `unsafe` code.
-We will look at [mutating static data](../unsafe/mutable-static-variables.md) in the chapter on Unsafe Rust.
+Because `static` variables are accessible from any thread, they have more ownership constraints than data on the stack or the heap. 
+For instance, they cannot be safely mutated. We will look at [mutating static data](../unsafe/mutable-static-variables.md) in the chapter on Unsafe Rust..
```
qwandor, line 50: One full stop is enough.

src/basic-syntax/static-and-const.md
```diff
@@ -38,18 +42,31 @@ fn main() {
...
 * `static`, on the other hand, is much more similar to a `const` or mutable global variable in C++.
 * `static` provides object identity: an address in memory and state as required by types with interior mutability such as `Mutex<T>`.
 * It isn't super common that one would need a runtime evaluated constant, but it is helpful and safer than using a static.
+* `thread_local` data can be created with the macro `std::thread_local`.
+
+### Properties table: 
+
+| Property | Static | Constant |
+|---|---|---|
+| Allocated at runtime | Yes | No |
```
qwandor, line 64: What do you mean by this? Statics are generally allocated at compile-time, in the rodata section of the binary. Constants aren't 'allocated' at all, they are inlined.

src/basic-syntax/static-and-const.md
```diff
@@ -38,18 +42,31 @@ fn main() {
...
+| Lives for the entire duration of the program | Yes | No |
+| Can be mutable | Yes (unsafe) | No |
+| Evaluated at compile time | No | Yes |
```
qwandor, line 67: Static initialisers are also evaluated at compile time.

src/basic-syntax/static-and-const.md
```diff
@@ -38,18 +42,31 @@ fn main() {
 }
 ```
 
-As noted in the [Rust RFC Book][1], these are not inlined upon use and have an actual associated memory location.  This is useful for unsafe and embedded code, and the variable lives through the entirety of the program execution.
+As noted in the [Rust RFC Book][1], these are not inlined upon use and have an actual associated memory location.  This is useful for unsafe and 
+embedded code, and the variable lives through the entirety of the program execution.
 When a globally-scoped value does not have a reason to need object identity, `const` is generally preferred.
 
-Because `static` variables are accessible from any thread, they need to be guarded, for example by a [`Mutex`](https://doc.rust-lang.org/std/sync/struct.Mutex.html), or accessed using `unsafe` code.
```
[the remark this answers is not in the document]
qwandor: Agreed. What do you think of my wording in #1019?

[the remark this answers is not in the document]
sakex: Great, merged it and reverting my own change

qwandor (approved):

PR #603, opened 2023-04-29
Cleanup references to "Day 4"

mgeisler:
This PR replaces references to the old "Day 4" with references to Android, bare-metal, or concurrency as needed. Each commit should work on its own and I took care to update the redirects in `book.toml` along the way.


djmitche (approved):
Looks great!

PR #604, opened 2023-05-02
Move closures to the top level

djmitche:
I've gone looking for where we discuss closures, and without this change they do not appear in the TOC until you twirl down "Operators", which is a weird place for them.

Also, I looked at the list of "Important Traits" and, sure, I can get through those before lunch.  But these traits take quite a while to understand!


mgeisler (approved):
Sure, sounds good!

PR #605, opened 2023-05-03
Order Send/Sync before Arc/Mutex

djmitche:
The Arc/Mutex chapters mention Send/Sync in the speaker notes, and in fact serve as good illustrations of the traits, so let's define the traits before referencing them.


mgeisler:
> The Arc/Mutex chapters mention Send/Sync in the speaker notes, and in fact serve as good illustrations of the traits, so let's define the traits before referencing them.

The way I've approached this is to talk about `Arc`/`Mutex` as tools: people already know these tools from other languages, so I find that I can introduce them without reference to `Send` and `Sync`.

Later, we get into more into the theory of how Rust prevents certain values from moving between threads.

We could try your way too: explain the theory before the practical applications — I would be curious about how it works for you.

mgeisler (approved):
Let's try this way and then we'll change back if it doesn't work well.

djmitche:
I'll try it today (before merging) and report back.

djmitche:
I did like this ordering - Send/Sync feels like a dive into the "technicalities" and then Arc and Mutex are pretty light and familiar in comparison. It also seemed to match other parts of the course where we talk about general stuff and then present a list of useful things in the standard library. So I'm leaning 60/40 toward merging this. Plz click "Merge" or "Close" depending on which way you're leaning :)

mgeisler:
> So I'm leaning 60/40 toward merging this. Plz click "Merge" or "Close" depending on which way you're leaning :)

Thanks for testing it! I'm happy to merge this too and I look forward to trying it out at some point!

PR #609, opened 2023-05-04
Improve phrasing of comment

qwandor:
Some people interpreted the previous comment as meaning the RTC itself needed to be initialised or started somehow, but this is not the case.


mgeisler (approved):
Sounds good!

PR #612, opened 2023-05-04
Fix pluralisation inconsistency

expectocode:


mgeisler (approved):
Thanks a lot, this has annoyed me every single time I reach this slide! :smile:

Issue #537, opened 2023-03-30
Italian translation


mgeisler:

mgeisler:
Awesome, thanks @detro!

PR #613, opened 2023-05-04
Italian: Google Translate API submission

detro:
Context: #537 

In my first pass I started translating stuff manually: it's definitely smarter to begin with an auto-generated translation, and then tweak from there. So I finally gave https://github.com/mgeisler/cloud-translate a go.

I did notice that `cloud-translate` didn't try to translate blocks of code, so that's a good sign: those will require a bit more focus, as the code should remain in English, but the comments inside of it should be translated.

I decided to submit the auto-translated content, and will then proceed to make more PRs building up from this.

## A note about how I'm doing it

Going back to my time at university in Italy, we learned programming using a mix of Italian and English. English is usually preserved for terms that are highly topical and that, translated, would completely confuse an Italian reader (i.e. "keywords"). English is usually accompanying the Italian translation.

For example, `"Copying and Cloning"` becomes `"Copiare (Copy) e Clonare (Clone)"`. This way the strong connection with the actual language key terms (in this case `Copy` and `Clone` traits) is preserved, while using Italian (`Clonare` and `Copiare`) to help the reader understand.

Hope it makes sense. And if there are other Italians reading this, I'd love to hear your opinion.

### DIFF
--- a/po/it.po
+++ b/po/it.po
@@ -2,7 +2,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Comprehensive Rust 🦀\n"
 "POT-Creation-Date: \n"
-"PO-Revision-Date: 2023-05-04 19:07+0100\n"
+"PO-Revision-Date: 2023-05-04 19:16+0100\n"
 "Last-Translator: \n"
 "Language-Team: \n"
 "Language: it\n"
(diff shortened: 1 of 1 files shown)


detro:
@mgeisler: "great!" about the translation approach, and "yes" about the fact that auto-translated are fuzzy. I'm using [POEdit](https://poedit.net/) and it correctly highlights those, so it makes it even easier to proceed.

PR #616, opened 2023-05-05
da: Sync translation with c45b350

mgeisler:
This is the result of running

    msgmerge --update po/da.po po/messages.pot


mgeisler:
@thedataking I can sync again after #632, it's a completely [mechanical process](https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md#updating-an-existing-translation) which I propose we every few weeks if the translation drifts apart from the source texts.

thedataking (approved):

PR #618, opened 2023-05-05
Point students to the git repo containing exercise code

djmitche:
One of the students in my session mentioned that a Git repo containing the exercise code would be more useful to them than a ZIP file. This seems pretty easy to provide!

/cc @rbehjati in case of conflicts with other changes to the concurrency exercise.

### DIFF
--- a/src/exercises/bare-metal/compass.md
+++ b/src/exercises/bare-metal/compass.md
@@ -20,8 +20,9 @@ Hints:
   [nRF52833 datasheet](https://infocenter.nordicsemi.com/pdf/nRF52833_PS_v1.5.pdf) if you want, but
   it shouldn't be necessary for this exercise.
 
-Download the [exercise template](../../comprehensive-rust-exercises.zip) and look in the `compass`
-directory for the following files.
+Find the following files either in `compass/` in the [exercise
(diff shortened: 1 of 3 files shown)


qwandor (changes requested):
The directories in the git repository you've linked to are for the solutions, not the templates. This is not what students should be starting from.

djmitche:
Oh -- where are the templates, then?

djmitche:
If I understand correctly, based on reading that PR: for the bare-metal exercises there are files in the crate that are not included and thus not written to the ZIP file.  I didn't realize that when I put the async exercise together, so it only includes the "unsolved" version.

qwandor:
The solutions are checked into this repository, the template is generated from the Markdown files, with some functionality to include specific files or parts of files from the solutions. See the `{{#include ... }}` directives in e.g. https://github.com/google/comprehensive-rust/blob/main/src/exercises/bare-metal/compass.md, and the corresponding `// ANCHOR` and `// ANCHOR_END` comments in e.g. https://github.com/google/comprehensive-rust/blob/main/src/exercises/bare-metal/compass/src/main.rs .

PR #619, opened 2023-05-06
Fix mistakes in book-library.rs

Saphyel:
I just signed for the CLA and I fixed some errors I saw in the exercise.

### DIFF
--- a/src/exercises/day-1/book-library.rs
+++ b/src/exercises/day-1/book-library.rs
@@ -112,7 +112,7 @@ impl Library {
 // method signatures, including the "self" parameter! You may
 // also need to update the variable bindings within main.
 fn main() {
-    let mut library = Library::new();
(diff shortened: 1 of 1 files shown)


src/exercises/day-1/book-library.rs
```diff
@@ -112,7 +112,7 @@ impl Library {
 // method signatures, including the "self" parameter! You may
 // also need to update the variable bindings within main.
 fn main() {
-    let library = Library::new();
+    let mut library = Library::new();
```
mgeisler, line 115: People are supposed to run into this mistake by themselves. The exercise is there to let people work with `mut` and method receivers.

```suggestion
    let library = Library::new();
```

mgeisler (approved):
Thanks, the update of `{book}` is great, it's something I forgot in an earlier PR!

PR #627, opened 2023-05-10
Beginnings of an async chat exercise

djmitche:
@rbehjati here's what I came up with this afternoon. It has a basic solution (using a broadcast channel) and not-very-good error handling.

We could probably make this a little more challenging -- for example, by making clients not receive their own messages. WDYT?


rbehjati:
> We could probably make this a little more challenging -- for example, by making clients not receive their own messages. WDYT?

This is currently more like an echo server. Do you mean to actually make various clients talk to each other? There will be some design considerations, but either way I think it won't get too complicated, which is good.

djmitche:
Echo and a simple broadcast-chat are very similar! I think it's a 10-minute task for students to understand the code they see, find out how broadcast channels work, and use one to send messages between clients. Do we need to make it harder than that?

rbehjati:
> Echo and a simple broadcast-chat are very similar! I think it's a 10-minute task for students to understand the code they see, find out how broadcast channels work, and use one to send messages between clients. Do we need to make it harder than that?

Yes. I think. That is what I was trying to say in my previous comment :)

rbehjati:
Or actually, I am rather thinking of a broadcast version, and a non-broadcast version; where the client says which other client it wants to connect to and send messages to. Do you think that would be too complicated? I was thinking that should be doable within the 1-hour exercise session.

djmitche:
/cc @mauvealerts since they'll be teaching this soon

Ideas for slightly-harder variants:
 - Avoid "echoing" messages back to the client that sent them (so, could still use a broadcast channel, but filter what gets sent back to each client, or could loop over all clients except self and send the message)
 - Support "DM's" with some special syntax
 - Support joining and leaving rooms

The last two seem _too long_ to me, by comparison to the dining philosophers.

rbehjati:
Yes. The difficulty is in identifying the clients somehow. We could make DMs and room support as optional extensions, but I guess we don't want to make it too open-ended again.

djmitche:
I think that there's a happy medium to an exercise where there's a basic "get it working" portion (so, in this case, just add a broadcast channel) and one "for further study" portion (maybe adding DMs).

I'd like to update both dining philosophers to the same:
 1. Get the philosophers eating, until they deadlock
 2. Fix the deadlock (with a link to wikipedia or something for a hint as to how)

We could suggest using https://crates.io/crates/random_name_generator to generate random, typeable names for clients?

rbehjati:
> I think that there's a happy medium to an exercise where there's a basic "get it working" portion (so, in this case, just add a broadcast channel) and one "for further study" portion (maybe adding DMs).
> 
> I'd like to update both dining philosophers to the same:
> 
> 1. Get the philosophers eating, until they deadlock
> 2. Fix the deadlock (with a link to wikipedia or something for a hint as to how)

+1. This also seems more aligned with @mgeisler's idea of bite-sized exercise.  

> We could suggest using https://crates.io/crates/random_name_generator to generate random, typeable names for clients?

I think a simple counter works too :)

mauvealerts:
As written, this could almost(?) as easily be done using `TcpStream` and `BufReader` (plus `Ext` traits). I think the websocket version is nicer, but for comparison:
```rust
let (raw_in, mut raw_out) = socket.split();
let mut buf_in = BufReader::new(raw_in);
let mut line = String::new();
match buf_in.read_line(&mut line).await {
  //...
}
```

> > We could suggest using https://crates.io/crates/random_name_generator to generate random, typeable names for clients?
> 
> I think a simple counter works too :)

I was initially leaning towards `rand::random::<u8>()`, but after thinking a bit there's probably enough ways to make a counter work for students to come up with one of them (counter before spawn, mutex, atomics).

Maybe it's worth suggesting a syntax for DMs, to remove a silly decision? `@id some message` feels especially easy since it lends itself to `split()` and `strip_prefix()`.  `/dm id some message` probably leads to folks tripping over `splitn()` (which returns an iterator).

djmitche:
I thought of using raw TCP, but I think it's nice to show off how easy HTTP things can be in Rust. It's also handy to have both a client and server for students to fiddle with.

mgeisler:
About the formatting error:

image

I'm not sure why we get this error. The [`rustfmt.toml`](https://github.com/google/comprehensive-rust/blob/main/rustfmt.toml) file does not specify `imports_granularity` and the option is [unstable](https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#imports_granularity) and defaults to `Preserve` in any case.

I would actually like to enable `imports_granularity`: people will get a warning if they're on the stable channel, but that seems harmless if we can ensure a consistent style across all our code snippets.

djmitche:
@rbehjati would you like me to continue to develop / debug this PR, or do you want to pick up here?

rbehjati:
> @rbehjati would you like me to continue to develop / debug this PR, or do you want to pick up here?

Let's have a chat offline.

djmitche:
Per a conversation just now, I will rebase and fix the formatting on this, and merge it. Then @rbehjati will work on incorporating it into the markdown, splitting the prompt from the solution, and adding some instructions to the exercise.

mgeisler:
> ```
> E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/c/cross-toolchain-base/libc6-dev-arm64-cross_2.35-0ubuntu1cross3_all.deb  Unable to connect to azure.archive.ubuntu.com:http:
> ```
> 
> so I assume we'll have to wait that out :(

I've seen that happen intermittently over the last few days. Try rerunning the workflow.

PR #632, opened 2023-05-12
da: Update translation

thedataking:
A few more translations in order for me to get my feet wet.


po/da.po
```diff
@@ -97,11 +97,11 @@ msgstr "Referencer"
 
 #: src/SUMMARY.md:29
 msgid "Dangling References"
-msgstr ""
+msgstr "Forældede referencer"
```
mgeisler, line 100: I have forgotten how people talk about pointers in Danish, but I would say that "Forældede" has a slightly different meaning.

Asking for a translation on Google lead me to [dinglende](https://ordnet.dk/ddo/ordbog?aselect=dingle&query=dingle). That also sound a bit silly to me :smile:

mgeisler (approved):
Cool, thanks a lot! This is a great start!

po/da.po
mgeisler:
```suggestion
msgstr "Dinglende referencer"
```

:smile:

mgeisler:
Feel free to merge with or without my suggestion about "dinglende" above. We'll get to update this much more I'm sure.

po/da.po
thedataking: Yeah, I agree that `Forældede` is slightly off. I think `Dinglende` is slightly better albeit silly. I wonder what you think about `Hængende`? Let me know if I'm nitpicking too much here :)

voss (approved):
LGTM herfra.

po/da.po
mgeisler: Yeah, `dinglende` sounds quite silly :smile:  I'm happy with `Hængende` as well. I'm sure the Danish Rust community will send us patches if it ends up bothering someone :smile:

mgeisler: You can merge it when you like — I let authors with write access merge their own pull requests so that people can do small fixups like this.

PR #634, opened 2023-05-15
Run `apt update` before `apt install`

mgeisler:
The VMs on GitHub have updated APT caches.


qwandor:
Are you sure this isn't going to cause problems in future if the cache gets out of date?

djmitche (approved):
LGTM

PR #637, opened 2023-05-15
Enable `imports_granularity` for `rustfmt`

mgeisler:
This requires us to run a [nightly Rust for our formatting check][1]. Other users will see a warning when invoking `rustfmt`:

  Warning: can't set `imports_granularity = Module`, unstable features
  are only available in nightly channel.

This should be harmless overall and help us ensure better consistency in our import formatting.

[1]: https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#imports_granularity

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Format Rust code
+      - name: Install nightly rustfmt
+        run: |
+          rustup default nightly
+          rustup component add rustfmt
+
+      - name: Check Rust formatting
         run: cargo fmt --all -- --check
 
   cargo:
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,2 +1,2 @@
-# imports_granularity = "module"
+imports_granularity = "module"  # Please use nightly for this setting.
 max_width = 90


.github/workflows/build.yml
```diff
@@ -16,7 +16,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Format Rust code
+      - name: Install nightly Rust
+        runs: rustup default nightly
```
mgeisler, line 20:
```suggestion
        run: rustup default nightly
```

.github/workflows/build.yml
```diff
@@ -16,7 +16,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Format Rust code
+      - name: Install nightly rustfmt
+        run: |
+          rustup default nightly
```
djmitche, line 21: Instead of changing the default, you can use `cargo +nightly fmt ...` in the next step. I think there's a similar thing for `rustup` to add a component to a specific version.

mgeisler: Sure, one can do that... :slightly_smiling_face:  However, I don't like how the `+foo` deviates from the normal command line syntax.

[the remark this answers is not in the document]
mgeisler: I actually find it nice to have a unified default for a job — in this case `nightly`. The alternative requires me to decide on a line by line basis (and has the potential of me forgetting to add `+foo` on a line).

But since you both seem to suggest this approach, I updated the PR to match. I'll eventually ditch this for something like `dprint` instead to get unified formatting of everything.

qwandor (approved):

PR #642, opened 2023-05-16
Additional Italian translation

detro:
While at it, I updated `.gitignore` so we don't accidentally commit a `po/*.po~` file, that is generated every time we update the translation from source.

### DIFF
--- a/po/it.po
+++ b/po/it.po
@@ -2,7 +2,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Comprehensive Rust 🦀\n"
 "POT-Creation-Date: \n"
-"PO-Revision-Date: 2023-05-04 19:16+0100\n"
+"PO-Revision-Date: 2023-05-16 20:09+0100\n"
 "Last-Translator: \n"
 "Language-Team: \n"
 "Language: it\n"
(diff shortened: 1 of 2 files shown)


po/it.po
```diff
@@ -242,925 +244,795 @@ msgid "Designing a Library"
...
-#: src/SUMMARY.md:291
+#: src/SUMMARY.md:292
 #, fuzzy
 msgid "Async Traits"
 msgstr "Tratti asincroni"
 
 #: src/SUMMARY.md:295
 #, fuzzy
+msgid "Elevator Operations"
+msgstr "# Funzionamento dell'ascensore"
```
nicomazz, line 959: Nit: is it correct to have the # prefix?

po/it.po
```diff
@@ -1668,7 +1608,8 @@ msgid ""
 "The [Concurrency Deep Dive](../concurrency.md) is a full day class on classical\n"
 "as well as `async`/`await` concurrency."
 msgstr ""
-"Il [Concurrency Deep Dive](../concurrency.md) è un corso di un'intera giornata sulla musica classica\n"
+"Il [Concurrency Deep Dive](../concurrency.md) è un corso di un'intera giornata sulla musica "
```
nicomazz, line 1611: Classical music?

nicomazz (approved):
It's difficult to review 13k lines, but I skimmed thought it and it seems good. I'm assuming the format of the file is correct. Probably there are still a few automatic translations that could be improved, but overall it looks a good improvement!

po/it.po
mgeisler: Good catch, it should go away since it'll end up generating a Markdown heading.

The `fuzzy` marker saves it: it means that the translations here won't actually be used for anything. All fuzzy translations need proof-reading by a human (who will take away the marker).

mgeisler:
> It's difficult to review 13k lines,

Yes, that's something you can only skim briefly :smile: 

> but I skimmed thought it and it seems good. I'm assuming the format of the file is correct.

We have a test for that now! We didn't always have one and the format broke easily. The [latest run](https://github.com/google/comprehensive-rust/actions/runs/4995877964/jobs/8948319198) says

```
msgfmt --statistics -o /dev/null po/it.po
202 translated messages, 1187 fuzzy translations, 343 untranslated messages.
```

> Probably there are still a few automatic translations that could be improved, but overall it looks a good improvement!

I'm happy to merge any improvement. The most critical thing is to "protect" the non-fuzzy entries. The fuzzy marker is the bit that that protects our sanity — the fuzzy ones still need proof reading, so they can be as weird as they like :smile:

po/it.po
detro: Yeah, this is an auto-generated (by GCP Translate API) that it's still `fuzzy`-marked, meaning it's not "active". 

I'm using the brilliant [Poedit](https://poedit.net/) for this, and it marks all the fuzzy translations clearly (in orange) to make sure I know what's final and what not.

Screenshot 2023-05-17 at 21 08 18

po/it.po
detro: LOL - yeah, this is still fuzzy.

Everything "fuzzy" is what was auto-translated by GCP Translate API.

detro:
I think I should do this: I will not update the _source_ file (`messages.pot`) until we have done a full pass and got rid of all the fuzzy. Yes, is going to take a while.

By doing that though, I'd make the review easier, as one can load the giant file, and then do a side-by-side scanning, only looking at the changes.

I didn't think it through before updating the source file.

The tradeoff though is that we risk spending months on the translation, removing all the fuzzy, and then we update the source file and there is a large delta :(

I'm not sure what's best here.

Opinions? Ultimately we would benefit from agreeing to a way that makes reviews doable/sustainable.

detro:
Meanwhile, should I merge?

We could discuss what I wrote above in a proper "Discussion" here on GH.

mgeisler:
> Meanwhile, should I merge? 

Yeah, please go ahead and merge it! You're doing great with managing the fuzzy bits and that's really the most important thing here :slightly_smiling_face: 

> We could discuss what I wrote above in a proper "Discussion" here on GH.

Sounds good, there are probably other translators who run into the same problems and so it would be great to share the solutions too.

PR #644, opened 2023-05-18
da: Additional translations

thedataking:
We probably need to bikeshed whether we want to use "Flyttesemantik" or "Overførselssemantik" for "Move Semantics".

### DIFF
--- a/po/da.po
+++ b/po/da.po
@@ -126,7 +126,7 @@ msgstr "Metoder"
 
 #: src/SUMMARY.md:36
 msgid "Overloading"
-msgstr ""
+msgstr "Funktionsoverlæsning"
 
 #: src/SUMMARY.md:37 src/SUMMARY.md:66 src/SUMMARY.md:90 src/SUMMARY.md:119
 #: src/SUMMARY.md:148 src/SUMMARY.md:177 src/SUMMARY.md:204 src/SUMMARY.md:225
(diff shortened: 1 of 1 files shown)


po/da.po
```diff
@@ -160,7 +160,7 @@ msgstr "static & const"
 
 #: src/SUMMARY.md:46
 msgid "Scopes and Shadowing"
-msgstr ""
+msgstr "Omfang og overskyggede variabler"
```
mgeisler, line 163:
```suggestion
msgstr "Virkefelter og overskyggede variabler"
```

Jeg har relativt ofte hørt folk oversætte "scope" til ["virkefelt"](https://ordnet.dk/ddo/ordbog?entry_id=12007009&def_id=21095680&query=virkefelt).

mgeisler:
> We probably need to bikeshed whether we want to use "Flyttesemantik" or "Overførselssemantik" for "Move Semantics".

Oh, good question! :smile: I see there is a Danish translation of the Rust Book (https://github.com/rust-lang/book/issues/3079, by @DanKHansen), but it is brand new and "move semantic" is not yet translated.

po/da.po
mgeisler: Ah, jeg fandt et dokument som også bruger "virkefelt": https://homes.cs.aau.dk/~normark/c-prog-06/pdf/all.pdf.

mgeisler (approved):
I'm happy to get this in as a great base for further PRs!

po/da.po
thedataking: Ah, jeg var på dybt vand mht. scope - virkefelt er en bedre oversættelse – tak! Today I learned.

PR #645, opened 2023-05-19
Publish the in-progress Italian translation

mgeisler:
This makes the translation available on the official site. It's not linked yet, but having it available there should make it easier to review the state.


mgeisler:
Hey folks, if you think it would be useful, we can start publishing the in-progress translation already now.

PR #647, opened 2023-05-19
Add missing newlines around `<details>`

mgeisler:
The newlines makes the text more uniform when extracted into the POT file (spotted while doing #646).


mgeisler:
> Could we lint for this?

We could, though I expect to fix this more long-term with the next version of mdbook-i18n-helpers.

djmitche:
Ah, right, of course :)

PR #656, opened 2023-05-22
Fix inconsistent code in unbounded.md

bryanhitc:
Fix inconsistency between concurrency `unbounded.md` and `bounded.md` by using implicitly named arguments for both.


mgeisler (approved):
Thank you, that has annoyed me a few times but I never got around to fixing it :smile:

mgeisler (approved):
Thank you, that has annoyed me a few times but I never got around to fixing it :smile:

PR #658, opened 2023-05-23
Update rustc and cargo version information in running-locally.md

thoriumdesign:
Also improves verbiage of a sentence, from "With this is in place, then follow..." to "With this in place, follow..."


mgeisler (approved):
Thanks a lot for fixing my English! :smile: 

I was thinking that we could leave the version numbers alone, but then again -- the text just above does way "latest stable Rust release".

Perhaps we could rephrase the whole thing to say something like "The course is tested with Rust 1.61.0 and should also work with later versions." We should then of course actually test with this version in CI.

PR #660, opened 2023-05-23
Update lifetimes.md

primeagen-rustaceans:
Normally developers are aware of the simpler language of implicit vs explicit. By replacing elided with implicit readability of the article might improve.

### DIFF
--- a/src/ownership/lifetimes.md
+++ b/src/ownership/lifetimes.md
@@ -2,7 +2,7 @@
 
 A borrowed value has a _lifetime_:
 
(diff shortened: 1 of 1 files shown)


mgeisler (approved):
Thank you, I agree that this is easier to understand.

djmitche:
This is referred to as [elision](https://doc.rust-lang.org/nomicon/lifetime-elision.html) in Rust. Perhaps it would be good to reference that terminology as well?

PR #672, opened 2023-05-24
Update default-methods.md

freeaion:
This commit updates the example code to make NotEqual a super trait for Equals. The example code implements "trait NotEqual: Equals", which makes Equals a super trait for NotEqual.

### DIFF
--- a/src/traits/default-methods.md
+++ b/src/traits/default-methods.md
@@ -33,13 +33,18 @@ fn main() {
   implement themselves. Methods with default implementations can rely on required methods.
 
 * Move method `not_equal` to a new trait `NotEqual`.
-
-* Make `NotEqual` a super trait for `Equal`.
     ```rust,editable,compile_fail
-    trait NotEqual: Equals {
+    trait NotEqual {
         fn not_equal(&self, other: &Self) -> bool {
             !self.equal(other)
         }
+    }
+    ```
+
+* Make `NotEqual` a super trait for `Equal`.
+    ```rust,editable,compile_fail
+    trait Equals: NotEqual {
+        fn equal(&self, other: &Self) -> bool;
     }
     ```
 


edliaw:
Hm, it seems quite convoluted like this, and `self.equal` will not be available to `NotEqual`.  Maybe instead just fix the wording to be "Make `Equal` a super-trait of `NotEqual`".

freeaion:
The current example is as follows;

- Move method not_equal to a new trait NotEqual.
- Make NotEqual a super trait for Equal.

trait NotEqual: Equals {
    fn not_equal(&self, other: &Self) -> bool {
        !self.equal(other)
    }
}

I think it's better to update both example and phrase because the context for the example is to cover the blanket implementation where any types implementing Equal will have the impl of NotEqual. By making NotEqual a super trait of Equal, we can have the same implementation. If I misunderstood the context for the example, please let me know. Thank you.

edliaw:
That does make sense.  I think we'd need to leave the implementation of `not_equal` in `Equal` and remove the default implementation in `NotEqual` to make it work.

freeaion:
Hello @edliaw and @mgeisler, it's not clear on what to change in my commit for PR. (github's PR is not as good as critique in google, which makes me hard to accomodate what @edliaw was recommending.) Please let me know what needs to be changed to get LGTM. Thank you.

freeaion:
Hello @edliaw, as you pointed out, super trait cannot use a method in its sub trait, which results in a compilation error. (Oops.. :) ) I think the example doesn't allow us to show default impl for super/sub traits because it has deps btn them. I think the only way to resolve this issue is to revise the wording. Let me cancel this PR and initiate another one w/ the re-wording. github's PR is too inconvenient compared to CL/critique and/or gerrit. Thank you @edliaw.

PR #677, opened 2023-05-24
ru: Create ru.po file

baltuky:
Initial setup forRussian translation. This is a freshly generated empty `ru.po` file.


mgeisler (approved):
Excellent, thanks very much!

mgeisler:
I can see in the CI output that the file is indeed valid and completely empty — so even without speaking Russian, I think I can approve this one :smile:

PR #687, opened 2023-05-25
Updates Brazilian Portuguese translation to add new content (with cloud-translate)

rastringer:
will need fine-tuning!


mgeisler:
Thanks for the refresh, @rastringer!

A bunch of the changes are because of a change in how you wrap the strings in the PO file. Before, it seemed like your PO editor allowed arbitrary long lines, now it wraps at 80 columns or so. An example is this:

image

An old-fashioned apostrophe (`'`) got turned into a fancy curly one (`’`). Apart from that, the other changes are because the editor decided to wrap the lines. It's not critical to avoid this, but it inflates the size of the diff, which makes it hard to see what's going on.

Perhaps it would be useful for you to do use `msgmerge --no-wrap` or an equivalent option in Poedit?

We can also merge this as-is: you and the other translators should just agree long-term about a setting so that you can avoid flip-flopping back and forth :smile:

rastringer:
Thanks @mgeisler! Great points. I think I fixed the wrapping issue (using --no-wrap). Can take a look at commas as a follow up perhaps?

mgeisler:
> Can take a look at commas as a follow up perhaps?

I'm not sure what you mean by "commas"?

But yeah, I think this change needs to go in as-is and then you and the other translators will have to iterate on it. There are still just under 1k translated strings with this PO file — and about 500 fuzzy ones to proof read.

mgeisler (approved):
The sync looks good from a structural point of view and I think we should iterate more on this after merging it.

rastringer:
Thanks Martin! I meant 'apostrophe(')' vs comma, good catch. Great, we will iterate on this.

PR #689, opened 2023-05-25
ru: Run msgmerge with --no-wrap to avoid unnecessary diffs

baltuky:
Run `msgmerge --no-wrap --update po/ru.po po/messages.pot` to avoid unnecessary diffs.


mgeisler (approved):
That's a good idea!

In general, the changes to these files are hard to do in a nice way... I'm working on changes to mdbook-i18n-helpers to make it ignore things like newlines in the Markdown when extracting the text. That should make this kind of change a no-op (at some point in the future).

Issue #324, opened 2023-02-02
Chinese (Simplified) translation


hilbert-yaa:
Hi Martin @mgeisler ! I've had a conversation with you via email about this translation a week ago. I'm working on this issue rn.

Edit by @mgeisler: See https://github.com/google/comprehensive-rust/issues/324#issuecomment-1562626220 below for instructions.

mgeisler:
For reference: https://github.com/KaiserY/trpl-zh-cn.

kiwiwarmnfuzzy:
Hey @mgeisler: I saw that there's a machine-translated version by running the i18n script, could you run the script for the zh_CN locale for us? 

(I think Hilbert had a similar script but we noticed some off-by-one errors..., so just wanted to see if these errors are specific to zh_CN locale.) 

Thanks!

mgeisler:
Hi @kiwiwarmnfuzzy, the script it here: https://github.com/google/comprehensive-rust/pull/291. I asked the other translators, and they tell me that they haven't seen the same off-by-one errors.

Here is the output from a fresh run of that script: [zh_CN.zip](https://github.com/google/comprehensive-rust/files/10697720/zh_CN.zip).

mgeisler:
> Oh no worries I wasn't using your script.

Great to hear that!

mgeisler:
Hi there, just a small note: please rename the file to `zh-CN.po` so that the translation will eventually appear under `.../zh-CN/`. The `_` to `-` change should give us the right kind of language tag and this is the style we used for the Brazilian Portuguese translation in #381.

mgeisler:
From reading https://developers.google.com/search/docs/specialty/international/localized-versions, I'm thinking we should name the Simplified Chinese translation `zh-Hans` like @taotieren does in #436.

mgeisler:
> Got it. I've been occupied by my COVID and ISCA rebuttal lately.

Ugh, sorry to hear that! I had COVID myself last year and it was not fun at all... Good luck with the conference! :slightly_smiling_face:

est7:
@Hilbert-Yaa any update? can't wait to see it

kiwiwarmnfuzzy:
@xusiwei Thanks for reaching out! 

> I think called Chinese Simplified is better.

Could you explain a little why 'Chinese Simplified' is better? Is the difference between Mandarin v.s. Simplified about speaking v.s. writing? (I'm not very experienced in internationalization and various terminologies, so I'm mildly curious.)

[...]

PR #696, opened 2023-05-26
Update CODEOWNERS add @wnghl for Chinese (Simplified)

wnghl:
For #324


mgeisler (approved):
Thanks, welcome onboard!

PR #701, opened 2023-05-26
Update CODEOWNERS to add victorhsieh as zh-Hant reviewer

victorhsieh:


mgeisler (approved):
Thanks Victor for helping out!

Issue #684, opened 2023-05-25
Chinese (Traditional) translation


mgeisler:
Discuss a Chinese (Traditional) translation of the course. For reference, here is a translation of the Rust Book: https://github.com/rust-tw/book-tw.

mgeisler:
Just like for Chinese (Simplified), I suggest coordinating the work on this translation via a spreadsheet.

1. First, we need someone to put up a PR to get a `zh-Hant.po` file committed
2. After that, **use [this spreadsheet to coordinate](https://docs.google.com/spreadsheets/d/1bVSJEUGkI9gcqmmA4va4y-I7LzgQUqNwnZYJsQNMBGE/edit?usp=sharing)**.
    - _Please add yourself on the first page!_
    - Pick a file from the second sheet and add your Github username next to it to claim it.
        - Change the state to "In progress" and
        - Add today's date to the last change column.
    - The spreadsheet is editable by everybody (both inside and outside of Google) so you can add/remove things, comment on it and generally update it to suit your needs.

Please send small PRs! I suggest translating the titles in the `SUMMARY.md` file as you translate the content itself, but that will naturally lead to contention on that file. Perhaps you want to instead agree upfront about the wording for the titles? Our [discussions](https://github.com/google/comprehensive-rust/discussions) would be a good place to do this.

See #683 for a just-started discussion about where to coordinate a translation table. You might want to create a similar table — in the spreadsheet, in the PO file directly, or in a discussion post.

Let me know if I've forgotten something or can help in any other way!

mgeisler:
> Hi all, we need someone to put up an empty (or nearly empty) `po/zh-Hant.po` file using the [translation instructions](https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md#creating-and-updating-translations).
> 
> Please let me know if you run into any trouble with this!

I went ahead and created #697 for this since I already have everything setup locally.

mgeisler:
> Thanks for getting this started!

Thanks for the review on #697! You have an eye for detail and I invite you to add yourself to [CODEOWNERS](https://github.com/google/comprehensive-rust/blob/main/.github/CODEOWNERS) if you're interested in reviewing more changes :smile:

hueich:
Sent https://github.com/google/comprehensive-rust/pull/702

edong:
[...]

PR #702, opened 2023-05-26
Add hueich to CODEOWNERS

hueich:


hueich:
Adding myself (hueich) to CODEOWNERS for zh-hant to help with reviewing traditional chinese changes.
Context: #684

PR #703, opened 2023-05-26
Add kuanhungchen to CODEOWNERS of zh-Hant

kuanhungchen:
Add @kuanhungchen to CODEOWNERS of zh-Hant to work on reviewing traditional Chinese translation.


mgeisler (approved):
Thanks!

PR #711, opened 2023-05-27
Format files with `dprint`

mgeisler:
The dprint formatter is a flexible system which will use sandboxed WebAssembly formatters to format our code (mostly: it calls out to `rustfmt` for Rust code).

A particularly interesting feature is that dprint can format Rust code blocks in the Markdown files. However, before we turn that on, we need to have a way to [normalize the Markdown text][1] as it is extracted. That is so that the work put into the translations is kept after the reformatting.

[1]: https://github.com/google/mdbook-i18n-helpers/issues/19

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,8 +21,8 @@ jobs:
           rustup default nightly
           rustup component add rustfmt
 
-      - name: Check Rust formatting
-        run: cargo fmt --all -- --check
+      - name: Check formatting
+        uses: dprint/check@v2.2
 
   cargo:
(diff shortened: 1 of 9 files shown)


mgeisler:
> Was everything under `src/` correctly formatted already?

Nope, it's explicitly excluded for now -- since we need https://github.com/google/mdbook-i18n-helpers/issues/19 to be able to reformat this without losing the translations



> Also, we probably shouldn't be reformatting third-party source..

Ah, good idea! It hadn't occurred to me to exclude this too :smile:

djmitche (approved):

PR #715, opened 2023-05-28
Add @edong to CODEOWNERS for po/zh-Hant.po

edong:
Add @edong to CODEOWNERS for po/zh-Hant.po.
Also, sort the usernames on that row.


mgeisler (approved):
Thanks Eric!

PR #717, opened 2023-05-29
Clarify Default trait and operators

mauvealerts:
The description of `Default` sounded (to me) like it was referring to derive macros or something. I don't feel strongly about the verb (provides vs produces), but I think 'produces' suggests creating a value (which is usually the case).

I feel less strongly about the proposed changes to the 'operators' slide...
* the existing bullet point just needed clarification during the class
* implementing operators for different types might be too niche to bother with

### DIFF
--- a/src/traits/default.md
+++ b/src/traits/default.md
@@ -1,6 +1,6 @@
 # The `Default` Trait
 
-[`Default`][1] trait provides a default implementation of a trait.
+[`Default`][1] trait produces a default value for a type.
 
 ```rust,editable
 #[derive(Debug, Default)]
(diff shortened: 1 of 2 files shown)


src/traits/default.md
```diff
@@ -1,6 +1,6 @@
 # The `Default` Trait
 
-[`Default`][1] trait provides a default implementation of a trait.
+[`Default`][1] trait produces a default value for a type.
```
mgeisler, line 3: Yes, your phrasing is much better!

mgeisler (approved):
Thanks, this is a nice improvement!

PR #723, opened 2023-05-29
zh-Hant: translation for src/generics.md

kuanhungchen:
Part of #684.

### DIFF
--- a/po/zh-Hant.po
+++ b/po/zh-Hant.po
@@ -2,14 +2,15 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Comprehensive Rust 🦀\n"
 "POT-Creation-Date: \n"
-"PO-Revision-Date: 2023-05-30 02:08+0800\n"
+"PO-Revision-Date: 2023-05-31 06:33+0800\n"
 "Last-Translator: Hank Chen <kuanhungchen@google.com>\n"
 "Language-Team: Chinese (Traditional) <zh-l10n@lists.linux.org.tw>\n"
+"Language: zh_TW\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"Language: zh-Hant\n"
 "Plural-Forms: nplurals=1; plural=0;\n"
+"X-Generator: Poedit 3.3.1\n"
 
 #: src/SUMMARY.md:3
 msgid "Welcome to Comprehensive Rust 🦀"
@@ -7257,7 +7258,7 @@ msgstr ""
 
 #: src/generics.md:1
 msgid "# Generics"
-msgstr "# 泛型"
+msgstr "# 泛型（Generics）"
 
 #: src/generics.md:3
 msgid ""
@@ -7265,7 +7266,8 @@ msgid ""
 "sorting)\n"
 "over the types used in the algorithm."
 msgstr ""
-"Rust支援泛型，因此可以抽象化一個演算法（例如：排序）而不需要依賴特定型別。"
+"Rust支援泛型（Generics），因此可以抽象化一個演算法（例如：排序）而不需要依賴"
+"特定型別。"
 
 #: src/generics/data-types.md:1
 msgid "# Generic Data Types"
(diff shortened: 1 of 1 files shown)


po/zh-Hant.po
```diff
@@ -7257,14 +7257,15 @@ msgstr ""
...
-msgstr ""
+msgstr "# 泛型"
 
 #: src/generics.md:3
 msgid ""
 "Rust support generics, which lets you abstract an algorithm (such as "
 "sorting)\n"
 "over the types used in the algorithm."
 msgstr ""
+"Rust支援泛型，因此可以抽象化一個演算法（例如：排序）而不需要依賴特定型別。"
```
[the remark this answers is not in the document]
kuanhungchen: Thanks! Just added the English name here.

edong: zh-TW style:
* There should be a space between Chinese characters and English letters or numbers. -> `Rust 支援`
* Half-width parentheses `()` should be used instead of full-width parentheses `（）`, even when surrounding Chinese text or surrounded by Chinese text. There shouldn't be spaces between the parentheses and the text within them, but there should be spaces outside of them if adjacent to Chinese characters. -> `算法 (例如：排序) 而`

kuanhungchen: Updated, PTAL. Thanks!

po/zh-Hant.po
```diff
@@ -2,14 +2,15 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Comprehensive Rust 🦀\n"
 "POT-Creation-Date: \n"
-"PO-Revision-Date: 2023-05-26 15:20+0200\n"
-"Last-Translator: Martin Geisler <mgeisler@google.com>\n"
+"PO-Revision-Date: 2023-05-31 09:11+0800\n"
+"Last-Translator: Hank Chen <kuanhungchen@google.com>\n"
 "Language-Team: Chinese (Traditional) <zh-l10n@lists.linux.org.tw>\n"
+"Language: zh_TW\n"
```
hueich, line 8: Is this change intended?  Is there a reference for why it's zh_TW instead of zh-Hant?

po/zh-Hant.po
```diff
@@ -7257,14 +7258,16 @@ msgstr ""
 
 #: src/generics.md:1
 msgid "# Generics"
-msgstr ""
+msgstr "# 泛型 (Generics)"
```
hueich, line 7261: Not sure if there's a style guide about whether to include the English name in titles.

po/zh-Hant.po
```diff
@@ -7257,14 +7258,16 @@ msgstr ""
...
 
 #: src/generics.md:3
 msgid ""
 "Rust support generics, which lets you abstract an algorithm (such as "
 "sorting)\n"
 "over the types used in the algorithm."
 msgstr ""
+"Rust 支援泛型 (Generics)，因此可以抽象化一個演算法 (例如：排序) 而不需要依賴"
```
hueich, line 7269: The "抽象化" should probably be after.  E.g.:

因此可以將一個演算法 (例如：排序) 抽象化而不需要依賴

po/zh-Hant.po
kuanhungchen: No. The "Language" and "X-Generator" fields were somehow generated by the Poedit.
They are removed now.

po/zh-Hant.po
kuanhungchen: Updated.

po/zh-Hant.po
kuanhungchen: Updated. Thanks!

po/zh-Hant.po
kuanhungchen: In some languages (e.g. ja, de) they include X-Generator while others (e.g. zh-Hans, el) don't. Not sure if we want to have it included.

mgeisler: As far as I know, nothing in our pipeline uses the metadata here — we derive the URL directly from the name of the PO file. So you can change this as you like :slightly_smiling_face:

hueich (approved):
Looks good, thanks!

hueich (approved):
Looks good, thanks!

kuanhungchen:
Thanks for all the review!

Issue #652, opened 2023-05-22
Japanese translation


mgeisler:
The [Rust Book has been translated into Japanese](https://github.com/rust-lang-ja/book-ja). Let us see if we can reuse some of their terminology.

mgeisler:
Hey @momotaro1105, @keiichiw, @CoinEZ-JPN and other people interested in the Japanese translation.

Just a small note: I've created spreadsheets to help organize some of the other translations, see #684 for an example. That might be useful for you too.

kantasv:
Hi, I've read your misc-jp post and would love to contribute to this as a native Japanese speaker :) 
Do we already have the spreadsheet for the JA project? Or should I make one?
https://github.com/google/comprehensive-rust/issues/652#issuecomment-1562812496

CoinEZ-JPN:
hello, just checking - the next PR should come directly from my branch post fetch origin, correct?
and still to this #652?

keiichiw:
@CoinEZ-JPN If the branch on your forked repository is based on this repository's main branch and your commit message includes `#652`, that should be fine.

Note that it would be easier if one pull request translates only one chapter. That's said it'd be nice if you have branches such as `japanese-translate-ch-3` or `japanese-translate-ch-4` for each chapter.

CoinEZ-JPN:
@keiichiw hi thanks instructions
Is it alright if I commit the intro (ch.1 and 2) in a single PR?
The rest I will separate into its own branch.

@kantasv apol missed, not aware of what misc-jp is, but would be nice to work together - what do you need me to do?

CoinEZ-JPN:
@keiichiw pls ignore first comment, just created PR for just ch.0, the welcome page from `japanese-translation-ch-0`

keiichiw:
@mgeisler @CoinEZ-JPN @kantasv 
I just created [a spreadsheet](https://docs.google.com/spreadsheets/d/1qyu_59IlmxEHX4e5YnsxsjFWDY2o3YwIqE5PzMgKwnA/edit#gid=0) to gather the information on who is now working on which sections. Please take a look at it and put your name to the files you want to work on.

keiichiw:
@kantasv Thank you so much for offering to help! We've started translation. 
Please take a look at [po/ja.po](https://github.com/google/comprehensive-rust/blob/main/po/ja.po) and start translating whatever section you want!

Note that misc-jp is something Google internal. So, please avoid mentioning it in public places like here.

[...]

PR #668, opened 2023-05-24
Fix a couple typos on the welcome page 

gregl83:
- welcome page typos corrected

PR #729, opened 2023-05-30
ja: Chapter 0

CoinEZ-JPN:
- `mdbook serve` does NOT reflect translations in the following areas in both localhost and prod
Screenshot 2023-05-30 at 20 27 23
- same behavior as above observed in prod for Brazilian and Korean
- minor edits to table of contents for consistency in first commit

Part of #652.

### DIFF
--- a/po/ja.po
+++ b/po/ja.po
@@ -2,7 +2,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Comprehensive Rust 🦀\n"
 "POT-Creation-Date: \n"
-"PO-Revision-Date: 2023-05-30 19:59+0900\n"
+"PO-Revision-Date: 2023-05-30 20:20+0900\n"
 "Last-Translator: Kenta Aratani <kentaaratani@coinez.jp>\n"
 "Language-Team: Japanese <translation-team-ja@lists.sourceforge.net>\n"
 "Language: ja\n"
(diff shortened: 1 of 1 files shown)


mgeisler:
> * same behavior as above observed in prod for Brazilian and Korean

Strange, but I can reproduce this as well! Let me open a dedicated issue for this.

keiichiw (approved):
Thank you for the PR! LGTM for the translation. 

I guess we can ignore the `mdbook serve` issue for now and merge this PR, as @mgeisler will file another issue.

mgeisler:
> > > * same behavior as above observed in prod for Brazilian and Korean
> > 
> > 
> > Strange, but I can reproduce this as well! Let me open a dedicated issue for this.
> 
> ok, thank you

Ah, I tested this again!

The problem is that the paragraph has changed in a subtle way in the source material in #668. This means that the PO file is slightly outdated.

You have to refresh the `ja.po` file to incorporate the latest changes. You [use `msgmerge` to do this](https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md#updating-an-existing-translation). It will match the old English text with the new paragraph and re-use your translation. However, the whole entry is marked as "fuzzy", which means that someone (you) should proof-read it before removing the fuzzy marker.

This step is a bit annoying: `msgmerge` will cause a lot of diffs since it likely formats the file differently compared to your PO editor and compared to the format used by `mdbook-xgetttext`. Someone (probably me) needs to add some normalization code to the `mdbook-i18n-helpers` so that we can have more stable output from these tools.

For now, I would just ignore this and keep translating the PO file. You can decide when to synchronize the `ja.po` file when you feel there has been [enough changes to `src/`](https://github.com/google/comprehensive-rust/commits/main/src) to make it worth the trouble.

mgeisler:
@CoinEZ-JPN are you able to merge this PR now that it has been approved? I'm not actually sure if you can do that without having write access. If not, one of @keiichiw or @chikoski should merge it.

CoinEZ-JPN:
@mgeisler sorry, it doesn't look like I can

s/s↓
Screenshot 2023-06-03 at 10 41 16

mgeisler:
> @mgeisler sorry, it doesn't look like I can
> 
> s/s↓ Screenshot 2023-06-03 at 10 41 16

Thanks for testing, I'll merge it now!

PR #730, opened 2023-05-30
Link to incomplete French, German, and Japanese translations

mgeisler:
These translations all have more than 200 translated messages and they’ve been updated in the last month.

I suggest we update the list here every few weeks based on input from the translators: we don’t need a super strict rule here, we just need interested translators who would like to see their work celebrated.

I suggest we only link completed translations in the language picker since we have limited space there to differentiate the different levels of completeness.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -19,7 +19,7 @@ concurrency:
 env:
   CARGO_TERM_COLOR: always
   # Update the language picker in index.hbs to link new languages.
-  LANGUAGES: da it pt-BR ko
+  LANGUAGES: da de fr it ja ko pt-BR
 
 jobs:
   publish:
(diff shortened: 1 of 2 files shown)


mgeisler:
Hi @KookaS, @Throvn, and @CoinEZ-JPN. What do you think of this idea? I would like to showcase your work (:tada:) by making it possible to find the in-progress translations :slightly_smiling_face:

Throvn:
I like the idea.
Unfortunately, I have a lot on my plate right now. Therefore, I cannot promise to keep up with the pace of changing translations. That being said, I've just opened another PR with all of my changes so far. Whenever I continue the translations I will just open up another PR. If this is too little to get listed, I fully understand that.

That being said, a self imposed goal is to do one PR a week now :).

qwandor (approved):

HAMMERABA (approved):

PR #735, opened 2023-05-30
Continued german translation

Throvn:
Translated roughly until `3. Why Rust?`.


ronaldfw:
@Throvn Thanks for your additional work here. Just want to clarify: What is the plan for the PR? This still seems to be work-in-progress, but you've also opened this for review. IMO, you should aim to have smaller parts reviewed and actually merged before adding additional translation work in additional commits (which then would go into another PR).

ronaldfw:
> Since there was no activity I was just commiting on top of it. But I get that smaller PRs are better to review. I will declutter the PR and notify you once it's ready. Would one PR per chapter be small enough?

Yeah, probably.

mgeisler:
> Since there was no activity I was just commiting on top of it. But I get that smaller PRs are better to review.

The diff seems large because of `msgmerge` — if I were you, I would avoid merging the `main` branch into your on-going work. What way the many many file locations will stay stable while you work and the diff should shrink a lot.

Once the translation is 50% or 100% caught up, then I would do a `msgmerge` and fix the fuzzy entries. Until then, it creates more noise since the typo fixes are mixed with real translations.

Throvn:
Just to make sure that I get it: 
1. The best way to go about it is to only `msmerge` once at the beginning of every PR.
2. I should open up a seperate PR for each chapter (~60 chapters so far) to keep the changes small.

Or do you mean to translate the entire book at the current point and time, and then maintain it by `msmerge`ing the changes that were made during that time?

mgeisler:
> 1. The best way to go about it is to only `msmerge` once at the beginning of every PR.

Even more rarely: I would run `msgmerge` every 2-3 weeks or so until the translation is done.

When you run `msgmerge`, I would do it as the only thing in the PR. That way the reviewers know that there is nothing to review in the large diff.

When the translation is complete or nearly completely, then I agree with you: running `msgmerge` as the first thing makes sense. It will bring in the latest changes (hopefully not very many) and they can then be translated in one go.

mgeisler (approved):
Danke für die grosse Arbeit! Ich habe es so Stichprobe-mässig durchgelesen und es sieht gut aus — aber ich bin Däne, so mein Deutsch ist nicht immer vertrauenswürdig :smile: 

Let us merge this and then we can see the results on https://google.github.io/comprehensive-rust/de/ and fix anything that might need fixing.

mgeisler:
@Throvn, you should now be able to make smaller PRs going forward. This simplifies your workflow too: no merging, no `msgmerge`, etc. You just have to edit the `de.po` file with a consistent setup of your PO editor.

Throvn:
Got it and thank you for letting this one slide haha! It won't happen again. I will always work on it for at least an hour on Sunday so on Monday you should have a clean PR of my progress.

mgeisler:
> Got it and thank you for letting this one slide haha! It won't happen again. I will always work on it for at least an hour on Sunday so on Monday you should have a clean PR of my progress.

Thanks for all the hard word! :smile:

PR #737, opened 2023-05-31
zh-Hans: Translation for src/why-rust/compile-time.md

sylxjtu:
Part of https://github.com/google/comprehensive-rust/issues/324


po/zh-Hans.po
```diff
@@ -2269,12 +2278,17 @@ msgid ""
 "* In fact, some will consider infinitely populating a collection a memory\n"
 "  leak and Rust does not protect from those."
 msgstr ""
+"* 可以使用 [`Box::leak`] 来泄漏一个指针。该方法可以用于得到在运行时决定大小和初始化的静态变量\n"
+"* 可以使用 [`std::mem::forget`] 来让编译器“忘记”一个值（即其析构函数不会被执行）。\n"
+"* 可以使用 `Rc` 或 `Arc` 意外创建一个 [循环引用]。\n"
```
wnghl, line 2283: Here's a link of "reference cycle", if change it to Chinese the link will become a raw text:
image
image
```suggestion
"* 可以使用 `Rc` 或 `Arc` 意外创建一个循环引用（[reference cycle]）。\n"
```

wnghl (changes requested):

wnghl (approved):

po/zh-Hans.po
sylxjtu: I wonder is there any other solution other than keep the original text to be able to use these links? I searched for "reference cycle" but cannot find that in other places in the po file.

mgeisler: @sylxjtu, it's tricky! I wrote a bit about the problem in our [style guide](https://github.com/google/comprehensive-rust/blob/main/STYLE.md#translations). In short, we only extract the text part from the Markdown files, so the PO files don't have the link _definition_. They only have the link _usage_, so you have to be careful to align the use with the definition.

This is clearly something that could be improved in https://github.com/google/mdbook-i18n-helpers/, but I'm not sure exactly how.

wnghl: @mgeisler Oh I got it, there are some hidden line not shown in markdown preview mode which represents the link, and we need to add it to our po file like below: 
image
But I found it doesn't work in zh-Hans.po, pt-BR.po and es.po like this: 
image
I wonder how to fix this.

mgeisler: @wnghl, yes, you got it! The translation system is really a very indirect way of letting you construct a Markdown document. This means that you can inject arbitrary Markdown in your translations (such as `**foo**` and `[foo](http://example.com)`) and it also means that you have to keep the Markdown rules vaguely in mind.

Here the problem is in the pt-BR translation: they should not have translated the reference link (@rastringer and @hugojacob, FYI) since it refers to a link definition which is outside the PO file.

I'll send a PR for this next since it's easy to fix when you know how (and easy to fix even without speaking Portuguese) :smile:

mgeisler: I originally used the short

```markdown
foo [bar] baz

bar: http://example.net/
```

style links since I think they look great in the English source text. However, because of how they make translations harder, I'm thinking of switching them all to the inline style:

```markdown
foo [bar](http://example.net/) baz
```

What would make it much harder for the links to go out of sync in the translation and it would also allow the translator to replace the link with a different one, e.g., to the local Wikipedia page or so.

PR #739, opened 2023-05-31
Update lifetimes.md

FlorianvdEnt:
Don't think elidied is a word, nothing comes up if I google it. I think omitted works here.


qwandor-google:
I think 'elided' is the word that was intended. See https://doc.rust-lang.org/reference/lifetime-elision.html for comparison.

qwandor:
I think 'elided' is the word that was intended. See https://doc.rust-lang.org/reference/lifetime-elision.html for comparison.

FlorianvdEnt:
Ah I see, that makes more sense. Changed it.

qwandor (approved):

Issue #323, opened 2023-02-02
French translation


mgeisler:

mgeisler:
Hi @dblackstar, are you able to assign this issue to yourself?

mgeisler:
Here is a Cloud Translate version of the `messages.po` file: [fr.zip](https://github.com/google/comprehensive-rust/files/10570187/fr.zip). I hope that can work as a good starting point.

mgeisler:
> Oh yes it help a lot,

Excellent! Also, I just noted that there is a French translation of the Rust Book. That might be useful too: https://github.com/Jimskapt/rust-book-fr.

> i was figuring out how to host the translated book.

My plan is to publish the translations with the English version. We already publish a very incomplete Danish version here: https://google.github.io/comprehensive-rust/da/ (I'll add links to it when we have a more complete version).

dblackstar:
>> Excellent! Also, I just noted that there is a French translation of the Rust Book. That might be useful too: https://github.com/Jimskapt/rust-book-fr.
>
>Okay
> 
> >My plan is to publish the translations with the English version. We already publish a very incomplete Danish version here: https://google.github.io/comprehensive-rust/da/ (I'll add links to it when we have a more complete version).
>
>Okay it's clear, i will just PR with the translated fr.po file

ZoeThivet:
@dblackstar I'd be happy to help by reviewing the translation :)
I'm a native French speaker like you and I'm willing to provide an "inclusiveness" review so that the labels are gender neutral

dblackstar:
> @dblackstar I'd be happy to help by reviewing the translation :) I'm a native French speaker like you and I'm willing to provide an "inclusiveness" review so that the labels are gender neutral

@ZoeThivet Yes sure it would be nice to work on that togother. Fork my Repo and for the guide on translation we follow https://github.com/Jimskapt/rust-book-fr

mgeisler:
Hi @dblackstar and @ZoeThivet, I would very much encourage working together on this — there are nearly 30k words to translate so it's a big task!

The machine generated translation can perhaps help to get your started:

> Here is a Cloud Translate version of the `messages.pot` file: [fr.zip](https://github.com/google/comprehensive-rust/files/10570187/fr.zip).

mgeisler:
> hey I'd like to help with the translation too, how can I participate? I'm a native french speaker

Hi @KookaS, that would be amazing!

[...]

PR #743, opened 2023-06-01
fr: Updating French translation file, and a few messages.

yohcop:
Re-generated a new po/messages.pot, and merged into po/fr.po.

Updated a handful of message translations as well, but this small PR is mostly to 'get my feet wet' with the whole process.

Part of #323.

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -14,7 +14,7 @@ msgstr ""
 
 #: src/SUMMARY.md:3
 msgid "Welcome to Comprehensive Rust 🦀"
-msgstr "Bienvenue à Comprehensive Rust(le guide complet de Rust)🦀"
+msgstr "Bienvenue à Comprehensive Rust (le guide complet de Rust) 🦀"
 
 #: src/SUMMARY.md:4
 msgid "Running the Course"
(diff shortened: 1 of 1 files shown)


mgeisler:
Thanks @yohcop! I added a reference to the French translation issue (that way the PR shows up on the issue).

yohcop:
I'm not sure if I did this right. The file is so big, as most line numbers have changed, that github doesn't want to load the diff. Let me know if I should do things differently.

sakex (approved):
This is a huge CL, we can merge it like this for now. I will have a longer look on this next week

sakex:
Also, it doesn't let me add my review comments for some reason

mgeisler:
> I'm not sure if I did this right. The file is so big, as most line numbers have changed, that github doesn't want to load the diff. Let me know if I should do things differently.

You did it right! The problem is that `msgmerge` formats the PO file differently than our `mdbook-xgettext` plugin. I should fix that so I filed https://github.com/google/mdbook-i18n-helpers/issues/32 for it.

mgeisler:
I merged this, thanks for looking it over!

@sakex, you should generally merge PRs when approving them since most people won't have write access to the repository.

PR #748, opened 2023-06-01
Mentioned the XKCD source of the joke in enums.md

TPReal:
XKCD is under CC BY-NC 2.5 licence, so it is fair to at least link to the source when reusing a joke.


djmitche (commented):
Fair point, but let's make it a full-line comment on the line before the `4`?

TPReal:
Reworded, hopefully you like it better.

djmitche (approved):
Haha, looks good :)

PR #749, opened 2023-06-02
zh-TW: translation for src/generics/methods.md

kuanhungchen:
Part of #684.


po/zh-TW.po
```diff
@@ -7343,6 +7343,13 @@ msgid ""
 "      * `Point` is still generic and you can use `Point<f64>`, but methods "
 "in this block will only be available for `Point<u32>`."
 msgstr ""
+"* *問題：* 為什麼 `T` 在 `impl<T> Point<T> {}` 中被使用了兩次？\n"
```
mingyc, line 7346: > 被使用了兩次?

`specified twice` -> `被寫出了兩次`

`Isn’t that redundant?` is missing from the translation. May be update to `被重複寫出了兩次` or add a direct translation

victorhsieh: Hmm... 使用 is easier to understand to me than 寫出

kuanhungchen: How about 出現了兩次? So that we don't need to write in a passive way.
(To me, 被寫出 and 出現 are somehow similar. Both of them is better than 被使用 as we haven't actually "使用" the T yet.)

kuanhungchen: Updated.

mingyc (approved):
LGTM

PR #755, opened 2023-06-02
zh-TW: translation for src/generics/monomorphization.md

kuanhungchen:


po/zh-Hant.po
```diff
@@ -2,7 +2,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Comprehensive Rust 🦀\n"
 "POT-Creation-Date: \n"
-"PO-Revision-Date: 2023-06-01 22:43+0800\n"
+"PO-Revision-Date: 2023-06-02 14:43+0800\n"
```
mgeisler, line 5: A small detail: I've configured my own editor to stop updating this field. It is from a forgotten era where people liked to add such identifiers inline — now we can use our version control system instead :smile:

po/zh-Hant.po
```diff
@@ -7346,11 +7346,13 @@ msgstr ""
...
 #: src/generics/monomorphization.md:1
 msgid "# Monomorphization"
-msgstr ""
+msgstr "# 單型化"
 
 #: src/generics/monomorphization.md:3
 msgid "Generic code is turned into non-generic code based on the call sites:"
 msgstr ""
+"Rust 在編譯時進行單型化 (Monomorphization)，將泛型程式碼轉換成實際型別的程式"
+"碼："
```
mingyc, line 7355: How do you decide when to wrap a line? It looks like the above line is shorter than the enlish message. Also, I only found lineWidth = 80 set [here](https://github.com/google/comprehensive-rust/blob/1d730b0e1bf6c1e1cc8546cff2666ae41101e678/dprint.json#L2).

po/zh-Hant.po
kuanhungchen: I see. I'll stop uploading this.

po/zh-Hant.po
kuanhungchen: I don't wrap lines. The editor does it for me.

kuanhungchen: I set the editor to wrap at 80 now. Thanks!

po/zh-TW.po
```diff
@@ -7392,6 +7394,8 @@ msgid ""
 "had\n"
 "hand-coded the data structures without the abstraction."
 msgstr ""
+"這種抽象化不會影響程式碼執行時的效能：單型化的結果，等同於寫出資料結構的實際"
+"型別而不使用抽象化。"
```
mingyc, line 7398: nit: I feel it's more fluent with "等同於不使用抽象化並手動寫出資料結構的實際型別"

po/zh-TW.po
```diff
@@ -7392,6 +7394,8 @@ msgid ""
 "had\n"
 "hand-coded the data structures without the abstraction."
 msgstr ""
+"這種抽象化不會影響程式碼執行時的效能：單型化的結果，等同於寫出資料結構的實際"
```
mingyc, line 7397: > 這種抽象化不會影響程式碼執行時的效能

While this translation may be better to understand, it largely differs from the meaning of "zero-cost abstraction".

"零成本抽象" is also a thing in 
https://github.com/search?q=repo%3Arust-tw%2Fbook-tw+%E9%9B%B6%E6%88%90%E6%9C%AC&type=code. We should probably also use it here.

po/zh-Hant.po
mgeisler: Just a side note: since you're writing Markdown here, the rendered HTML is the same regardless of where you wrap the lines.

However, for consistency and for somewhat-readable diffs, it's great if the team decides on somewhat similar settings here.

As a secondary side-note, I've been playing with using [`msgfilter`](https://www.gnu.org/software/gettext/manual/html_node/msgfilter-Invocation.html) in combination with [`dprint`](https://dprint.dev/) to reformat the Markdown inside automatically. It was slow, so I didn't spend much time on it. But it's something to consider if teams would like to have auto-formatted `msgstr` entries.

po/zh-TW.po
kuanhungchen: Updated, thanks!

po/zh-TW.po
kuanhungchen: Updated.

po/zh-TW.po
```diff
@@ -7346,11 +7346,13 @@ msgstr ""
 
 #: src/generics/monomorphization.md:1
 msgid "# Monomorphization"
-msgstr ""
+msgstr "# 單型化"
 
 #: src/generics/monomorphization.md:3
 msgid "Generic code is turned into non-generic code based on the call sites:"
 msgstr ""
+"Rust 在編譯時進行單型化 (Monomorphization)，將泛型程式碼轉換成實際型別的程式"
```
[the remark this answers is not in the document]
kuanhungchen: Updated.

victorhsieh (approved):
LGTM as all comments are resolved.

mingyc:
LGTM

PR #762, opened 2023-06-02
Move the xkcd credit to a footnote

djmitche:
Since this is a joke, delivery is important, and putting the credit in the middle of the joke kind of ruins that delivery. Hopefully this duly credits Munroe while still preserving the humor!

/cc @TPReal who suggested the credit initially in #748.

### DIFF
--- a/src/enums.md
+++ b/src/enums.md
@@ -5,8 +5,7 @@ different variants:
 
 ```rust,editable
 fn generate_random_number() -> i32 {
-    // Implementation based on https://xkcd.com/221/
-    4  // Chosen by fair dice roll. Guaranteed to be random.
+    4  // Chosen by fair dice roll. Guaranteed to be random. [1]
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


TPReal:
My first thought to the suggestion of putting the reference "on the line before the `4`" was that it would kind of ruin the joke. But then I phrased it as "Implementation based on ..." and for me that is one-and-a-half jokes at the cost of one, and I liked it. But I'm fine with the change if you really think it's better this way.

mgeisler:
> But then I phrased it as "Implementation based on ..." and for me that is one-and-a-half jokes at the cost of one, and I liked it.

I actually read it the same way: the comment is extra funny now because it pretends to be a "serious" comment :smile: Similar to when an academic paper for some RNG or crypto implementation.

At first I felt that having a comment would take away from the joke, but this comment is actually nice.

djmitche:
ah, ok, let's leave it then!

PR #770, opened 2023-06-05
Add suggestion to use `rustup doc`

metaspace:
Suggest to use `rustup doc std::fmt` to open the documentation for `std::fmt`.


metaspace:
> @metaspace, could you sign the CLA so we can merge this?

Done! Had to check with employer how to handle CLA first.

djmitche:
Thanks!

PR #776, opened 2023-06-05
it: Further misc Italian translations

detro:
As per title.

I'm trying to get through as many "short" or "title" translations, before I engage with the big parts.

This time I haven't updated the based document, to avoid any spurious change caused by an update in the source `.po` file.

### DIFF
--- a/po/it.po
+++ b/po/it.po
@@ -2,7 +2,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Comprehensive Rust 🦀\n"
 "POT-Creation-Date: \n"
-"PO-Revision-Date: 2023-05-16 20:12+0100\n"
+"PO-Revision-Date: 2023-06-05 22:30+0100\n"
 "Last-Translator: \n"
 "Language-Team: \n"
 "Language: it\n"
@@ -10,7 +10,7 @@ msgstr ""
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
-"X-Generator: Poedit 3.2.2\n"
+"X-Generator: Poedit 3.3.1\n"
 
 #: src/SUMMARY.md:3
 msgid "Welcome to Comprehensive Rust 🦀"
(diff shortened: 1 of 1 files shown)


mgeisler:
> This time I haven't updated the based document, to avoid any spurious change caused by an update in the source `.po` file.

Grazie! The diff looks great from that point of view 🙂

po/it.po
```diff
@@ -824,219 +824,176 @@ msgid "Scoped Threads"
...
 #: src/SUMMARY.md:274 src/SUMMARY.md:294
-#, fuzzy
 msgid "Dining Philosophers"
 msgstr "Filosofi a tavola"
 
 #: src/SUMMARY.md:275
-#, fuzzy
 msgid "Multi-threaded Link Checker"
-msgstr "Verifica link multi-thread"
+msgstr "Verificatore di Link a thread multipli"
```
nicomazz, line 872: This sounds a bit weird, but I wouldn't know how to translate it better

po/it.po
```diff
@@ -824,219 +824,176 @@ msgid "Scoped Threads"
...
 #: src/SUMMARY.md:277
-#, fuzzy
 msgid "Concurrency: Afternoon"
 msgstr "Giorno 1: Pomeriggio"
 
 #: src/SUMMARY.md:279
-#, fuzzy
 msgid "Async Basics"
-msgstr "Nozioni di base sull'asincronismo"
+msgstr "Nozioni di base su asincronia (Async)"
```
nicomazz, line 880: Optional nit: Not sure how to translate it better, but "asincronia" sounds pretty bad. Maybe "Async 101"?

nicomazz (approved):
I have to say that some automatic translations were more fun than expected. Feel free to ignore the nits i've added.

po/it.po
detro: I know. But I couldn't come up with anything better.

detro:
I pushed changes to the branch (attempts to address what @nicomazz highlighted) but... https://www.githubstatus.com/incidents/1g1gkh0qpyvs

I was really confused when I pushed but the PR wasn't changing: my luck I guess!

NoahDragon (approved):
Changes in `po/zh-CN.po` LGTM

PR #777, opened 2023-06-06
ja: Ch.3 #652

CoinEZ-JPN:
short chapter, did not separate to sections

Translation not reflected on Ch.3.0:
"They should ask questions ... you are free to skip them as you like. "


keiichiw (approved):
LGTM!

PR #781, opened 2023-06-07
zh-TW translation for src/control-flow.md

kuanhungchen:
Part of #684.


po/zh-TW.po
```diff
@@ -5867,6 +5867,9 @@ msgid ""
 "similarly\n"
 "in Rust."
 msgstr ""
+"如同我們所見，`if` 是 Rust 中的一種表達式。它是用來根據條件判斷兩個區塊之一的"
```
victorhsieh, line 5870: I don't have the context, but does it make any better sense to say 在 Rust 中 `if` 是（or 其實也是）一種表達式？
它是 -> 它可以？
判斷 -> 執行 (並

victorhsieh, line 5870: I don't have the context, but does it make any better sense to say 在 Rust 中 `if` 是（or 其實也是）一種表達式？
它是 -> 它可以？
判斷 -> 執行 (並傳回)？

kuanhungchen: Here `if` is a thing that already appears in some examples before this page. To me it sounds more natural to start the sentence with `if` (which is what it wants to introduce here).
For the other suggestions, I updated accordingly. Thanks a lot!

victorhsieh (approved):

PR #789, opened 2023-06-09
Warn against running `msgmerge` too often

mgeisler:


mgeisler:
> Approved, but I admit I don't understand the issue here. For folks not as accustomed to Git and this process, maybe you could be more specific than "often"? Should it be run once before starting on a PR? Once at the end of some translation work, before submitting the PR?

You're right... let me be more prescriptive! I'm trying to gently warn people against producing huge PRs like in #735 which mixes both `msgmerge` and new translations.

I'll try to rework the text a bit.

TRANSLATIONS.md
```diff
@@ -104,6 +104,11 @@ updated messages are marked "fuzzy". A fuzzy entry will reuse the previous
 translation: you should then go over it and update it as necessary before you
 remove the fuzzy marker.
 
+> **Note:** You should not run `msgmerge` too often. The reason is that it often
+> creates a very large diff, which is hard or impossible to review. Focus on
+> finishing the translation first and only then start incorporating new changes
+> with `msgmerge`.
```
mgeisler, line 110:
```suggestion
> **Note:** Until the translation is complete, you should either run `msgmerge` or
> update the PO file for your language. Avoid mixing the two since it often creates a
> very large diff, which is hard or impossible to review. When the translation is complete,
> run `msgmerge` to incorporate changes in the English text.
```

djmitche (approved):
Much better!

PR #786, opened 2023-06-08
zh-TW: Translate Concurrency morning

mgeisler:
Hi all,

This translation was prepared by our in-house localization team. Since I don't speak Chinese, please look it over and add suggestions.

Let me know how good the translation is — if it is good, I can send them bigger parts to translate!

Thanks for your help! Part of #684.

PR #792, opened 2023-06-10
Rephrase when `send` blocks

mgeisler:
From a discussion with @hueich: https://github.com/google/comprehensive-rust/pull/786#discussion_r1224850499.

### DIFF
--- a/src/concurrency/channels/bounded.md
+++ b/src/concurrency/channels/bounded.md
@@ -30,5 +30,6 @@ fn main() {
     
 * Calling `send` will block the currnet thread until there is space in the channel for the new message. The thread can be blocked indefinitely if there is nobody who reads from the channel.
 * A call to `send` will abort with an error (that is why it returns `Result`) if the channel is closed. A channel is closed when the receiver is dropped.
+* A bounded channel with a size of zero is called a "rendezvous channel". Every send will block the current thread until another thread calls `read`.
     
 </details>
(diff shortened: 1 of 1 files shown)


mgeisler:
Hi @hueich, I hope this makes the sentence a little clearer.

hueich (approved):
Looks clearer, thanks!

hueich:
Btw, is it possible to create a zero-length bounded channel (like in Go)?  If so, what's the behavior there?

mgeisler:
> Btw, is it possible to create a zero-length bounded channel (like in Go)? If so, what's the behavior there?

Yes, that's possible and while reading up on channels, I learned that this is called a "rendezvous channel". Every `send` will block until a `read` is done. I can add a line about this.

src/concurrency/channels/bounded.md
```diff
@@ -25,3 +25,10 @@ fn main() {
     }
 }
 ```
+
+<details>
+    
+* Calling `send` will block the currnet thread until there is space in the channel for the new message. The thread can be blocked indefinitely if there is nobody who reads from the channel.
+* A call to `send` will abort with an error (that is why it returns `Result`) if the channel is closed. A channel is closed when the receiver is dropped.
```
mgeisler, line 32:
```suggestion
* A call to `send` will abort with an error (that is why it returns `Result`) if the channel is closed. A channel is closed when the receiver is dropped.
* A bounded channel with a size of zero is called a "rendezvous channel". Every send will block the current thread until another thread calls `read`.
```

PR #795, opened 2023-06-11
zh-TW: fix style in src/testing.md

kuanhungchen:
According to the localization guide, colons, in documentation, should be 全形.


kuanhungchen:
@johnathan79717
I've been referring to the [Microsoft localization guide](https://learn.microsoft.com/en-us/globalization/localization/ministyleguides/mini-style-guide-trad-chinese#4-punctuation) while translating.

PR #799, opened 2023-06-12
zh-TW: translation for src/control-flow/break-continue.md

kuanhungchen:
I couldn't find a good translation for "break" and "continue", so I just followed Korean version for not translating them.

Part of #684.

### DIFF
--- a/po/zh-TW.po
+++ b/po/zh-TW.po
@@ -6268,7 +6268,7 @@ msgstr ""
 
 #: src/control-flow/break-continue.md:1
 msgid "# `break` and `continue`"
-msgstr ""
+msgstr "# `break` and `continue`"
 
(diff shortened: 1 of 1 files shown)


po/zh-TW.po
```diff
@@ -6268,7 +6268,7 @@ msgstr ""
 
 #: src/control-flow/break-continue.md:1
 msgid "# `break` and `continue`"
-msgstr ""
+msgstr "# `break` and `continue`"
```
mingyc, line 6271: > `break` and `continue`

nit: `break` 和 `continue`

mingyc (approved):
LGTM with a nit

po/zh-TW.po
kuanhungchen: Nice catch, thanks!

PR #802, opened 2023-06-12
Add pages about entry point and exception handling on APs

qwandor:
Also clarified issues with Rust Raspberry Pi OS tutorial code.

### DIFF
--- a/src/bare-metal/aps/entry-point.md
+++ b/src/bare-metal/aps/entry-point.md
@@ -29,5 +29,7 @@ Before we can start running Rust code, we need to do some initialisation.
   GiB of address space for devices, the next 1 GiB for DRAM, and another 1 GiB higher up for more
   devices. This matches the memory layout that QEMU uses.
 * We also set up the exception vector (`vbar_el1`), which we'll see more about later.
+* All examples this afternoon assume we will be running at exception level 1 (EL1). If you need to
+  run at a different exception level you'll need to modify `entry.S` accordingly.
 
 </details>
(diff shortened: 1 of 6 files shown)


src/bare-metal/aps/entry-point.md
```diff
@@ -0,0 +1,33 @@
...
+  * If it were running in a VM, this can lead to cache coherency issues. The problem is that the VM
+    is accessing memory directly with the cache disabled, while the host has cachable aliases to the
+    same memory. Even if the host doesn't explicitly access the memory, speculative accesses can
+    lead to cache fills, and then changes from one or the other will get lost when the cache is
+    cleaned or the VM enables the cache. (Cache is keyed by physical address, not VA or IPA.)
+* For simplicity, we just use a hardcoded pagetable (see `idmap.S`) which identity maps the first 1
+  GiB of address space for devices, the next 1 GiB for DRAM, and another 1 GiB higher up for more
+  devices. This matches the memory layout that QEMU uses.
+* We also set up the exception vector (`vbar_el1`), which we'll see more about later.
+
```
schultetwin1, line 32: This may not be the right place for it, but what about adding a note about "we assume we will be running at EL1" here? I see you mention it in exceptions.md but it maybe nice to bring up earlier.

```suggestion
* All examples in this course assume we will be running at exception level 1 (EL1). If you need to run at a different exception level you will need to modify entry.S to suite your needs.
```

schultetwin1 (approved):
Looks great!

src/bare-metal/aps/entry-point.md
qwandor: Done.

HAMMERABA (approved):

PR #810, opened 2023-06-13
Fix typo in mutex.md

Millione:


src/concurrency/shared_state/mutex.md
```diff
@@ -32,7 +32,7 @@ implementation.
     * It is not possible to forget to acquire the mutex before accessing the protected data.
 * You can get an `&mut T` from an `&Mutex<T>` by taking the lock. The `MutexGuard` ensures that the
   `&mut T` doesn't outlive the lock being held.
-* `Mutex<T>` implements both `Send` and `Sync` iff `T` implements `Send`.
+* `Mutex<T>` implements both `Send` and `Sync` if `T` implements `Send`.
```
mgeisler, line 35: This is not actually a typo, instead it's a shorthand for "if and only if":

```suggestion
* `Mutex<T>` implements both `Send` and `Sync` iff (if and only if) `T` implements `Send`.
```

mgeisler (approved):
Thanks for flagging this and reminding me to explain it!

PR #822, opened 2023-06-15
Make User pub to avoid unused code warnings when running tests

qwandor:
When the tests are run without this on Playground, it gives a bunch of warnings about unused methods on `User`, because they are not part of the public API of the library. Making `User` public avoids most of them, and seems like a reasonable approach for people to take.


djmitche (approved):
Some of the other exercises have `#![allow(dead_code)]` to address these warnings. `pub User` is fine though, and makes sense in this case.

mgeisler (approved):

PR #826, opened 2023-06-18
Document the historical reference in the newtypes example

pwnall:
I hope this helps the folks running the course feel a bit more prepared / less surprised 😄


mgeisler (approved):
Thanks! It's always a lot of fun to teach this slide since there will be a few people in the audience who smiles when they get the reference :smile:

PR #833, opened 2023-06-19
"Strings and Iterators" solution that uses let-else

pwnall:
This PR is intended as the start of a discussion -- I'm not entirely sure it's an improvement as-is. I came up with the solution below for the Day 2 Afternoon exercise.

I was looking for something conceptually similar to the current solution -- `zip()`, but extending one of the iterators. So, I think the current solution teaches a trick that will be useful in practice, and should stand. I think my solution may be a bit easier to follow.

### DIFF
--- a/src/exercises/day-2/strings-iterators.rs
+++ b/src/exercises/day-2/strings-iterators.rs
@@ -15,23 +15,25 @@
 // ANCHOR: prefix_matches
 pub fn prefix_matches(prefix: &str, request_path: &str) -> bool {
     // ANCHOR_END: prefix_matches
-    let prefixes = prefix.split('/');
-    let request_paths = request_path
-        .split('/')
-        .map(|p| Some(p))
-        .chain(std::iter::once(None));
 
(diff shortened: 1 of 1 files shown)


mgeisler:
Hi @pwnall,

Looking at the solution I wrote again, I definitely think it could benefit from some comments! Your solution is actually simpler to understand since it is more explicit with the call to `next()`.

I would prefer to not include two full solutions (they'll go out of sync and they overload the solution slide). So could we instead use your solution and then explain the current "trick" in a comment? The double `Option<Option<&str>>` type is a bit weird and so I never truly liked the trick :smile:

pwnall:
Sorry, for the slow reply. I'll turn this around in the next couple of days.

pwnall:
@mgeisler Please take another look?

mgeisler (approved):
Yeah, that's perfect! Thanks!

pwnall:
@mgeisler Thank you very much for the quick review!

PR #830, opened 2023-06-19
Run builds on both Mac OS and Linux

mgeisler:
This would have helped us catch #570.

PR #843, opened 2023-06-21
Testing only -- speculative fix for #830

pwnall:


domenukk:
First (stupid) question, why are you creating `d_name` as a `CStr` and then `OsStr::from_bytes` on `d_name.to_bytes()` and not create `os_str` from `d_name` directly?
Next, if you have alignment issues, the safest option might be to create a local `let d_name = [0_u8; 1024]` and copy the values into that local array, before using it in other rust APIs. The local variable should be aligned properly.

mgeisler:
Hi @pwnall, I'll close this now that #857 is merged.

pwnall:
Thank you very much, apologies I forgot to clean up!

mgeisler:
> Thank you very much, apologies I forgot to clean up!

It's no problem at all, thanks for helping out!

PR #845, opened 2023-06-21
Add PR labeler action to label translation-related PRs

mingyc:


.github/labeler.yml
```diff
@@ -0,0 +1,47 @@
...
+- po/bn.po
+
+'translation/da':
+- po/da.po
+
+'translation/de':
+- po/de.po
+
+'translation/el':
+- po/el.po
```
mgeisler, line 14: Nice idea! I'm worried about having to maintain the list of languages here, but I guess we can add them on a best-effort basis?

mgeisler (approved):

.github/labeler.yml
mingyc: Yes. [This action](https://github.com/actions/labeler#inputs) does not support programmatically adding labels. The configuration here falls back to `translation` label by default for new languages.

mgeisler: Okay, sounds good. Could you add a line to `TRANSLATIONS.md` to ask people to update the configuration when adding a new translation?

mingyc: Done.

mgeisler: Thank you!

PR #848, opened 2023-06-21
Fix MacOS CI

domenukk:
This extends #830 with MacOS fixes

### DIFF
--- a/src/exercises/day-3/safe-ffi-wrapper.rs
+++ b/src/exercises/day-3/safe-ffi-wrapper.rs
@@ -45,7 +45,7 @@ mod ffi {
         pub d_reclen: u16,
         pub d_namlen: u16,
         pub d_type: u8,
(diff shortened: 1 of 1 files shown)


domenukk:
This tries to address CI failures in #830

domenukk:
Seems like it's no longer crashing, but the results are wrong...

Maybe related
https://github.com/rust-lang/libc/issues/414 ?

domenukk:
I'm not sure if `addr_of!` is still needed, or if the crash was related to the false struct definition (?) but it's probably safer to leave it in

mgeisler:
> I'm not sure if `addr_of!` is still needed, or if the crash was related to the false struct definition (?) but it's probably safer to leave it in

Yeah, not 100% sure about this myself. Let me merge it back into my branch and then I'll try both ways.

PR #857, opened 2023-06-22
Unified FFI setup for for readdir() on macOS

pwnall:
This PR switches all macOS code to use the modern (introduced in macOS 10.5) ABI. I think this is a better approach than forking on the macOS platform because

1. This makes macOS x86_64 closer to macOS aarch64, so CI (which only has x86_64) is more likely to catch bugs that would manifest on aarch64.
2. Commenting this code lets us sneak in enough details (such as `readdir$INODE64`) so interested folks can research further.

### DIFF
--- a/src/exercises/day-3/safe-ffi-wrapper.rs
+++ b/src/exercises/day-3/safe-ffi-wrapper.rs
@@ -16,7 +16,7 @@
 mod ffi {
     use std::os::raw::{c_char, c_int};
     #[cfg(not(target_os = "macos"))]
-    use std::os::raw::{c_long, c_ulong, c_ushort};
+    use std::os::raw::{c_long, c_ulong, c_ushort, c_uchar};
 
     // Opaque type. See https://doc.rust-lang.org/nomicon/ffi.html.
     #[repr(C)]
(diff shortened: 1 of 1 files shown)


src/exercises/day-3/safe-ffi-wrapper.rs
```diff
@@ -25,43 +25,46 @@ mod ffi {
...
+    // Layout according to the Linux man page for readdir(3), where ino_t and
+    // off_t are resolved according to the definitions in
+    // /usr/include/x86_64-linux-gnu/{sys/types.h, bits/typesizes.h}.
     #[cfg(not(target_os = "macos"))]
     #[repr(C)]
     pub struct dirent {
-        pub d_ino: c_long,
-        pub d_off: c_ulong,
+        pub d_ino: c_ulong,
+        pub d_off: c_long,
```
mgeisler, line 35: Just out of curiosity, how did you find this? I traced this way back and it's super likely that I misread something when pretending to a be a C preprocessor in my head :smile:

mgeisler (approved):
Thanks, this is great and I love your point about triggering a discussion about the `link_name` attribute :+1:

src/exercises/day-3/safe-ffi-wrapper.rs
```diff
@@ -16,7 +16,7 @@
 mod ffi {
     use std::os::raw::{c_char, c_int};
     #[cfg(not(target_os = "macos"))]
-    use std::os::raw::{c_long, c_ulong, c_ushort};
+    use std::os::raw::{c_long, c_ulong, c_ushort, c_uchar};
```
mgeisler, line 19: I notice that `uchar` was introduced in Rust 1.64. We don't have a strict lower bound for the Rust compiler version in the course — but there is often someone who tries to use the one shipped in our gLinux Debian install. I should probably just clean up the installation instructions to be clear that people need to install from rustup.rs.

djmitche: It might be worth mentioning "MSRV" in those instructions -- it's a good concept to be aware of anyway.

src/exercises/day-3/safe-ffi-wrapper.rs
pwnall: https://android.googlesource.com/platform/prebuilts/gcc/linux-x86/host/x86_64-linux-glibc2.15-4.8/+/a5ef72f/sysroot/usr/include/x86_64-linux-gnu/sys/types.h has off_t --> __off_t / __off64_t and something similar for ino_t.

I looked at the includes, and had a lucky guess that bits/types.h would be useful.

https://android.googlesource.com/platform/prebuilts/gcc/linux-x86/host/x86_64-linux-glibc2.15-4.8/+/a5ef72f/sysroot/usr/include/x86_64-linux-gnu/bits/types.h has the decoder ring for __ULONGWORD_TYPE / __SLONGWORD_TYPE / __SLONGWORD_TYPE / etc., which show up below, and suggests that __off_t / __off64_t etc. match __OFF_T_TYPE  / __OFF64_T_TYPE -- I assumed that __STD_TYPE is some form of typedef.

https://android.googlesource.com/platform/prebuilts/gcc/linux-x86/host/x86_64-linux-glibc2.15-4.8/+/a5ef72f/sysroot/usr/include/x86_64-linux-gnu/bits/typesizes.h defines __OFF_T_TYPE, __OFF64_T_TYPE, etc. in terms of __ULONGWORD_TYPE etc. introduced  in types.h.

At a higher level, while looking at the macOS crash on CI, I stumbled upon something like https://stackoverflow.com/a/48169793/537046 which told me POSIX mandates that `ino_t` is unsigned. This was my hint that I should investigate :)

pwnall: Also, those links are what Google search gave me when I asked for ino_t and off_t on Linux.

mgeisler: Awesome, thanks!

PR #859, opened 2023-06-23
Tweak speaker note about move closures and FnOnce

### LINKED ISSUE #651: Factual error: course says 'move' closures can only be FnOnce, not true
Hey, amazingly written course!

In section 25.8 (Important Traits / Closures ...), it is stated that:

> move closures only implement FnOnce.

That is not true. What move closures do is that they move their captured values into the closure itself instead of just borrowing them (immutably or mutably). It gives the closure an exclusive access to the captures, but the closed over values only get deallocated when the closure itself gets deallocated. Thus it's perfectly fine to have a move closure be FnMut or even Fn.

Here's an example of a move closure being FnMut ([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=79f2d83d4020ec222a121069dbd49b9b))

```rust
fn call_3_times(mut f: impl FnMut()) {
    f();
    f();
    f();
}

fn main() {
    let mut v = vec![];
    let f = move || {
        v.push(1);
        println!("{:?}", v);
    };
    call_3_times(f);
}
```

mauvealerts:
Fixes #651. I was going to just delete the note, but the disconnect is interesting. The current draft highlights it.

Example of a `move` closure implementing `Fn`
```rust
let x = String::new();
let x_len = move || x.len();
x_len();
x_len(); // OK
```

Example of a non-`move` closure which implements only `FnOnce`
```rust
let x = String::new();
let just_drop = || drop(x); // implicitly captures by value
just_drop();
just_drop(); // error
x.len(); // error
```

### DIFF
--- a/src/traits/closures.md
+++ b/src/traits/closures.md
@@ -13,26 +13,49 @@ fn apply_with_log(func: impl FnOnce(i32) -> i32, input: i32) -> i32 {
 
 fn main() {
     let add_3 = |x| x + 3;
-    let mul_5 = |x| x * 5;
-
     println!("add_3: {}", apply_with_log(add_3, 10));
-    println!("mul_5: {}", apply_with_log(mul_5, 20));
+    println!("add_3: {}", apply_with_log(add_3, 20));
+
+    let mut v = Vec::new();
+    let mut accumulate = |x: i32| {
+        v.push(x);
+        v.iter().sum::<i32>()
+    };
+    println!("accumulate: {}", apply_with_log(&mut accumulate, 4));
+    println!("accumulate: {}", apply_with_log(&mut accumulate, 5));
+
+    let multiply_sum = |x| x * v.into_iter().sum::<i32>();
+    println!("multiply_sum: {}", apply_with_log(multiply_sum, 3));
 }
 ```
 
 <details>
 
-If you have an `FnOnce`, you may only call it once. It might consume captured values.
+An `Fn` neither consumes nor mutates captured values, or perhaps captures nothing at all, so it can
+be called multiple times concurrently.
 
 An `FnMut` might mutate captured values, so you can call it multiple times but not concurrently.
 
-An `Fn` neither consumes nor mutates captured values, or perhaps captures nothing at all, so it can
-be called multiple times concurrently.
+If you have an `FnOnce`, you may only call it once. It might consume captured values.
 
 `FnMut` is a subtype of `FnOnce`. `Fn` is a subtype of `FnMut` and `FnOnce`. I.e. you can use an
 `FnMut` wherever an `FnOnce` is called for, and you can use an `Fn` wherever an `FnMut` or `FnOnce`
 is called for.
 
-Both `move` and non-`move` closures might only implement `FnOnce`.
+The compiler infers some other traits for closures. For example, `add_3` is `Copy` and
+`multiply_sum` is `Clone`.
+
+By default, closures will capture by reference if they can. The `move` keyword makes them capture
+by value.
+```rust,editable
+fn make_greeter(prefix: String) -> impl Fn(&str) {
+    return move |name| println!("{} {}", prefix, name)
+}
+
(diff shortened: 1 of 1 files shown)


mgeisler (commented):
Thanks for bringing this up! This slide is one I don't like very much when teaching the class — the topic is complicated and needs much more time than we devote to it. So I would love to see it improved somehow (not sure how).

> `move` closures only implement `FnOnce`.

Now that I see this again, I also think it should go. I _think_ I was thinking of a situation where you `move` a variable into the closure _and_ you call a method/function which takes ownership. Like your `drop` example. However, my misunderstand was that `move` is not necessary then: Rust will automatically capture the variable with a `move` in that case.

Instead of the weaker (but correct!) statement that both `move` and non-`move` closures can be `FnOnce`, I think we should show a concrete example where `move` is needed.

What normally bites me is that I want to return a closure which only needs to borrow the captured variable — but because I return the closure, I need to `move` the variable explicitly. I would appreciate getting that example into the notes to make it clearer for everybody how to trigger the different cases.

djmitche:
I agree that we give this topic too little attention. It's a good place to synthesize students' understanding of move vs. copy, lifetimes, borrows, and auto-Deref.

Would it make sense to write an exercise for this?  Something with, at its core, "write Fn, FnOnce, and FnMut closures with and without `move`".  Or maybe the exercise gives six examples of callbacks and asks students to describe which Fn* types they satisfy and why.

Maybe I've just been reading too many math textbooks lately :)

mauvealerts:
Inspired by the feedback, I reworked things quite a bit:
* Expanded example to include closures of each type
* Reordered speaker notes to like up with examples
* Added notes about
  *  Inferred traits
  * `move`

I recall presenting this slide and, despite the complexity of the topic, not having much to "play with." This might be too much, though.

mauvealerts:
Oh, and I'm fine with adding an exercise around closures. I just haven't thought much about it yet.

mauvealerts:
If we want to stick to one slide and trim speaker notes a bit, I think I'd cut either/both:
* Note about `Copy` and `Clone`. It's a small detail in general. I mostly included to explain why `add_3` can be passed twice
* Clarification about subtyping ("I.e. you can use an..."). It's a lot of text and presumably folks are familiar with subtyping

If we want to split this into multiple slides, I suggest adding a short "Closures" section with a split like:
* `fn()` pointers vs `Fn` - emphasis on compatability
* Captures and `move`
* `Fn`, `FnMut`, `FnOnce`

mgeisler (approved):
Awesome, this is a much more well-rounded explanation of this :smile:

mgeisler:
> If we want to stick to one slide and trim speaker notes a bit

As a general rule, I consider the speaker notes to have "unlimited" space since they're not shown during the class. Even if they become long, people who read them outside of a class can pop them open in a separate window and thus have the notes next to the slide itself.

Also, as you can tell from this, the notes are called _Speaker Notes_, but I hope they are useful for people who read the course to learn Rust too. That dual-purpose is probably making us write strange notes at times and it's something I'm not 100% sure how to solve. Now that I write that, perhaps I should start by changing the header to _Notes_ instead... :thinking:

mgeisler:
> If we want to split this into multiple slides, I suggest adding a short "Closures" section with a split like:
> 
> * `fn()` pointers vs `Fn` - emphasis on compatability
> * Captures and `move`
> * `Fn`, `FnMut`, `FnOnce`

That sounds like a very good idea, actually! I am always in trouble when I get to the current slide: I am running out of time and don't want to get into this big topic. But... of course I do and people have lots of questions. So I end up trying to explain the topic using on-the-fly code and by showing people the types from the stdlib documentation.

So that's probably the worst of both worlds :sweat_smile: Having dedicated slides would be great since I can then skip them if we're out of time. @randomPoison and @fw-immunant, you've both been teaching the class repeatedly. Feel free to jump in with your feedback here!

PR #865, opened 2023-06-23
zh-TW: translate enums

mgeisler:


po/zh-TW.po
```diff
@@ -5197,6 +5296,137 @@ msgid ""
...
+"     \n"
+"     macro_rules! dbg_bits {\n"
+"         ($e:expr, $bit_type:ty) => {\n"
+"             println!(\"- {}: {:#x}\", stringify!($e), transmute::<_, "
+"$bit_type>($e));\n"
+"         };\n"
+"     }\n"
+"\n"
+"     // 在 2^n Some() 中包裝值的巨集，n 為「@」符號的數字。\n"
+"     // 評估此巨集時，必須提高遞迴限制。\n"
```
johnathan79717, line 5392: Do we translate comments in code?

po/zh-TW.po
```diff
@@ -5197,6 +5296,137 @@ msgid ""
...
+"             Some(Some($value))\n"
+"         };\n"
+"         ($value:expr, @ $($more:tt)+) => {\n"
+"             many_options!(many_options!($value, $($more)+), $($more)+)\n"
+"         };\n"
+"     }\n"
+"\n"
+"     fn main() {\n"
+"         // TOTALLY UNSAFE. Rust provides no guarantees about the bitwise\n"
+"         // representation of types.\n"
```
johnathan79717, line 5405: What about other comments in code such as this one?

johnathan79717 (changes requested):

po/zh-TW.po
mgeisler: Ideally, we should try to minimize the amount of comments in the code. The whole example here is a little too complicated for my taste, so I would like to refactor or remote it completely.

PR #871, opened 2023-06-23
zh-TW: translate cargo

mgeisler:
Part of #684.


po/zh-TW.po
```diff
@@ -1611,56 +1611,70 @@ msgid ""
...
 msgid "#### Debian"
-msgstr ""
+msgstr "#### Debian"
 
 #: src/cargo.md:20
 msgid ""
 "On Debian/Ubuntu, you can install Cargo, the Rust source and the [Rust "
 "formatter][6] with"
 msgstr ""
+"您可以在 Debian/Ubuntu 上使用下列程式碼，安裝 Cargo、Rust 來源和 [Rust 格式設"
```
johnathan79717, line 1657:
```suggestion
"您可以在 Debian/Ubuntu 上使用下列程式碼，安裝 Cargo、Rust 原始碼和 [Rust 格式設"
```

po/zh-TW.po
```diff
@@ -1699,6 +1716,15 @@ msgid ""
 "`rustup`\n"
 "  will let you switch between them as needed."
 msgstr ""
+"* `rustc`：Rust 編譯器，可將 `.rs` 檔案轉換成二進位檔和其他中繼格式。\n"
+"\n"
+"* `cargo`：Rust 依附元件管理工具和建構工具。Cargo 知道如何下載 <https://"
+"crits.io> 上代管的依附元件，且會在建構您的專案時，將這些依附元件傳送至 "
```
johnathan79717, line 1722:
```suggestion
"crates.io> 上代管的依附元件，且會在建構您的專案時，將這些依附元件傳送至 "
```

po/zh-TW.po
```diff
@@ -1832,12 +1905,20 @@ msgid ""
...
+"% cargo --version\n"
+"cargo 1.69.0 (6e9a83356 2023-04-12)\n"
+"```"
 
 #: src/cargo/running-locally.md:15
 msgid ""
 "With this in place, follow these steps to build a Rust binary from one\n"
 "of the examples in this training:"
 msgstr ""
+"完成上述步驟後，請按照下列步驟操作，在本訓練課程的任一範例中建立 Rust 二進位"
```
johnathan79717, line 1920: We may want to unify how "build" is translated.  In earlier of this file, it was translated to "建構".

johnathan79717 (changes requested):

po/zh-TW.po
johnathan79717: Actually only replace only when it's translated from "build", not from "create".

johnathan79717 (approved):

po/zh-TW.po
johnathan79717: I pushed a commit to fix this.

mgeisler: Thanks! I cannot read Chinese, so good that you caught this :sweat_smile:

PR #876, opened 2023-06-24
Correct print statement in library exercise

0xChin:
I was going through the first exercise of day 1 afternoon, and I got confused with the prints of the emptiness status:

```sh
The library is empty: true
The library is no longer empty: false
```

I don't know if it was on purpose to show *the output of calling the `is_empty` method*, but in case it wasn't this change should solve it c:

### DIFF
--- a/src/exercises/day-1/book-library.rs
+++ b/src/exercises/day-1/book-library.rs
@@ -119,7 +119,7 @@ fn main() {
     //library.add_book(Book::new("Lord of the Rings", 1954));
     //library.add_book(Book::new("Alice's Adventures in Wonderland", 1865));
     //
-    //println!("The library is no longer empty: {}", library.is_empty());
+    //println!("The library is no longer empty: {}", !library.is_empty());
     //
     //
     //library.print_books();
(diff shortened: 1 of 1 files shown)


mgeisler:
> I don't know if it was on purpose to show _the output of calling the `is_empty` method_, but in case it wasn't this change should solve it c:

No, I don't think there was any deep thought behind this so I'm happy to make it more smooth :smile:

Perhaps we could make it more clear by writing out both the code and the English text. Something like

> The library is empty: library.is_empty(): true
>
> The library is no longer empty: library.is_empty(): false

That would be more explicit and I think explicit is good for people who are new to anything.

0xChin:
sounds good, change commited :)

Tnation888:
Felt even better i noticed you are more experienced chiiin.

Sent from Yahoo Mail on Android 
 
  On Sun, Jun 25, 2023 at 1:05 PM, ***@***.***> wrote:   


sounds good, change commited :)

djmitche (commented):
Maybe just shorten these to `"library.is_empty() = {}"'?

0xChin:
> Maybe just shorten these to `"library.is_empty() = {}"'?

I think removing the english text might make it inconsistent with the other prints, maybe indicating the result with an arrow, like this:

`"The library is empty: library.is_empty() -> true"`

mgeisler (approved):
Thanks!

PR #879, opened 2023-06-27
pt-BR: Completes and updates translation of ch.2 and ch.3

henrif75:
Part of #317


rastringer:
Hi @henrif75, thank you for all your work on this and the other PR! Did you use poedit please? Most of the 3,000+ changes appear to be line truncations so it makes it difficult to isolate the updates to the translation.
Eg in the image 
Screenshot 2023-06-28 at 8 37 07 AM

henrif75:
Hi @rastringer, yes I used poedit, I'm not sure what might have happened. I'll figure it out and will circle back, thanks.

mgeisler:
@rastringer, can you share how you've configured Poedit (if you changed the defaults)? You're probably the one with the most experience in Poedit, so it would be great if you could share that in #800.

rastringer:
> 

Hi @mgeisler, good idea. I actually never changed the defaults (on Mac). Looking at the settings I wonder if this part may be key:

Screenshot 2023-06-29 at 11 12 53 AM

henrif75:
I did some experiments and I believe the "Wrap at" option is the offender. I disabled that and compared the files and it seems to get rid of this annoying problem. I'll delete my PRs and will create another one today.

PR #882, opened 2023-06-29
Suggest to run day 3 afternoon exercise locally

schultetwin1:
Based on feedback from running the class, students found developing the day 3 afternoon exercise locally was a better experience than using Playground.

### DIFF
--- a/src/exercises/day-3/afternoon.md
+++ b/src/exercises/day-3/afternoon.md
@@ -2,9 +2,8 @@
 
 Let us build a safe wrapper for reading directory content!
 
-For this exercise, we suggest attempting to use a local dev environment instead
-of using Playground if possible. This will allow you to attempt to run your
-binary on your own machine.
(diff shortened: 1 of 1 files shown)


src/exercises/day-3/afternoon.md
```diff
@@ -2,6 +2,14 @@
 
 Let us build a safe wrapper for reading directory content!
 
+For this exercise, we suggest attempting to use a local dev environment instead
+of using Playground if possible. This will allow you to attempt to run your
+binary on your own machine.
```
mgeisler, line 7: I think we can make the text more direct here:

```suggestion
For this exercise, we suggest using a local dev environment instead
of the Playground. This will allow you to run your binary on your own machine.
```

mgeisler (approved):
This is a great point and something we should encourage more.

I normally encourage people to install Rust already on the first day, and I repeat this every day... by Day 3, most people have it installed :smile:

Issue #625, opened 2023-05-09
Improvements to exerciser


mgeisler:
The recent #623 reminded me of something I've noticed regarding the exerciser tool, namely that I think we can improve it. Today, I think the workflow has some drawbacks:

1. It requires people to install another tool, a tool which they mostly need when publishing pages
2. It requires manual steps to actually become useful from the served pages
3. It makes the output folder more cluttered than before with the `html/` and `exerciser/` subdirectories
4. The new style is inconsistent with the format we already use for Android: notice how the filenames are [italic in the Android chapter](https://google.github.io/comprehensive-rust/android/build-rules/binary.html) and [monospace in the bare-metal chapter](https://google.github.io/comprehensive-rust/exercises/bare-metal/rtc.html).

I would actually have liked us to a tool which we can run only as part of the publish workflow. I asked for that [way back when the exerciser was introduced out of the blue](https://github.com/google/comprehensive-rust/pull/538#discussion_r1158619862).

Another bad side-effect of the tool is that it requires us to mention lots and lots of files in the course which the students shouldn't actually have to look at. See all the _you shouldn’t need to change this_ comments on https://google.github.io/comprehensive-rust/exercises/bare-metal/rtc.html.

@qwandor, @djmitche, @rbehjati, and others: what are your thoughts?

rbehjati:
I agree with you @mgeisler. The process is complicated and not very well described. I had to look at GitHub actions to figure out how to generate the exercises. 

The tooling aside, is it even a good idea to have such large exercises with so much template code? I think the exercises should be designed so that the students can understand what is expected of them in no more than 4-5 mins. I have not yet done the elevator exercise from async nor the bare-metal exercise, but based on just a quick look, it feels like there is too much content to go through before starting with the exercise. I think that is also the feedback from @djmitche from teaching the async part.

Can the exercises be simplified, and some of the template code be moved to a crate that the students could just add a dependency to? I am assuming that the students can use most of the template code without having to actually read through the code. #618 might be related, but there is a discussion of the solution vs templates that I did not quite understand :)

qwandor:
The only people who have to install the tool is people who want to modify exercises and preview them locally. They already have to install the other mdbook tools, so is one more really an issue? We could add a script to install all publishing tools if that would be helpful.

[...]

PR #884, opened 2023-06-29
Add missing binaries in TRANSLATIONS.md instructions.

yohcop:
With these instructions, I can start from a clean fork, and get running quickly.

### DIFF
--- a/TRANSLATIONS.md
+++ b/TRANSLATIONS.md
@@ -27,17 +27,24 @@ GNU Gettext utilities below.
 
 ## Preparation
 
-We use two helpers for the translations:
+You will need the [Gettext] utilities (`msginit`, `msgmerge`). Under Ubuntu, you
+can install with:
+
+```shell
+$ sudo apt install gettext
+```
+
+We use `mdbook` and two helpers for the translations:
 
 - `mdbook-xgettext`: This program extracts the English text. It is an mdbook
   renderer.
 - `mdbook-gettext`: This program translates the book into a target language. It
   is an mdbook preprocessor.
 
-Install both helpers with:
+Install everything with:
 
 ```shell
-$ cargo install mdbook-i18n-helpers
+$ cargo install mdbook mdbook-i18n-helpers
 ```
 
 ## Creating and Updating Translations
--- a/TRANSLATIONS.md
+++ b/TRANSLATIONS.md
@@ -27,8 +27,8 @@ GNU Gettext utilities below.
 
 ## Preparation
 
-You will need the [Gettext] utilities (`msginit`, `msgmerge`). Under Debian and Ubuntu, you
-can install with:
+You will need the [Gettext] utilities (`msginit`, `msgmerge`). Under Debian and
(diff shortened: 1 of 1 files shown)


henrif75:
I'm just realizing that the whole "Preparation" section is redundant since it's already on the README file. Perhaps you could add the Ubuntu info to the README and replace the "Preparation", for a note asking people to ensure they've followed instructions there?

TRANSLATIONS.md
```diff
@@ -27,17 +27,24 @@ GNU Gettext utilities below.
 
 ## Preparation
 
-We use two helpers for the translations:
+You will need the [Gettext] utilities (`msginit`, `msgmerge`). Under Ubuntu, you
```
mgeisler, line 30: We use a Debian-based distribution at work, so I would like us to mention Debian here:

```suggestion
You will need the [Gettext] utilities (`msginit`, `msgmerge`). Under Debian and Ubuntu, you
```

mgeisler:
> I'm just realizing that the whole "Preparation" section is redundant since it's already on the README file. Perhaps you could add the Ubuntu info to the README and replace the "Preparation", for a note asking people to ensure they've followed instructions there?

I tried to cover two use-cases here: README is for people who would like to read the course offline. We've managed to make the translation infrastructure mostly orthogonal to that usecase, so the README doens't need to talk about the Gettext tools. It _almost_ doesn't have to talk about mdbook-i18n-helpers: things work if you don't install them, but you get a warning from `mdbook build`. So that's why mdbook-i18n-helpers is added to the README (I would like to remove the mdbook-exerciser, see #625).

I see TRANSLATIONS as the main document for people who want to contribute to the translations. So it needs to mention Gettext and the rest of the infrastructure. As for installing mdbook, we could refer back to the README here. That would be better since we then keep the instructions in one place.

How does that sound?

yohcop:
Yes, I think I understand it the same way as @mgeisler . TRANSLATIONS.md is instructions for translators, README for users/contributors to the course.

TRANSLATIONS.md
```diff
@@ -27,17 +27,24 @@ GNU Gettext utilities below.
...
   renderer.
 - `mdbook-gettext`: This program translates the book into a target language. It
   is an mdbook preprocessor.
 
-Install both helpers with:
+Install everything with:
 
 ```shell
-$ cargo install mdbook-i18n-helpers
+$ cargo install mdbook mdbook-i18n-helpers
```
mgeisler, line 47: I think we should leave this as it is and instead refer back to the README for how to build the book (as @henrif75 suggested). So the instructions here could start with something like:

> First, ensure you can build the book and that `mdbook serve` works (follow the instructions in the README).

Then everything will be nicely split.

yohcop: Ok, let me know if that works!

henrif75 (approved):

mgeisler (approved):

mgeisler:
Thanks!

Issue #894, opened 2023-07-02
Runtime Guarantees might me incorrect.


sak96:
Integer overflow is defined part in runtime guarantee might be incorrect.
For rust program.
```rust
fn main() {
    let x = i32::MAX /2 ;
   println!("x={x};\n2x+3={}",x * 2 + 3 );
}
```

When compiled with optimization there is no error.
```console
$ rustc -O a.rs
$ ./a
x=1073741823;
2x+3=-2147483647
```

It does check overflow in non-optimized mode.
```console
$ rustc a.rs
$ ./a
thread 'main' panicked at 'attempt to add with overflow', a.rs:3:31
stack backtrace:
   0: rust_begin_unwind
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/std/src/panicking.rs:579:5
   1: core::panicking::panic_fmt
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/panicking.rs:64:14
   2: core::panicking::panic
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/panicking.rs:114:5
   3: a::main
   4: core::ops::function::FnOnce::call_once
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

mgeisler:
Your analysis is spot on: the behavior depends on the optimization mode (or more precisely, it depends on the [`overflow-checks` option](https://doc.rust-lang.org/rustc/codegen-options/index.html#overflow-checks), which is `on` by default in debug builds and `off` in optimized builds).

This is why https://google.github.io/comprehensive-rust/why-rust/runtime.html says that overflow is "defined". The speaker notes explain that the behavior is a panic or wrap-around.

We should update the main text to spell this out — perhaps a small edit is enough:

> * Integer overflow is defined (panic or wrap-around).

We could then link to the relevant option in the speaker notes.

Would you be up for sending a PR for this?

sak96:
Sorry i think it is now fixed.
There are speaker notes explaining the issue.

Should i still need to provide the pr.
or we can close the issue.

mgeisler:
> Should i still need to provide the pr. or we can close the issue.

I think it would be good to update the text like I suggest above. I can review it if you open a PR.

sak96:
[...]

PR #896, opened 2023-07-03
Update description for runtime overflow check

sak96:
Fixes issue #894


mgeisler:
> Looks like there is extra procedure for CLA.

Yes, that is correct. Are you not able to sign the agreement for some reason?

mgeisler:
> i will if there is larger contribution i am making. for now it seems like bit too much for one liner contribution. i know i might seem annoying, but sorry.

Yeah, that's fair! I'll do the changes myself based on our chat here.

PR #898, opened 2023-07-03
Update points-polygons.rs Polygon::length solution to use iterators

edemir:
Update points-polygons.rs Polygon::length solution to use iterators for pairwise distance calculation

### DIFF
--- a/src/exercises/day-2/points-polygons.rs
+++ b/src/exercises/day-2/points-polygons.rs
@@ -84,18 +84,9 @@ impl Polygon {
     }
 
     pub fn length(&self) -> f64 {
-        if self.points.is_empty() {
-            return 0.0;
-        }
-
-        let mut result = 0.0;
-        let mut last_point = self.points[0];
-        for point in &self.points[1..] {
-            result += last_point.dist(*point);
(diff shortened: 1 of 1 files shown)


mgeisler (changes requested):
Hi @edemir, thanks for the elegant solution!

Since this exercise is for beginners, I would suggest we keep the simpler solution as the default. Your solution could go into a comment instead as an alternative.

mgeisler:
Hi @edemir, thanks for updating the code to use a comment — is there a particular reason you closed the PR?

edemir:
Hi @mgeisler I tried to squash two commits and bungled it up. Created a fresh pull requests here: https://github.com/google/comprehensive-rust/pull/950

mgeisler:
> Hi @mgeisler I tried to squash two commits and bungled it up. Created a fresh pull requests here: #950

Okay, thanks for putting up the new PR! Just so you know, we squash-merge all PRs in this repository so you don't have to be very careful with the commit history in a PR. We do this to lower the barrier to entry: we only need to ensure the PRs have a nice title since that's what will be preserved in the history.

PR #899, opened 2023-07-03
ru: Update po/ru.po file with new messages from the updated version of the original

baltuky:
Update `po/ru.po` file with new messages from the updated version of the original


zvonden:
I guess it is not only new messages, but the updated version of the original. Can you please amend the description?

baltuky:
> I guess it is not only new messages, but the updated version of the original. Can you please amend the description?

Done!

zvonden (approved):

PR #900, opened 2023-07-04
Clarify that each philosopher should think/eat 100 times

pwnall:
Folks who have hazy memories of the philosopher dining problem may interpret the original instructions as "make each philosopher think and eat once". This interpretation loses a critical detail, because the resulting code is highly unlikely to deadlock in practice, even without breaking the symmetry.


mgeisler (approved):
Thanks, that's a great point!

pwnall:
Thank you very much for the quick review!

PR #902, opened 2023-07-04
Update C manual memory management example

devdri:
In C (as opposed to C++) the explicit cast from void* to int* is not required. It is also not idiomatic to do so in C code.

Actual C codebase would use `malloc()` without the cast, and a C++ one (when not using abstractions) a `new int[n]` - both a bit cleaner and less verbose than this example.


mgeisler (approved):
Awesome, thanks for helping out! As you can tell, I'm not a hardcore C developer :smile:

djmitche (approved):

Issue #175, opened 2023-01-17
Look at mdbook-keeper


mgeisler:
We currently have a number of code blocks marked with `compile_fail`: this is often because they depend on third-party crates which aren't available when running `mdbook test`.

We should look into using https://github.com/tfpk/mdbook-keeper/ to handle testing instead. It explicitly supports using third-party crates in code snippets.

mgeisler:
This is of course a bit of an open ended issue :slightly_smiling_face:  What I hope for is that someone will look at what `mdbook keeper` can do for us:

- can it run our current tests?
- can we use it to run tests with more dependencies?

The goal would be remove the `compile_fail` from our code blocks.

If you can get this working locally, then we can work on a updating the GitHub action which runs the tests for us on every PR.

mgeisler:
Thanks @mauvealerts! That's already a very valuable analysis! For example, I had not realized that you run the tests all the time... so we would definitely only enable it on demand like we do for the [`mdbook-xgettext` renderer](https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md#generating-the-po-template).

mauvealerts:
I have something that I'm fairly happy with, when combined with all of these changes to `mdbook-keeper`
* tfpk/mdbook-keeper#2
* tfpk/mdbook-keeper#3
* tfpk/mdbook-keeper#4

I have things running comparable to `mdbook test`. IMO the steps necessary to make a mergable PR are:
- [ ] Decide where the skeleton project should go (currently `mdbook-keeper-deps`)
- [ ] Don't run tests by default for `mdbook serve`
- [ ] `mdbook-keeper` has released a version with the changes
- [ ] Prove out the usefulness by testing some snippets

mgeisler:
Thanks for pushing this!

> * Decide where the skeleton project should go (currently `mdbook-keeper-deps`)

I think this should be the root of the repository, or perhaps the `src/` folder. We already have a Cargo workspace setup so that `cargo check` does something useful for the code in our exercises. I hope we can build on this for the book itself: perhaps we can move up the `src/exercises/Cargo.toml` file to `src/` and then use its dependencies across the examples?

mauvealerts:
[...]

PR #904, opened 2023-07-05
Fix solution in Link Checker in Concurrency Morning exercises

### LINKED ISSUE #816: Concurrency morning solution does not contain a solution for Link Checker
Hi! 
Really enjoyed the course so far, I noticed something missing though so I'm opening an issue:

The [concurrency morning solutions file](https://github.com/google/comprehensive-rust/blob/main/src/exercises/concurrency/solutions-morning.md#L1) doesn't contain the solution for the [link checker task](https://github.com/google/comprehensive-rust/blob/main/src/exercises/concurrency/link-checker.md). I thought to check in [link-checker.rs](https://github.com/google/comprehensive-rust/blob/main/src/exercises/concurrency/link-checker.rs), but although it contains extra code it doesn't look like it solves the multi-threaded part of the problem

pwnall:
This change fixes the following issues with the current solution:

1. It is not listed on the "solutions" page.
2. It is not multi-threaded and does not use channels.

Fixes #816.

### DIFF
--- a/src/exercises/concurrency/link-checker.md
+++ b/src/exercises/concurrency/link-checker.md
@@ -57,12 +57,13 @@ Your `src/main.rs` file should look something like this:
 ```rust,compile_fail
 {{#include link-checker.rs:setup}}
 
-{{#include link-checker.rs:extract_links}}
+{{#include link-checker.rs:visit_page}}
 
 fn main() {
+    let client = Client::new();
     let start_url = Url::parse("https://www.google.org").unwrap();
-    let response = get(start_url).unwrap();
-    match extract_links(response) {
+    let crawl_command = CrawlCommand{ url: start_url, extract_links: true };
+    match visit_page(&client, &crawl_command) {
         Ok(links) => println!("Links: {links:#?}"),
         Err(err) => println!("Could not extract links: {err:#}"),
     }
--- a/src/exercises/concurrency/link-checker.rs
+++ b/src/exercises/concurrency/link-checker.rs
@@ -12,9 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::{sync::Arc, sync::Mutex, sync::mpsc, thread};
+
 // ANCHOR: setup
-use reqwest::blocking::{get, Response};
-use reqwest::Url;
+use reqwest::{blocking::Client, Url};
 use scraper::{Html, Selector};
 use thiserror::Error;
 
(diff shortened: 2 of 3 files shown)


pwnall:
@mgeisler This is a lot of code, and I'm pretty new at Rust. I won't be upset if you merge and change / rewrite, if that's easier than writing up a lot of feedback 😄

djmitche (approved):
This looks good! I added a few rustdoc comments. There are some interesting things to discuss here, such as using a Mutex to make a "mpmc" channel.

pwnall:
@djmitche Thank you very much for the comment improvements!

In case it helps, I came up with the Arc+Mutex while trying to follow the exercise instructions at https://github.com/google/comprehensive-rust/blob/4413b6a28f6d0be5bf2d76564432015b072e2f5b/src/exercises/concurrency/link-checker.md?plain=1#L80-L81. In particular, I understood "send the URLs [...] to a channel" as asking me to figure out how to use a single mpsc::channel to coordinate between the "main" thread and the threads checking URLs.

I also considered the `crossbeam` crate, which offers "mpmc" channels. I thought that solution would be less compatible with the spirit of the exercise, because the instructions already include a list of crates for us to use.  By contrast, the proposal above ties in a few concepts introduced in the morning session.

I wrote my thoughts above hoping they help us come up with the best possible prompt text for future students.

mgeisler:
> @mgeisler This is a lot of code, and I'm pretty new at Rust. I won't be upset if you merge and change / rewrite, if that's easier than writing up a lot of feedback smile

Thanks a lot for putting this up — people have been writing me asking for a solution :grimacing: Infact, we have #816 open for this :smile: 

Fun fact, to test this, I downloaded the code and tried to crawl https://google.github.io/comprehensive-rust/ instead — I was curious if we have broken links there! That soon resulted in

```
Got crawling error: bad http response: 429 Too Many Requests
```

errors and a bit later, I was blocked from posting this comment... luckily, the rate limit was lifted a few minutes later.

I noticed one thing:

```
Checking https://google.github.io/comprehensive-rust/print.html#assumptions
Checking https://google.github.io/comprehensive-rust/print.html#running-the-course
Checking https://google.github.io/comprehensive-rust/print.html#course-structure
Checking https://google.github.io/comprehensive-rust/print.html#deep-dives
```

It would be great to clean the URLs of any query string and anchor texts before inserting them into `visited_pages`.

mgeisler:
Another thing I've noticed when doing this exercise in class: it's _super hard_ and I wish I had provided much more scaffolding for people to use. Your PR is already a great improvement, but I feel we should

1. Write comments above the functions we give to the students, and
2. Consider giving the major building blocks as well — with comments explaining what we expect people to fill in.

I'm an old-school web developer, but I've had people in my classes who've never done any web development whatsoever. HTTP status codes are foreign to them, HTML is foreign to them, URL fragments are foreign to them. Adding more scaffolding would probably help a lot there to make the exercise useful people from more backgrounds.

After this PR is merged, I would also like to include a diagram or two which shows the intended interaction between the crawler threads, the channels, and the main thread.

I've had people in my classes who took the old link extraction logic and ran that in a multi-threaded way — when that is a purely local and super fast operation. So we would need more detail explaining that it is the IO operations that are slow and which we seek to parallelize here.

0951459527 (approved):

0951459527 (commented):
R Tao Channel

pwnall:
@mgeisler I fully agree with your thoughts above! Would you be OK if I copied your thoughts (and my reply) into a separate Issue?

I see this PR as making a first step in what is hopefully the right direction. This PR brings the solution in line with the requirements outlined in the problem, with minimal scaffolding changes. I think that, once we agree that there are no obvious simplifications to the solution, we can conclude that the problem is too difficult. Once we're there, we can think of possible fixes.

I would be most interested in prototyping a scaffolding change where we give students a `spawn_crawler_threads` that works with a single crawler thread, and ask them to figure out how to support multiple crawler threads. I worry that this may make things too simple, but I figure I can iterate once I see what that looks like.

WDYT?

jmpfar:
My extremely unedited solution with a single channel and no mutex
It's extremely full of .unwrap() and far from my proudest moment, but might be useful if you want to try adding that kind of solution
https://gist.github.com/jmpfar/4f6f0c4258037c0e4dda981a6d5dc6dc

mgeisler:
> @mgeisler I fully agree with your thoughts above! Would you be OK if I copied your thoughts (and my reply) into a separate Issue?

Yes, definitely!

> I see this PR as making a first step in what is hopefully the right direction. This PR brings the solution in line with the requirements outlined in the problem, with minimal scaffolding changes. I think that, once we agree that there are no obvious simplifications to the solution, we can conclude that the problem is too difficult. Once we're there, we can think of possible fixes.

I completely agree, we should merge this now since it's a huge improvement on what we have.

> I would be most interested in prototyping a scaffolding change where we give students a `spawn_crawler_threads` that works with a single crawler thread, and ask them to figure out how to support multiple crawler threads. I worry that this may make things too simple, but I figure I can iterate once I see what that looks like.

My experience is that people find this to be a hard exercise — most find it too hard for the 30-40 minutes we normally have at this point in the course. I think less than 25% of the people in a class manage to write a working solution in that time.

mgeisler:
> My extremely unedited solution with a single channel and no mutex It's extremely full of .unwrap() and far from my proudest moment, but might be useful if you want to try adding that kind of solution https://gist.github.com/jmpfar/4f6f0c4258037c0e4dda981a6d5dc6dc

Hi @jmpfar, thanks for posting this! Your solution is not bad: it delegates the I/O-heavy operations to new threads and does the CPU-intensive work in the main thread. Many people don't get to that stage.

A small problem with your solution is that you have an unbounded number of threads: if a page has 100 outgoing links, you spawn 100 threads. To solve that, you more or less have to create a multi-producer-multi-consumer abstraction: you need this to have a bounded number of threads read from a single channel.

mgeisler:
> Anyway from my side I can attest this was the hardest exercise in the course (minus android and baremetal which I didn't do) and it took me significant time (but was fun all in all :)

Thanks, that matches my impression too. I'm glad you liked it despite it being hard :smile:  Perhaps it would be good to call it a "take home exercise" instead since it's really a small projects and not a real _exercise_ with a clear answer.

mgeisler (approved):
Let's get this merged since it's an excellent start for further updates. @fw-immunant has put up some PRs to restructure the course so he might have further input down the line.

mgeisler:
Hi @pwnall, the build fails because a code block is not marked with `compile_fail`:

```
---- /tmp/mdbook-uKXC6H/exercises/concurrency/solutions-morning.md - Concurrency_Morning_Exercise::Link_Checker (line 108) stdout ----
error[E0433]: failed to resolve: use of undeclared crate or module `reqwest`
  --> /tmp/mdbook-uKXC6H/exercises/concurrency/solutions-morning.md:126:5
```

When we run `mdbook test`, all code blocks are tested — but they're tested without dependencies. So we need to mark them as `compile_fail` if they depend on anything outside the standard library. @mauvealerts is looking to see if we can use mdbook-keeper (#175) to get more test coverage here.

pwnall:
Thank you for decoding the error for me!

Let me figure out how to modify this chain to make the build green. I haven't updated a branch with someone else's commits before 😄

pwnall:
@mgeisler Thank you again for the tips! The build is green now :)

mgeisler:
> @mgeisler Thank you again for the tips! The build is green now :)

Yeah, looks good! Just for your information, we squash-merge all PRs in this repository. This makes it a bit easier for people since they can just add more and more commits to fix review comments. The commits all disappear from the main history when the branch is squash-merged into `main`. (But force-pushing like you did is also completely fine: that's the workflow you would use if you do a normal merge and want to preserve a nice history.)

mgeisler:
> Thank you for decoding the error for me!

A small note: it feels terribly to have large code snippets that say `compile_fail` — how can we know that it still works? In this case we're saved by the fact that the code is included from an external `link-checker.rs` file which in turn is referenced by our [`Cargo.toml`](https://github.com/google/comprehensive-rust/blob/main/src/exercises/Cargo.toml#L39-L41) file. This means that we _do_ compile the link checker as part of our CI, but we do it outside of `mdbook test`.

PR #911, opened 2023-07-05
Rename 'designing a library' example

fw-immunant:
Library design in Rust is its own topic that this course does not cover; to avoid confusion, let's stick to describing things in terms of books.


qwandor (approved):

mgeisler:
> Library design in Rust is its own topic that this course does not cover; to avoid confusion, let's stick to describing things in terms of books.

True... though the name is a bit tongue-in-cheek and I've always had the impression that people liked the joke. But perhaps it was too subtle?

PR #912, opened 2023-07-05
Clarify a couple minor points

fw-immunant:


src/basic-syntax/static-and-const.md
```diff
@@ -39,9 +39,11 @@ fn main() {
 ```
 
 As noted in the [Rust RFC Book][1], these are not inlined upon use and have an actual associated memory location.  This is useful for unsafe and embedded code, and the variable lives through the entirety of the program execution.
+When a globally-scoped value does not have a reason to need its own address in memory, `const` is generally preferred.
 
 
 We will look at mutating static data in the [chapter on Unsafe Rust](../unsafe.md).
+Because `static` variables are accessible from any thread, mutable static variables require manual, unsafe, synchronization of accesses.
```
qwandor, line 46: This could link to https://google.github.io/comprehensive-rust/unsafe/mutable-static-variables.html which talks about mutable statics a bit more.

src/basic-syntax/static-and-const.md
```diff
@@ -39,9 +39,11 @@ fn main() {
 ```
 
 As noted in the [Rust RFC Book][1], these are not inlined upon use and have an actual associated memory location.  This is useful for unsafe and embedded code, and the variable lives through the entirety of the program execution.
+When a globally-scoped value does not have a reason to need its own address in memory, `const` is generally preferred.
```
qwandor, line 42: ... and doesn't have interior mutability. E.g. making a static `Mutex<MyType>` into a const wouldn't work.

henrif75:
Hey @fw-immunant thank you for your suggestions. Would you like to check @qwandor 's comments before we're able to merge your PR?

src/basic-syntax/static-and-const.md
fw-immunant: Would saying "object identity" instead of "its own address in memory" capture both of these requirements, and then we could specify "(e.g. requiring its own address in memory or its own dynamic state such as types like `Mutex<T>` which have interior mutability)"? Or is that too wordy? Maybe the latter bit could go in presenter notes.

src/basic-syntax/static-and-const.md
fw-immunant: Maybe it would make more sense to change the link in the previous sentence:
```
We will look at [mutating static data](../unsafe/mutable-static-variables.md) in the chapter on Unsafe Rust.
```

src/basic-syntax/static-and-const.md
qwandor: Sounds reasonable.

src/basic-syntax/static-and-const.md
qwandor: Sure, go for it.

fw-immunant:
Pushed an update and also clarified the first sentence of the consts/statics page which implied that `const`s relate to global *state*, which is specifically false.

qwandor (approved):

PR #913, opened 2023-07-06
Reorder material on first two days

### LINKED ISSUE #510: Methods used in exercise before they are introduced
Day one's afternoon exercises use methods, but these aren't described until day 2.  The afternoon exercises use traits as well (Iterator, IntoIterator), but perhaps those are more easily explained away with references to future slides.

fw-immunant:
This will require some discussion and some changes to exercises, but hopefully should be useful as a PR at least to get something concrete to discuss.

Basically, my thoughts are:
- Morning of day 1 still introduces the language and its high-level goals/value proposition, and starts with the built-in data types Rust provides, and how you define a function. I'm thinking the implicit-conversions exercise could be expanded to cover more about overflow/underflow/divide-by-zero behavior and maybe checked operations, to emphasize what Rust does differently than other languages (in particular C and C++) here. The arrays and for loops exercise can lead in to the next topic, which will (after reorg) be control flow.
- Afternoon of day 1 gets a frontloading of the basic control flow structures in Rust but not the more exotic ones. This lets us introduce enums prior to match, motivating `match`. Then we get into patterns, which we can give a mix of `match`/`if let`/`while let`/`let ... else`. The exercises for day 1 afternoon will be the Luhn algorithm (where we can match on digits and enums such as `Option` and practice working with variables) and some other pattern-matching exercise, perhaps one where we dig deep into a nested data structure with matching. I sometimes do this synchronously with students while going over the pattern matching slides at present, but it could be split out into a proper exercise.
- Morning of day 2 still has discussion of memory management, but we take out the "Points and Polygons" in favor of something simpler that still works out structs/ownership/method receivers. I'm not yet sure what that looks like, but the expanded "Health Statistics" exercise (#909) is partway there. Implementing an iterator is a great exercise but seems like too much alongside all the rest of what Points+Polygons does, so maybe we could strip it down to just defining an iterator.
- This does deflate day 2 afternoon a fair bit, so I think some more shuffling or just moving the day divisions might help to keep things balanced. As mentioned, this is a WIP, and mostly to inspire discussion at this point.

Feedback welcome!

Fixes #510.


mgeisler:
Thanks @fw-immunant!

When discussing with @gribozavr and @scentini, we talked about moving more of the difficult stuff to the morning slots: the thinking was that people would be more energized there compared to an afternoon slot. So moving almost all of the syntax slides to Day 1 afternoon seems good!

src/SUMMARY.md
```diff
@@ -40,10 +40,43 @@
...
+
 - [Variables](basic-syntax/variables.md)
   - [Type Inference](basic-syntax/type-inference.md)
   - [static & const](basic-syntax/static-and-const.md))
   - [Scopes and Shadowing](basic-syntax/scopes-shadowing.md)
+- [Enums](enums.md)
+  - [Variant Payloads](enums/variant-payloads.md)
+  - [Enum Sizes](enums/sizes.md)
+
+- [Novel Control Flow](advanced-control-flow.md)
```
djmitche, line 59: I like this distinction -- "control flow you've heard of" vs. "new forms of control flow" (that they're not really _new_, but fall under the "modern language features" part of the "Why Rust" slide).

src/SUMMARY.md
```diff
@@ -40,10 +40,43 @@
...
+  - [if let expressions](control-flow/if-let-expressions.md)
+  - [while let expressions](control-flow/while-let-expressions.md)
+  - [match expressions](control-flow/match-expressions.md)
+
+- [Pattern Matching](pattern-matching.md)
```
djmitche, line 64: I like putting this on day 1 - it's "cool" and powerful and thus motivating, but not as mind-blowing as borrows or move semantics.

djmitche (commented):
Left some minor comments here, but I suspect that PRs aren't the best medium for discussing a reorganization of the course. I'm not sure what is, though!

src/SUMMARY.md
mgeisler: Me too, I think this will be a nice feature to demonstrate on the first day.

mgeisler (commented):
I paged through the course locally and I like it! Some small notes:

- Please update the small course overviews on
    - [x] `running-the-course/course-structure.md`
    - [x] `welcome-day-1.md`, and
    - [x] `welcome-day-2.md`
- [x] Please create a `control-flow/novel.md` file for the new chapter.
- [ ] Can we add a small pattern matching exercise which asks students to evaluate an expression tree using a recursive function? Something like [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6543aea82e1304c314fe44b475e3eb67). All the boxing might be rather confusing, though?

    Alternatively, we should make the link a placeholder like in my other comment.
- [x] Please remember to update the `redirect` table in `book.toml` if you move any file (doesn't seem to be the case here)

mgeisler:
Hey @fw-immunant, I like this a lot! I would like you to already think about how we can remove the "Basic Syntax" section (or at least rework it). It doesn't make so much sense after this PR since we cover most of it again right after the break:

image

So we should have a PR to fix that right after merging this one :smile:

src/SUMMARY.md
```diff
@@ -40,10 +40,43 @@
 
 # Day 1: Afternoon
 
+- [Control Flow](control-flow.md)
```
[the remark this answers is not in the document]
fw-immunant: We could. The reason I didn't do this initially is that I thought that keeping enums right after variables made sense as it gives an interesting class of values that you can store in variables, and I wanted advanced control flow (including pattern matching) to follow right after enums as it shows how to consume them.

fw-immunant:
From CI:
```
Run mdbook build
2023-08-10 20:24:56 [INFO] (mdbook::book): Book building has started
2023-08-10 20:24:56 [INFO] (mdbook::book): Running the exerciser backend
2023-08-10 20:24:56 [INFO] (mdbook::renderer): Invoking the "exerciser" renderer
2023-08-10 20:24:56 [INFO] (mdbook::book): Running the html backend
Error: -10 20:24:58 [ERROR] (mdbook::utils): Error: Rendering failed
Error: -10 20:24:58 [ERROR] (mdbook::utils): 	Caused By: Unable to emit redirects
Error: -10 20:24:58 [ERROR] (mdbook::utils): 	Caused By: Not redirecting "/home/runner/work/comprehensive-rust/comprehensive-rust/book/html/exercises/day-1/book-library.html" to "../day-2/book-library.html" because it already exists. Are you sure it needs to be redirected?
```

But this PR does remove `day-1/book-library.md`, so presumably we need to add a redirect there. Am I missing something?

src/SUMMARY.md
```diff
@@ -68,42 +101,17 @@
   - [Storing Books](exercises/day-1/book-library.md)
```
mgeisler, line 101: re https://github.com/google/comprehensive-rust/pull/913#issuecomment-1673898090: This is the line that makes `mdbook` output a `day-1/book-library.html` file.

src/SUMMARY.md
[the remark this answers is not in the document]
mgeisler: Okay, let's go with this ordering then. You, Nicole, and the other instructors will tell us how it works in a few weeks.

mgeisler:
Hi @fw-immunant, I think this just needs a little rebase on top of current `main` and then it should be good to go :smile:

src/SUMMARY.md
fw-immunant: Thanks, fixed!

djmitche (approved):
Let's get this landed :)

src/SUMMARY.md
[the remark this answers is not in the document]
mgeisler: This has to be an empty link until we add the page — `mdbook` creates a grayed-out heading in the sidebar for empty links, but it creates an empty page for actual links:

```suggestion
- [Novel Control Flow]()
```

We'll fill out this page in a followup pr.

mgeisler (approved):
Thanks for much, I'm excited to see this in action!

mgeisler:
@fw-immunant please merge this at your convenience! :rocket:

Let me push a small commit to your branch with a page for the novel control flow.

PR #918, opened 2023-07-07
fr: Clean update of messages

yohcop:
No changes to translated messages, just updating the file.

This is the result of running `msgmerge --update po/fr.po po/messages.pot` after generating a new messages.pot.

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -53,7 +53,7 @@ msgid "Day 1: Morning"
 msgstr "Jour 1 : Matin"
 
 #: src/SUMMARY.md:18 src/SUMMARY.md:75 src/SUMMARY.md:128 src/SUMMARY.md:185
-#: src/SUMMARY.md:211 src/SUMMARY.md:259
+#: src/SUMMARY.md:211 src/SUMMARY.md:261
 msgid "Welcome"
 msgstr "Bienvenue"
 
@@ -131,7 +131,7 @@ msgstr "Surcharge"
 
 #: src/SUMMARY.md:37 src/SUMMARY.md:66 src/SUMMARY.md:90 src/SUMMARY.md:119
 #: src/SUMMARY.md:148 src/SUMMARY.md:177 src/SUMMARY.md:204 src/SUMMARY.md:225
-#: src/SUMMARY.md:251 src/SUMMARY.md:273 src/SUMMARY.md:293
+#: src/SUMMARY.md:253 src/SUMMARY.md:275 src/SUMMARY.md:296
 msgid "Exercises"
 msgstr "Exercices"
 
@@ -240,8 +240,9 @@ msgid "Lifetimes in Data Structures"
 msgstr "Durées de vie dans les structures de données"
 
 #: src/SUMMARY.md:67
-msgid "Designing a Library"
-msgstr "Conception d'une bibliothèque"
+#, fuzzy
+msgid "Storing Books"
+msgstr "String"
 
 #: src/SUMMARY.md:68
 msgid "Iterators and Ownership"
@@ -279,7 +280,7 @@ msgstr "Tailles d'énumération"
 msgid "Method Receiver"
 msgstr "Récepteur de méthode"
 
-#: src/SUMMARY.md:84 src/SUMMARY.md:159 src/SUMMARY.md:272
+#: src/SUMMARY.md:84 src/SUMMARY.md:159 src/SUMMARY.md:274
 msgid "Example"
 msgstr "Exemple"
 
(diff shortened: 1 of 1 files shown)


yohcop:
Sorry, I had another one of those, but I mistakenly deleted my branch, thinking that it had been merged. So did another clean update.

po/fr.po
```diff
@@ -240,8 +240,9 @@ msgid "Lifetimes in Data Structures"
 msgstr "Durées de vie dans les structures de données"
 
 #: src/SUMMARY.md:67
-msgid "Designing a Library"
-msgstr "Conception d'une bibliothèque"
+#, fuzzy
+msgid "Storing Books"
+msgstr "String"
```
yohcop, line 245: This is really weird. Where does the translation come from? I did not change anything after running msgmerge. (there are a couple of those), just leaving them for a second, in case the reviewer knows what's going on.

po/fr.po
```diff
@@ -745,253 +746,267 @@ msgid "Application Processors"
...
 
-#: src/SUMMARY.md:239 src/SUMMARY.md:241
+#: src/SUMMARY.md:240 src/SUMMARY.md:242
 msgid "Using It"
 msgstr "En l'utilisant"
 
 #: src/SUMMARY.md:243
+#, fuzzy
+msgid "Exceptions"
+msgstr "Fonctions"
```
yohcop, line 791: Another weird automatic (?) translation

po/fr.po
```diff
@@ -745,253 +746,267 @@ msgid "Application Processors"
...
-#: src/SUMMARY.md:292
+#: src/SUMMARY.md:294
 #, fuzzy
 msgid "Async Traits"
 msgstr "Traits asynchrones"
 
 #: src/SUMMARY.md:295
+#, fuzzy
+msgid "Cancellation"
+msgstr "Traductions"
```
yohcop, line 947: and another one...

po/fr.po
```diff
@@ -5605,8 +5637,8 @@ msgstr ""
 
 #: src/exercises/day-1/book-library.md:1
 #, fuzzy
-msgid "# Designing a Library"
-msgstr "# Concevoir une bibliothèque"
+msgid "# Storing Books"
+msgstr "# Chaîne"
```
yohcop, line 5641: another weird update

po/fr.po
yohcop: I see the "Desigining a Library" string was moved at the end of the file, in a commented string?

po/fr.po
```diff
@@ -15855,6 +16087,99 @@ msgstr ""
 "* Exécutez l'exemple dans QEMU avec `make qemu_logger` sous `src/bare-metal/"
 "aps/examples`."
 
+#: src/bare-metal/aps/exceptions.md:1
+#, fuzzy
+msgid "# Exceptions"
+msgstr "# Les fonctions"
```
yohcop, line 16093: Another Exceptions -> Fonctions added translation?

yohcop:
I'm leaving those comments / weird diffs open for a bit, as I am looking for feedback on why this is happening. (https://github.com/google/comprehensive-rust/discussions/919)

po/fr.po
mgeisler: Hi @yohcop, what you see here is `msgmerge` looking at the old and new strings and trying it's best (it's not very clever) to find an existing translation for the new changed `msgid`.

As you know, the `msgid` fields change as we evolve the English source text. So when we fix a typo (#916), the `msgid` changes when you regenerate `messages.pot`. When you run `msgmerge` the `msgid` in your `fr.po` file is compared with the ones in `messages.pot` and here `msgmerge` was able to correctly match up the updated English text with the existing French translations on [line 10,920](https://github.com/google/comprehensive-rust/pull/918/files#diff-940623802973a9e80a24d1c2e258a73d0e9c532b9feb52d2607681dca35f83c6L10918-R10994).

For a typo fix, the French translation will still work (unless someone copied the typo when translating :smile:). All that's left to do is to remove the `fuzzy` marker that `msgmerge` inserts. This marker says "The translation here is _fuzzy_ and needs a review." We don't use fuzzy translations at all, so it's important that you do a sweep of the `fr.po` file afterwards to clean them up.

In this case, you see the strange guess is marked fuzzy as well. So it's weird, but ultimately harmless.

Finally, what you see at the bottom is a kind of recycle bin for old translations — translations that were made obsolete since their `msgid` no longer matches any `msgid` in the `messages.pot` file:

```po
#~ msgid "Designing a Library"
#~ msgstr "Conception d'une bibliothèque"

#, fuzzy
#~ msgid "# Designing a Library"
#~ msgstr "# Concevoir une bibliothèque"
```

(cc @henrif75 since this is good stuff to know when dealing with translations)

sakex (approved):

Issue #893, opened 2023-07-01
pt-BR: Coordinating updates and additional translations


henrif75:
I'm creating this issue to coordinate additional work on the pt-BR translation.
Continuation of #317

henrif75:
I'll be working on ch. 6 to 10 for the next week.

mgeisler:
Cc: @rastringer, @hugojacob, and @joaovicmendes, if you haven't seen it.

joaovicmendes:
Awesome idea, @henrif75!
I'll probably work on it mostly Friday, so I think I can work on chap 11 to 17 (day 2 morning, basically).

henrif75:
Thanks @joaovicmendes! 
I'll tackle now the day 2 afternoon, ch. 18 to 21.

joaovicmendes:
Thanks @henrif75! This week I'll try to work on Chapters 22 - 26 (Day 3 Morning)

henrif75:
> Thanks @henrif75! This week I'll try to work on Chapters 22 - 26 (Day 3 Morning)

Sounds good! I'll work on Day 3 afternoon (Ch. 27 - 30)

henrif75:
@joaovicmendes @mgeisler @rastringer @hugojacob 
Hey folks, I'm planning to send a PR only to normalize the format of the pt-br.po file through msgmerge et. al. I anticipate it will be large but mostly changing the lovely line wrapping - as we've been discussing in #800

henrif75:
I'll begin reviewing/updating the Android section.

henrif75:
Great, thanks @rastringer for the PR #1120 
I'll work on removing the fuzzy entries for the next day.

henrif75:
Closing this issue since we're done with the core translation work.

PR #921, opened 2023-07-07
Re-syncing pt-BR.po with freshly extracted messages.pot

henrif75:
Re-syncing pt-BR.po with freshly extracted messages.pot through msgmerge.
This commit **doesn't change any translation**, it only updates the original English messages in the pt-BR.po file.
Those changes are almost all caused by changes in line wrapping.
Current pt-BR.po contains original messages wrapped at 80 columns.
This PR will make easier to check future diffs for the pt-BR translations.

Related issues: #317 and #893


henrif75:
Never mind, please disregard. Actually the extraction is doing the other way around, which we should discuss in a different context.

PR #950, opened 2023-07-10
Adds a comment for an iterator based implementation for Polygon::length

edemir:
Outlines an alternative implementation for Polygon::length.  This exercise is aimed at novice users and we want to keep the implementation simple. For users familiar with iterator concepts this comment points to an alternative.


djmitche (approved):
Nice! This will read to the eager student as "here's a take-home exercise" and to the more exhausted student as "cool extra info".

mgeisler:
Thanks!

Issue #334, opened 2023-02-03
Make the shell snippets easier to copy/paste (e.g. cargo/running-locally.md)?


gendx:
The code blocks of `mdbook` provide a "Copy to clipboard" button - however pages like https://google.github.io/comprehensive-rust/cargo/running-locally.html contain shell blocks like follows:

```
$ cargo new exercise
     Created binary (application) `exercise` package
```

Clicking copies the whole block, including the dollar sign and the shell output. It would be nice to be able to only copy the command, e.g. `cargo new exercise`. We could achieve that by:
1. Creating a separate block for the output.
2. Removing the dollar sign. This would however make it less obvious that it's a shell command. Unless `mdbook` provides a way to display a dollar sign before the content, but exclude it from the copying?

WDYT?

mgeisler:
Yeah, that's definitely an annoyance. It's reported upstream here: https://github.com/rust-lang/mdBook/issues/864.

People tried to fix it repeatedly: https://github.com/google/comprehensive-rust/pull/25 and https://github.com/google/comprehensive-rust/pull/29. So we should probably do something about it...

My idea would be to fix it upstream, but I haven't taken the time to do that.

gendx:
Glad to see it's already reported upstream :)

mgeisler:
Our style guide actually talks about this: https://developers.google.com/style/code-syntax#prompt.

The system used there knows to not copy the `$`, even for examples that contain multiple commands. I think we should enhance `mdbook` with a similar capability.

mgeisler:
Fixed via #951.

mgeisler:
If `mdbook` ever gets support for not copying the prompt, then we can consider adding the prompts back. For now, it's probably better to remove them since this has been suggested repeatedly :smile:

sharunkumar:
Oh alright I could check it out

sharunkumar:
> We could separate the output from the shell commands (like you did in your PRs)

@mgeisler when doing [this](https://github.com/google/comprehensive-rust/pull/955/commits/17d8358be26c6ae2c6ed04456aa66c51bef191e9), the test build [fails](https://github.com/google/comprehensive-rust/actions/runs/5525362750/jobs/10078892609?pr=955#step:11:175) on the PR

I noticed this when I tried it in the previous PR itself, not sure how to fix it

mgeisler:
> > We could separate the output from the shell commands (like you did in your PRs)
> 
[...]

PR #955, opened 2023-07-11
remove $ from code-blocks in build_all.sh

sharunkumar:
Follow up for https://github.com/google/comprehensive-rust/issues/334#issuecomment-1630276602

### DIFF
--- a/src/android/aidl/client.md
+++ b/src/android/aidl/client.md
@@ -22,6 +22,6 @@ Build, push, and run the client on your device:
 {{#include ../build_all.sh:birthday_client}}
 ```
 
-```shell
+```text
 Happy Birthday Charlie, congratulations with the 60 years!
 ```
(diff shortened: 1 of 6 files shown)


src/android/aidl/client.md
```diff
@@ -20,5 +20,8 @@ Build, push, and run the client on your device:
 
 ```shell
 {{#include ../build_all.sh:birthday_client}}
+```
+
+```shell
```
mgeisler, line 25: Nice! Could you remove the `shell` language here and elsewhere — the output is just text and a stray `$` should not be syntax highlighted.

mgeisler: Sorry, I only saw [your comment now](https://github.com/google/comprehensive-rust/issues/334#issuecomment-1631580582) — please change `shell` to `text`. The default language is `rust` and so `mdbook test` will try to treat the code block as Rust code if you don't explicitly make it `text`.

sharunkumar: Done ✌️

mgeisler (approved):
Thanks for the help!

PR #961, opened 2023-07-13
zh-CN: translate traits

mgeisler:
Part of #324.

### DIFF
--- a/po/zh-CN.po
+++ b/po/zh-CN.po
@@ -9104,12 +9104,12 @@ msgstr ""
 
 #: src/traits.md:1
 msgid "# Traits"
-msgstr ""
+msgstr "# 特征"
 
 #: src/traits.md:3
 msgid ""
 "Rust lets you abstract over types with traits. They're similar to interfaces:"
-msgstr ""
+msgstr "Rust 让您可以依据特征对类型进行抽象化处理。特征与接口类似："
 
 #: src/traits.md:5
 msgid ""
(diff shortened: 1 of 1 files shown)


mgeisler:
Hey all, please take a look and leave suggestions if there is something here I should fix.

po/zh-CN.po
```diff
@@ -9153,13 +9153,13 @@ msgstr ""
 
 #: src/traits/trait-objects.md:1
 msgid "# Trait Objects"
-msgstr ""
+msgstr "# 特征对象"
```
wnghl, line 9156:
```suggestion
msgstr "# 特征（Trait）对象"
```
Keywords in the language or specific concepts are often not translated, such as in https://rustwiki.org/zh-CN/book/ch10-02-traits.html.

po/zh-CN.po
```diff
@@ -9153,13 +9153,13 @@ msgstr ""
...
 
 #: src/traits/trait-objects.md:3
 msgid ""
 "Trait objects allow for values of different types, for instance in a "
 "collection:"
-msgstr ""
+msgstr "特征对象可接受不同类型的值，举例来说，在集合中会是这样："
```
wnghl, line 9162:
```suggestion
msgstr "特征（Trait）对象可接受不同类型的值，举例来说，在集合中会是这样："
```

po/zh-CN.po
mgeisler: Nice, I would do the same for a Danish translation :slightly_smiling_face:

mgeisler:
@wnghl applied a bunch of suggestions — @henrif75, perhaps you can rubber stamp this so we can get it merged? We'll need to run a `msgmerge` on the file eventually and go over the many fuzzy messages after that.

henrif75 (approved):

PR #965, opened 2023-07-13
zh-CN: translate generics

mgeisler:
Part of #324.

### DIFF
--- a/po/zh-CN.po
+++ b/po/zh-CN.po
@@ -9015,7 +9015,7 @@ msgstr ""
 
 #: src/generics/methods.md:1
 msgid "# Generic Methods"
-msgstr "# 通用方法"
+msgstr "# 泛型方法"
 
 #: src/generics/methods.md:3
 msgid "You can declare a generic type on your `impl` block:"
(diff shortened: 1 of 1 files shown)


mgeisler:
Hey all, please take a look and leave suggestions if there is something here I should fix.

po/zh-CN.po
```diff
@@ -9014,11 +9015,11 @@ msgstr ""
 
 #: src/generics/methods.md:1
 msgid "# Generic Methods"
-msgstr ""
+msgstr "# 通用方法"
```
wnghl, line 9018:
```suggestion
msgstr "# 泛型方法"
```
Maybe '泛型' is a more common translation of word 'generic'

po/zh-CN.po
```diff
@@ -9052,14 +9053,20 @@ msgid ""
...
 
 #: src/generics/monomorphization.md:1
 msgid "# Monomorphization"
-msgstr ""
+msgstr "# 单态化"
 
 #: src/generics/monomorphization.md:3
 msgid "Generic code is turned into non-generic code based on the call sites:"
-msgstr ""
+msgstr "通用代码会根据调用站点转换为非通用代码："
```
wnghl, line 9069:
```suggestion
msgstr "泛型代码根据调用位置转换为非泛型代码："
```

po/zh-CN.po
mgeisler: Let me commit your suggestion for now, then I suggest discussion the best translation on #324 or similar.

wnghl (approved):

PR #967, opened 2023-07-13
zh-CN: translate structs

mgeisler:
Part of #324.


anlunx (approved):
I think the keypoints of this section is left untranslated by accident. Although we can have another PR for that

PR #969, opened 2023-07-13
zh-TW: translate structs

mgeisler:
Part of ##684.


po/zh-TW.po
```diff
@@ -19499,3 +19501,30 @@ msgid ""
...
+"  *  Rust generally doesn’t like inexplicit things, like automatic "
+"unwrapping or for instance using booleans as integers.\n"
+"  *  Operator overloading is discussed on Day 3 (generics).\n"
+"* The example is a subtle reference to the [Mars Climate Orbiter](https://en."
+"wikipedia.org/wiki/Mars_Climate_Orbiter) failure."
+msgstr ""
+"* 如要對原始類型中值的額外資訊進行編碼，Newtypes 是絕佳的方式，舉例來說：\n"
+" * 此數字會採用某些測量單位：在上例中為 `Newtons`。\n"
+" * 此值在建立時已通過某些驗證，因此往後不必在每次使用 'PhoneNumber(String)` "
+"或 `OddNumber(u32)` 時再次驗證。\n"
```
johnathan79717, line 19524:
```suggestion
" * 此值在建立時已通過某些驗證，因此往後不必在每次使用時再次驗證。例如："
"'PhoneNumber(String)` 或 `OddNumber(u32)` 時。\n"
```

po/zh-TW.po
```diff
@@ -19499,3 +19501,30 @@ msgid ""
...
+"* 示範了如何透過存取 newtype 中的單一欄位，將 “f64” 值新增至 `Newtons` 類"
```
johnathan79717, line 19525:
```suggestion
"* 示範如何透過存取 newtype 中的單一欄位，將 “f64” 值新增至 `Newtons` 類"
```

johnathan79717 (approved):
I notice some untranslated speaker notes such as L5353-5354.  Did you intentionally leave them out?

mgeisler:
> I notice some untranslated speaker notes such as L5353-5354. Did you intentionally leave them out?

Nope, I did not exclude anything on purpose — I just mechanically split the translation based on the source of each message (using `msgrep --location`).

po/zh-TW.po
```diff
@@ -19499,3 +19501,30 @@ msgid ""
 "}\n"
 "```"
 msgstr ""
+
+#: src/structs/tuple-structs.md:37
```
mgeisler, line 19505: Hi @henrif75, you'll notice that this translation causes merge conflicts for other PRs (such as #971).

I used `msgcat --use-first po/zh-TW.po newly-translated-zh-TW.po -o po/zh-TW.po` to update the file and this makes left-over translations go to the end. Some potential questions:

- Why did the message go to the end? The `msgid` wasn't found in the `zh-TW.po` file we had at the moment.
- Why was it not found? The `zh-TW.po` file has not been updated with a `msgmerge` in a while, so it has outdated `msgid` entries.
- Is this bad? No, not at all. The `msgid` is from a freshly generated `messages.pot` file, which means that it has a new and useful `msgid` entry. When `msgmerge --update` is run later, the entry will go to its correct spot.
- How can we avoid merge conflicts? I've moved messages like this one to its approximate correct location by hand in other PRs. Basically, if the PR had a message all the way at the end, it was a smell and I used the `#:` source line to move it back to an earlier spot in the file.

I guess that's more than you ever wanted to know about all this :smile:

PR #976, opened 2023-07-14
pt-BR: Review new fuzzy entries after .po file refresh

henrif75:
Part of #317 and #893


joaovicmendes (approved):
I tried my best to *look* through everything, but I did not do a proper detailed review. If someone else would also like to check, that would be nice :)

PR #978, opened 2023-07-14
Update mutex.md to correct spelling

vanyabrucker:


qwandor:
As the parentheses after explain, "iff" means "if and only if". It is not a misspelling.

PR #987, opened 2023-07-17
Rust isn't functional, but it does have functional elements

domenukk:
Add information about this

### DIFF
--- a/src/hello-world.md
+++ b/src/hello-world.md
@@ -37,7 +37,7 @@ Key points:
   the scope they are used in. Rust macros are actually only
   [partially hygienic](https://veykril.github.io/tlborm/decl-macros/minutiae/hygiene.html).
 
-* While it is not a functional language, it includes a range of
-  [functional concepts](https://doc.rust-lang.org/book/ch13-00-functional-features.html).
+* Rust is multi-paradigm. For example, it has powerful [object-oriented programming features](https://doc.rust-lang.org/book/ch17-00-oop.html),
+  and, while it is not a functional language, it includes a range of [functional concepts](https://doc.rust-lang.org/book/ch13-00-functional-features.html).
 
 </details>


src/hello-world.md
```diff
@@ -37,4 +37,7 @@ Key points:
   the scope they are used in. Rust macros are actually only
   [partially hygienic](https://veykril.github.io/tlborm/decl-macros/minutiae/hygiene.html).
 
+* While it is not a functional language, it includes a range of
+  [functional concepts](https://doc.rust-lang.org/book/ch13-00-functional-features.html)
```
mgeisler, line 41:
```suggestion
  [functional concepts](https://doc.rust-lang.org/book/ch13-00-functional-features.html).
```

src/hello-world.md
```diff
@@ -37,4 +37,7 @@ Key points:
   the scope they are used in. Rust macros are actually only
   [partially hygienic](https://veykril.github.io/tlborm/decl-macros/minutiae/hygiene.html).
 
+* While it is not a functional language, it includes a range of
```
sakex, line 40: I don't think we want to say that, we could make the point about any paradigm. Ie it's not OO but has aspects of object orientation

mgeisler (approved):

src/hello-world.md
mgeisler: You're right, but I've seen lots of people call Rust somewhat functional.

Could we include what you just wrote as well? 🙂 It's a common question in class! People like to try and classify Rust and it's neither a strict functional language nor is it a strict OO language.

domenukk: Yes maybe add OO as well? I'm fine with dropping functional again here as well, or just saying "Mutli-Paradigm"(as wikipedia does)

domenukk: I have added that, please take another look

sakex (approved):

PR #989, opened 2023-07-17
Move to more legible types

domenukk:
There is some Clippy lint (cannot find it right now) that suggests this change, and I think it's reasonable.
`321i32` vs `321_i32` - I got a clear favorite.

### DIFF
--- a/src/basic-syntax/scalar-types.md
+++ b/src/basic-syntax/scalar-types.md
@@ -40,4 +40,7 @@ There are a few syntaxes which are not shown above:
   }
   ```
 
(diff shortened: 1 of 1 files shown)


fw-immunant:
I think it might be better to show both this form and the allmashedtogether one that it replaces. Students may see both forms and showing the type suffices separately from cosmetic underscores in addition to in combination demonstrates the orthogonality of the features.

domenukk:
Is it enough to "show" that in the speaker notes?

domenukk:
@fw-immunant added more information about the underscores into the speaker notes

henrif75 (approved):

PR #991, opened 2023-07-17
Add note about doc tests

domenukk:
We discuss them later so it's not _needed_ but it might still be nice to talk about here

### DIFF
--- a/src/basic-syntax/rustdoc.md
+++ b/src/basic-syntax/rustdoc.md
@@ -30,4 +30,7 @@ idiomatic to document all public items in an API using this pattern.
 * Inner doc comments are discussed later (in the page on modules) and need not
   be addressed here.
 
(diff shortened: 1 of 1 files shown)


djmitche (approved):

src/basic-syntax/rustdoc.md
```diff
@@ -30,4 +30,7 @@ idiomatic to document all public items in an API using this pattern.
 * Inner doc comments are discussed later (in the page on modules) and need not
   be addressed here.
 
+* Rustdoc comments can contain code snippets that we can run and test using `cargo test`.
+  We will discuss these tests in the [Testing section](../testing/doc-tests.html).
```
mgeisler, line 34: Just a small note, we can (and should) link to the Markdown files instead. That makes the links work both in GitHub and in the rendered HTML (`mdbook` will replace `.md` with `.html` in the link).

PR #993, opened 2023-07-17
Fix links to FromIterator, HashSet

domenukk:
The links were a bit misguided


djmitche (approved):
Those fragment references have been giving us some trouble lately, but this at least gets to the right page!

PR #995, opened 2023-07-17
Static variables don't need unsafe

domenukk:
Thix fixes a slight inaccuracy introduced in #912


henrif75 (approved):

mgeisler:
FYI, @fw-immunant and @qwandor.

qwandor:
I think it's still worth mentioning `Mutex`, it's common to use it for safe interior mutability of static variables.

PR #998, opened 2023-07-17
Memory Management: Rework Scope-based Cons

domenukk:
Trying to also capture Swift's ARC, I'm not sure if it's really scope-based but it's very close(?)

BTW: I have no idea how use-after-free can happen here?


domenukk:
After talking to @mgeisler, I've removed the UAF reference, since this is not inherent to the method but occurs when people work around it by leaking a pointer

henrif75 (approved):

PR #999, opened 2023-07-17
Use non-generic default

domenukk:
This is what clippy (albeit pedantic) suggests (see https://rust-lang.github.io/rust-clippy/master/index.html#/default_trait_access ) and it feels less redundant to me.

### DIFF
--- a/src/structs/field-shorthand.md
+++ b/src/structs/field-shorthand.md
@@ -56,11 +56,11 @@ fn main() {
      }
      fn create_default() {
          let tmp = Person {
-             ..Default::default()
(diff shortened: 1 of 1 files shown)


henrif75 (approved):

mgeisler:
> This is what clippy (albeit pedantic) suggests (see https://rust-lang.github.io/rust-clippy/master/index.html#/default_trait_access ) and it feels less redundant to me.

Awesome, this is also what I would write normally — much clearer with `Foo::default()` than the mysterious `Default::default()`.

PR #1005, opened 2023-07-18
Error is not yet available in core on stable

domenukk:
It's "good practice", unless when it isn't :)


mgeisler (approved):

proski:
It's available in Rust 1.81 released today.

mgeisler:
Thanks @proski, I created #2369 to revert the changes here. From reading the release notes, it seems it is `core::error::Error` that is now stable, so should the page talk about that type instead?

PR #1010, opened 2023-07-20
New is implied in square

domenukk:
It's a minor nitpick, but I got the impression that in rust you generally don't use `new_something` but just `something`, the `new` is implied by the return type.


djmitche (commented):
Hm, I came here to say "no, I like `new_square`" but scanning the stdlib, you're right. There are some `new_..` methods but they are variants like `new_in(allocator)`. Things that construct specific variants of a type tend to not include a `new_` prefix.

I'm curious what other course maintainers think.

mgeisler:
> Hm, I came here to say "no, I like `new_square`" but scanning the stdlib, you're right.

Me too :smile: Dominik sits right next to me and I hate to admit it, but he's probably right...

mgeisler:
The [Rust API guidelines](https://rust-lang.github.io/api-guidelines/naming.html) mention `Foo::new` and `Foo::with_more_details`.

mgeisler (approved):
People seem to agree with this name, so let's get this in!

PR #1015, opened 2023-07-21
Remove unneeded macro from Enum Sizes slide

domenukk:
We don't need to introduce macros here :)

### DIFF
--- a/src/enums/sizes.md
+++ b/src/enums/sizes.md
@@ -3,10 +3,8 @@
 Rust enums are packed tightly, taking constraints due to alignment into account:
 
 ```rust,editable
-use std::{
-    any::type_name,
-    mem::{align_of, size_of},
-};
+use std::any::type_name;
+use std::mem::{align_of, size_of};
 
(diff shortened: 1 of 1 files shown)


src/enums/sizes.md
```diff
@@ -3,13 +3,14 @@
...
-
-macro_rules! dbg_size {
-    ($t:ty) => {
-        println!("{}: size {} bytes, align: {} bytes",
-                 stringify!($t), size_of::<$t>(), align_of::<$t>());
-    };
+use std::{
+    any::type_name,
+    mem::{align_of, size_of},
+};
```
mgeisler, line 9: We try to use module-level imports (but I haven't figured out how to run `rustfmt` on inline code snippets yet):

```suggestion
use std::any::type_name;
use std::mem::{align_of, size_of};
```

mgeisler (approved):
This is so much better, thanks a lot!

djmitche (approved):

PR #1018, opened 2023-07-23
Update while-let-expressions.md

shaurya947:
Update code snippet description to use `v.into_iter()` instead of `v.iter()` since the former is what's actually used in the code snippet.


henrif75 (approved):
Thank you!

PR #1019, opened 2023-07-24
Clarify distinction between interior mutability and mutable statics.

qwandor:
Follow-up to #995, which confused the two a bit.


mgeisler:
Hi @domenukk, please take a look at this. Thanks!

sakex (approved):

PR #1020, opened 2023-07-24
Move `thread::spawn` to separate function

mgeisler:
This might make it clearer why the thread cannot borrow from the string.

@djmitche, do you think this is worth the extra vertical space?

### DIFF
--- a/src/concurrency/scoped-threads.md
+++ b/src/concurrency/scoped-threads.md
@@ -5,12 +5,15 @@ Normal threads cannot borrow from their environment:
 ```rust,editable,compile_fail
 use std::thread;
 
-fn main() {
+fn foo() {
     let s = String::from("Hello");
-
     thread::spawn(|| {
         println!("Length: {}", s.len());
     });
+}
+
+fn main() {
+    foo();
 }
 ```
 


djmitche:
Yes, I think that's worthwhile. Does the playground run `#[test]` blocks if given in this context? That might save space, and I don't think there's anything in the test harness that "kills" threads between tests.

mgeisler:
> Yes, I think that's worthwhile. Does the playground run `#[test]` blocks if given in this context? That might save space, and I don't think there's anything in the test harness that "kills" threads between tests.

The [full Playground](https://play.rust-lang.org/) supports running tests: the `Run` button changes to `Test` when it notices a unit test.

djmitche (approved):
```
error[E0601]: `main` function not found in crate `playground`
  --> src/main.rs:12:2
   |
12 | }
   |  ^ consider adding a `main` function to `src/main.rs`
```
so looks like `#[test]` won't work..

mgeisler:
Yeah, unit tests only work in the [full playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3d2c89d5699af6187b312feb2177bbe9). The embedded playground is not smart enough to detect the unit tests. It would be nice to extend it to be as clever as the full playground.

PR #1024, opened 2023-07-24
Align dining-philosophers-async.rs with sync version

mgeisler:
This updates the version to use `std::mem::swap` like the synchronous version.

### DIFF
--- a/src/exercises/concurrency/dining-philosophers-async.rs
+++ b/src/exercises/concurrency/dining-philosophers-async.rs
@@ -72,12 +72,15 @@ async fn main() {
         let mut philosophers = vec![];
         let (tx, rx) = mpsc::channel(10);
         for (i, name) in PHILOSOPHERS.iter().enumerate() {
-            let left_fork = forks[i].clone();
-            let right_fork = forks[(i + 1) % PHILOSOPHERS.len()].clone();
+            let left_fork = Arc::clone(&forks[i]);
+            let right_fork = Arc::clone(&forks[(i + 1) % PHILOSOPHERS.len()]);
+            if i % 2 == 1 {
+                std::mem::swap(&mut left_fork, &mut right_fork);
+            }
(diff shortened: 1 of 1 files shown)


src/exercises/concurrency/dining-philosophers-async.rs
```diff
@@ -72,12 +72,15 @@ async fn main() {
         let mut philosophers = vec![];
         let (tx, rx) = mpsc::channel(10);
         for (i, name) in PHILOSOPHERS.iter().enumerate() {
-            let left_fork = forks[i].clone();
-            let right_fork = forks[(i + 1) % PHILOSOPHERS.len()].clone();
+            let left_fork = Arc::clone(&forks[i]);
+            let right_fork = Arc::clone(&forks[(i + 1) % PHILOSOPHERS.len()]);
+            if i % 2 == 1 {
```
mgeisler, line 77: I'm not actually sure why this is done for every _second_ philosopher? Why not just swap it for a single philosopher like the synchronous version?

rbehjati: Good point. That should work. Do you want me to fix it, or will you prefer to fix it in this PR?

rbehjati (approved):
Thanks.

src/exercises/concurrency/dining-philosophers-async.rs
mgeisler: > Do you want me to fix it, or will you prefer to fix it in this PR?

Cool, I think this should do it:

```suggestion
            // To avoid a deadlock, we have to break the symmetry
            // somewhere. This will swap the forks without deinitializing
            // either of them.
            if i  == 0 {
```

rbehjati: Nice. Thanks.

PR #1027, opened 2023-07-24
`std::mem::drop` is in the prelude

mgeisler:


rbehjati (approved):

qwandor:
I think this change makes the distinction between `std::mem::drop` and `Drop::drop` less clear.

rbehjati:
> I think this change makes the distinction between `std::mem::drop` and `Drop::drop` less clear.

I agree with that :) 

I approved the change because I thought maybe there is an established convention that did not know of for shortening `std::mem::drop` to `drop`.

qwandor:
> > I think this change makes the distinction between `std::mem::drop` and `Drop::drop` less clear.
> 
> I agree with that :)
> 
> I approved the change because I thought maybe there is an established convention that did not know of for shortening `std::mem::drop` to `drop`.

In code, sure. In the explanatory text here, I think it's more clear to be explicit.

src/traits/drop.md
```diff
@@ -34,7 +34,7 @@ fn main() {
 Discussion points:
 
 * Why doesn't `Drop::drop` take `self`?
-    * Short-answer: If it did, `std::mem::drop` would be called at the end of
+    * Short-answer: If it did, `drop` would be called at the end of
```
rbehjati, line 37: Perhaps change "`drop`" to "`drop` (std::mem::drop)" to be clear.

mgeisler:
> > > I think this change makes the distinction between `std::mem::drop` and `Drop::drop` less clear.
> > 
> > 
> > I agree with that :)
> > I approved the change because I thought maybe there is an established convention that did not know of for shortening `std::mem::drop` to `drop`.
> 
> In code, sure. In the explanatory text here, I think it's more clear to be explicit.

Good point, I had not thought of distinguishing between the code and the text.

mgeisler:
Oh, my bad, it's already merged... Let me fix in a new PR.

PR #1028, opened 2023-07-24
Fix `self` shorthand syntax

VitoKingg:
According to [The Rust Reference](https://doc.rust-lang.org/nightly/reference/items/associated-items.html#methods), the keyword `self` is a shorthand for `self: Self`, not `self: &Self`.

### DIFF
--- a/src/methods.md
+++ b/src/methods.md
@@ -31,11 +31,11 @@ Key Points:
 * It can be helpful to introduce methods by comparing them to functions.
   * Methods are called on an instance of a type (such as a struct or enum), the first parameter represents the instance as `self`.
   * Developers may choose to use methods to take advantage of method receiver syntax and to help keep them more organized. By using methods we can keep all the implementation code in one predictable place.
-* Point out the use of the keyword `self`, a method receiver. 
-  * Show that it is an abbreviated term for `self:&Self` and perhaps show how the struct name could also be used. 
+* Point out the use of the keyword `self`, a method receiver.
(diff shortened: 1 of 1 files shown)


mgeisler (approved):
Thanks for catching this!

src/methods.md
```diff
@@ -31,11 +31,11 @@ Key Points:
 * It can be helpful to introduce methods by comparing them to functions.
   * Methods are called on an instance of a type (such as a struct or enum), the first parameter represents the instance as `self`.
   * Developers may choose to use methods to take advantage of method receiver syntax and to help keep them more organized. By using methods we can keep all the implementation code in one predictable place.
-* Point out the use of the keyword `self`, a method receiver. 
-  * Show that it is an abbreviated term for `self:&Self` and perhaps show how the struct name could also be used. 
+* Point out the use of the keyword `self`, a method receiver.
+  * Show that it is an abbreviated term for `self: Self` and perhaps show how the struct name could also be used.
```
qwandor, line 35: I think you fixed the wrong thing: in the example above, the method signature uses `&self` (which is indeed equivalent to `self: &Self`) not just `self`.

VitoKingg: In the code example, it uses `&self`, but in the speaker notes, it demonstrates the keyword `self` not `&self`. I suggest adding " `&'lifetime self` is an abbreviated term for `self: &'lifetime Self` " as well to avoid the ambiguity.

PR #1031, opened 2023-07-25
Revert "`std::mem::drop` is in the prelude"

mgeisler:
Reverts google/comprehensive-rust#1027

We discussed it and it seems nicer to spell out the full function name in the text.

### DIFF
--- a/src/traits/drop.md
+++ b/src/traits/drop.md
@@ -34,7 +34,7 @@ fn main() {
 Discussion points:
 
 * Why doesn't `Drop::drop` take `self`?
-    * Short-answer: If it did, `drop` would be called at the end of
+    * Short-answer: If it did, `std::mem::drop` would be called at the end of
         the block, resulting in another call to `Drop::drop`, and a stack
         overflow!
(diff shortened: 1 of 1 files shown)


src/traits/drop.md
```diff
@@ -34,7 +34,7 @@ fn main() {
 Discussion points:
 
 * Why doesn't `Drop::drop` take `self`?
-    * Short-answer: If it did, `drop` would be called at the end of
+    * Short-answer: If it did, `std::mem::drop` would be called at the end of
```
rbehjati, line 37:
```suggestion
    * Short-answer: If it did, `drop` (`std::mem::drop`) would be called at the end of
```

rbehjati (approved):
Thanks.

src/traits/drop.md
qwandor: Actually, reading this again, it's not really true: `std::mem::drop` wouldn't be called, there's nothing special about it. The value would just be dropped.

qwandor: Well it's not called directly, it's called as part of dropping the value.

PR #1029, opened 2023-07-25
Introduce Rust Fundamentals on welcome page

mgeisler:
The split between Day 1 to 3 and the deep dives was never explicitly explained.

PR #1032, opened 2023-07-25
Explain what "Android team" means

mgeisler:
I hope this makes it a little clearer where the course comes from.

A followup to #1029.


djmitche (approved):
:)

PR #964, opened 2023-07-13
zh-TW: translate std

mgeisler:
Part of #684.

PR #1036, opened 2023-07-27
zh-TW: fix line breaks in lists

mgeisler:
I looked through the recent PRs for the Taiwanese translations and found that some lists items had extra line breaks (without the necessary indentation). This PR should fix the rendering.

Part of #684.


mgeisler:
@johnathan79717, thanks for flagging this in your review of #964.

po/zh-TW.po
```diff
@@ -9028,16 +9028,9 @@ msgid ""
 "  you can iterate over a vector with `for i in some_vec { .. }` but\n"
 "  `some_vec.next()` doesn't exist."
 msgstr ""
-"*`Iterator` 特徵會對集合實作許多常見的函式程式操作，\n"
-"例如 `map`、`filter`、`reduce` 等等。您可以藉由此特徵找出\n"
-"所有相關的說明文件。在 Rust 中，這些函式會產生程式碼，且應與對應的命令式\n"
-"實作項目一樣有效率。\n"
+"*`Iterator` 特徵會對集合實作許多常見的函式程式操作，例如 `map`、`filter`、`reduce` 等等。您可以藉由此特徵找出所有相關的說明文件。在 Rust 中，這些函式會產生程式碼，且應與對應的命令式實作項目一樣有效率。\n"
```
johnathan79717, line 9031: In this case, I think the problem is not the newline, but the lack of space between "*" and the list item content. Also let's try to keep lines short for people with smaller screens.

```suggestion
"* `Iterator` 特徵會對集合實作許多常見的函式程式操作，例如 `map`、`filter`、"
"  `reduce` 等等。您可以藉由此特徵找出所有相關的說明文件。在 Rust 中，這些"
"  函式會產生程式碼，且應與對應的命令式實作項目一樣有效率。\n"
```

johnathan79717 (changes requested):
Thanks for the change!  I think it's still worth keeping the lines short for people with smaller screens, as long as we make sure that the lists are displayed correctly.

johnathan79717 (approved):

mgeisler:
> Thanks for the change! I think it's still worth keeping the lines short for people with smaller screens, as long as we make sure that the lists are displayed correctly.

Yeah, I realized afterwards that I could just have removed the `\n` characters but kept the lines the same in the PO file :smile:

Eventually, I'll update mdbook-i18n-helpers to extract the text item-by-item and also turn `'\n'` into `' '` in the PO files. It will be the same Markdown text, but this normalization will remove a lot of noise in the PO files.

PR #1038, opened 2023-07-27
Default trait slides cleanup

domenukk:
This is a follow up on #999 and also adds additional notes to the default slide


src/traits/default.md
```diff
@@ -43,7 +43,9 @@ fn main() {
   * Standard Rust types often implement `Default` with reasonable values (e.g. `0`, `""`, etc).
   * The partial struct copy works nicely with default.
   * Rust standard library is aware that types can implement `Default` and provides convenience methods that use it.
+  * the `..` syntax is called [`struct update syntax`][2]
```
mgeisler, line 46: I think this should be in a normal font (it's not code):

```suggestion
  * the `..` syntax is called [struct update syntax][2]
```

mgeisler (approved):
Thanks!

Issue #847, opened 2023-06-21
Make `/welcome.html` redirect to `/`


mgeisler:
We currently have both

- `https://google.github.io/comprehensive-rust/` (there must be an `index.html` page here)
- `https://google.github.io/comprehensive-rust/welcome.html`

online. The latter is mostly seen when you navigate forward and then back to the start page.

We should only have one page and we should make `https://google.github.io/comprehensive-rust/` the [canonical page](https://developers.google.com/search/docs/crawling-indexing/canonicalization).

Fixing this might involve working on `mdbook` to have it generate links back to the `/` page instead of to `/index.html`.

Make sure this works for the translations as well.

CYBWEBALI:
Can I work on this issue?

mgeisler:
> Can I work on this issue?

Sure, go right ahead! :smile:

lorinaX:
Could I work on this or is already taken by someone else? 😄

mgeisler:
Hey @lorinaX, you could discuss with @CYBWEBALI on how to solve this. I don't know what the plan is here, but I would be curious to hear :smile: 

We also have another issue related to linking pages: https://github.com/google/mdbook-i18n-helpers/issues/35. That would also require changing `mdbook` or (probably easier) write a tool which can add the necessary HTML after the book has been rendered.

mgeisler:
Hi @CYBWEBALI and @lorinaX, please let me know if you have an idea for how to solve this :smile: It would be good to discuss it before either of you invest a lot of time into it.

lorinaX:
well since the other guy didn't grab this issue, I'll go ahead and work on it right now I have a few ideas on how I can do this, I will make sure to update you after I have finished a few ideas, shouldn't take me too long!

Tnation888:
Okay I'll be waiting to be updated 

Sent from Yahoo Mail on Android 
 
  On Sun, Jun 25, 2023 at 8:39 AM, Lorina ***@***.***> wrote:   


well since the other guy didn't grab this issue, I'll go ahead and work on it right now I have a few ideas on how I can do this, I will make sure to update you after I have finished a few ideas, shouldn't take me too long!

lorinaX:
Can you give me some exact steps for me to replicate this, I did some changes and I think it should be fine now.

Tnation888:
What do you think was the issue and what steps did you take for you to make the changes?

Sent from Yahoo Mail on Android 
 
[...]

PR #1039, opened 2023-07-28
Rename `welcome.md` to `index.md`

mgeisler:
When building the book, mdBook will always generate an `index.html` page for first page of the book. This meant that we had the same content available under two different names:

- `welcome.html`: this is what the TOC link to, and
- `index.html` or simply `/`: this is what search engines link to.

Renaming the page and setting up a redirect should fix this confusion. We still don’t have a good way of avoiding links to the `index.html` page, but this should fix the first half of the problem.

I tested this for translations as well by building the output in a subdirectory and serving the parent directory.

Part of #847.


qwandor (approved):

mgeisler:
Thanks, let me deploy this now so I can test it before going home... :smile:

mgeisler:
It looks good, `https://google.github.io/comprehensive-rust/ko/welcome.html` sends me to `https://google.github.io/comprehensive-rust/ko/` as expected.

PR #1044, opened 2023-08-01
Fixed link

Arman-sm:


sakex (approved):

sakex:
Thanks for this

mgeisler:
> Now that #1045 is fixed, let me try rerunning the tests so we can merge this!

Nope, it still fails. I'm a bit surprised by this since my understanding was that we're testing the merge commit that would be created for the PR. I see the action checking out `refs/remotes/pull/1044/merge`.

I tried fetching this locally and now I see the problem: the merge commit is made at the time of the PR:

image

We might be able to get this unstuck by adding another commit to the branch, which I'll try next.

src/basic-syntax/static-and-const.md
```diff
@@ -50,7 +50,7 @@ Because `static` variables are accessible from any thread, they must be `Sync`.
 is possible through a [`Mutex`](https://doc.rust-lang.org/std/sync/struct.Mutex.html), atomic or
 similar. It is also possible to have mutable statics, but they require manual synchronisation so any
 access to them requires `unsafe` code. We will look at
-mutable statics](../unsafe/mutable-static-variables.md) in the chapter on Unsafe Rust.
+[mutable statics](../unsafe/mutable-static-variables.md) in the chapter on Unsafe Rust.
```
mgeisler, line 53: Spurious change to make tests pick up fix for #1045 (we'll eventually auto-format all our Markdown files so the space is not a huge concern):

```suggestion
[mutable statics](../unsafe/mutable-static-variables.md) in the chapter on Unsafe Rust. 
```

mgeisler:
Adding a new commit worked. This means we can advice others to do the same: either add a new commit or simply `git merge` the latest `main` into their branch. That will ensure they have the new changes and the tests should work again.

src/basic-syntax/static-and-const.md
```diff
@@ -50,7 +50,7 @@ Because `static` variables are accessible from any thread, they must be `Sync`.
...
```
mgeisler, line 53:
```suggestion
[mutable statics](../unsafe/mutable-static-variables.md) in the chapter on Unsafe Rust.
```

Issue #1047, opened 2023-08-02
Print with speaker notes


mgeisler:
### Discussed in https://github.com/google/comprehensive-rust/discussions/1041

<div type='discussions-op-text'>

<sup>Originally posted by **armenic** July 30, 2023</sup>
Dear developers, community,

Is there a way to print the book with the speaker notes open? When I click print book icon then the pages are rendered with speakers notes appearing as a collapsed `Details >` and it is not clickable in the PDF. It seems one needs to go page by page expand each speaker note so that they are printed.

Thanks so much for your help!</div>

sakex:
I have some time to do this now, I'm assigning myself

sakex:
- I was not able to reproduce locally either with `mdbook serve` or `mdbook build` -> `python3 -m http.server 8080`.
- On the  public version, the first `details` seemed to be opened while the rest are closed.
- At first I thought it was a race condition issue. The actual culprit is actually is that we are checking for `window.location.pathname == "/print.html"` when the actual page in prod is `comprehensive-rust/print.html`. It is easily missed because `mdbook serve` serves directly into `/`.
-  Fix https://github.com/google/comprehensive-rust/pull/1048

PR #1048, opened 2023-08-02
Fix print details toggle

sakex:
Fixes https://github.com/google/comprehensive-rust/issues/1047

### DIFF
--- a/speaker-notes.js
+++ b/speaker-notes.js
@@ -230,7 +230,8 @@
       markDefunct();
       break;
     default:
-      if (window.location.pathname == "/print.html") {
+      const printPageRegexp = /print\.html?/;
+      if (window.location.pathname.match(printPageRegexp)) {
         setupPrintPage();
         return;
       }
--- a/speaker-notes.js
+++ b/speaker-notes.js
@@ -230,8 +230,7 @@
       markDefunct();
       break;
     default:
-      const printPageRegexp = /print\.html?/;
-      if (window.location.pathname.match(printPageRegexp)) {
+      if (window.location.pathname.endsWith("/print.html")) {
         setupPrintPage();
         return;
       }


sakex:
https://github.com/google/comprehensive-rust/issues/1047#issuecomment-1662121869

speaker-notes.js
```diff
@@ -230,7 +230,8 @@
       markDefunct();
       break;
     default:
-      if (window.location.pathname == "/print.html") {
+      const printPageRegexp = /print\.html?/;
+      if (window.location.pathname.match(printPageRegexp)) {
```
mgeisler, line 234: Ah, thanks for finding the problem so quickly!

Since we want to check if the string [ends with](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith) `/print.html`, could we do this instead:

```suggestion
      if (window.location.pathname.endsWith("/print.html")) {
```

This way we won't get a false negative on a page named `ups-print.html`.

mgeisler (changes requested):

speaker-notes.js
sakex: My biggest concern would be for instance `webassembly/print.html`

sakex:
Fixes https://github.com/google/comprehensive-rust/issues/1047

speaker-notes.js
mgeisler: > Sounds good, I wasn't aware of that method

Yeah, I didn't know about it either :smile: 

> My biggest concern would be for instance `webassembly/print.html`

Right, but that should only happen if we make a `print.md` file ourselves — and in that case we will get an error from mdbook.

mgeisler:
Thanks!

PR #1049, opened 2023-08-02
Update memory management comparison

mgeisler:
This is a follow-up to #998, which I feel took out too much of the key points of the scope-based memory management section.

The points made in each section should compare the technique against the other techniques. It should do so based on the languages listed (or languages that work the same).

### DIFF
--- a/src/memory-management/comparison.md
+++ b/src/memory-management/comparison.md
@@ -24,12 +24,12 @@ Here is a rough comparison of the memory management techniques.
   * Double-frees.
   * Memory leaks.
 * Automatic like Java:
+  * Runtime overhead.
   * Garbage collection pauses.
   * Destructor delays.
 * Scope-based like C++:
-  * Complex, opt-in by programmer (on C++).
-  * Circular references can lead to memory leaks
-  * Potential runtime overhead
+  * Complex, opt-in by programmer.
+  * Potential for use-after-free.
 * Compiler-enforced and scope-based like Rust:
   * Some upfront complexity.
   * Can reject valid programs.


src/memory-management/comparison.md
```diff
@@ -24,12 +24,12 @@ Here is a rough comparison of the memory management techniques.
...
   * Memory leaks.
 * Automatic like Java:
+  * Runtime overhead.
   * Garbage collection pauses.
   * Destructor delays.
 * Scope-based like C++:
-  * Complex, opt-in by programmer (on C++).
-  * Circular references can lead to memory leaks
-  * Potential runtime overhead
+  * Complex, opt-in by programmer.
```
domenukk, line 31: This is a very c++-centric point, and not true for other languages

src/memory-management/comparison.md
```diff
@@ -24,12 +24,12 @@ Here is a rough comparison of the memory management techniques.
   * Double-frees.
   * Memory leaks.
 * Automatic like Java:
+  * Runtime overhead.
   * Garbage collection pauses.
   * Destructor delays.
 * Scope-based like C++:
-  * Complex, opt-in by programmer (on C++).
-  * Circular references can lead to memory leaks
```
domenukk, line 31: This one is very important I think

domenukk:
Left some comments, what you'll do with it is up to you :)

src/memory-management/comparison.md
```diff
@@ -24,12 +24,12 @@ Here is a rough comparison of the memory management techniques.
...
-  * Potential runtime overhead
+  * Complex, opt-in by programmer.
+  * Potential for use-after-free.
```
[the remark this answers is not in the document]
domenukk: Maybe move it to speaker notes and add an explanation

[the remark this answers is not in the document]
mgeisler: The problem is that you can call [`std::unique_ptr::<T>::get`](https://en.cppreference.com/w/cpp/memory/unique_ptr/get) to get a raw pointer. So you frequently end up having to take responsibility for the lifetime of the pointers.

I see this as an important problem with this approach: the approach works 98% of the way, but last few percent will still cause memory issues.

src/memory-management/comparison.md
mgeisler: I removed it because it applies to all the techniques.

In CPython, reference counting is used for garbage collection. In addition, there is a [garbage collector](https://docs.python.org/3/library/gc.html) which runs once in a while to detect and clean up reference cycles. As described on that page, you can disable this garbage collector if you know you never create cycles. It used to be the case that cycles between objects with a destructor (a `__del__` method) would be uncollectable, but this is apparently no longer true after [PEP 442](https://peps.python.org/pep-0442/). If I understand correctly, it is now only C-level types which can become uncollectable.

In any case, the argument about reference cycles leaking memory also applies to manual memory management and it even applies to Rust. So I don't think we need to highlight it here.

src/memory-management/comparison.md
mgeisler: You mentioned ARC in Swift at some point, so I guess you might be thinking of this? Similar to the automatic reference counting done by Python, I'm not sure if ARC falls under the heading of scope-based memory management...

To be more precise: using `std::unique_ptr` in C++ comes with _zero overhead_ over using raw pointers. The compiler will insert precisely the necessary instructions to determine when to free the pointer, even if there are early returns etc. You can implement this in any language that has a concept of a destructor.

I'm not sure how to classify reference counting (manuel in C, `std::shared_ptr` in C++, ARC in Swift, and Python) in this sense.

src/memory-management/comparison.md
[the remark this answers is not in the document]
mgeisler: Right — I suppose that applies to all techniques except garbage collection like in Java?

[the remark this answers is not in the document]
sakex: Here is an example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4d5b587876ea626c7c2cc6cdea7763b9

The code can be fixed by manually implementing Drop but it is annoying to deal with

src/memory-management/comparison.md
domenukk: Yes I'm thinking of `std::shared_ptr`, I don't think you can write complete software with `unique_ptr` alone, can you?

src/memory-management/comparison.md
domenukk: It is inherent to reference counting: you cannot build reference cycles in (safe) rust at all, unless if you use `Rc` (which is the same thing as a `shared_ptr`/ARC again), and a GC will also clean up cycles.

src/memory-management/comparison.md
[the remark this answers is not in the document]
domenukk: That's still a very c++ problem, and, hence, I'd personally put it in the speaker notes

src/memory-management/comparison.md
domenukk: Ok read up on what scope-based memory management actually is, it's pretty much just stack allocated variables in C, and a bit more beef in C++? In this case the main drawback should be that it only works for simple non-dynamic programs, right?

sakex: @domenukk Reading your comment I realize that the course equates automatic memory management to garbage collection (in summary.md, automatic memory management is called garbage collection).

In my mental model, automatic memory management could be thought of as an interface for the programmer that allows them to not care about the memory lifecycles. While GC vs RC is an implementation detail.

Scope based forces the programmer to care about the lifecycles through move and reference semantics while abstracting away the actual allocations and destructions. So the interface is different for the programmer.

Maybe the class doesn't need to mention Garbage Collection or reference counting at all and just focus on automatic vs manual vs RAII?

src/memory-management/comparison.md
sakex: @domenukk Reading your comment I realize that the course equates automatic memory management to garbage collection (in summary.md, automatic memory management is called garbage collection).

In my mental model, automatic memory management could be thought of as an interface for the programmer that allows them to not care about the memory lifecycles. While GC vs RC is an implementation detail.

Scope based forces the programmer to care about the lifecycles through move and reference semantics while abstracting away the actual allocations and destructions. So the interface is different for the programmer.

Maybe the class doesn't need to mention Garbage Collection or reference counting at all and just focus on automatic vs manual vs RAII?

djmitche:
This section of slides has always felt like an awkward fit to me: much of it is not about Rust, and it contains a bunch of assertions and value judgements about other languages that feel like they are inviting an argument. I admit I've never had an argument -- in fact, I've never had any questions or discussion on this section. But, the discussion above is getting pretty long and has a lot of details about C++ that aren't really relevant to this course.

Could we simplify this section down to one slide? I think the key point is that Rust picks a particular memory management strategy that is similar to C++, and enforces that you don't "cheat" without a good reason. So maybe:

* Memory management designs differ mostly on how memory is freed
  * Up to the programmer to call `free(..)` at an appropriate place (ex: C).
  * Using reference counting to identify un-used data at the moment it is no longer used (ex: Python).
  * Using garbage collection to identify un-used data some time after it is no longer used (ex: Java).
  * Using "smart pointers" and variable scope to identify un-used data at the moment it goes out of scope (ex: C++).

* Rust chooses the last option, and enforces correct usage with the compiler.
  * It is possible to do "clever" things (such as the equivalent of `std::unique_ptr::get`) but these must be flagged with the "unsafe" keyword, must still produce correct behavior, are subject to heavy review.

I'm working on v2 of the course, which is adding a bunch of new content, so making existing things more concise is going to be important. If we can get through "memory management" in 2 minutes instead of 10, that helps a lot.

src/memory-management/comparison.md
domenukk: Yes maybe GC and RefCounting overcomplicate things then, sorry for the confusion.

As non-c++ person, thinking of RAII as a memory management model was non-obvious.
To make my confusion worse, the concept of RAII can be implemented using RefCounting, and wiki isn't talking about memory at all...
https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization#Reference_counting

domenukk:
Maybe, for now, just move the use-after-free example to notes or add an explanation to notes, then simplify this later as suggested by [djmitche](https://github.com/djmitche)

randomPoison:
My general thoughts on this from having run a few classes: I personally feel like this summary page is the most redundant part of the memory management section. I think the current slides going over each of the strategies is enough of a review for most of the students. So far I haven't had anybody really ask questions or express confusion, and this slide just repeats things that should have already been said on the previous few slides.

mgeisler:
Hey folks, I decided to take the page out completely — I agree with @randomPoison that it feels redundant.

djmitche (approved):
Thanks! This is shrunk down quite a bit in v2, too.

PR #1051, opened 2023-08-03
Expand "iff" to "if and only if"

mgeisler:
There has been a few PRs which wanted to change "iff" to "if". Expanding it should help, but we might also consider using less heavy math jargon.

### DIFF
--- a/src/concurrency/shared_state/arc.md
+++ b/src/concurrency/shared_state/arc.md
@@ -28,8 +28,8 @@ fn main() {
 
 * `Arc` stands for "Atomic Reference Counted", a thread safe version of `Rc` that uses atomic
   operations.
-* `Arc<T>` implements `Clone` whether or not `T` does. It implements `Send` and `Sync` iff
-  (if and only if) `T` implements them both.
+* `Arc<T>` implements `Clone` whether or not `T` does. It implements `Send` and `Sync` if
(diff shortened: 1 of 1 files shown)


src/concurrency/shared_state/arc.md
```diff
@@ -28,8 +28,8 @@ fn main() {
 
 * `Arc` stands for "Atomic Reference Counted", a thread safe version of `Rc` that uses atomic
   operations.
-* `Arc<T>` implements `Clone` whether or not `T` does. It implements `Send` and `Sync` iff `T`
-  implements them both.
+* `Arc<T>` implements `Clone` whether or not `T` does. It implements `Send` and `Sync` iff
+  (if and only if) `T` implements them both.
```
qwandor, line 32: Rather than using the short then explaining it, let's just use the long form. I.e. "It implements Send and Sync if and only if T implements them both."

mgeisler, line 32:
```suggestion
* `Arc<T>` implements `Clone` whether or not `T` does. It implements `Send` and `Sync` if
  and only if `T` implements them both.
```

mgeisler: Okay, that's simpler indeed.

qwandor (approved):

PR #1052, opened 2023-08-03
Expand "if" to "if and only if" in pl031.rs

mgeisler:
Similar to #1051.

### DIFF
--- a/src/exercises/bare-metal/rtc/src/pl031.rs
+++ b/src/exercises/bare-metal/rtc/src/pl031.rs
@@ -86,8 +86,8 @@ impl Rtc {
 
     /// Returns whether there is currently an interrupt pending.
     ///
-    /// This should be true iff `matched` returns true and the interrupt is
-    /// masked.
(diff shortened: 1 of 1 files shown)


src/exercises/bare-metal/rtc/src/pl031.rs
```diff
@@ -86,8 +86,8 @@ impl Rtc {
 
     /// Returns whether there is currently an interrupt pending.
     ///
-    /// This should be true iff `matched` returns true and the interrupt is
-    /// masked.
+    /// This should be true iff (if and only if) `matched` returns true and the
```
qwandor, line 89: Again, let's omit "iff" entirely and just write "if and only if".

mgeisler: Okay, sure:

```suggestion
    /// This should be true if and only if `matched` returns true and the
```

qwandor (approved):

PR #1053, opened 2023-08-05
ja: Translate chapter 48 (Channels)

kantasv:
Hi, this is a ja translation ( google#652 ) for the chapter 48 (Channels). Could you review and suggest changes if necessary? Thank you!

**Note:** I left some `msgid`s blank, because those lines are outdated in the original EN content. IMO, we can update the original content in the separate MR in the JA translation project :) 

**Example** of updates in the original content which should be addressed in a separate MR:
`po/ja.po`
https://github.com/google/comprehensive-rust/blob/c23da457ae62815477454290dde86d5b598c28df/po/ja.po#L13842
`src/concurrency/channels/bounded.md`
https://github.com/google/comprehensive-rust/blob/c23da457ae62815477454290dde86d5b598c28df/src/concurrency/channels/bounded.md?plain=1#L3


po/ja.po
```diff
@@ -13798,14 +13800,18 @@ msgid ""
 "or\n"
 "  `Receiver` is dropped and the channel is closed."
 msgstr ""
+"* `mpsc` は Multi-Producer, Single-Consumerの略称です。 `Sender` と `SyncSender` から `Clone`が実装"
+"される（よって複数のproducerが作成可能）のですが、`Receiver`についてはそうではありません。\n"
```
keiichiw, line 13804: 「`Sender`と`SyncSender`は`Clone`を実装している」

This sentence means that you can find `impl Clone for Sneder` [here](https://doc.rust-lang.org/std/sync/mpsc/struct.Sender.html#impl-Clone-for-Sender%3CT%3E).

keiichiw (approved):
LGTM % comments

kantasv:
Hi @keiichiw , thank you for your feedback as always -- a great learning opportunity for me!
I've added fix regarding the comments.  Btw, `LGTM % comments` does it mean "Looks good except comments?"

keiichiw:
Thanks for the fix!

Ah, yeah. I meant LGTM once the comments are resolved, but it might not be a so correct phrase.

PR #1054, opened 2023-08-05
ja: refresh original EN content, google#652

kantasv:
Hi, recently I noticed a lot of content in `ja.po` have been outdated as original english content have been updated a lot. I updated the original conten in `ja.po`. Could you review the diff? google#652

Unfortunately, the diff is so large. I'm not sure if the `msmerge` command provides consistent output in different platforms. In my case, I use MacBook OS Monterey, FYI. 

The diff on `ja.po` comes from the following command, as instructed by [TRANSLATING.md]:(https://github.com/google/comprehensive-rust/blob/c23da457ae62815477454290dde86d5b598c28df/TRANSLATIONS.md?plain=1#L108C6-L108C6).

```
msgmerge --update po/xx.po po/messages.pot
```


mgeisler (approved):
Hi @kantasv, thanks for putting this up!

The diff looks good — what you see is that `msgmerge` normalizes the PO file to a standard line width and that is refreshes all the source locations. The rate of churn will go down as we make fewer and fewer changes to the source material — and it will also go down if you run `msgmerge` more frequently.

Since this is a mechanical change, we can just commit it and then make smaller PRs afterwards which remove the `fuzzy` markers and fix the translations.

kantasv:
Hi, thanks @mgeisler ! Okay, then can you or @chikoski , or @keiichiw merge this diff? (since I'm not a code owner of the original repo. )

keiichiw (commented):
Thanks! LGTM

Issue #823, opened 2023-06-15
Add examples of how to use CXX


mgeisler:
We need to expand the chapter on [interoperability with C++](https://google.github.io/comprehensive-rust/android/interoperability/cpp.html) and add concrete examples of how to use CXX.

sakex:
I used Cxx heavily in the past, but that was some time ago, so I might have forgotten some details. It is a pretty complex crate and that does two very complex things:

1. It provides engineers with a set of Rust and C++ classes and methods that map to their counterparts in the other language (for instance `std::vector` in C++ becomes `CxxVector` in Rust.)
2. It generates C++ headers from Rust sources.

This is an amazing tool that vastly facilitates working with FFIs between Rust and C++. However, it is also complex to use and to set up. I think we should definitely cover this subject in detail as teams that are already using C++ and are looking to move to Rust will greatly profit from this tool. However, given its complexity, I think we should reconsider the scope to allocate to Cxx.

I tried to think of a list of quirks of using Cxx I could remember:

- It doesn't handle building the program for you so you will need to edit build.rs and build/link your c++ libraries yourself
- You cannot own most C++ types in Rust because of the destructor
- std::vector (CxxVector) cannot be owned from Rust unless in a unique_pointer because of the unstable ABI.
- Maps, Option, Sets, Arc, etc are not supported
- No generic methods or async
- No overloading
- Need to define exported Rust types and imported C++ types in a bridge module with a proc macro at the top so they can be translated to C++ headers
- Some derive macros are supported (serde for instance, which was made by the same author)


PS: Google also have a crate built on top of Cxx called [AutoCXX](https://github.com/google/autocxx) that I don't know much about but could be interesting to investigate.

mgeisler:
> However, given its complexity, I think we should reconsider the scope to allocate to Cxx.

When I've been teaching the class, I normally spend ~45 minutes on the crate by going through the tutorial (via the documentation, not via code) and by talking about the mapping between types.

The section is part of the Android class since we expect Android engineers to need it first. I feel we could in principle pull the interoperability topics out of the Android class and instead make them their own class. 

> * It doesn't handle building the program for you so you will need to edit build.rs and build/link your c++ libraries yourself

That is mitigated by the Android build system where you just [copy some build definitions](https://source.android.com/docs/setup/build/rust/building-rust-modules/android-rust-patterns#rust-cpp-interop-using-cxx) and then things should work out of the box.

sakex:
[...]

PR #1060, opened 2023-08-10
Add speaker notes for CXX page

mgeisler:
While we don’t have slides showing how to use the crate (#823), we can at least explain at a high level what the instructor should focus on when walking through the tutorial.


randomPoison (commented):
I can't approve because I'm not a collaborator on the repo yet but LGTM 👍

mgeisler:
> I can't approve because I'm not a collaborator on the repo yet but LGTM +1

Thanks, I've added @sakex who has opinions on CXX as well :slightly_smiling_face:

PR #1061, opened 2023-08-10
Tweaks from student feedback

djmitche:
* Improve guidance for the Luhn exercises -- many students either missed the "right to left" bit, or got bogged down with all of the iterator methods and accomplished nothing in the allotted time.
* Show an example of a method-specific type.

### DIFF
--- a/src/exercises/day-2/luhn.md
+++ b/src/exercises/day-2/luhn.md
@@ -18,7 +18,7 @@ following to validate the credit card number:
 
 Copy the code below to <https://play.rust-lang.org/> and implement the function.
 
-Try to solve the problem the "simple" way first, using for loops an integers.
+Try to solve the problem the "simple" way first, using `for` loops and integers.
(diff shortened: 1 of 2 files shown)


src/exercises/day-2/luhn.md
```diff
@@ -16,8 +16,10 @@ following to validate the credit card number:
 
 * The credit card number is valid if the sum ends with `0`.
 
-Copy the following code to <https://play.rust-lang.org/> and implement the
-function:
+Copy the code below to <https://play.rust-lang.org/> and implement the function.
+
+Try to solve the problem the "simple" way first, using for loops an integers.
```
mgeisler, line 21: It seems we normally put the `for` as code:

```suggestion
Try to solve the problem the "simple" way first, using `for` loops and integers.
```

mgeisler (approved):
Thanks!

Issue #1040, opened 2023-07-29
Latest mdbook and mdbook-svgbob are not compatible


star-du:
I follow the instruction to install `mdbook` and `mdbook-svgbob` in README. 
However, as of today (July 29th), when I run `cargo install mdbook mdbook-svgbob`, it reports:
```
error[E0277]: the trait bound `Cfg: From<&toml::map::Map<std::string::String, toml::value::Value>>` is not satisfied
  --> /Users/star/.cargo/registry/src/index.crates.io-6f17d22bba15001f/mdbook-svgbob-0.2.1/src/preprocessor.rs:58:42
   |
58 | ...     .and_then(|map| map.try_into().map_err(|err| error!("{}", err))....
   |                             ^^^^^^^^ the trait `From<&toml::map::Map<std::string::String, toml::value::Value>>` is not implemented for `Cfg`
   |
   = note: required for `&toml::map::Map<std::string::String, toml::value::Value>` to implement `Into<Cfg>`
   = note: required for `Cfg` to implement `TryFrom<&toml::map::Map<std::string::String, toml::value::Value>>`
   = note: required for `&toml::map::Map<std::string::String, toml::value::Value>` to implement `TryInto<Cfg>`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `mdbook-svgbob` (bin "mdbook-svgbob") due to previous error
error: failed to compile `mdbook-svgbob v0.2.1`, intermediate artifacts can be found at `/var/folders/v5/36d3013s7f56pv4d4kpftd5c0000gn/T/cargo-install2cKq85`
     Summary Successfully installed mdbook! Failed to install mdbook-svgbob (see error(s) above).
error: some crates failed to install
```

It seems this version of `mdbook-svgbob` does not like my `mdbook v0.4.32`, which is the version in [crates.io](https://crates.io/crates/mdbook) registry.

I see we are using `--locked` flag in `.github/workflows/install-mdbook/action.yml` to specify the desired mdbook version (v0.4.28), which works for me. Should we update the README.md as well, so new users will not be frustrated by similar build failures?

star-du:
It's also mentioned in another issue in `mdbook-svgbob` [here](https://github.com/boozook/mdbook-svgbob/issues/25) , but sadly the discussion there was not immediately helpful to me.

mgeisler:
Hi @star-du,

[...]

PR #1066, opened 2023-08-11
Update mdbook to 0.4.34

mgeisler:
I also updated `book.js` and `index.hbs` to the latest versions while keeping our local modifications.

I tested this locally with `mdbook serve` for both English and Brazilian Portuguese and everything seems to work fine.

Inspired by #1040.


djmitche (approved):
[RUBBER-STAMP ✔️]

mgeisler:
I'll merge this now and give is a second check before stopping for today...

mgeisler:
The site looks normal so I think we're good :slightly_smiling_face:

PR #1073, opened 2023-08-11
Comprehensive Rust v2

### LINKED ISSUE #1082: Create more and better exercises
One of the more consistent asks from course participants is more exercises. People like the interactive form, but they also want to play with Rust themselves.

The current exercises have some problems:

- They sometimes rely on content not fully covered at the time (Day 1 afternoon exercises uses methods which are not fully introduced at this point)
- They sometimes ask you to solve a problem where the problem itself is challenging (the multi-threaded link checker in the Concurrency section is more of a mini-project than an exercise).

To solve the above, we should write new exercises. The exercises should be trivial: the participants should be able to come up with a correct solution in 30 seconds — the challenge should be "How to implement this in Rust?" The Luhn exercise is a good example: it's easy to see how to solve it in a C-like way where you just iterate through the characters one by one and ignore most error cases. So students can start with that version and iterate on it to make it nicer.

Exercises can also be less free-form than the current exercises. If we provide more scaffolding for the students, we can guide them better and ask them to implement a bigger task. This is something we should do for the link checker exercise. We should provide most of the solution and just let students fill in 20-30 missing lines. That reflects how engineers typically work: you read a lot more code than you write and you are often asked to work in an existing system.

### LINKED ISSUE #1465: Version 2 reshuffle and refresh
@djmitche has been working on a major reshuffle and refresh of the course for a few months now: #1073.  The branch is getting to a state where we need to think about how to land this.

The branch is

- Adding new content
- Moving existing content
- Removing some existing content

From a read through, I think this is a significant improvement. It covers things in a much better order and it gives many new details!

There are some things I believe we need to address to merge it:

- How does this interfere with existing classes organized at Google? We have multiple overlapping classes right now, with the next open slot being Nov 22nd.
  - I feel we need to run this with real participants before we merge it. I would prefer an in-person class since those have more engagement in my experience.
  - We need to align merging this with the existing instructors.
- How does this interact with with translations?
  - #1463: I tried to summarize the impact there. The v2 branch touches around 500 of the ~2k messages in the course.
  - Because of the snapshot publishing system, it's should be okay to merge the branch.
  - Overall, I'm impressed how much of the existing text has been preserved!
- How does this interact with existing pages?
  - #489: We don't have a good understanding of the number of 404 hits we have on the course.
  - #488: Nor do we know how ofte the redirection pages are used.
  - I'm afraid that we won't notice problems here if we just land the branch without data on this.
- Size and style of the new slides
  - #1464: I tried to write down my thoughts there.
  - I'm worried about being able to reign this in if we suddenly let several of the slides become 50% larger. When I've talked with people about this in the past, I've sometimes been met with a "well it fits on my screen"-style reply. Which doesn't solve the actual problem I observe in class and leaves me to pick up the pieces.

djmitche:
I've taken some work by @fw-immunant and others on the new organization of the course and condensed it into a form amenable to a text editor and some computational analysis. You can see the inputs in `course.py` but the interesting bits are the output: `outline.md` and `slides.md`.

The idea is to break the course into more, smaller segments with exercises at the ends and breaks in between. So `outline.md` lists the segments, their duration, and sums those durations up per-day. It shows we're about an hour too long right now! There are more details of the segments in `slides.md`, or you can see mostly the same stuff in `course.py`.  

This now contains all of the content from the v1 course, ensuring both that we've covered everything and that we'll have somewhere to redirect every page.

I'd appreciate feedback on:
 * The order of topics, at the level of segments (don't get too deep yet, this is still very rough)
 * Whether the 1080-minute time limit is appropriate
 * If so, what we might think of cutting (about 5%)

Fixes #1082.
Fixes #1465.


djmitche:
This PR now contains the "new" course in `src/`, as derived through some automation. That automation is repeatable, and we should be able to rebase with it fairly easily, even where `git` doesn't see a file as having moved.

There's still a lot of editing to do, and I'll start in on that from day 1 forward.

djmitche:
I've got Day 1 morning and a bit of the afternoon ("Tuples and Arrays") edited together now. So far, rebasing hasn't been bad, as the upstream changes have been relatively straighforward and the sequence of commits in this PR helps guide Git to carry those changes through to the final product.

*However*, I'm starting to lose the faith that this is going to work out:
 * This PR is essentially un-reviewable, except by loading up the result in `mdbook serve` and flipping through it.
 * I'm the only one looking at this, so it's certain to be of a rough level of quality at best, and need some work before being taught
 * This is almost certainly going to be a nightmare for translators

I can keep working on this, and get to the end, but I'm worried that either it will be merged with "@djmitche did a lot of work so let's just merge it" resulting in a poor quality course until we can collectively improve it; or that review will take a long time and we'll end up basically forking the course.

I can see a few fixes for this:
 * Maybe it's just OK and I should just keep working.
 * Maybe we can find a way for a few people to collaborate so that the quality is higher when it lands.
 * Maybe we should embrace the fork and eventually land this in a `v2` branch that we can teach from and improve for a while before making the switch.

Thoughts?

djmitche:
Now rebased onto main, and with all of day-1 complete. That was probably ~8 hours work (mostly on planes to/from RustConf), so it could be a while until I finish the other two days. That said, days 2 and 3 changed less, so perhaps a little faster.

djmitche:
Regarding translations, I think that should be "not too bad" -- I've tried to not unnecessarily rewrite text that is appropriate to the new slide. So I would hope that many of the translated messages carry over without issue. Maybe as the work here finishes (with all three days complete) we can look at the number of mismatches, fuzzy matches, and exact matches and make a plan from there.

src/hello-world/variables-and-values.md
```diff
@@ -1,4 +1,25 @@
...
+```rust,editable
+fn main() {
+    let x: i32 = 10;
+    println!("x: {x}");
+    // x = 20;
+    // println!("x: {x}");
+}
+```
+
+## Scalar Values
```
fw-immunant, line 22: This heading seems wrong for a table of types and literals. Maybe we should make this slide "Variables, Values, and Types"?

src/hello-world/variables-and-values.md
```diff
@@ -18,6 +39,12 @@ The types have widths as follows:
 
 <details>
 
+Uncomment the `x = 20` to demonstrate that variables are immutable by default.
+Add the `mut` keyword to allow changes.
+
+Due to type inference the `i32` is optional. We will gradually show the types
+less and less as the course progresses.
```
fw-immunant, line 46: The `i32` fallback isn't really inferred from anything in the context here, but is the fallback that Rust's type inference uses for integral types when they're underconstrained after inference. A discussion of inference is worthwhile but should probably make this distinction and mention the f64 fallback for floats, especially as the two differ in byte size which may be surprising.

src/hello-world/arithmetic.md
```diff
@@ -0,0 +1,39 @@
...
+minutes: 5
+---
+
+# Arithmetic
+
+```rust,editable
+fn main() {
+    let a: i32 = 12;
+    let b: i32 = 35;
+    let c: i32 = 37;
```
fw-immunant, line 11: Since we've introduced the fallback type, maybe we should use it here to show more idiomatic code.

src/hello-world/arithmetic.md
```diff
@@ -0,0 +1,39 @@
...
+    if a.pow(2) + b.pow(2) == c.pow(2) {
```
fw-immunant, line 12: I think it's better not to introduce methods here if we're trying to focus on arithmetic (where the built-in operators give us plenty to teach in one go, and various interesting math methods like `sin` and `count_leading_zeros` don't quite fall under arithmetic).

src/hello-world/exercise.md
```diff
@@ -0,0 +1,27 @@
...
+
+If you haven't used the site before, you will see a simple hello-world program.
+Try running it.
+
+- Add some variables and print them using the `{..}` notation seen in the
+  lessons.
+- Under "Tools", use the `rustfmt` option to format your code in the "standard"
+  way.
+- If you're interested, use "ASM" under "..." to see the generated assembly
+  code.
```
fw-immunant, line 27: I'm not sure this exercise has enough of a goal condition to be productive for most students. It's good to introduce the playground, but without a purpose the students don't get a sense of achievement and may not have anything they particularly want to try. If we want to cover formatting syntax in detail or demonstrate playground features, those seem like things that are better to do as directed teaching than hoping students stumble upon them in an open-ended exercise.

src/language-features/casting.md
```diff
@@ -0,0 +1,28 @@
+---
+minutes: 3
+---
+
+# Casting
+
+Rust has no _implicit_ type conversions, but does support explicit casts with
+`as`. These generally follow C semantics where those are defined.
```
fw-immunant, line 8: `as` conversions are a pretty common source of bugs and are discouraged where possible. It would be good if we could introduce these at the same time as the Into/From/TryInto traits so we can discuss best practices. At the very least we should warn that `as` can change the meaning of values (`0xffu8 as i8`) and lose data (`256u16 as u8`).

src/SUMMARY.md
```diff
@@ -12,178 +12,152 @@
...
+  - [Functions](control-flow-basics/functions.md)
+  - [Useful Macros](control-flow-basics/macros.md)
+  - [Exercise: Collatz Sequence](control-flow-basics/exercise.md)
+    - [Solution](control-flow-basics/solution.md)
+- [Language Features](language-features.md)
+  - [Type Inference](language-features/inference.md)
+  - [Casting](language-features/casting.md)
+  - [Documentation](language-features/docs.md)
+  - [Static and Const](language-features/static-and-const.md)
+  - [Compiler Lints and Clippy](language-features/lints.md)
```
fw-immunant, line 40: "Language features" sounds like a grab-bag, and this isn't improved by having both things possibly best described as "features" for developers (e.g. docs and lints) grouped with core language semantics (`static`, `const`, type inference, and casting).

Maybe we could move "type inference" and "casting" closer to arithmetic and move "static and const" nearer to "variables and values"? Then this section can focus on developer workflow features.

src/user-defined-types/let-control-flow.md
```diff
@@ -0,0 +1,122 @@
...
+
+# Let Control Flow
+
+Rust has a few control flow constructs which differ from other languages. They
+are used for pattern matching:
+
+- `if let` expressions
+- `while let` expressions
+- `match` expressions
+# `if let` expressions
```
fw-immunant, line 17:
```suggestion
# `let-else` expressions
```

We've already introduced `match`, and presumably the commented-out case here should be about let-else which is something of a minor complication to `let` that makes sense after seeing if-let nesting's rightward drift.

src/std-types/exercise.md
```diff
@@ -0,0 +1,56 @@
...
+minutes: 5
+---
+
+# Exercise: Hash Set
+
+In this exercise you will build a very simple hash set that stores `u32` values.
+The hash set will have a fixed size, and buckets should be selected with a
+simple modulus operation (`i % num_buckets`). Use a `Vec` to represent each
+bucket (separate chaining). Hint: a vector can contain other vectors:
+`Vec<Vec<u32>>`.
```
fw-immunant, line 11: I don't think this is quite the right level of exercise to test/develop knowledge of the stdlib. We haven't really explained generics, ownership, or borrowing yet, and students coming from high-level languages may not be familiar with terminology for hash table implementations like "buckets" and "chaining".

What specific takeaways do we want students to derive from this exercise?

fw-immunant (commented):
Reviewed day 1; lots of comments inline.

mgeisler:
> Regarding translations, I think that should be "not too bad" -- I've tried to not unnecessarily rewrite text that is appropriate to the new slide. So I would hope that many of the translated messages carry over without issue.

Awesome, that is a relief to hear! I have a suggestion here: we should extract the messages from this branch ~weekly and `msgmerge` the result into the existing translations (or into the most active translations). That way @henrif75 can ask our Localization team (and community translators) to translate the messages while you work with @fw-immunant and others on the content.

I'm reluctant to add special code for this into `mdbook-xgettext` since the translation infrastructure is used by multiple projects now. But we could update all translations weekly and tell people not to do further `msgmerge` runs themselves.

src/hello-world/exercise.md
djmitche: I replaced this with a slide introducing the playground. It gives an instructor a chance to encourage students to keep it open and noodle around during the course, which I think helps students learn and ask good questions.

sakex:
Hey Martin, I'm waiting on the review of the latest changes 😅

mgeisler:
> Hey Martin, I'm waiting on the review of the latest changes 😅

Oh, I was not aware! I added comments now.

djmitche:
OK, I _think_ this is basically ready for review. I'll put some more comments to guide review on Monday after having had a look myself to find and silly mistakes.

djmitche:
OK! I think this is ready for review!

I'd like to take a strong bias toward landing this and following up. The `NOTES.txt` file can turn into a punch-list for that followup, so feel free to suggest changes to that file for things you'd like on the list (and it's already seeded with a bunch!). The things we need to identify now are things that would either be hard to fix in a followup, or things that would make the course impossible to teach in its current form.

I suspect the best way to do this review is to use `mdbook serve` and page through the 3-day course in order, checking for missing pieces, bad concept ordering, or incorrect information. Please try to avoid editing the content -- this was hard for me to avoid in the reorganization process, too! For any pages that need editing, let's make a list of those in NOTES.txt and do so in followup PRs.

I'd like to land this by Nov 6, so please pace your reviews accordingly. Of course, if this is a disaster we can change that date :)

src/android/interoperability/cpp/overview.svg
mgeisler, line 1: This was a symlink which accidentally got converted to an image.

src/smart-pointers/box.md
```diff
@@ -0,0 +1,115 @@
...
+minutes: 10
+existing course material:
+- std/box.md
+- std/box-recursive.md
+- std/box-niche.md
+---
+
+# Box<T>
+
+# `Box`
```
mgeisler, line 11: This seems to be double:
```suggestion
# `Box<T>`
```

src/android/bpfmt.sh
mgeisler, line 1: This script seems to have lost its execute bit?

src/android/build_all.sh
mgeisler, line 1: Same here with this script.

count.dat
mgeisler, line 1: What is this mysterious file for? :smile:

src/methods-and-traits/welcome.md
mgeisler, line 1: Duplicates `methods-and-traits.md`.

src/memory-management/welcome.md
mgeisler, line 1: Duplicates `memory-managemetn.md`

timClicks:
I am very late to this PR, but wanted to add some general observations before reviewing the changes in depth. This comment is made _without_ seeing the change, but with a high degree of enthusiasm for the course.

Hopefully it'll be useful to read a few points based on my [experience teaching the class recently](https://github.com/accelerant-dev/comprehensive-rust/tree/2023-10) and the changes that I made to the core repository. 

One thing to note is that my course spans 6 days, with 4h each session. With 24h of teaching over 2 weeks, there's a lot of opportunity to spend more time doing exercises and discussing problems/ideas with learners.

## Interactivity and dynamism is essential

One thing I love about this course is that its material is designed to be adjusted on the fly. You can see here I describe a way to make the suggested solution code more general purpose by showing learners how to explore the official Rust documentation for the `char` type and its `is_whitespace()` method: 

Implementing the Luhn algorithm

Accordingly, version 2 should be _too_ prescriptive. If the format gets too locked down, then it will lose some of its magic. 

## Less emphasis on C++

Much of the current material makes reference to prior knowledge of C++, but that wasn't appropriate for my cohort. 

## Changes I made

In general, I removed lots of the text from what's shown on screen and moved it into the speaker notes. That way it's available for reference, but learners don't get intimidated by walls of text.

You can also see that I tinkered with the theme a little bit.

default


## More examples, more frequently

I've added several examples and inter-mixed them with text to break up the day. That way the course is currently structured is to present a block of material, then move to a large example. Here's a snippet from my repository's `SUMMARY.md`:

```md
- [Welcome](welcome-session-2.md)

- [Recap](recap-session-2.md) <!-- CLI app -->

- [Structs](structs/structs.md)
  - [Tuple Structs](structs/tuple-structs.md)
  - [Field Shorthand Syntax](structs/field-shorthand.md)

- [Methods](methods.md)
  - [Method Receiver](methods/receiver.md)
  - [Example](methods/example.md)

- [Exercise: Thumbnail Generator](exercises/session-2/thumbnail-generator.md)

```

## Adding space for non-coding activities

The course notes should probably include space in the agenda for other exercises. The mode shift enables learners to decompress a bit.

Here's an example of our group looking up the standard library for text types and then placing them into a Venn diagram of traits.

text-types-in-rust

## Recaps

At the start of each day, I added a recap exercise which covered most of the material that we touched on the day earlier. I'm hoping this provided a bit of spaced repetition and also a chance for people to mentally warm up for the day.

mgeisler:
> ## Adding space for non-coding activities
> The course notes should probably include space in the agenda for other exercises. The mode shift enables learners to decompress a bit.
> 
> Here's an example of our group looking up the standard library for text types and then placing them into a Venn diagram of traits.

Oh, this is smart! Breaking the flow and getting people to engage in group activities like this sounds like a great addition!

src/android/bpfmt.sh
djmitche: I reverted all of the changes to src/android -- they were just side-effects of the conversion script.

count.dat
djmitche: This comes from one of the exercises. I thought I had gitignored it!

src/methods-and-traits/welcome.md
djmitche: It doesn't, actually. The welcome talks about the whole day, while the `methods-and-traits.md` is just the content of this first segment. But those are in the wrong order. I'll fix this up to put the per-day welcomes into their own slides (without a corresponding segment).

src/memory-management/welcome.md
djmitche: (same as above)

djmitche:
@mgeisler:

> Afternoon: I would turn 8.5 Type Aliases into a note — but that's partly due to personal preference since I don't use the feature myself.

This is pretty common, so I think students should see it on a slide. It's short :)

> I think people need more time than 15 minutes for the evaluator.

Maybe -- I think all of the exercises will require some fine-tuning with experience. The exercise only requires writing `eval`, which is mostly about pattern matching. So I'm optimistic that students could get something up and running in a reasonable timeframe. I'm hesitant to allocate more time to exercises -- we already have nearly 4 hours of exercises!

> Those days [days 1-2] do use types like String and HashMap which likely will bring up questions about how to deep-copy those values.

I'm interested to see where this starts to trip students up on those days. We can add an extra loop of the spiral with light coverage of move and copy semantics at that time, maybe just suggesting using `.clone()` liberally as necessary. I'm just not sure where that will occur without experience teaching the course.

> The exercises are really ramping up here! I think people will appreciate this, if we give them sufficient time (both should have 30 minutes instead of 15 minutes).

There are six exercises in day 3, so I'm not sure what you mean by "both"?

> I think these [deep-dives] did not change, right?

Correct.

@timClicks:

> Much of the current material makes reference to prior knowledge of C++, but that wasn't appropriate for my cohort.

Yeah, partly that's the history of who wrote the course, but within Google people have backgrounds in various languages, so we should reduce that a bit. I think this PR makes some progress on that reduction.

> In general, I removed lots of the text from what's shown on screen and moved it into the speaker notes

This makes good sense for a presentation of the material, especially with a longer timeframe.  Lots of folks also self-study from the course, though, and I think the split of on-slide content (the main thing to learn from the slide) and speaker notes (explanation, background, side-notes, and more-to-explore) helps those people to focus their learning.  In self-study, it's pretty intimidating to have every slide consist of a text box with some strange-looking code in it!

> Non-coding Activities

I agree, these sound fascinating, and maybe they would be good replacements for a few of the exercises? We would need to find a reliable way to host them. It looks like you have some tools you use?

> Recaps

This is a really good idea.

djmitche:
I've accommodated the above suggestions, and rebased on top of the latest from `upstream/main`.  I added the recaps idea to NOTES.txt, which will be the source of a bunch of issues for later follow-up.

mgeisler:
> > The exercises are really ramping up here! I think people will appreciate this, if we give them sufficient time (both should have 30 minutes instead of 15 minutes).
> 
> There are six exercises in day 3, so I'm not sure what you mean by "both"?

Good point :smile: I was mostly thinking of _Exercise: Protobuf Parsing_ and _Exercise: Rewriting with Result_. They just seem big because of the setup code.

It normally takes 30-45 minutes for people to solve the old FFI exercise. Of course, the FFI exercise is strange in its own right: it requires a lot of research and not so much code. I guess your exercises has less of that — but there is a lot of code to read and understand.

I think the exercises are great: it's perfect that we make people read more real-world code! It's something people have been requesting a lot so I think this is a huge leap forward. I just think we should expect it to take 30+ minutes to solve them.

src/iterators/intoiterator.md
```diff
@@ -0,0 +1,86 @@
...
+- exercises/day-2/iterators-and-ownership.md
+---
+
+# `IntoIterator`
+
+The `Iterator` trait tells you how to _iterate_ once you have created an
+iterator. The related trait
+[`IntoIterator`](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html)
+defines how to create an iterator for a type. It is used automatically by the
+`for` loop.
```
mgeisler, line 13: This applies to many of the new pages: most of the old pages (at least those I wrote) followed a scheme of

1. A single line of text to introduce the slide
2. A small code block
3. Perhaps 2-3 lines of extra text before the notes start

You see that in [6.3. References](https://google.github.io/comprehensive-rust/basic-syntax/references.html) and the next few slides.

Wny do they look like this? Becuase that's who I've been writing all my slides for years :smile: 

I don't know if you're aware of the "aspect ratio helper"? Uncomment it in `book.toml` and it'll draw a big rectangle on top of the pages. It shows that code blocks with more than 15 lines are too large — as soon as you execute them and get 1-2 lines of results, then people lose the overview:

image

deavid:
I am reading the book from beginning to end to see if I can spot anything wrong. I'll add here everything I found. These might be nits and okay to fix in a different PR or might need additional discussion later on. These might also exist in the original version too.

6.1. Loops: While is not explained properly on text. "Like other languages" isn't a self-contained explanation. The for loop explanation seems also very superficial, and starting with an inclusive range seems odd. I would prefer to have inclusive and non-inclusive examples, as well as an array too. (for example, just add a comment "// you can also try 0..4 and [3,6,7] here")

6.6. Macros: It doesn't explain what a Macro is. Why println!() is a macro? couldn't it be a function? Also it only talks about function macros but the title suggests Macros in general; so either the title changes to "Function Macros", or we add some light explanation about the other types of macros.

6.7. Exercise: Collatz Conjecture: not everyone has enough math background to understand the problem the way it is defined. An example would be good here.

7.4. Destructuring: There's no definition for the word/concept destructuring. Few programming languages have this. Also in translations we will likely want to add a definition anyway, so it would be better if there's a definition in the source material.

8.1. Shared references: "Rust’s references provide a way to refer to another value" - this definition is cyclical. Also the other part of the phrase compares it to C++. So unless one has prior experience with pointers, the presenter will definitely need to explain a lot here. It would help to have an initial explanation and not assume that people know what a pointer or reference is.
* Speaker notes: "A shared reference does not allow modifying the valueit refers to" -> "valueit" should be two words.
* And in my opinion, the fact that shared references are read-only should be in the slide, not in speaker notes.
8.2. Exclusive References: At no point it is said that we are talking about &mut, it has to be guessed from the example.

9.3. Enums: says 5 minutes, but enums in Rust are special. This will likely be more like 15 minutes. Also, there's lots of developers that do not have that much experience with enums in general; either because their language doesn't support it, or because they're not that needed for what they typically code.

9.4. Deriving: What is a trait? What is an implementation? This section comes before explaining these. Also it doesn't mention that these are macros too.

9.5. Type Aliases: The example shows how to declare, but on how it looks in practice is not clear enough (there is a use of one type alias but it's hidden in a complex type). Ideally I would add some function signatures or custom structs that showcase the cleanliness of this.

9.7. Let Control Flow: The example for let-else, isn't an actual example of let-else. It's showcasing a regular if let with regular else blocks. I also feel that Result<T,E> and Option<T> should have been introduced by now and I don't recall reading it so far.

I'm gonna stop here, this is for day 1. Will continue later for day 2.
As I said above, I'm not sure which of these problems were introduced on this PR, or even whether they are actual problems or if it was done like that purposefully.

count.dat
mgeisler: Thanks for cleaning it up!

src/references/shared.md
```diff
@@ -0,0 +1,64 @@
...
+* Rust does not automatically create references for you - the `&` is always
+  required.
+
+* Rust will auto-dereference in some cases, in particular when invoking
+  methods (try `r.count_ones()`). There is no need for an `->` operator
+  like in C++.
+
+* In this example, `r` is mutable so that it can be reassigned (`r = &b`).
+
+* A shared reference does not allow modifying the valueit refers to, even if
```
ahomescu, line 54:
```suggestion
* A shared reference does not allow modifying the value it refers to, even if
```

src/references/shared.md
```diff
@@ -0,0 +1,64 @@
...
+  be much smaller than the thing they point to.
+
+* Rust does not automatically create references for you - the `&` is always
+  required.
+
+* Rust will auto-dereference in some cases, in particular when invoking
+  methods (try `r.count_ones()`). There is no need for an `->` operator
+  like in C++.
+
+* In this example, `r` is mutable so that it can be reassigned (`r = &b`).
```
ahomescu, line 52: Minor suggestion (please ignore if it's not what you're looking for): might be useful to mention that this re-binds the reference (`r` becomes a pointer to something else), in contrast to C++ where an assignment to a reference changes the value of the pointee.

src/hello-world/example.md
```diff
@@ -62,17 +68,17 @@ Despite just 29 lines of code, this C example contains serious bugs in at least
 9. Unchecked cases in the `switch` statement (line 11)
 10. Unchecked return values of `stat` and `fopen` (lines 18 and 26)
 
-_Shouldn't these bugs be obvious even for a C compiler?_  
+_Shouldn't these bugs be obvious even for a C compiler?_
 No, surprisingly this code compiles warning-free at the default warning level, even in the latest GCC version (13.2 as of writing).
 
-_Isn't this a highly unrealistic example?_  
+_Isn't this a highly unrealistic example?_
 Absolutely not, these kind of bugs have lead to serious security vulnerabilities in the past. Some examples:
```
ahomescu, line 75:
```suggestion
Absolutely not, these kind of bugs have led to serious security vulnerabilities in the past. Some examples:
```

ahomescu: This typo is not in any of your changes, so feel free to ignore it.

deavid:
I'm continuing on the review of Day 2. Same disclaimer as before: unsure if these are relevant or specific to the V2.

11.3. Trait Objects: Talks about memory management and Box<T> but that's on Day 3 morning.
13. Standard Library Types: I would think if this can fit in Day 1 instead. There's no need for generics to understand this. Suffice to say that Vec<String> is a Vec of strings and Option<i32> is an optional integer.

Day 3:

19. Memory Management: I'm not particularly convinced that this belongs to Day 3. I'm sure a lot of people will be having problems with ownership from Day 1.
24.3. Static and Const: It's inside the "Unsafe" chapter. People skimming trough the materials might get into the impression that Static and Const are unsafe constructs.

And with that I finished reviewing the book up to Day 3. I get the impression that Traits and Standard Library traits could be moved for later, while Standard Library Types and Memory Management could be earlier.

djmitche:
@mgeisler :
> I think the exercises are great: it's perfect that we make people read more real-world code! It's something people have been requesting a lot so I think this is a huge leap forward. I just think we should expect it to take 30+ minutes to solve them.

Let's gather some data on this. We can start by adjusting the time for the exercises based on that data, and then look at how that affects the course schedule. We can then either scale back or replace the exercises, or remove other material from the course.

Regarding the aspect ratios:

I think this will make a good follow-up. I'll add it to NOTES. We've discussed the possibility of using more, simpler slides.

@deavid:
> 6.1. Loops: While is not explained properly on text. "Like other languages" isn't a self-contained explanation. The for loop explanation seems also very superficial, and starting with an inclusive range seems odd. I would prefer to have inclusive and non-inclusive examples, as well as an array too. (for example, just add a comment "// you can also try 0..4 and [3,6,7] here")

Done

> 6.6. Macros: It doesn't explain what a Macro is. Why println!() is a macro? couldn't it be a function? Also it only talks about function macros but the title suggests Macros in general; so either the title changes to "Function Macros", or we add some light explanation about the other types of macros.

I added a speaker note. This slide is early on the first day, so I'd rather be brief.

> 6.7. Exercise: Collatz Conjecture: not everyone has enough math background to understand the problem the way it is defined. An example would be good here.

Done
 
> 7.4. Destructuring: There's no definition for the word/concept destructuring. Few programming languages have this. Also in translations we will likely want to add a definition anyway, so it would be better if there's a definition in the source material.

Done.
 
> 8.1. Shared references: "Rust’s references provide a way to refer to another value" - this definition is cyclical. Also the other part of the phrase compares it to C++. So unless one has prior experience with pointers, the presenter will definitely need to explain a lot here. It would help to have an initial explanation and not assume that people know what a pointer or reference is.

Done. I think instructors will need to approach this a few ways anyway. And this is only the first pass on references -- day 3 covers them in much more depth.

> * Speaker notes: "A shared reference does not allow modifying the valueit refers to" -> "valueit" should be two words.
> * And in my opinion, the fact that shared references are read-only should be in the slide, not in speaker notes.

Done.

>   8.2. Exclusive References: At no point it is said that we are talking about &mut, it has to be guessed from the example.

Done

> 9.3. Enums: says 5 minutes, but enums in Rust are special. This will likely be more like 15 minutes. Also, there's lots of developers that do not have that much experience with enums in general; either because their language doesn't support it, or because they're not that needed for what they typically code.

Let's see how this turns out. This isn't the only slide on enums, it just happens to have accumulated a ton of speaker notes.
 
> 9.4. Deriving: What is a trait? What is an implementation? This section comes before explaining these. Also it doesn't mention that these are macros too.

I'll change it to "behaviors", and expand the speaker notes a bit. This is another example of "spiraling", where we use a concept lightly, then add detail later.
 
> 9.5. Type Aliases: The example shows how to declare, but on how it looks in practice is not clear enough (there is a use of one type alias but it's hidden in a complex type). Ideally I would add some function signatures or custom structs that showcase the cleanliness of this.

I'm not sure what you mean here. Let's leave it to a followup and you can make a PR.  I've added it to NOTES.

> 9.7. Let Control Flow: The example for let-else, isn't an actual example of let-else. It's showcasing a regular if let with regular else blocks. I also feel that Result<T,E> and Option should have been introduced by now and I don't recall reading it so far.

Check the speaker notes :)


> 11.3. Trait Objects: Talks about memory management and Box but that's on Day 3 morning.

Added to NOTES.txt.

> 13. Standard Library Types: I would think if this can fit in Day 1 instead. There's no need for generics to understand this. Suffice to say that Vec is a Vec of strings and Option is an optional integer.

An earlier revision moved this out to day 2. There's a cycle here, we have to break it somehow, and the preferred solution is "spiraling", where we address something partially to start, and then circle back with a fuller description later.  This is the "fuller" description of these standard-library types, after they've been used more lightly on day 1. If there are spots in day 1 where we rely too heavily on the details of these types, or where a simple/partial explanation could help,  we should make that change.

> 19. Memory Management: I'm not particularly convinced that this belongs to Day 3. I'm sure a lot of people will be having problems with ownership from Day 1.

This course is deeply intertwined, and it's tempting to say that everything should be on day 1, but then day 1 lasts 3 days! We moved this to the end in hopes of getting folks up and running with the language before encountering the more difficult cases around borrow checking, ownership, and so on. Partly that's by using a lot of copy types, and partly by just using things in a way where ownership flows naturally down the stack. As we teach the course, if we find places that students get hung up on ownership, we should try to avoid that rather than dig in too early on ownership.

>     24.3. Static and Const: It's inside the "Unsafe" chapter. People skimming trough the materials might get into the impression that Static and Const are unsafe constructs.

Yeah, @fw-immunant and I identified a chapter as a "grab-bag" and this was the hardest slide to move out of that grab-bag chapter.  I couldn't figure out where else these fit naturally, and i put them here since you do need unsafe to modify a static. What do you think?

I've force-pushed to the PR because I rebased it on the latest `main`, but the updates from above are in the last commit.

randomPoison:
I've been reviewing the v2 version. I've got miscellaneous bits of feedback, but nothing that should block this PR. Personally I'm in favor of getting this merged in and doing further iteration through regular PRs, rather than continuing to give large piles of feedback on this one PR.

@djmitche how are you feeling about this? Are there any major changes that you're still planning to make?

mgeisler:
> @djmitche how are you feeling about this? Are there any major changes that you're still planning to make?

@randomPoison, would you be able to run your Nov 15 class using the new material? We'll email people to let them know that "Hey! :wave: You're getting a preview of the new cool v2 of the course!" :slightly_smiling_face: 

I think people will be okay with that and they'll know that the delivery will be a bit unrehearsed.

djmitche:
That's great to hear! No, no major changes are planned. The NOTES.txt in this PR indicate some minor changes planned for followup. For the most part, I think we need more data, so getting the PR merged and the course taught would be great. If you'd like to put your feedback into NOTES.txt, that'd be fine, but we could also wait and record it in the form of PRs or issues after merge.

And, teaching that in a "beta test" mode so students feel encouraged to provide feedback would be even better. In particular, I think we'll need students to give detailed feedback on the exercises -- too hard? too fast? etc. If they know they are beta testers, they may be less likely to assume they "just don't get it" when the exercise takes them 2x as long as I thought it would.

randomPoison:
> @randomPoison, would you be able to run your Nov 15 class using the new material? We'll email people to let them know that "Hey! 👋 You're getting a preview of the new cool v2 of the course!" 🙂

Yeah, I'm down to do that. Do you want to wait to merge this until after that, though? I don't think we should wait, personally. 

> If you'd like to put your feedback into NOTES.txt, that'd be fine, but we could also wait and record it in the form of PRs or issues after merge.

I'm leaning towards just making PRs after this is merged. There's a couple of minor ideas for reordering slides, but otherwise it's mostly minor improvements to wording or exercises, so not all of it is even v2-specific. Getting this merged in is a higher priority than trying to make sure we've gathered all feedback, imo.

deavid:
I would also vote to get this merged as is. If we keep patching it the translations will get out of sync more and more, and once merged they will get more broken as we fix more stuff here.

fw-immunant:
So, I have a lot of tiny nitpicks throughout, but I agree with @randomPoison that those needn't be blockers. On the other hand, after reviewing the material from Day 2 and Day 3 I have some concerns about ordering and timing:
- Day 2 morning covers methods and traits, generics, standard library types, and standard library traits. I usually spend a significant amount of time on each std type we cover, touring the API and pointing out idiosyncrasies of the API and of how Rust treats these concepts. After each slide I visit the std docs and highlight methods that developers should know, either because they are common or because they are unusual. I also go into the way many of these types "inherit" many of their methods via Deref, e.g. the way `String` obtains many methods from `Deref<Target=str>` and `Vec<T>` from `Deref<Target=[T]>` I do the same tour for the std traits, though admittedly there's generally less to say about traits than the types, so this goes a bit faster. Even so, this seems like a *lot* of material for half a day. Concretely, I spend more like 10 minutes on `String`/`&str` than 3, and probably 8 or so on `Vec<T>`/`&[T]`. Option is also longer than 3 minutes as it's worth covering the various combinator methods and why `as_ref`/`as_mut` are useful, and the ways to convert between Option and Result.
\
I also worry that if students haven't seen lifetimes yet, they will not be equipped to understand many of the type signatures in the stdlib, including ones that we really do want to discuss (such as those on the `Entry` API for `HashMap`--it's only because of borrowing that this API can exist).
- It seems strange for slices and lifetimes to be between error handling and unsafe Rust on the second half of Day 3. Both of the latter are in some sense "nitty gritty" concerns--error handling is important in real-world programs where things can and do go wrong, and unsafe Rust is necessary in situations where safe Rust fails us (in performance overhead or ability to do interop). Slices, meanwhile, are a primitive type provided by the standard library, and lifetimes are used pervasively in the standard library and in any nontrivial code working with borrowed data. I feel like it would be better to frontload lifetimes so that students get more experience with them and know the basics before they see more advanced uses (such as some of those in std).
- One last small ordering nit: I think categorizing `Drop` under "Smart Pointers" isn't really appropriate, but maybe it could just move into the prior section on "Memory Management" as it's related to Copy etc.

djmitche:
> * Day 2 morning covers methods and traits, generics, standard library types, and standard library traits. I usually spend a significant amount of time on each std type we cover, touring the API and pointing out idiosyncrasies of the API and of how Rust treats these concepts. After each slide I visit the std docs and highlight methods that developers should know, either because they are common or because they are unusual. I also go into the way many of these types "inherit" many of their methods via Deref, e.g. the way `String` obtains many methods from `Deref<Target=str>` and `Vec<T>` from `Deref<Target=[T]>` I do the same tour for the std traits, though admittedly there's generally less to say about traits than the types, so this goes a bit faster. Even so, this seems like a _lot_ of material for half a day. Concretely, I spend more like 10 minutes on `String`/`&str` than 3, and probably 8 or so on `Vec<T>`/`&[T]`. Option is also longer than 3 minutes as it's worth covering the various combinator methods and why `as_ref`/`as_mut` are useful, and the ways to convert between Option and Result.

This seems to be a frequently-raised point here. But I'm not sure how to address it. Day 2 already looks like this (noting that exercises are included in each segment's time):

* Day 2: 415 min (expected: 360 min)
  * 09:00: _Methods and Traits_ (50 min) (Exercise: GUI Library)
  * 09:50: break (10 min)
  * 10:00: _Generics_ (40 min) (Exercise: Generic `min`)
  * 10:40: break (10 min)
  * 10:50: _Standard Library Types_ (35 min) (Exercise: Counter)
  * 11:25: break (10 min)
  * 11:35: _Standard Library Traits_ (60 min) (Exercise: ROT13)
  * 12:35: lunch (60 min)
  * 01:35: _Iterators_ (30 min) (Exercise: Iterator Method Chaining)
  * 02:05: break (10 min)
  * 02:15: _Modules_ (35 min) (Exercise: Modules for the GUI Library)
  * 02:50: break (10 min)
  * 03:00: _Testing_ (55 min) (Exercise: Luhn Algorithm)
  * 03:55: end 

If we conservatively double the time for the `std` segments, that's another 95 minutes, or 520 minutes total. The other two days are each 365 minutes already. Factor in that @mgeisler has indicated all of the exercises probably need twice the time (so, add 15m per segment), and we're basically looking at a 4th day.

This refactor knowingly added a bunch of content that we knew was missing from v1, and also more exercises on request. So perhaps this is just what it costs to have nice things? Or, is there some significant amount of content we could cut as "out of scope" -- even cutting unsafe (40m) and iterators (30m) wouldn't be enough, really.

>   I also worry that if students haven't seen lifetimes yet, they will not be equipped to understand many of the type signatures in the stdlib, including ones that we really do want to discuss (such as those on the `Entry` API for `HashMap`--it's only because of borrowing that this API can exist).

References are spiraled, so they have been introduced in day 1 afternoon (segment "References"). We could add another "loop" of that spiral here to cover lifetime annotations without the full borrow-checker coverage that comes later.  The signature for `HashMap::entry` is particularly obtuse, using `'_`, but maybe that could be explained away as "only usable for a short time" or something like that.

> * It seems strange for slices and lifetimes to be between error handling and unsafe Rust on the second half of Day 3. Both of the latter are in some sense "nitty gritty" concerns--error handling is important in real-world programs where things can and do go wrong, and unsafe Rust is necessary in situations where safe Rust fails us (in performance overhead or ability to do interop). Slices, meanwhile, are a primitive type provided by the standard library, and lifetimes are used pervasively in the standard library and in any nontrivial code working with borrowed data. I feel like it would be better to frontload lifetimes so that students get more experience with them and know the basics before they see more advanced uses (such as some of those in std).

It's hard to do lifetimes before memory management, and we decided earlier to push memory management near the end of the course, so the range of positions where we can place slices and lifetimes are limited. But, looking at the second half of day-2 forward, as it currently stands:

* Day 2 (afternoon)
  * _Iterators_ (30 min) (Exercise: Iterator Method Chaining)
  * _Modules_ (35 min) (Exercise: Modules for the GUI Library)
  * _Testing_ (55 min) (Exercise: Luhn Algorithm)

* Day 3 (morning)
  * _Memory Management_ (50 min) (Exercise: Builder Type)
  * _Smart Pointers_ (40 min) (Exercise: Binary Tree)
  * _Borrowing_ (45 min) (Exercise: Health Statistics)

* Day 3 (afternoon)
  * _Error Handling_ (35 min) (Exercise: Rewriting with Result)
  * _Slices and Lifetimes_ (55 min) (Exercise: Protobuf Parsing)
  * _Unsafe Rust_ (40 min) (Exercise: FFI Wrapper)

"Modules" and "Testing" are also nitty-gritty, the way "Error Handling" and "Unsafe Rust" are, so maybe those could move to the end of the course, with some of the day-3 content moving to the end of day-2 to compensate. Then "Slices and Lifetimes" follows right after "Borrowing". Something like:

* Day 2 (afternoon)
  * _Iterators_ (30 min) (Exercise: Iterator Method Chaining)
  * _Memory Management_ (50 min) (Exercise: Builder Type)
  * _Smart Pointers_ (40 min) (Exercise: Binary Tree)

* Day 3 (morning)
  * _Borrowing_ (45 min) (Exercise: Health Statistics)
  * _Slices and Lifetimes_ (55 min) (Exercise: Protobuf Parsing)
  * _Modules_ (35 min) (Exercise: Modules for the GUI Library)

* Day 3 (afternoon)
  * _Testing_ (55 min) (Exercise: Luhn Algorithm)
  * _Error Handling_ (35 min) (Exercise: Rewriting with Result)
  * _Unsafe Rust_ (40 min) (Exercise: FFI Wrapper)

What do you think?

> * One last small ordering nit: I think categorizing `Drop` under "Smart Pointers" isn't really appropriate, but maybe it could just move into the prior section on "Memory Management" as it's related to Copy etc.

Yeah, that's a good point! I've made that change (the timings above in this comment are before the change, though).

timClicks:
> > Day 2 morning covers methods and traits, generics, standard library types, and standard library traits. [...] Even so, this seems like a lot of material for half a day.
>
> [...] and we're basically looking at a 4th day.

4 comfortable days is likely to produce a better result than 5 rushed days. However, now the course feels like "nearly a week" rather than "half a week".

I also agree that merging as-is then touching up details with follow-up PRs is sensible.

mgeisler:
Hey all, I created #1465 to collect the outstanding issues which I think we should at least look at before merging this PR. As Dustin writes there, the new content is ~30% larger than what we had before, so we should really make the class a 4-day class to do it justice. Let's discuss that more over on the issue.

mgeisler:
I added a link to the outline in the first comment:

> **Please consult [the outline](https://github.com/djmitche/comprehensive-rust/blob/cr2-organization/cr2-organization/outline.md)** to quickly see what goes where!

As Dustin rebases the PR, this should stay up to day and it's a great way to get an overview of this work.

djmitche:
The outline is now generated via an mdbook preprocessor (mdbook-course) This means it's not visible in the PR, but I think that's the right decision long-term - see #1254.

djmitche:
UPDATED:
 * Reorganized to put borrowing and slices closer together
 * Doubled time for most exercises (based on feedback)
 * Increased time for std-{types,traits} (based on feedback)
 * Updated course guidelines to suggest 6 instructional hours per day (the course now has more breaks and more time for exercises, so this is very much not 6 hours of lecture)
 * Split into 4 days

The new outline has the following timings:

```
Fundamentals: 21 hours and 35 minutes: 2 hours and 25 minutes shorter than target 24 hours
Sessions:
  Day 1 Morning: 3 hours: right on time
    Welcome: 5 minutes
    Hello, World: 20 minutes
    Types and Values: 1 hour and 5 minutes
    Control Flow Basics: 1 hour
  Day 1 Afternoon: 2 hours and 55 minutes: 5 minutes shorter than target 3 hours
    Welcome: 0 minutes
    Tuples and Arrays: 1 hour
    References: 50 minutes
    User-Defined Types: 50 minutes
  Day 2 Morning: 3 hours and 15 minutes: 15 minutes OVER TARGET 3 hours
    Welcome: 3 minutes
    Pattern Matching: 50 minutes
    Methods and Traits: 1 hour and 5 minutes
    Generics: 45 minutes
  Day 2 Afternoon: 3 hours: 2 minutes shorter than target 3 hours
    Welcome: 0 minutes
    Standard Library Types: 1 hour and 10 minutes
    Standard Library Traits: 1 hour and 40 minutes
  Day 3 Morning: 2 hours and 15 minutes: 45 minutes shorter than target 3 hours
    Welcome: 3 minutes
    Memory Management: 1 hour and 10 minutes
    Smart Pointers: 45 minutes
  Day 3 Afternoon: 2 hours and 20 minutes: 40 minutes shorter than target 3 hours
    Welcome: 0 minutes
    Borrowing: 1 hour
    Slices and Lifetimes: 1 hour and 10 minutes
  Day 4 Morning: 3 hours: 4 minutes shorter than target 3 hours
    Welcome: 3 minutes
    Iterators: 45 minutes
    Modules: 45 minutes
    Testing: 55 minutes
  Day 4 Afternoon: 2 hours: 1 hour and 5 minutes shorter than target 3 hours
    Welcome: 0 minutes
    Error Handling: 45 minutes
    Unsafe Rust: 1 hour and 5 minutes
```
I'm comfortable with the distribution of content like this, until we've actually taught the course and gotten some harder data for how long each segment takes, at which point we can consider moving things around a bit more to even out the days. In particular, I've left the last day an hour short in case things slip!

mgeisler (approved):
Congratulations Dustin! Amazing work here!

PR #1079, opened 2023-08-14
Move thiserror::error

sakex:
It is slightly confusing to import a symbol called `Error` twice

### DIFF
--- a/src/error-handling/dynamic-errors.md
+++ b/src/error-handling/dynamic-errors.md
@@ -6,10 +6,9 @@ all the different possibilities. `std::error::Error` makes this easy.
 ```rust,editable,compile_fail
 use std::fs;
 use std::io::Read;
-use thiserror::Error;
 use std::error::Error;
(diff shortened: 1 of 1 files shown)


qwandor (commented):
I disagree. The point of the `thiserror::Error` derive macro is that it derives the standard `Error` trait, so using the same name for both is intentional, and consistent with the other derive macros.

sakex:
@qwandor I changed to how it's done [in the book](https://doc.rust-lang.org/rust-by-example/error/multiple_error_types/boxing_errors.html)

qwandor:
I still don't really see why you want the trait to be named differently to the derive macro. If anything, this is a good opportunity to mention to students that traits and derive macros usually have matching names.

djmitche (changes requested):
Agreed that the old version of this diff is more idiomatic. Maybe add a speaker-note comment about this?

PR #1081, opened 2023-08-14
Iterate over array ref instead of array.

dyoo:
This revises the example to iterate over `&array` to dodge a few issues.


djmitche (approved):
Thanks!

PR #1084, opened 2023-08-15
Binary tree exercise

sakex:
@mgeisler  What do you think about Leetcode-style exercises for https://github.com/google/comprehensive-rust/issues/1082 ? Tree data structures are great at showing how to use `Option<Box>`, `Rc`, etc.

Also, what do you think about adding intermediary steps solutions in spoilers?

### DIFF
--- a/speaker-notes.js
+++ b/speaker-notes.js
@@ -145,18 +145,11 @@
   // Create headers on the print page.
   function setupPrintPage() {
     for (const notes of document.querySelectorAll("details")) {
-      let summaryContent;
-      if (notes.getAttribute("solution") !== null) {
-        notes.open = false;
-        summaryContent = "Solution";
-      } else {
-        notes.open = true;
-        summaryContent = "Speaker Notes";
-      }
-      const summary = document.createElement("summary");
+      notes.open = true;
+      let summary = document.createElement("summary");
       notes.insertBefore(summary, notes.firstChild);
-      const h4 = document.createElement("h4");
-      h4.append(summaryContent);
+      let h4 = document.createElement("h4");
+      h4.append("Speaker Notes");
       summary.append(h4);
     }
   }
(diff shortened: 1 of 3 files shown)


mgeisler:
> @mgeisler What do you think about Leetcode-style exercises for #1082 ? Tree data structures are great at showing how to use `Option<Box>`, `Rc`, etc.

Small exercises like that could be great for illustrating these types. Though, note #1082: we should avoid creating exercises which are hard to solve. The challenge should be "How to do this in Rust?", not "How to do this?".

> Also, what do you think about adding intermediary steps solutions in spoilers?

I'm sure this can work well for small exercises — though there is a danger that people open the solutions quickly when they're right there on the page. Currently, a large number of the course participants don't even realize that the solutions are there at the end of the course.

djmitche:
I think students generally appreciate the time to get "hands-on" writing code, rather than answering open-response questions. We actually have decent exercises for the first day, where students might not have all of the information necessary to write code.

sakex:
The issue we had last week was that we didn't have time to cover the exercises really. That's why I was thinking more of a quiz format where we could ask questions to the students and then discuss the answer together (many times these discussions happened organically). It would allow the tutor to gauge what the students are struggling with.

Small iterative exercises might also be nice for students to do at their own pace. Having intermediary solutions allows them to get unblocked if needed.

That's my rationale for my 2 suggestions here. In the end, I think that exercises format should coexist and be used in different situations.

mgeisler:
> The issue we had last week was that we didn't have time to cover the exercises really.

It is hard to balance the time between exploration via live-coding, talking, answering questions, etc. All that should be done, while also leaving time for the exercises (30-45 minutes per morning/afternoon block). I've noticed myself that I sometimes run out of time when teaching the class.

PR #1090, opened 2023-08-18
Add @javisorribes as a reviewer for ES

JaviSorribes:


.github/CODEOWNERS
```diff
@@ -3,7 +3,7 @@ po/bn.po @mirwasi @sudiptachatterjee
 po/da.po @mgeisler @voss @thedataking
 po/de.po @ronaldfw @fechu
 po/el.po @Mandragorian
-po/es.po @deavid
+po/es.po @deavid @javisorribes
```
djmitche, line 6:
```suggestion
po/es.po @deavid @JaviSorribes
```

djmitche (changes requested):

djmitche (approved):
Hm, GitHub is still saying it's invalid, perhaps because you don't have write access to the repo. @mgeisler may need to fix that :)

henrif75:
> Hm, GitHub is still saying it's invalid, perhaps because you don't have write access to the repo. @mgeisler may need to fix that :)

I gave @JaviSorribes write permission.

henrif75:
@JaviSorribes please accept the invite from GitHub. You'll also need to accept the Google CLA at https://cla.developers.google.com/cla

mgeisler:
> You'll also need to accept the Google CLA at https://cla.developers.google.com/cla

Small note: this is not necessary for Googlers — I guess the page will say the same :smile:

henrif75:
> > You'll also need to accept the Google CLA at https://cla.developers.google.com/cla
> 
> Small note: this is not necessary for Googlers — I guess the page will say the same 😄

Got it, should the test be changed to account for it?

mgeisler:
> Got it, should the test be changed to account for it?

It should already be green when people use their `@google.com` email address in the commits. There has been a few cases where people accidentally used their private addresses and we had to fix this by rewriting the commits. We don't have much control over this check — it's added via a global setting on the `google` GitHub organization.

JaviSorribes:
Hi, sorry but I didn't have notifications configured (fixed) and the invitation is now expired. Could you please send again?

PR #1094, opened 2023-08-21
Update for-loops.md to remove misleading sentence

dyoo:
Drops the comment about usage in the other loop.  It's not related to the focus on copy semantics, and so we want to eliminate that possible confusion.

### DIFF
--- a/src/exercises/day-1/for-loops.md
+++ b/src/exercises/day-1/for-loops.md
@@ -85,7 +85,7 @@ preview of issues of ownership that will come later in the afternoon.
 Without the `&`...
 * The loop would have been one that consumes the array.  This is a
   change [introduced in the 2021
-  Edition](https://doc.rust-lang.org/edition-guide/rust-2021/IntoIterator-for-arrays.html)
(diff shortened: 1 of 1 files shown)


src/exercises/day-1/for-loops.md
```diff
@@ -85,9 +85,8 @@ preview of issues of ownership that will come later in the afternoon.
 Without the `&`...
 * The loop would have been one that consumes the array.  This is a
   change [introduced in the 2021
-  Edition](https://doc.rust-lang.org/edition-guide/rust-2021/IntoIterator-for-arrays.html), and ...
-* since the array is also accessed in the second loop, an implicit
-  array copy would have occured; since `i32` is a copy type, then
+  Edition](https://doc.rust-lang.org/edition-guide/rust-2021/IntoIterator-for-arrays.html)
```
mgeisler, line 88: I like to end my list items with a `.`:
```suggestion
  Edition](https://doc.rust-lang.org/edition-guide/rust-2021/IntoIterator-for-arrays.html).
```

mgeisler (approved):
Thanks, I agree that this is clearer and more to-the-point.

PR #1121, opened 2023-08-25
Update deep-dive titles in course structure

mgeisler:
I'm trying to align on these names for the additional material:

- Rust in Android
- Bare-Metal Rust
- Concurrency in Rust

It's not perfectly reflected everywhere, but this brings us a bit closer to that.

### DIFF
--- a/src/running-the-course/course-structure.md
+++ b/src/running-the-course/course-structure.md
@@ -15,7 +15,7 @@ specialized topics:
 
 ### Android
 
-The [Android Deep Dive](../android.md) is a half-day course on using Rust for
+The [Rust in Android](../android.md) deep dive is a half-day course on using Rust for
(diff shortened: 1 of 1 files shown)


src/running-the-course/course-structure.md
```diff
@@ -33,7 +33,7 @@ commands it runs and make sure they work when you run them by hand.
 
 ### Bare-Metal
```
mgeisler, line 34:
```suggestion
### Bare-Metal Rust
```

src/running-the-course/course-structure.md
```diff
@@ -44,7 +44,7 @@ page](../bare-metal.md).
 
 ### Concurrency
```
mgeisler, line 45:
```suggestion
### Concurrency in Rust
```

PR #1122, opened 2023-08-25
da: Fix all fuzzy entries

mgeisler:
Part of #286.

### DIFF
--- a/po/da.po
+++ b/po/da.po
@@ -957,7 +957,7 @@ msgstr "Asynkrone egenskaber (eng. Traits)"
 
 #: src/SUMMARY.md:297 src/async/pitfalls/cancellation.md:1
 msgid "Cancellation"
-msgstr "Afbrydelser"
+msgstr "Annulering"
 
 #: src/SUMMARY.md:300 src/exercises/concurrency/chat-app.md:1
 #: src/exercises/concurrency/solutions-afternoon.md:119
--- a/po/da.po
+++ b/po/da.po
@@ -245,7 +245,7 @@ msgstr "Livstider i datastrukturer"
 
 #: src/SUMMARY.md:68 src/exercises/day-1/book-library.md:1
 msgid "Storing Books"
-msgstr "Gemning af bøger"
+msgstr "Lagring af bøger"
 
 #: src/SUMMARY.md:69 src/exercises/day-1/iterators-and-ownership.md:1
 msgid "Iterators and Ownership"
@@ -1826,7 +1826,7 @@ msgid ""
 "This is the first day of Rust Fundamentals. We will cover a lot of ground "
 "today:"
 msgstr ""
-"Dette er den første dag af Fust Fundamentals. Vi kommer til at dække en "
+"Dette er den første dag af Rust Fundamentals. Vi kommer til at dække en "
 "masse terræn i dag:"
 
 #: src/welcome-day-1.md:6
@@ -4381,7 +4381,7 @@ msgstr ""
 
 #: src/ownership/double-free-modern-cpp.md:1
 msgid "Extra Work in Modern C++"
-msgstr "Ekstra arbejde i moderne C++"
+msgstr "Defensiv kopiering i moderne C++"
 
 #: src/ownership/double-free-modern-cpp.md:3
 msgid "Modern C++ solves this differently:"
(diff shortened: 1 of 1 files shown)


mgeisler:
Hejsa, kan jeg få et review her? :slightly_smiling_face:

po/da.po
```diff
@@ -244,9 +244,8 @@ msgid "Lifetimes in Data Structures"
 msgstr "Livstider i datastrukturer"
 
 #: src/SUMMARY.md:68 src/exercises/day-1/book-library.md:1
-#, fuzzy
 msgid "Storing Books"
-msgstr "String og str"
+msgstr "Gemning af bøger"
```
thedataking, line 248: nit: "Lagring af bøger"? Gemning er fint men jeg havde ikke set det ord før nu :)

po/da.po
```diff
@@ -960,9 +956,8 @@ msgid "Async Traits"
 msgstr "Asynkrone egenskaber (eng. Traits)"
 
 #: src/SUMMARY.md:297 src/async/pitfalls/cancellation.md:1
-#, fuzzy
 msgid "Cancellation"
-msgstr "Installation"
+msgstr "Afbrydelser"
```
thedataking, line 960: nit: Cancellation is singular, Afbrydelser is plural.

po/da.po
```diff
@@ -1830,12 +1822,11 @@ msgid "Welcome to Day 1"
 msgstr "Velkommen til Dag 1"
 
 #: src/welcome-day-1.md:3
-#, fuzzy
 msgid ""
 "This is the first day of Rust Fundamentals. We will cover a lot of ground "
 "today:"
 msgstr ""
-"Dette er den første dag af Comprehensive Rust. Vi kommer til at dække en "
+"Dette er den første dag af Fust Fundamentals. Vi kommer til at dække en "
```
thedataking, line 1829: Rust Fundamentals

po/da.po
```diff
@@ -4390,9 +4380,8 @@ msgstr ""
 "```"
 
 #: src/ownership/double-free-modern-cpp.md:1
-#, fuzzy
 msgid "Extra Work in Modern C++"
-msgstr "Dobbeltfrigivelser i moderne C++"
+msgstr "Ekstra arbejde i moderne C++"
```
thedataking, line 4384: I'm not sure I have a better translation but I like that its not clear what kind of work we're talking about without additional context. It is not the programmer who has to do more work here; it is the machine. If we can change the English text, I'd use "Additional duplication in modern C++"; if not, I'd translate that to Danish for additional clarity.

OTOH, I don't know if you're comfortable with the additional "distance" between the English and Danish wording.

po/da.po
```diff
@@ -18673,9 +18665,8 @@ msgid ""
 msgstr ""
 
 #: src/exercises/concurrency/chat-app.md:127
-#, fuzzy
 msgid "Running the binaries"
-msgstr "Afvikling af kurset"
+msgstr "Kørsel af binære filer"
```
thedataking, line 18669: Can we keep "Afvikling"? I much prefer that to "Kørsel".

po/da.po
mgeisler: Nej, jeg har nok bare glemt hvordan man taler dansk :see_no_evil: 
```suggestion
msgstr "Lagring af bøger"
```

po/da.po
mgeisler: Sikke noget fusk!

```suggestion
"Dette er den første dag af Rust Fundamentals. Vi kommer til at dække en "
```

po/da.po
mgeisler: > If we can change the English text, I'd use "Additional duplication in modern C++"; if not, I'd translate that to Danish for additional clarity.

We can (and _should!_) update the English text to fix this kind of awkwardness. There are lots of text like this which can be expanded just a little to make it easier to understand in both English and other languages.
 
> OTOH, I don't know if you're comfortable with the additional "distance" between the English and Danish wording.

I'm not an expert, but my gut feeling would be to minimize the distance. In this case, I will try to fix both at the same time: I can update the English text (I created #1192) and I can preemptively translate the new heading. It will be marked fuzzy when the PR is merged and then I can proof-read it again.

mgeisler:
```suggestion
msgstr "Defensiv kopiering i moderne C++"
```

po/da.po
mgeisler: Yeha, certainly!

```suggestion
msgstr "Afvikling af binære filer"
```

po/da.po
mgeisler: Hmm, good point... perhaps _Annulering_ is a better translation?

```suggestion
msgstr "Afbrydelser"
```

thedataking (approved):

Issue #330, opened 2023-02-03
GitHub action to check formatting of Markdown files?


gendx:
In https://github.com/google/comprehensive-rust/pull/329, I'm adding some text which makes a line exceed the 80-column standard formatting of markdown files. I'm waiting for the workflows to be approved to see if they will catch the formatting mistake.

If not, it would be good to have a way to automatically format and check the markdown files (ala `cargo fmt` and `cargo fmt --check`).

mgeisler:
Yes, that's definitely something we should add — I'm thinking that we might not need to make it a check for merging (since I want to encourage people to make small drive-by commits). Instead I simply want to reformat things once in a while.

Reformatting everything is waiting on #318 since that will allow us to change the formatting of the Markdown text without affecting the translations.

fw-immunant:
I believe this was fixed by #1157.

Issue #326, opened 2023-02-02
Russian translation


mgeisler:

mgeisler:
For reference, there is a Russian translation of the Rust Book: https://github.com/rust-lang-ru/book.

baltuky:
@mgeisler Would appreciate if you could assign this issue to me

mgeisler:
> @mgeisler Would appreciate if you could assign this issue to me

There you go! :smile:

mgeisler:
Hi @baltuky, would you be able to add a nearly-empty PO file for Russian? Just translate a few entries — that will get us started and it will make it easier for others to start contributing!

baltuky:
Hey Martin,
Yes, I've created a PO file for Russian. I will create a PR for that later today.

baltuky:
@mgeisler I've created a [PR](#677) that brings a fresh `ru.po` file, PTAL

atmsk:
I would like to contribute. How do I get started?

baltuky:
> I would like to contribute. How do I get started?

Hey 👋 
It makes sense to start by reading the [Translations of Comprehensive Rust 🦀](https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md) page. 

I've prepared a draft of `Day 1: Morning` part and working on `Day 1: Afternoon`. So, feel free to pick any other chapters.

istolga:
also want to contribute

mgeisler:
Hey all! There seems to be great momentum here! I propose that you **[split the work via a spreadsheet](https://docs.google.com/spreadsheets/d/1T9l_Gd4O7qe3rA5EnK4Rk-7pRO4TutukCuJdakUHDss/edit?usp=sharing).** That seems to work well for the two Chinese translations.

baltuky:
> Hey all! There seems to be great momentum here! I propose that you **[split the work via a spreadsheet](https://docs.google.com/spreadsheets/d/1T9l_Gd4O7qe3rA5EnK4Rk-7pRO4TutukCuJdakUHDss/edit?usp=sharing).** That seems to work well for the two Chinese translations.
> 
> Edit it all you want, use as much or as little as you like!

That is a great idea. 💯  I've updated the spreadsheet with my assignments. Now the translation process should be more transparent.

starostinak:
Hey, I also want to contribute 👋 Will look though the translations guide and pick something unclaimed from the spreadsheet.

starostinak:
[...]

Issue #282, opened 2023-01-26
Translations


mgeisler:
* #262

dblackstar:
Is there a translation in French? If not i would like to do it.

mgeisler:
> Is there a translation in French? If not i would like to do it.

Hey @dblackstar, nobody has signed up for that yet! It would be awesome to have a French translation too. Let me quickly create an issue for that.

baltuky:
Hey 👋 Let’s translate it in Russian. I can do that

mgeisler:
> Working on the Chinese Mandarin translation of the book, thanks for your prior instructions! Can you help create an issue for that?

Thanks @Hilbert-Yaa for creating the issue, I've added it to the list above!

> Hey wave Let’s translate it in Russian. I can do that

Awesome @baltuky, I've created #326 for that.

mgeisler:
I've attached Cloud Translate versions of the `messages.pot` to the new issues. Please let me know if you need me to update anything else. You can also generate those yourself via #291.

mgeisler:
Hi all, if you haven't already, please take a look at the Rust Book translations: https://doc.rust-lang.org/book/appendix-06-translation.html. When possible, I suggest using the same terminology as they use.

pierd:
Hi @mgeisler, I could a Polish translation.

mgeisler:
> Hi @mgeisler, I could do a Polish translation.

Hi @pierd, thanks a lot, that would be super cool! If you create an issue for it, then I'll link it here!

pierd:
Created the issue for Polish translation: https://github.com/google/comprehensive-rust/issues/399

wfranck:
could I help with this issue? I had not seen that the ticket for Spanish was open and I made progress with the translation.

pierd:
Do we generally translate source code snippets? IMHO it definitely makes sense for the comments and maybe string literals but not really for variable/function/etc names.

I was checking other translations for some examples. Brazilian Portuguese version translates some string literals and some comments but not all ([modules](https://google.github.io/comprehensive-rust/pt-BR/modules.html), [some exercise](https://google.github.io/comprehensive-rust/pt-BR/exercises/day-1/book-library.html)) while Korean version leaves the string literals as is but translates all comments ([modules](https://google.github.io/comprehensive-rust/ko/modules.html), [the same exercise](https://google.github.io/comprehensive-rust/ko/exercises/day-1/book-library.html)).

[...]

PR #1130, opened 2023-08-25
ru: Refresh .po file after mdbook-i18n-helpers 0.2.2

henrif75:
Part of #330, #282, #326


mgeisler (approved):
Thanks! The new fuzzy entries should be fixed in subsequent PRs.

PR #1135, opened 2023-08-28
Fix duplicate Day 2 Afternoon exercise page

mgeisler:
Spotted by @gregpaton08 in #913.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -97,18 +97,16 @@
   - [Lifetimes](ownership/lifetimes.md)
   - [Lifetimes in Function Calls](ownership/lifetimes-function-calls.md)
   - [Lifetimes in Data Structures](ownership/lifetimes-data-structures.md)
-- [Exercises](exercises/day-2/morning.md)
-  - [Storing Books](exercises/day-2/book-library.md)
-  - [Iterators and Ownership](exercises/day-2/iterators-and-ownership.md)
-
 - [Structs](structs.md)
   - [Tuple Structs](structs/tuple-structs.md)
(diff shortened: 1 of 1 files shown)


mgeisler:
This was found here: https://github.com/google/comprehensive-rust/commit/d3a90373b0007d7b90c940aef69e9a8b1eb98440#r125548839.

mgeisler:
@fw-immunant, we'll probably merge the before you get online, but please skim it too to ensure I haven't forgotten anything.

mgeisler:
Hmm, this isn't completely right:

> # Day 2: Morning Exercises
> We will look at implementing methods in two contexts:
> - Simple struct which tracks health statistics.
> - Multiple structs and enums for a drawing library.

The exercise named "A Simple GUI Library" is on Day 3 Morning.

fw-immunant (approved):
LGTM, thanks for catching this.

Edit: Ah, I see. Let me take a closer look and figure out what exactly should be where.

mgeisler:
> Edit: Ah, I see. Let me take a closer look and figure out what exactly should be where.

Thanks a lot! I started looking, but got a bit lost myself :smile:

djmitche:
FWIW there's already a subsequent reorg of this content planned, so I don't think it's worth spending too much time on this. Let's get it to a spot where it can be taught, and then focus on the v2.

mgeisler:
Thanks for putting up #1136.

> FWIW there's already a subsequent reorg of this content planned, so I don't think it's worth spending too much time on this. Let's get it to a spot where it can be taught, and then focus on the v2.

Agreed. We'll close this and merge #1136 to get the loop fixed.

PR #1139, opened 2023-08-29
ja: Translate chapter 49.2 (bounded channels)

kantasv:
Hi, JA translators! (#652) I've added translations for "bounded channels". This is a tiny MR, but I'm open to suggestions. Have a good day!

### DIFF
--- a/po/ja.po
+++ b/po/ja.po
@@ -16587,6 +16587,8 @@ msgstr ""
 msgid ""
 "With bounded (synchronous) channels, `send` can block the current thread:"
 msgstr ""
+"Bounded（かつ同期的）なチャネルを用いたとき、`send`は現在のスレッドをブロック"
+"できます："
 
 #: src/concurrency/channels/bounded.md:5
 msgid ""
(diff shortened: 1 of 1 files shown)


po/ja.po
```diff
@@ -16621,18 +16623,26 @@ msgid ""
 "channel for the new message. The thread can be blocked indefinitely if there "
 "is nobody who reads from the channel."
 msgstr ""
+"`send`を呼ぶことによって、チャネルに新たなメッセージのための空きができるま"
+"で、現在のスレッドはブロックされます。もし誰もチャネルから値を読み取らない場"
+"合は、このスレッドは無期限にブロックされることがあります。"
```
HidenoriKobayashi, line 16628: I would say something like: "Sendを呼んだときにチャネルに空きがなければ、チャネルに..."
although that will not be a 1-to-1 translation

mgeisler: The English original can be improved in many cases — as translators, you're in a unique position to do so since you'll be reading the text more carefully than most :smile: 

You can always fix small issues directly in the translation. It would then be helpful to "port" those fixes back to the English text. When you `msgmerge` the PO file next, you will see that the translation is marked "fuzzy". This signals that it needs proof-reading: in this case there will be nothing to do except remove the fuzzy marker again. This kind of de-fuzzing should be done repeatedly, ideally weekly or so.

mgeisler: Oh, second small point: GitHub has a great suggestion tool which saves a lot of time in discussions. It's the left-most button in the toolbar and it looks like this when I add what you wrote above (no idea what it says):

```suggestion
"Sendを呼んだときにチャネルに空きがなければ、チャネルに"
```

People can commit suggestions with a few clicks, which saves a lot of time.

kantasv:
Thank you @HidenoriKobayashi and @mgeisler for suggestions and feedback! I've incorporated changes in the latest commit :)

To @keiichiw or @chikoski if this looks okay, could you merge the MR?

keiichiw (approved):
LGTM. Thank you!

PR #1140, opened 2023-08-30
Insert `<link rel="canonical">` elements

mgeisler:
This prevents duplicate search results in case copies of the material is being left visible on the web. See the [Search Central documentation](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls) for details.

The implementation here is very simple and should go away when https://github.com/google/mdbook-i18n-helpers/issues/70 is implemented.

### DIFF
--- a/theme/index.hbs
+++ b/theme/index.hbs
@@ -142,6 +142,33 @@
             }
         </script>
 
+        {{! Move to template code after fixing this issue:
+            https://github.com/google/mdbook-i18n-helpers/issues/70 }}
+        <script>
+          (function () {
+              // See these pages for details:
+              // https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls
+              // https://developers.google.com/search/docs/crawling-indexing/javascript/javascript-seo-basics
+              let base = "https://google.github.io/comprehensive-rust";
+              {{#if (eq language "en")}}
+              let canonical_href = `${base}/{{ path }}`;
+              {{else}}
+              let canonical_href = `${base}/{{ language }}/{{ path }}`;
+              {{/if}}
+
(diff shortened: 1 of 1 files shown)


theme/index.hbs
```diff
@@ -29,6 +29,10 @@
         <meta name="viewport" content="width=device-width, initial-scale=1">
         <meta name="theme-color" content="#ffffff" />
 
+        {{! Replaced with a <link rel="canonical"> element in
+            publish.yml. Remove when after fixing this issue:
```
djmitche, line 33:
```suggestion
            publish.yml. Remove after fixing this issue:
```

djmitche (approved):
Is the idea that forks of this repo should point to the canonical location? Doing this substitution in the workflow means that someone running `mdbook build` and uploading it to a test location won't get the substitution.

mgeisler:
> Is the idea that forks of this repo should point to the canonical location? Doing this substitution in the workflow means that someone running `mdbook build` and uploading it to a test location won't get the substitution.

That is a good point... maybe I should use JavaScript instead then! It's supported, but I got the impression that it's a bit [discouraged](https://developers.google.com/search/docs/crawling-indexing/javascript/javascript-seo-basics#properly-inject-canonical-links). But it's important that people get the tags even if they run `mdbook build` locally, so we should go with that route instead. Thanks!

theme/index.hbs
mgeisler: Thanks!

mgeisler:
Thanks for catching the problem with putting the logic into `publish.yml`. I downloaded the tarball for Danish and checked locally that we generate correct canonical URLs now.

PR #1172, opened 2023-09-06
Add solution for Health Statistics exercise

fw-immunant:


src/exercises/day-2/health-statistics.rs
mgeisler, line 1: This doesn't follow the system (eh, more like "guidelines") of the other solutions. That is, I normally tried to reuse the same Rust file for both the exercise and the solution. For the exercise, I would include only part of the file — the solution is then the full file.

It's tedious, but it ensures some level of consistency between the function signatures in the exercise and the solution.

The solution is the then also added to the `src/exercises/Cargo.toml` file to ensure it gets built by `cargo check` in CI.

mgeisler (changes requested):
Thanks for cleaning up the order of the exercises in the SUMMARY... we forgot that when we closed #1135.

We should just make sure that the solution is incorporated correctly in the rest of the course.

src/exercises/day-2/health-statistics.rs
fw-immunant: I think I've gotten this aligned with how the other solutions are done.

mgeisler (approved):
Thanks, it aligned! I realize now how messy this system is... :grimacing:

src/exercises/day-2/health-statistics.rs
mgeisler: You did! It looks a bit ugly, but that's not your fault :smile:

It would be interesting if we could use a better template engine here as well (like in https://github.com/google/mdbook-i18n-helpers/issues/70).

PR #1180, opened 2023-09-06
fr: Update the fr.po file with new English messages (msgmerge).

AdrienBaudemont:


mgeisler:
Hey @lb034582341 and @sakex, can you take a look here? There's actually nothing to do: just observe that the `msgmerge` output looks correct.

sakex (approved):

PR #1182, opened 2023-09-06
Add test case for invalid non digit cc number

varunsomani:
Adding this test case to handle case where we have non-numeric cc number along with a valid cc number. This is discovered when we filter with is_numeric() instead of !is_whitespace() all test case does pass. To handle such scenario adding this test case.


mgeisler (approved):
Thanks, looks great!

PR #1194, opened 2023-09-10
Corrected Luhn's Algorithm Step 3

abba-m:
In the exercise guideline for the Luhn's algorithm (Day 1, Afternoon exercise), I made a correction to step 3 to provide a more accurate explanation. Instead of stating that doubling 7 becomes 14, I clarified that doubling 7 results in 14, and we should sum the individual digits of 14 to get the correct value of 5.


djmitche (approved):
Thanks! This problem definition has definitely been challenging.

PR #1195, opened 2023-09-10
Update dependencies in chat-app exercise

Gelbpunkt:
Hello there :wave: I'm the author of the `tokio-websockets` crate and today we published version `0.4.0` of the crate, which finally makes the public API consistent and aligns it with expectations.

`WebsocketStream` now implements the `Stream` trait, so calling `next()` has to be done via `StreamExt` now. `connect()` returns a tuple of `(WebsocketStream, Response)`, so that has to be deconstructed now. And finally, `as_text()` now returns an `Option<&str>` instead of `Result<&str, Error>`.

And, *very* importantly: The old `WebsocketStream::next` was **not** cancellation safe, the new `Stream` implementation is. The exercise uses `select!`, which very well might have caused misbehavior.

Since we also emptied the default features, I've added the very minimal ones required to compile the exercise.

This is my first contribution here, so feel free to point me to some things if I'm missing anything. I wasn't really sure what to do about the translations, I guess I should leave them as is and they'll be updated by translators in other PRs?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,10 +3,25 @@
 version = 3
 
 [[package]]
-name = "aho-corasick"
-version = "0.7.20"
+name = "addr2line"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cc936419f96fa211c1b9166887b38e5e40b19958e5b895be7c1f93adec7071ac"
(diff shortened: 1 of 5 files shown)


djmitche (approved):
Wow, thank you!

The course talks about cancellation safety, so it's amusing that I made exactly that mistake in the exercise. Thanks for fixing it!

Issue #284, opened 2023-01-26
Spanish translation


mgeisler:
For reference: https://github.com/thecodix/book and https://github.com/ManRR/rust-book-es.

mgeisler:
Here is a Google Translate version to get you started: [es.zip](https://github.com/google/comprehensive-rust/files/10555902/es.zip)

wfranck:
could I help with this issue? I had not seen that the ticket for Spanish was open and I made progress with the translation.

mgeisler:
Hi @wfranck, yes please! I haven't heard of any progress being made here, so I suggest you add a link to your repository/branch here. Then perhaps @antodellanzo can end up helping you :smile:

wfranck:
Thank you.  At the moment I only have it local and installed the project to see the changes following the instructions of the project. I'll upload changes later.

wfranck:
I forked the project and started with the first commit translated until prior to the start of the 'Days' of the course
https://github.com/wfranck/comprehensive-rust/blob/main/po/es.po

wfranck:
I will continue reading the course in English without doing the translation or, working and have this in my own repository. I'm sure someone will be able to translate it better.

mgeisler:
> I will continue reading the course in English without doing the translation or, working and have this in my own repository. I'm sure someone will be able to translate it better.

Okay, thanks for the work you did! I will talk to the people who can read Spanish and try to merge in the parts you already translated. Then I hope more people from the community will jump in and help.

mgeisler:
@henrif75, when you get the translation back from the Localization team, you will need to normalize it using `mdbook-i18n-normalize`. After that, I suggest splitting it into a PR per top-level directory (using `msggrep`).

@antodellanzo, I'll remove you as the assignee here, but feel free to send us PRs if you find the time!

mgeisler:
Hi all, I just learnt about https://rustlanges.github.io/ from https://github.com/rust-lang/book/issues/661#issuecomment-1730766276. Perhaps it could be useful for people here.

mgeisler:
It would be great if we could reach out to them and see if there is a way to collaborate!

mgeisler:
Hi all! I wanted to let you know about #1460. This is an update to mdbook-i18n-helpers which will remove ~6k lines from your PO file. It changes the way code blocks are processed: we now extract strings and comments directly.

[...]

PR #1196, opened 2023-09-11
Standardize language menu format and adds Spanish

henrif75:
Spanish translation #284


djmitche (approved):

mgeisler:
Thanks, I love that you put the English names first — we should have done that right away.

PR #1203, opened 2023-09-12
da: refresh translation

mgeisler:
Stats before:

    312 translated messages, 2155 untranslated messages.

and after:

    320 translated messages, 2136 untranslated messages.

Part of #286.

### DIFF
--- a/po/da.po
+++ b/po/da.po
@@ -51,8 +51,8 @@ msgstr ""
 msgid "Day 1: Morning"
 msgstr "Dag 1: Formiddag"
 
-#: src/SUMMARY.md:19 src/SUMMARY.md:76 src/SUMMARY.md:130 src/SUMMARY.md:187
-#: src/SUMMARY.md:213 src/SUMMARY.md:263
+#: src/SUMMARY.md:19 src/SUMMARY.md:79 src/SUMMARY.md:134 src/SUMMARY.md:192
+#: src/SUMMARY.md:218 src/SUMMARY.md:268
 msgid "Welcome"
 msgstr "Velkommen"
 
@@ -120,7 +120,7 @@ msgstr "Funktioner"
 msgid "Rustdoc"
 msgstr ""
 
-#: src/SUMMARY.md:36 src/SUMMARY.md:83 src/basic-syntax/methods.md:1
+#: src/SUMMARY.md:36 src/SUMMARY.md:102 src/basic-syntax/methods.md:1
 #: src/methods.md:1
 msgid "Methods"
 msgstr "Metoder"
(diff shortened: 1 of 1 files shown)


mgeisler:
Hey @thedataking, just a small refresh to get the latest strings into the PO file.

po/da.po
```diff
@@ -3601,3797 +3633,3507 @@ msgstr ""
...
-":                           :     :                               :\n"
-":    s1                     :     :                               :\n"
-":   +-----------+-------+   :     :   +----+----+----+----+----+  :\n"
-":   | ptr       |   o---+---+-----+-->| H  | e  | l  | l  | o  |  :\n"
-":   | len       |     5 |   :     :   +----+----+----+----+----+  :\n"
-":   | capacity  |     5 |   :     :                               :\n"
-":   +-----------+-------+   :     :                               :\n"
-":                           :     `- - - - - - - - - - - - - - - -'\n"
-"`- - - - - - - - - - - - - -'\n"
-"```"
```
mgeisler, line 4054: Hi @dyoo, this shows a (minor) problem with the skipping logic: we skip the ASCII art drawing because it doesn't contain `"` or `//`. So the logic should probably be made conditional on the programming language of the code block.

po/da.po
```diff
@@ -151,876 +151,879 @@ msgstr "Arrays og for-løkker"
...
+msgid "Variant Payloads"
+msgstr "Nyttelast i varianter"
+
+#: src/SUMMARY.md:58 src/enums/sizes.md:1
+msgid "Enum Sizes"
+msgstr "Størrelse af enumerationer"
+
+#: src/SUMMARY.md:60 src/control-flow/novel.md:1
+msgid "Novel Control Flow"
+msgstr "Nyskabende forgreninger"
```
thedataking, line 212: I had to go an look at the English version to see what was meant by 'nyskabende'. It seems we want to talk about control flows not commonly found in other language so how about any of `usædvanlige/unikke/specielle forgreninger`? One thing more thing to keep in mind is that match isn't that unusual now that python has it and it has been in functional languages forever so I'm not sure how well the English version will stand the test of time.

thedataking (approved):
LGTM. 

I reviewed the Danish translations in the top and scrolled through the rest to see if there were any more. Since the diff is so large, it is possible I missed some. It would be easier to review if the manual changes were a separate commit from the larger automated change but I'm not suggesting you change your workflow as we can always fix any papercuts in a separate commit.

po/da.po
mgeisler: > One thing more thing to keep in mind is that match isn't that unusual now that python has it and it has been in functional languages forever so I'm now even sure the English version will stand the test of time.

That is a good point, thanks! @fw-immunant and @djmitche, something to take into account, especially for the v2 of the course.

mgeisler: I think I like your suggestion:

```suggestion
msgstr "Usædvanlige forgreninger"
```

mgeisler:
> Since the diff is so large, it is possible I missed some. It would be easier to review if the manual changes were a separate commit from the larger automated change but I'm not suggesting you change your workflow as we can always fix any papercuts in a separate commit.

Yeah, this continues to be a problem... I'm thinking #1205 could help here to cleanly separate `msgmerge` from fixing fuzzy entries.

Issue #836, opened 2023-06-20
Ukrainian translation


mgeisler:
Thanks @git-user-cpp for starting this in #835. This issue is to track the translation. Please read the instructions linked in #282, in particular, please read the [translation instructions](https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md) and the [style guide](https://github.com/google/comprehensive-rust/blob/main/STYLE.md).

The Rust book has been [translated into Ukrainian](https://github.com/rust-lang-ua/rustbook_ukrainian) and it would be good to use its terminology where possible.

If there are many translators, consider using [this sheet](https://docs.google.com/spreadsheets/d/1-EeAaa_pDkBM1j2mOIv5UXngNOwrq4AH8EO_vIakOYU/template/preview) to coordinate the work. This has been helpful for the Chinese and Japanese translations.

PetroProtsyk:
Hi there, I'd be happy to help with translation

mgeisler:
Thanks @PetroProtsyk and @yaremam! The first thing to do would be to skim through #835 so we can get it merged. The first commit is always huge, so don't worry about that — we just need the ~73 translated messages proofread and someone needs to click the "Approve" button (anybody can do this).

If you end up being many translators here, then I suggest using the spreadsheet template linked above to coordinate. But it's really up to you — comments on this issue could also be more than enough. I'll let you discuss this with @git-user-cpp.

dyeroshenko:
Merged the initial PR and we can start working on improvements in the next iterations. FYI - there is an amazing [vocabulary](https://github.com/rust-lang-ua/rustbook_ukrainian/blob/master/DICTIONARY.md) (proudly borrowed from [Rustbook Ukrainian repo](https://github.com/rust-lang-ua/rustbook_ukrainian)) we can reuse & contribute to.

yaremam:
Is anyone actively working on this?  I created the spreadsheet from the template above - https://docs.google.com/spreadsheets/d/10is7_b4dTzHTzzAJB067vG2LGEY-VekJFsaVjV1j6DU/ 
Please request access and pick your files to translate. I am going to start with welcome.md

mgeisler:
Hi @yaremam, welcome! I don't think anybody is working actively on this.

Consider using https://github.com/mgeisler/cloud-translate to seed the translation with a machine translation. The results are sometimes hilarious, but sometimes helpful :smile:

mgeisler:
[...]

PR #1208, opened 2023-09-12
uk: Continued Translation - Finished 1-200, Continued with 201-400

yaremam:
part of #836


yaremam:
@dyeroshenko, can you please take a look? Should we add more owners to the file maybe?

dyeroshenko (approved):
LGTM

Issue #671, opened 2023-05-24
Persian translation


mgeisler:
Hi @mbpouri, here is an issue for you!

I don't see any translation on https://doc.rust-lang.org/book/appendix-06-translation.html, but perhaps there is a Persian Rust community somewhere where you can find translations of common words :smile: 

The translation is live at https://google.github.io/comprehensive-rust/fa/.

mgeisler:
Let me know if/when you run into https://github.com/rust-lang/mdBook/issues/1486 — then we'll have to work on adding good support for right-to-left languages in mdbook.

mbpouri:
sure, on it. thanks pal @mgeisler

mgeisler:
@DannyRavi, I tried out the translation in #1042 and the text looks like this out of the box:

![image](https://github.com/google/comprehensive-rust/assets/89623/6012220c-9925-404e-a1f8-cbf03d78dd48)

Adding `dir="rtl"` to the `html` tag makes it look like this:

![image](https://github.com/google/comprehensive-rust/assets/89623/d3f41579-4983-4448-a1ce-bacbc263500b)

Does this look more correct to you? I cannot read it, so it's hard for me to judge how it looks :slightly_smiling_face: 

Ideally, we should fix this in the upstream mdbook project (https://github.com/rust-lang/mdBook/issues/1486), but we can perhaps prototype a fix for them here and then contribute it upstream. I'm thinking we can simply use different `index.hbs` template depending on the writing direction.

DannyRavi:
Before starting the translation of the texts. I noticed that the book has a problem with right-to-left texts according to links [#1486](https://github.com/rust-lang/mdBook/issues/1486) and [#671](https://github.com/google/comprehensive-rust/issues/671)
After translating some texts and testing the program on a local server, I realized that the right-to-left texts are not correct and the texts are difficult to understand for Persian speakers. Finally, I had no idea for right-to-left texts.

mgeisler:
Hi @DannyRavi, could you test locally with `mdbook serve` after modifying `theme/index.hbs` like this:

```patch
modified   theme/index.hbs
@@ -1,5 +1,5 @@
 <!DOCTYPE HTML>
-<html lang="{{ language }}" class="sidebar-visible no-js {{ default_theme }}">
+<html lang="{{ language }}" dir="rtl" class="sidebar-visible no-js {{ default_theme }}">
     <head>
         <!-- Book generated using mdBook -->
         <meta charset="UTF-8">
[...]

PR #1209, opened 2023-09-13
fa: add more Persian

DannyRavi:
some Persian translations have been added.

Part of #671.

### DIFF
--- a/po/ja.po
+++ b/po/ja.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: Comprehensive Rust 🦀\n"
-"POT-Creation-Date: 2023-09-24\n"
+"POT-Creation-Date: 2023-10-15T23:41:49Z\n"
 "PO-Revision-Date: 2023-06-06 13:18+0900\n"
 "Last-Translator: Kenta Aratani <kentaaratani@coinez.jp>\n"
 "Language-Team: Japanese <translation-team-ja@lists.sourceforge.net>\n"
@@ -52,8 +52,8 @@ msgstr "ローカル環境での実行"
 msgid "Day 1: Morning"
 msgstr "Day 1： AM"
 
-#: src/SUMMARY.md:19 src/SUMMARY.md:79 src/SUMMARY.md:137 src/SUMMARY.md:195
-#: src/SUMMARY.md:221 src/SUMMARY.md:271
+#: src/SUMMARY.md:19 src/SUMMARY.md:80 src/SUMMARY.md:135 src/SUMMARY.md:193
+#: src/SUMMARY.md:219 src/SUMMARY.md:269
 msgid "Welcome"
 msgstr "Welcome"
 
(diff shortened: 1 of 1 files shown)


henrif75:
Hi @DannyRavi thanks for your PR!
I think there's an issue with the sample code in one of your files. You can find it out by running `MDBOOK_BOOK__LANGUAGE=fa mdbook test` from the root of your project directory.

po/fa.po
```diff
@@ -1881,12 +1953,21 @@ msgid ""
 "Hello, world!\n"
 "```"
 msgstr ""
+"```shell\n"
+"$ cd exercise\n"
+"$ cargo run\n"
+"   Compiling exercise v0.1.0 (/home/mgeisler/tmp/exercise)\n"
+"    Finished dev [unoptimized + debuginfo] target(s) in 0.75s\n"
+"     Running `target/debug/exercise`\n"
+"Hello, world!"
```
mgeisler, line 1962: This code block seems to not be closed:

```suggestion
"Hello, world!"
"```"
```

mgeisler: Hey @DannyRavi, I cannot edit your branch, otherwise I would have applied this change myself. I checked locally that it makes `MDBOOK_BOOK__LANGUAGE=fa mdbook test` work without error.

mgeisler:
Hi @DannyRavi, can you please merge `main` into your branch? You will get a small merge conflict around the POT-Creation-Date and PO-Revision-Date fields. You should be able to resolve this by hand. The POT-Creation-Date field is updated when you `msgmerge` new strings into your PO file. It is set by the latest version of mdbook-i18n-helpers, so make sure to `cargo install mdbook-i18n-helpers` as well to get version 0.2.4.

Then follow the [preparation steps](https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md#preparation) for translations to install both `msgcat` and `dprint`.

This will let you run `dprint fmt` to auto-format the `fa.po` file. Push the whole thing to your branch so you can see that the checks pass.

DannyRavi:
Hi @mgeisler 
I made a small translation on the  "fa.po"  file and used the "dprint fmt" command, and the problem of line 1027 in the "fa.po" file was solved.
Also, there were problems to install and use "dprint fmt", which was solved by using a proxy.
thank  you.

mgeisler:
> Hi @mgeisler I made a small translation on the "fa.po" file and used the "dprint fmt" command, and the problem of line 1027 in the "fa.po" file was solved. Also, there were problems to install and use "dprint fmt", which was solved by using a proxy. thank you.

I'm glad you could figure it out! Let us know if we need to amend the setup instructions in any way.

I notice you closed the PR? Is that on purpose because you will open a new one? It's completely fine to open a new one — we should then get @javad-jafari added as a reviewer so the two of you can help each other out.

DannyRavi:
Hi @mgeisler 
I probably made a mistake when I closed this Pull request
I can also cooperate with @javad-jafari.

DannyRavi:
Also, I will do a review on @javad-jafari 's work and if it is approved, we will add more persian translations.

PR #1213, opened 2023-09-16
Replace `binary_search` example with `try_into`

mgeisler:
I don't think `binary_search` is the best example of using `Result` — it should actually use an `Either` type instead.

### DIFF
--- a/src/std/option-result.md
+++ b/src/std/option-result.md
@@ -8,7 +8,7 @@ fn main() {
     let first: Option<&i8> = numbers.first();
     println!("first: {first:?}");
 
-    let arr: Result<[i8, 3], Vec<i8>> = numbers.try_into();
+    let arr: Result<[i8; 3], Vec<i8>> = numbers.try_into();
     println!("arr: {arr:?}");
 }
 ```
--- a/src/std/option-result.md
+++ b/src/std/option-result.md
@@ -8,8 +8,8 @@ fn main() {
     let first: Option<&i8> = numbers.first();
     println!("first: {first:?}");
 
-    let idx: Result<usize, usize> = numbers.binary_search(&10);
-    println!("idx: {idx:?}");
+    let arr: Result<[i8, 3], Vec<i8>> = numbers.try_into();
+    println!("arr: {arr:?}");
 }
 ```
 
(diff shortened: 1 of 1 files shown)


src/std/option-result.md
```diff
@@ -8,8 +8,8 @@ fn main() {
     let first: Option<&i8> = numbers.first();
     println!("first: {first:?}");
 
-    let idx: Result<usize, usize> = numbers.binary_search(&10);
-    println!("idx: {idx:?}");
+    let arr: Result<[i8, 3], Vec<i8>> = numbers.try_into();
```
mgeisler, line 11: Ah, Rust is hard... :smile: 

```suggestion
    let arr: Result<[i8; 3], Vec<i8>> = numbers.try_into();
```

src/std/option-result.md
```diff
@@ -18,8 +18,8 @@ fn main() {
 * `Option` and `Result` are widely used not just in the standard library.
 * `Option<&T>` has zero space overhead compared to `&T`.
 * `Result` is the standard type to implement error handling as we will see on Day 3.
-* `binary_search` returns `Result<usize, usize>`.
-  * If found, `Result::Ok` holds the index where the element is found.
-  * Otherwise, `Result::Err` contains the index where such an element should be inserted.
+* `try_into` attempts to convert the vector into a fixed-sized array. This can fail:
+  * If the vector has the right size, `Result::Ok` is returned with the array.
+  * Otherwise, `Result::Err` is returend with the original vector.
```
mgeisler, line 23: Thanks typo-checker :smile: 

```suggestion
  * Otherwise, `Result::Err` is returned with the original vector.
```

randomPoison (approved):
I like this improvement, it was a bit non-obvious why `binary_search` was returning a `Result` so this should be clearer 🎉

mgeisler:
> I like this improvement, it was a bit non-obvious why `binary_search` was returning a `Result` so this should be clearer 🎉

Yeah, I don't know what I was thinking when I added `binary_search` originally :smile:

djmitche:
BTW, in v2 I split up Option and Result, and since references aren't introduced until day 3 I made different choices:

```rust,editable,should_panic
fn main() {
    let name = "Löwe 老虎 Léopard Gepardi";
    let mut position: Option<usize> = name.find('é');
    println!("find returned {position:?}");
    assert_eq!(position.unwrap(), 14);
    position = name.find('Z');
    println!("find returned {position:?}");
    assert_eq!(position.expect("Character not found"), 0);
}
```

I think this is a pretty typical use of Option as a return value -- either an answer or a claim there's no answer. And it doesn't involve references. And shows off `.expect` and `.unwrap`.

```rust,editable
use std::fs::File;
use std::io::Read;

fn main() {
    let file: Result<File, std::io::Error> = File::open("diary.txt");
    match file {
        Ok(mut file) => {
            let mut contents = String::new();
            if let Ok(bytes) = file.read_to_string(&mut contents) {
                println!("Dear diary: {contents} ({bytes} bytes)");
            } else {
                println!("Could not read file content");
            }
        },
        Err(err) => {
            println!("The diary could not be opened: {err}");
        }
    }
}
```

This also feels like a pretty typical `Result` when doing something clearly fallible. It also nicely sets up the indentation horror show that occurs if you don't have `let else` or `?`.

PR #1216, opened 2023-09-18
[DRAFT] Example

sakex:


src/index.md
```diff
@@ -4,6 +4,13 @@
 [![GitHub contributors](https://img.shields.io/github/contributors/google/comprehensive-rust?style=flat-square)](https://github.com/google/comprehensive-rust/graphs/contributors)
 [![GitHub stars](https://img.shields.io/github/stars/google/comprehensive-rust?style=flat-square)](https://github.com/google/comprehensive-rust/stargazers)
 
+<LanguagePicker>
```
sakex, line 7: Shows that it works inside of markdown too

book.toml
```diff
@@ -82,3 +82,22 @@ line-numbers = true
...
+optional = true
+
+[output.i18n.languages]
+"en" = "English"
+"es" = "Spanish (Español)"
+"ko" = "Korean (한국어)"
+"pt-BR" = "Brazilian Portuguese (Português do Brasil)"
+
+[output.tera-backend]
+after = "html"
```
sakex, line 98: Doesn't do anything. I will remove this. The order is actually alphabetical. I will open a PR on mdbook to document it when I have time

book.toml
```diff
@@ -82,3 +82,22 @@ line-numbers = true
...
+components = [
+    { name = "Macros", path = "src/components/macros.html" },
```
sakex, line 100: I put them in `src` but `theme` would probably be better

src/components/language_picker.html
```diff
@@ -0,0 +1,39 @@
+{% import "Macros" as macros %}
```
sakex, line 1: Macros work (import the name defined in the `book.toml` and not the file name. You could name it with the file name too)

src/components/language_picker.html
```diff
@@ -0,0 +1,39 @@
...
+<button id="language-toggle{{ counter }}" class="icon-button" type="button"
+    title="Change language" aria-label="Change language"
+    aria-haspopup="true" aria-expanded="false"
+    aria-controls="language-list{{ counter }}" style="{{ attributes | get(key="style", default="")  }}">
```
sakex, line 5: Reading attributes passed in HTML

src/components/language_picker.html
```diff
@@ -0,0 +1,39 @@
...
+    <i class="fa fa-globe"></i>
+</button>
+<ul id="language-list{{ counter }}" class="theme-popup" aria-label="Languages"
+    role="menu" style="left: auto; right: 10px;">
+    {% for identifier, language_name in get_context(key="output.i18n.languages") %}
+    <li role="none">
+        <a id="{{ identifier }}"
+            href="{{ macros::get_rendered_path(identifier=identifier) }}"
```
sakex, line 13: Calling the macro

src/components/language_picker.html
```diff
@@ -0,0 +1,39 @@
...
+            style="color: inherit;">
+            <button role="menuitem" class="theme {% if identifier == language %} theme-selected {% endif %}">
+                {% include "Included" %} {{ language_name }}
```
sakex, line 16: Include directive

src/components/macros.html
```diff
@@ -0,0 +1,3 @@
+{% macro get_rendered_path(identifier) %}
```
sakex, line 1: Defining macro

theme/index.hbs
```diff
@@ -165,59 +165,9 @@
...
-                          for (let lang of langList.querySelectorAll("a")) {
-                              if (lang.id == "en") {
-                                  lang.href = `${full_path_to_root}${path}`;
-                              } else {
-                                  lang.href = `${full_path_to_root}${lang.id}/${path}`;
-                              }
-                          }
-                        </script>
-
+                        <LanguagePicker style="color: red;">
```
sakex, line 169: `<LanguagePicker />` syntax is not supported

src/components/language_picker.html
```diff
@@ -0,0 +1,39 @@
...
+            </button>
+        </a>
+    </li>
+    {% endfor %}
+</ul>
+
+<script>
+    let langToggle = document.getElementById("language-toggle{{ counter }}");
+    let langList = document.getElementById("language-list{{ counter }}");
+    {% raw %}
```
sakex, line 26: We need a raw block for the Javascript block

src/index.md
sakex: Click events don't work, it is because of mdbook that prevents bubbling in nested HTML. We might be able to fix this with z-index. I don't think it is a priority right now.

sakex: We can click on code blocks, so there definitely is a way anyhow

book.toml
```diff
@@ -82,3 +82,22 @@ line-numbers = true
...
+    { name = "Included", path = "src/components/include_language_picker.html" },
+    { name = "LanguagePicker", path = "src/components/language_picker.html", dependencies = ["Included", "Macros"] },
```
sakex, line 102: We need to declare dependencies here to include components or import macros. The syntax is the same for both

theme/index.hbs
sakex: For now `tera` syntax is not supported in `index.hbs` but I will add it later, it is going to look a bit ugly though.

book.toml
mgeisler: I agree: `mdbook` considers `src/` it's own directory and it will for example copy files it finds there to the destination directory.

src/components/language_picker.html
mgeisler: I see! This is the difference in thinking: I was expecting the user to declare an include directory and then rely solely on the mechanisms in the template engine (import a file by it's relative path, for example).

I believe you've instead built more infrastructure to make it easier to use — but the problem with that is that the user now has to understand the custom logic here (and you and I have to document and maintain this).

src/components/language_picker.html
mgeisler: Okay, but then I think the `{{` below has a `{` too many?

src/components/language_picker.html
sakex: I want to support both options, but to do use a macro like this in `index.hbs`, you will need to write sometimes 4 `{` in a row `{{{{ attribute | get(key="style") }}}}`, I find that very clunky, when using HTML syntax is pretty easy to understand for people already writing HTML.

sakex: You might even have to write 6 in a row when I think about it

src/components/language_picker.html
sakex: Yes, that's a mistake actually (from copy pasting the HBS one), but still valid JS

book.toml
mgeisler: Why can you not just include all files from a directory directly? Make them all available to Tera?

src/components/language_picker.html
mgeisler: This should go away in a MVP — I would like to keep this super simple at first.

src/components/language_picker.html
mgeisler: > I want to support both options

Please don't: avoid building up your own vocabulary here since nobody but you will know about it :slightly_smiling_face: I would like us to have as little code to maintain as possible here.

src/index.md
mgeisler: I guess we can take this out?

theme/head.hbs
```diff
@@ -8,7 +8,7 @@
   gtag('js', new Date());
   gtag('config', 'G-ZN78TEJMRW');
 </script>
-
+
```
mgeisler, line 11:
```suggestion

```

book.toml
sakex: It's an outdated comment it's not like that anymore

PR #1217, opened 2023-09-18
Consolidate editor recommendations in main text

mgeisler:
This streamlines the text about editors by simply mentioning them. RustRover is mentioned last since it's new and I don't have any experience with it yet.


djmitche (approved):
Much nicer, thanks!

PR #1222, opened 2023-09-19
it: Refresh translation

detro:
As per https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md#refreshing-an-existing-translation.

There is a bunch of stuff to correct, but I wanted to start by submitting this as it contains plenty of changes already.


mgeisler (approved):
Grazie!

I looked it over with `gh pr diff 1222 | bat -l patch` locally and it looks good.

PR #1224, opened 2023-09-19
Add mdbook-course to handle parsing frontmatter

djmitche:
In v2 of the course, I'd like to include an estimate of the time to be spent on each segment in the Markdown file. I think a good place for such metadata is in the frontmatter.

For review purposes, though, I just want to display that information. So, this is a start at a new mdbook preprocessor that just separates out the frontmatter and includes it in a `<pre>` block. Eventually, I'd like to parse it and put the time in the speaker notes.

### DIFF
--- a/mdbook-course/README.md
+++ b/mdbook-course/README.md
@@ -1,7 +1,7 @@
 # mdbook-course
 
-This is an mdBook preprocessor to handle some specific details of the
-Comprehensive-Rust course.
+This is an mdBook preprocessor to handle some specific details of
+Comprehensive Rust.
 
 ## Frontmatter
 
--- a/mdbook-course/README.md
+++ b/mdbook-course/README.md
@@ -11,7 +11,7 @@ aid review of the new course, it places this content in a `<pre>` block.
 
 ## Future Work
 
-* Parse the `minutes` property from frontmatter and
-  * Generate a course timeline
-  * Include timing information in the speaker notes
-* Generate per-segment tables of contents.
+- Parse the `minutes` property from frontmatter and
+  - Generate a course timeline
+  - Include timing information in the speaker notes
+- Generate per-segment tables of contents.
--- a/mdbook-course/README.md
+++ b/mdbook-course/README.md
@@ -1,0 +1,17 @@
+# mdbook-course
+
+This is an mdBook preprocessor to handle some specific details of the
+Comprehensive-Rust course.
+
+## Frontmatter
+
+The preprocessor parses "frontmatter" -- YAML between `---` at the beginning of
+a Markdown file -- and removes it from the rendered result. At the moment, to
+aid review of the new course, it places this content in a `<pre>` block.
+
+## Future Work
+
+* Parse the `minutes` property from frontmatter and
+  * Generate a course timeline
+  * Include timing information in the speaker notes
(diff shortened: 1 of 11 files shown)


mgeisler:
Hey @djmitche, the new preprocessor seems similar to https://github.com/dvogt23/mdbook-yml-header (from [Third Party Plugins](https://github.com/rust-lang/mdBook/wiki/Third-party-plugins)), though that preprocessor seems to just throw away the front-matter.

mdbook-course/README.md
```diff
@@ -0,0 +1,17 @@
+# mdbook-course
+
+This is an mdBook preprocessor to handle some specific details of the
+Comprehensive-Rust course.
```
mgeisler, line 4: I'm trying to use "Comprehensive Rust" as a noun:

```suggestion
This is an mdBook preprocessor to handle some specific details of
Comprehensive Rust.
```

mdbook-course/src/main.rs
```diff
@@ -0,0 +1,46 @@
...
+
+fn main() {
+    pretty_env_logger::init();
+    let app = Command::new("mdbook-course")
+        .about("Comprehensive-Rust-specific mdbook preprocessor")
+        .subcommand(Command::new("supports").arg(Arg::new("renderer").required(true)));
+    let matches = app.get_matches();
+
+    if let Some(_) = matches.subcommand_matches("supports") {
+        // Support all versions of mdbook.
```
mgeisler, line 29: Tiny point: This is not actually about versions, it's about the [supporting the renderer](https://rust-lang.github.io/mdBook/for_developers/preprocessors.html#hooking-into-mdbook).

mdbook-course/src/main.rs
```diff
@@ -0,0 +1,46 @@
...
+use clap::{Arg, Command};
+use mdbook::preprocess::CmdPreprocessor;
+use mdbook_course::frontmatter::remove_frontmatter;
+use std::io::{stdin, stdout};
+use std::process;
+
+fn main() {
+    pretty_env_logger::init();
+    let app = Command::new("mdbook-course")
+        .about("Comprehensive-Rust-specific mdbook preprocessor")
```
mgeisler, line 24:
```suggestion
        .about("mdbook preprocessor for Comprehensive Rust")
```

mgeisler (approved):
I'm curious to see where you can take this.

djmitche:
> Hey @djmitche, the new preprocessor seems similar to https://github.com/dvogt23/mdbook-yml-header (from [Third Party Plugins](https://github.com/rust-lang/mdBook/wiki/Third-party-plugins)), though that preprocessor seems to just throw away the front-matter.

I looked at that crate, and yes -- it's basically the no-op preprocessor with regex-based discarding. I thought it better to use `matter` for that purpose, and I do intend to extend this implementation to actually parsing the frontmatter. So choosing `mdbook-yml-header` seemed a poor choice.

mdbook-course/src/main.rs
djmitche: Oh, I was confusing this with the renderer's version-support check!

mgeisler (approved):
Looks good, and I like that this should be safe for the existing pages since they don't have any front-matter.

PR #1231, opened 2023-09-21
Add a pattern-matching exercise

### LINKED ISSUE #1228: Create pattern matching exercise
We need an exercise for Day 1 afternoon, most likely something about pattern matching.

djmitche:
This adds a second day-1 afternoon exercise, drawn from the v2 work. It illustrates general use of an enum and also previews `Result`.

Fixes #1228.


mgeisler (approved):
Thank you, that's beautiful!

PR #1235, opened 2023-09-22
Revert "Update boot-library.md"

mgeisler:
Reverts google/comprehensive-rust#225. We normally include links at the start of each exercise blocks — but we don't include links in the individual exercises.

If we want to add such links, we should do it consistently across the entire course.


mgeisler:
This will fix the broken link reported by @njr0 in #1234.

djmitche (approved):

PR #1236, opened 2023-09-22
Build translations from `mdbook-i18-helpers`

sakex:
Supports `mdbook build` and `mdbook serve`. I'd recommend waiting a bit before submitting and let translators try it by manually patching this PR.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -41,19 +41,7 @@ jobs:
         run: mdbook build -d book
 
       - name: Zip exercise templates
-        run: cd book/exerciser && zip --recurse-paths ../html/comprehensive-rust-exercises.zip comprehensive-rust-exercises/
-
-      - name: Build all translations
-        run: |
-          for po_lang in ${{ env.LANGUAGES }}; do
-              echo "::group::Building $po_lang translation"
-              MDBOOK_BOOK__LANGUAGE=$po_lang \
-              MDBOOK_OUTPUT__HTML__SITE_URL=/comprehensive-rust/$po_lang/ \
-              mdbook build -d book/$po_lang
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -41,19 +41,7 @@ jobs:
...
-          for po_lang in ${{ env.LANGUAGES }}; do
-              echo "::group::Building $po_lang translation"
-              MDBOOK_BOOK__LANGUAGE=$po_lang \
-              MDBOOK_OUTPUT__HTML__SITE_URL=/comprehensive-rust/$po_lang/ \
-              mdbook build -d book/$po_lang
-              (cd book/$po_lang/exerciser && zip --recurse-paths ../html/comprehensive-rust-exercises.zip comprehensive-rust-exercises/)
-              mv book/$po_lang/html book/html/$po_lang
-              echo "::endgroup::"
-          done
+        run: cd book && find . -name "exerciser" -type d -exec bash -c 'zip --recurse-paths ./html/$(dirname {})/comprehensive-rust-exercises.zip {}/comprehensive-rust-exercises/' \;
```
randomPoison, line 44: I don't think this is doing the same thing that the original logic was doing. It looks like the original was putting the archive at `book/$po_lang/html/comprehensive-rust-exercises.zip`, but with the version you have here they would end up at `book/html/$po_lang/comprehensive-rust-exercises.zip`. Is this change intentional?

max-heller:
Likely relevant to this is that the publish workflow now passes additional configuration for each language to a mdbook backend (`mdbook-pandoc`) to render PDFs. We should be sure this maintains the same functionality:

https://github.com/google/comprehensive-rust/blob/3afdabf391c833f1a4a4785989fd93224dc3ff3f/.github/workflows/build.sh#L33-L42

PR #1237, opened 2023-09-22
Fix solution for Storing Books exercise

mgeisler:
As pointed out by @njr0 in #1233, we were making the exercise confusing by showing people code that cannot work — and then expecting the course participants to somehow fix this, without setting clear boundaries for what can and cannot be modified.

This PR should align the exercise with the other exercises in the course and avoid the “brain teaser” here.

This also has the advantage of having a full working solution, with no commented code which will bit-rot over time.

### DIFF
--- a/src/exercises/day-2/book-library.md
+++ b/src/exercises/day-2/book-library.md
@@ -25,14 +25,24 @@ Use this to model a library's book collection. Copy the code below to
     }
 
 {{#include book-library.rs:Library_len}}
+        todo!("Return the length of `self.books`")
+    }
 
 {{#include book-library.rs:Library_is_empty}}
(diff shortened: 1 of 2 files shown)


njr0:
Looks good to me. Thanks! (Super impressed at the responsiveness, too!)

djmitche (approved):
Nice!

mgeisler:
> Looks good to me. Thanks! (Super impressed at the responsiveness, too!)

Thanks! We're super reliant on people like you who go the extra mile to explain and debug problems in the course — it has grown to a size where I can no longer keep it all in my head 🙂 The extensive CI tests makes it easy to confidently update the code.

PR #1241, opened 2023-09-23
Add a "minimum wrong example" program in C

ColinFinck:
This example shows what kind of bugs easily slip into C code and are made impossible via Rust.

I have created this example for the motivation slides of my master thesis. Since then, the university institute has used it in their introduction to Rust.
I hereby make this part available to the Comprehensive Rust course and relicense it under the terms of the Apache 2.0 license.

Thank you @mgeisler for your invitation to contribute to this course!


mgeisler:
Oh, that's amazing! Thanks for writing this up @ColinFinck!

I'll look at it tomorrow when I'm back in the office — I want to try compiling the C code with our regular `clang` compiler options to see if it's also warning free in that configuration :smile:

mgeisler:
> I want to try compiling the C code with our regular `clang` compiler options to see if it's also warning free in that configuration 😄

The Android build system enables more warnings, so the problems are no longer completely undetected. I'll have to play a bit more with it to see how we can adapt the example into a real and useful one for our environment.

ColinFinck:
> > I want to try compiling the C code with our regular `clang` compiler options to see if it's also warning free in that configuration 😄
> 
> The Android build system enables more warnings, so the problems are no longer completely undetected. I'll have to play a bit more with it to see how we can adapt the example into a real and useful one for our environment.

Yes, some of these problems *can* be detected by a C/C++ compiler if you enable additional flags. And I'm not surprised that the build system of Android or any comparably large project comes with these flags enabled.

But what matters here are the compiler default settings.
Rust detects all problems of my example code by default. If you import any Rust project, no matter if small or large, in-house or foreign code, you can reasonably expect it to compile warning-free on the official Rust compiler with the default compiler flags. When that is the case, you can also be sure that the code is immune from the problems of my example.

In contrast to that, the Android build system compiler flags are only applied to existing C/C++ code in your Android tree. That surely raises the quality of in-house code that is already part of your tree. But what do you do when you import a foreign library into your tree and it raises a warning?
Sure, you could start fixing warnings in foreign code, but this [has caused security havoc in the past](https://research.swtch.com/openssl) and you are therefore discouraged to do it on your own. Furthermore, each C/C++ compiler has their own set of flags and warnings. Fixing all warnings for one compiler may not be sufficient for pleasing the next compiler. All in all, eliminating C/C++ compiler warnings can be a never-ending task due to the endless possible combinations of compilers and flags.

As a result, I usually see large C/C++ projects compile with lots of warnings by default. Consequently, when warnings are the norm, you cannot even be sure that the code is free from basic problems, like the ones in my example.

I hope that this explanation makes sense, and makes my example relevant to large projects like Android.

src/SUMMARY.md
```diff
@@ -24,6 +24,7 @@
   - [Compile Time Guarantees](why-rust/compile-time.md)
   - [Runtime Guarantees](why-rust/runtime.md)
   - [Modern Features](why-rust/modern.md)
+  - [An Example in C](why-rust/an-example-in-c.md)
```
mgeisler, line 27: I would like to start the section with the example. I think it will work very well as a fun and provocative example:

```suggestion
  - [An Example in C](why-rust/an-example-in-c.md)
  - [Compile Time Guarantees](why-rust/compile-time.md)
  - [Runtime Guarantees](why-rust/runtime.md)
  - [Modern Features](why-rust/modern.md)
```

mgeisler:
> Yes, some of these problems _can_ be detected by a C/C++ compiler if you enable additional flags. And I'm not surprised that the build system of Android or any comparably large project comes with these flags enabled.

You have good points, thanks!

Let us try including it as-is and then we'll adjust based on experience when @fw-immunant, @randomPoison, @sakex, and others teach the class next.

src/SUMMARY.md
ColinFinck: > I would like to start the section with the example. I think it will work very well as a fun and provocative example:

As you like it.
Feel free to even rename the section "A Provocative Example in C" or something like that. Honestly, no fitting adjective came to mind when I had to think about the section title.

PR #1243, opened 2023-09-24
Sync `src/` with date in PO file when publishing

mgeisler:
This will freeze translation in place: they will keep using the same English Markdown source files as the starting point until a new POT file is merged into the translation.

We still update all the files around the Markdown files: this allows us to fix things in the theme, for example.

Part of https://github.com/google/mdbook-i18n-helpers/issues/16. The logic here should eventually be moved to somewhere in mdbook-i18n-helpers, most likely to the renderer that @sakex is building in https://github.com/google/mdbook-i18n-helpers/pull/84.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -30,6 +30,9 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+        with:
+          # We need the full history below.
+          fetch-depth: 0
 
       - name: Setup Rust cache
         uses: ./.github/workflows/setup-rust-cache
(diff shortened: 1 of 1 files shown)


mgeisler:
Hi @sakex, I would like to have this PR merged ahead of https://github.com/google/mdbook-i18n-helpers/pull/84 — it will stop the bleeding of us constantly invalidating the work done by our translators.

As I write in the PR description, you should be able to implement the same logic in https://github.com/google/mdbook-i18n-helpers/pull/84 (and do it more robustly and cleanly since you can part the PO file more easily than my shell script).

sakex (approved):

mgeisler:
@henrif75, this change will have (good) implications for the translations. Translators will be able to freeze the English source files used when publishing the translation.

This means that they will effectively be working on a snapshot of the course — the snapshot is found based on the `POT-Creation-Date` header found in each PO file. This header is updated by `msgmerge` and it will soon be generated by `mdbook-xgettext`: https://github.com/google/mdbook-i18n-helpers/pull/87.

I will write this up in our translation documentation as well.

Issue #1249, opened 2023-09-24
Fix resolver setting


mgeisler:
Our CI runs are outputting the following every time `mdbook-course` and `mdbook-exerciser` are installed:

```
cargo install --path mdbook-exerciser --locked
  
warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
```

We should probably follow the advice given here to silence the warnings.

friendlymatthew:
Hi may I take on this issue?

mgeisler:
Hi @friendlymatthew (such a cool username!), of course, it's all yours!

friendlymatthew:
@mgeisler  Thanks :)

I specified the resolver within the base `Cargo.toml` file. I just signed the CLA so hopefully it gets approved quick!

In the mean time, I'll look for other work :)

mgeisler:
Awesome, thanks again @friendlymatthew! Perhaps you would be for fixing some annoyences in the mdbook-exerciser tool we use? See #625 for that. Otherwise just look around at the many open issues here and in [mdbook-i18n-helpers](https://github.com/google/mdbook-i18n-helpers) :smile:

PR #1253, opened 2023-09-25
Fix resolver setting

friendlymatthew:
Fixes: https://github.com/google/comprehensive-rust/issues/1249

Since `mdbook-exerciser` and `mdbook-course` are using Rust edition 2021, they naturally want to use the newer resolver (resolver = 2). 
Screen Shot 2023-09-25 at 9 57 32 AM


djmitche:
Thanks for fixing this!

PR #1255, opened 2023-09-26
Clarify distinction between Drop trait and drop function.

qwandor:
This clears up some misleading explanation from #246.


mgeisler (approved):
Great, thanks for expanding on this!

PR #1264, opened 2023-09-27
da: refresh translation and fix fuzzy entries

mgeisler:
There are two commits, I hope they're individually small enough that GitHub will show them.


mgeisler:
Yeah, the second commit works well in GitHub — the first commit is a clean `msgmerge` so it's fairly uninteresting.

thedataking (approved):
LGTM. Thanks for the ping and separate commits 👍

PR #1271, opened 2023-09-27
Actually print something in prime_squares example

djmitche:
As suggested in #1268.


mgeisler (approved):
Thanks, that's a good call :smile:

PR #1272, opened 2023-09-28
Automatically format YAML, CSS, and JS files

mgeisler:
Initially, I also formatted the templates in `theme/`, but this triggers https://github.com/prettier/prettier/issues/11834. So I exclude them for now.


mingyc (approved):
Thanks! LGTM.

PR #1276, opened 2023-09-28
Add unittest example to rustdoc

### LINKED ISSUE #1275: Give unit test example as part of the Markdown docs page
II think [the rustdocs page](https://google.github.io/comprehensive-rust/basic-syntax/rustdoc.html) should include a simple unit test as part of the doc. That slide is very short, yet giving a unit test example so well integrated into the documentation system early on would add a lot to the "WOW factor"

nyurik:
Fixes #1275

### DIFF
--- a/src/basic-syntax/rustdoc.md
+++ b/src/basic-syntax/rustdoc.md
@@ -6,6 +6,13 @@ All language items in Rust can be documented using special `///` syntax.
 /// Determine whether the first argument is divisible by the second argument.
 ///
 /// If the second argument is zero, the result is false.
+///
+/// # Example
+/// ```
+/// assert!(is_divisible_by(42, 2));
+/// assert!(!is_divisible_by(5, 3));
+/// assert!(!is_divisible_by(13, 0));
+/// ```
 fn is_divisible_by(lhs: u32, rhs: u32) -> bool {
     if rhs == 0 {
         return false;  // Corner case, early return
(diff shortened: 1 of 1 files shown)


src/basic-syntax/rustdoc.md
```diff
@@ -6,6 +6,13 @@ All language items in Rust can be documented using special `///` syntax.
 /// Determine whether the first argument is divisible by the second argument.
 ///
 /// If the second argument is zero, the result is false.
+///
+/// # Example
+/// ```
+/// assert!(is_divisible_by(42, 2));
+/// assert!(!is_divisible_by(5, 3));
+/// assert!(!is_divisible_by(13, 0));
```
mgeisler, line 14: Showing off a doctest is a great idea! However, I'm always afraid of losing vertical space, so let's just show it with a single line:

```suggestion
```

mgeisler (approved):
Thanks a lot, this is an awesome addition! I've used these kinds of tests for decades coming from Python, but many people have never seen them before and you're right that they go Wow! :exploding_head: when they see it :smile:

mgeisler:
Now that I see #1282, I guess the example here might be a little redundant? I guess we'll know after @fw-immunant and @randomPoison teaches the class next.

nyurik:
I was also thinking about it, but realized that it is actually for the better.  First slide gives a glimpse of the capability without details (the wow factor early on), and the second slide much later goes into details, like the hash and the use statement,   and could grow to include an example of hidden functions, etc.

nyurik:
P.S. In other words - first slide focus is on docs, and the second slide focuses on the testing

mgeisler:
> P.S. In other words - first slide focus is on docs, and the second slide focuses on the testing

Thanks, that's a good way to put it! :slightly_smiling_face:

I'm not a professional educator, but I'm told that repetition is important when learning something new. So gradually showing the capabilities is a good thing.

PR #1277, opened 2023-09-29
Introduce a glossary

mgeisler:
The goal of this is twofold: give translators a place to document how certain terms are translated as well as giving people a place to quickly find a definition of a term. The slides might not always give a quick definition the same way a glossary can.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -310,6 +310,7 @@
 ----
 
 - [Thanks!](thanks.md)
+- [Glossary](glossary.md)
 - [Other Resources](other-resources.md)
 - [Credits](credits.md)
 
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -1,0 +1,76 @@
+# Glossary
+
+The following is a glossary which aims to give a short definition of many Rust
+terms. For translations, this also serves to connect the term back to the
+English original.
+
+<!-- Translators: please list the English term in italic after your translation.
+For Danish, this would look like this: `allokere (_allocate_)`. -->
+
+| Term                  | Notes                                                                        |
+| --------------------- | ---------------------------------------------------------------------------- |
+| allocate              | Dynamic memory allocation on [the heap](memory-management/stack-vs-heap.md). |
+| argument              |                                                                              |
+| Bare-metal Rust       | See [Bare-metal Rust](bare-metal.md).                                        |
+| block                 | See [Blocks](control-flow/blocks.md) and _scope_.                            |
+| borrow                | See [Borrowing](ownership/borrowing.md).                                     |
+| borrow checker        | The part of the Rust compiler which checks that all borrows are valid.       |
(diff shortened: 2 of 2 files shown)


mgeisler:
@kochinc, @victorhsieh, @hueich from #1240: please let me know if you think this will be useful?

kochinc:
It will be helpful for both readers and translators. The glossary can be used as a reading aid for readers. Translators can put translations and their original terms side-by-side as a standard reference for anyone wanting to help with the translation.

However, there will be an issue of how it should be sorted in various languages. Do we always sort the terms based on the original English phrases or based on the translated phrases using the language's collation? Or. maybe this should another discussion among translators.

From a translator's point of view, for now the translators only need a table of terms and their translations to standardize the translations throughout the book.

victorhsieh:
+1 to what @kochinc said.

(and, if only the user can click the table header to sort by column... :)

mgeisler:
> However, there will be an issue of how it should be sorted in various languages. Do we always sort the terms based on the original English phrases or based on the translated phrases using the language's collation? Or. maybe this should another discussion among translators.

Good point, I had not thought of this... unfortunately, the translations will not be able to sort the table differently than the original — because of how we extract the text cell-by-cell.

We don't have it right now, but it would be possible to add some JavaScript to make the tables sortable.

> From a translator's point of view, for now the translators only need a table of terms and their translations to standardize the translations throughout the book.

Yeah, the question is really if it's helpful to move these glossaries into the course material. Several translation teams have developed them and we might just add extra friction by making them part of the Markdown source?

mgeisler:
Alright, let us try introducing this! I only filled out a few entries here, I hope to crowd-source the rest of the entries after merging it.

victorhsieh (approved):
nit: Do we have a convention/requirement for max column, e.g. 100?

I was thinking if we do, we could avoid changing all the lines / line history when we need to adjust the width.

mgeisler:
> nit: Do we have a convention/requirement for max column, e.g. 100?
> 
> I was thinking if we do, we could avoid changing all the lines / line history when we need to adjust the width.

I will eventually reformat the Markdown: #1157. This is waiting a bit because it will make the v2 rewrite harder.

That being said, line lengths are a concern for me: Markdown tables don't support wrapping of the table cell content. So there is very limited space for us to express ourselves in the English original.

I was considering using a list format instead because of this. It would look better if the English text becomes longer than a one line paragraph.

Now that I think of it, using list-loke formatting would be more in tune with how a paper dictionary looks... that's probably something to experiment with after merging this.

mgeisler:
> I was thinking if we do, we could avoid changing all the lines / line history when we need to adjust the width.

On re-reading this, you have a great point here as well: the auto formatting will indeed adjust the entire Markdown table to make it look nice — resulting in big diffs when we make any of the lines longer. A list would avoid this too.

PR #1282, opened 2023-09-29
Add hash to rustdoc example

nyurik:
Hashes are important (and somewhat confusing to novices) part of the rustdoc.

### DIFF
--- a/src/testing/doc-tests.md
+++ b/src/testing/doc-tests.md
@@ -6,7 +6,7 @@ Rust has built-in support for documentation tests:
 /// Shortens a string to the given length.
 ///
 /// ```
-/// use playground::shorten_string;
+/// # use playground::shorten_string;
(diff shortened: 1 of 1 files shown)


src/testing/doc-tests.md
```diff
@@ -17,4 +17,5 @@ pub fn shorten_string(s: &str, length: usize) -> &str {
 
 * Code blocks in `///` comments are automatically seen as Rust code.
 * The code will be compiled and executed as part of `cargo test`.
+* Adding `#` in the code will hide it from the docs, but will still compile/run it.
```
djmitche, line 20:
```suggestion
* Adding `# ` in the code will hide it from the docs, but will still compile/run it.
```
IIRC the space is important

mgeisler (approved):
Nice!

PR #1283, opened 2023-09-30
ja: Translate Chapter 24 (Generics)

kantasv:
Hi JP translation folks (#652). Here's a MR for Generics chapter. Could you take a look at it? Thank you!

### DIFF
--- a/po/ja.po
+++ b/po/ja.po
@@ -8527,10 +8527,13 @@ msgid ""
 "Rust support generics, which lets you abstract algorithms or data structures "
 "(such as sorting or a binary tree) over the types used or stored."
 msgstr ""
+"Rustはジェネリクス（generics）をサポートします。これにより、使用または保存す"
+"る型に関してアルゴリズムやデータ構造（ソートアルゴリズムや、二分木など）を抽"
+"象化することができます。"
 
 #: src/generics/data-types.md:3
 msgid "You can use generics to abstract over the concrete field type:"
-msgstr ""
+msgstr "ジェネリクスを使って、平面を表す具象型を抽象化することができます："
 
 #: src/generics/data-types.md:5
 msgid ""
@@ -8552,14 +8555,16 @@ msgstr ""
 #: src/generics/data-types.md:21
 msgid "Try declaring a new variable `let p = Point { x: 5, y: 10.0 };`."
 msgstr ""
+"次のような新しい変数を宣言してみてください `let p = Point { x: 5, y: 10.0 };"
+"`."
 
 #: src/generics/data-types.md:23
 msgid "Fix the code to allow points that have elements of different types."
-msgstr ""
+msgstr "異なる型の要素を持つ点を許容するように、コードを修正してください。"
 
 #: src/generics/methods.md:3
 msgid "You can declare a generic type on your `impl` block:"
-msgstr ""
+msgstr "`impl`ブロックを用いて、ジェネリック型を宣言することができます："
 
 #: src/generics/methods.md:5
 msgid ""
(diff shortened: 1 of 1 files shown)


po/ja.po
```diff
@@ -8527,10 +8527,13 @@ msgid ""
...
 "(such as sorting or a binary tree) over the types used or stored."
 msgstr ""
+"Rustはジェネリクス（generics）をサポートします。これにより、使用または保存す"
+"る型に関してアルゴリズムやデータ構造（ソートアルゴリズムや、二分木など）を抽"
+"象化することができます。"
 
 #: src/generics/data-types.md:3
 msgid "You can use generics to abstract over the concrete field type:"
-msgstr ""
+msgstr "ジェネリクスを使って、平面を表す具象型を抽象化することができます："
```
keiichiw, line 8536: This "field" is not "平面" but "a member of the struct"

po/ja.po
```diff
@@ -8552,14 +8555,16 @@ msgstr ""
...
 
 #: src/generics/data-types.md:23
 msgid "Fix the code to allow points that have elements of different types."
-msgstr ""
+msgstr "異なる型の要素を持つ点を許容するように、コードを修正してください。"
 
 #: src/generics/methods.md:3
 msgid "You can declare a generic type on your `impl` block:"
-msgstr ""
+msgstr "`impl`ブロックを用いて、ジェネリック型を宣言することができます："
```
keiichiw, line 8567:
```suggestion
msgstr "`impl`に対して、ジェネリックな型を宣言することもできます："
```

Here, "a generic type" in this sentence is not `struct Point<T>` but `T` in `impl<T> Point<T>`. So, "implブロックを用いて" is incorrect, I think.

po/ja.po
```diff
@@ -8587,30 +8592,40 @@ msgid ""
...
 msgstr ""
+"_Q:_ なぜ`T`は２回も `impl<T> Point<T> {}` において指定されたのでしょうか？冗"
+"長ではありませんか？"
 
 #: src/generics/methods.md:26
 msgid ""
 "This is because it is a generic implementation section for generic type. "
 "They are independently generic."
 msgstr ""
+"なぜなら、これはジェネリクスに対したジェネリックな実装の箇所だからです。それ"
```
keiichiw, line 8603:
```suggestion
"なぜなら、これはジェネリクスに対してのジェネリックな実装の箇所だからです。それ"
```

keiichiw (commented):
Thanks for the pull request! Here are my review comments

po/ja.po
kantasv: That makes sense, thanks!

po/ja.po
kantasv: Thanks! The word "field" seems like a member variable or data memeber in this context. I'll transliterate the word as "フィールド" since it's a technical term :)

kantasv:
Thanks @keiichiw, I applied changes based on your review. When you have some time could you review the diff once again? Have a good day.

kantasv:
Tiny modification: In the previous commit (0d20784), I translated `the concrete field type` -> `フィールドを持つ具象型` to make it sounds more natural.

po/ja.po
```diff
@@ -8527,10 +8527,14 @@ msgid ""
...
 "(such as sorting or a binary tree) over the types used or stored."
 msgstr ""
+"Rustはジェネリクス（generics）をサポートします。これにより、使用または保存す"
+"る型に関してアルゴリズムやデータ構造（ソートアルゴリズムや、二分木など）を抽"
+"象化することができます。"
 
 #: src/generics/data-types.md:3
 msgid "You can use generics to abstract over the concrete field type:"
 msgstr ""
+"ジェネリクスを使って、フィールドを持つ具象型を抽象化することができます："
```
keiichiw, line 8537: Hmm, this sentence doesn't make much sense to me. Also, "concrete field type" cannot be translated to "フィールドを持つ具象型". 
If you are not so confident in your understanding or translation of a certain sentence, I would be happy if you could write about it in the comments or ask me for advice in the comments.

```suggestion
"ジェネリクスを使って、具体的なフィールドの型を抽象化することができます："
```

ジェネリクスがない`struct Point { x: u32, y: u32 }` といったstructのフィールドは`u32`のような具体的な型を持ちますが (concrete field type)、それを`T`として抽象化しているという意味です。

kantasv: Hi, @keiichiw-san, thanks for correcting me regarding that with great examples. In the original paragraph, I was a bit confused with struct Point and its nested fields' types, but as you mentioned looks like generics abstract those fields types in the context. From next MR on, I'll list unclear points in advance so that I can get feedback about it. Thank you!

keiichiw: Thanks!

keiichiw (approved):

PR #1286, opened 2023-10-02
Minor rewording of safe-ffi-wrapper exercise.

njr0:
For readability.

### DIFF
--- a/src/exercises/day-3/safe-ffi-wrapper.rs
+++ b/src/exercises/day-3/safe-ffi-wrapper.rs
@@ -14,9 +14,6 @@
 
 // ANCHOR: solution
 // ANCHOR: ffi
-
-use tempfile::tempfile;  // requires tempfile = "3" in Cargo.toml dependencies
-
 mod ffi {
     use std::os::raw::{c_char, c_int};
     #[cfg(not(target_os = "macos"))]
--- a/src/exercises/day-3/safe-ffi-wrapper.rs
+++ b/src/exercises/day-3/safe-ffi-wrapper.rs
@@ -14,6 +14,9 @@
 
 // ANCHOR: solution
 // ANCHOR: ffi
+
+use tempfile::tempfile;  // requires tempfile = "3" in Cargo.toml dependencies
+
 mod ffi {
     use std::os::raw::{c_char, c_int};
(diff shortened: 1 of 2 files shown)


mgeisler (approved):
Sounds good, thanks!

src/exercises/day-3/safe-ffi-wrapper.rs
```diff
@@ -14,6 +14,9 @@
 
 // ANCHOR: solution
 // ANCHOR: ffi
+
+use tempfile::tempfile;  // requires tempfile = "3" in Cargo.toml dependencies
```
mgeisler, line 18: We test the solution using the [`Cargo.toml` in `src/exercises`](https://github.com/google/comprehensive-rust/blob/main/src/exercises/Cargo.toml) and this line is not actually needed — the test code below uses the `tempfile` crate directly.

If you would want to import `tempfile::TempDir` for the tests, then you should add the `use tempfile::TempDir` line _inside_ `mod tests`. This way it becomes possible to distinguish between regular dependencies and dev-dependencies. The latter are only used by `cargo test` and won't burden other users of the code.

T

src/exercises/day-3/safe-ffi-wrapper.rs
```diff
@@ -14,6 +14,9 @@
...
+
```
mgeisler, line 19: Let me take it out so we can merge the PR directly:

```suggestion
```

mgeisler, line 19: Let me take it out so we can merge the PR directly:

```suggestion
```

njr0:
Sorry, I see what you mean now. I added the `use` first, and that didn't fix it, so then I updated the Cargo.toml. But I didn't notice the warning:
```sh
warning: unused import: `tempfile::tempfile`
 --> src/main.rs:1:5
  |
1 | use tempfile::tempfile;
  |     ^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default
```

With you now!

mgeisler:
The only missing piece here is the Google CLA — we're not allowed to import code without a signed CLA.

PR #1288, opened 2023-10-02
Ex31 1 tempfile

njr0:
Add missing `use` in model solution

### DIFF
--- a/src/exercises/day-3/safe-ffi-wrapper.md
+++ b/src/exercises/day-3/safe-ffi-wrapper.md
@@ -2,7 +2,7 @@
 
 Rust has great support for calling functions through a _foreign function
 interface_ (FFI). We will use this to build a safe wrapper for the `libc`
-functions you would use from C to read the filenames of a directory.
+functions you would use from C to read the names of files in a directory.
 
 You will want to consult the manual pages:
 
--- a/src/exercises/day-3/safe-ffi-wrapper.rs
+++ b/src/exercises/day-3/safe-ffi-wrapper.rs
@@ -14,6 +14,9 @@
 
 // ANCHOR: solution
 // ANCHOR: ffi
+
+use tempfile::tempfile;  // requires tempfile = "3" in Cargo.toml dependencies
+
 mod ffi {
     use std::os::raw::{c_char, c_int};
     #[cfg(not(target_os = "macos"))]
--- a/src/exercises/day-3/safe-ffi-wrapper.rs
+++ b/src/exercises/day-3/safe-ffi-wrapper.rs
@@ -15,7 +15,7 @@
 // ANCHOR: solution
 // ANCHOR: ffi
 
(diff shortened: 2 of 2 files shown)


mgeisler:
Thanks! However, as explained in https://github.com/google/comprehensive-rust/pull/1286#discussion_r1342749331, this should not be necessary.

I triggered the test run, but as [expected it fails](https://github.com/google/comprehensive-rust/actions/runs/6381182120/job/17317597344?pr=1288):

```
error[E0432]: unresolved import `tempfile`
  --> src/exercises/day-3/safe-ffi-wrapper.rs:18:5
   |
18 | use tempfile::tempfile;   // requires tempfile = "3" in Cargo.toml dependencies
   |     ^^^^^^^^ use of undeclared crate or module `tempfile`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `comprehensive-rust` (bin "safe-ffi-wrapper") due to previous error
warning: build failed, waiting for other jobs to finish...
```

njr0:
I'm sure you're right, but I don't understand.

If I do `cargo new ex31_1`, replace the main with your model solution and run it, it doesn't work for me because it doesn't know about the `tempfile`. (It's slightly strange, but at the start of the course you say you want to encourage students to use cargo to solve the exercises, but in most of them you recommend going to a playground instead. I have always used cargo to do the exercises.)

mgeisler:
> I'm sure you're right, but I don't understand.
> 
> If I do `cargo new ex31_1`, replace the main with your model solution and run it, it doesn't work for me because it doesn't know about the `tempfile`.

That is correct! My thinking was that people would copy the code from [the exercise page](https://google.github.io/comprehensive-rust/exercises/day-3/safe-ffi-wrapper.html) and then expand that to solve the exercise.

The unit tests found in [the solution](https://google.github.io/comprehensive-rust/exercises/day-3/solutions-afternoon.html) are actually there for my own safety: they allow me to ensure that the solution works.

I see now that this is rather confusing! We've never explained anywhere that the solutions have dependencies outside of the standard library.

> (It's slightly strange, but at the start of the course you say you want to encourage students to use cargo to solve the exercises, but in most of them you recommend going to a playground instead. I have always used cargo to do the exercises.)

Yeah, that's also a good point! Originally, I wrote the entire course centered around the Rust Playground — mostly because I was unsure about how complicated it would be to have 20 course participants install and set up Rust on Linux and Mac laptops... I wanted to avoid having to deal with random problems of that setup. However, it is of course an unnatural way to write larger amounts of Rust code, so I later added the recommendation to install Rust locally. It seems to work okay in the classes we teach as well: people normally set this up ahead of time.

So in short, we should update the boiler plate text for how the exercises are done. I created #1289 for this.

mgeisler:
#1290 is about expanding the solutions to mention their dependencies.

PR #1291, opened 2023-10-02
Link blog post from README.md

mgeisler:
I hope the blog post can help shed more light on why we wrote the course and how it has been used at Google.

I'm not sure if "Press" is the best title, but I hope to link other articles later from this section at a later point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,11 @@ trigger when going through the code samples. We hope to improve on this via
 [speaker notes](https://github.com/google/comprehensive-rust/issues/53) and by
 [publishing videos](https://github.com/google/comprehensive-rust/issues/52).
 
+## Press
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,6 +28,11 @@ trigger when going through the code samples. We hope to improve on this via
 [speaker notes](https://github.com/google/comprehensive-rust/issues/53) and by
 [publishing videos](https://github.com/google/comprehensive-rust/issues/52).
 
+## Press
```
djmitche, line 31: I think "Press" is fine, if a tiny bit cheeky since you wrote the press. This is a place someone might look if they wanted to see how to publicize this on their socials.

djmitche (approved):

README.md
mgeisler: 🙂  Maybe "News" would be better?

mgeisler: Let me go with Press for now, then we can change it later as we add more here.

PR #1294, opened 2023-10-03
Avoid naming struct field like trait method

### LINKED ISSUE #1292: Fix `name` ambiguity in trait object diagram
> So if `<Dog as Pet>::name` refers to the field `name` in the `Dog` struct, what does `<Cat as Pet>::name` refer to?

Oh, sorry, I misunderstood above. `<Dog as Pet>::name` is supposed to refer to the `Dog::name` method.
 
> If you called the field `dogname` (in the Dog struct) I think it would be clearer what's going on. Then, presumably, the label would become `<Dog as Pet>::dogname`.
> 
> It would make more sense if the `::name` refers to the function (though the function isn't really on heap, presumably and the `name: Fido` would then be `dogname: Fido`.

Yeah, that is what the picture was supposed to show. I think "the heap" as "somewhere in memory". I guess the vtable is generated at compile-time and put somewhere in the binary — and then memory-mapped into the address space when the program is loaded from disk.

> Obviously, I'm confused, but I don't think adding this information to the notes would help, because the information still seems inconsistent to me. (Sorry.)

No, I actually think you got it!

I suggest we change the `name` method to `talk` or `speak` in the `Pet` trait. The implementations can then be something like

```rust
impl Pet for Cat {
    fn speak(&self) -> String {
        String::from("Miau!")
    }
}

impl Pet for Dog {
    fn speak(&self) -> String {
        format!("Woof, my name is {}!", self.name);
    }
}
```

That would remove the `name` ambiguity. We should then also write notes to explain each part of the diagram.

_Originally posted by @mgeisler in https://github.com/google/comprehensive-rust/discussions/1270#discussioncomment-7173183_

mgeisler:
The `name` struct field was confusing because it was named the same as the trait method.

This should clean things up a little. I also compressed the code a bit to make the unimportant structs take up less space.

Fixes #1292.

### DIFF
--- a/src/traits/trait-objects.md
+++ b/src/traits/trait-objects.md
@@ -3,39 +3,31 @@
 Trait objects allow for values of different types, for instance in a collection:
 
 ```rust,editable
+struct Dog { name: String, age: i8 }
+struct Cat { lives: i8 } // No name needed, cats won't respond anyway.
+
 trait Pet {
-    fn name(&self) -> String;
+    fn talk(&self) -> String;
 }
-
-struct Dog {
-    name: String,
-}
-
-struct Cat;
 
 impl Pet for Dog {
-    fn name(&self) -> String {
-        self.name.clone()
-    }
+    fn talk(&self) -> String { format!("Woof, my name is {}!", self.name) }
 }
 
 impl Pet for Cat {
-    fn name(&self) -> String {
-        String::from("The cat") // No name, cats won't respond to it anyway.
-    }
+    fn talk(&self) -> String { String::from("Miau!") }
 }
 
 fn main() {
     let pets: Vec<Box<dyn Pet>> = vec![
-        Box::new(Cat),
-        Box::new(Dog { name: String::from("Fido") }),
+        Box::new(Cat { lives: 9 }),
+        Box::new(Dog { name: String::from("Fido"), age: 5 }),
     ];
     for pet in pets {
-        println!("Hello {}!", pet.name());
+        println!("Hello, who are you? {}", pet.talk());
     }
 }
 ```
-
 
 Memory layout after allocating `pets`:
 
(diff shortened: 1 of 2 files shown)


mgeisler:
Hi @njr0, please let me know if the explanation here is clearer?

njr0:
Looks great! Thanks!

src/traits/trait-objects.md
```diff
@@ -47,32 +39,40 @@ Memory layout after allocating `pets`:
...
-                                  :     | |   '---->| "<Dog as Pet>::name" |    :
-                                  :     | |         +----------------------+    : 
-                                  :     | |                                     : 
-                                  :     | |   +-+                               :   
-                                  :     | '-->|\|                               :     
-                                  :     |     +-+                               :    
-                                  :     |                                       : 
-                                  :     |     +----------------------+          : 
-                                  :     '---->| "<Cat as Pet>::name" |          : 
+                                  :     | |   |     +----------------------+    :
```
djmitche, line 46: Could these be something like
```
vtable:
  talk: Dog::talk
```

src/traits/trait-objects.md
```diff
@@ -47,32 +39,40 @@ Memory layout after allocating `pets`:
 :   +-----------+-------+   :     :   +-----+-----+                             :
 :   | ptr       |   o---+---+-----+-->| o o | o o |                             :
 :   | len       |     2 |   :     :   +-|-|-+-|-|-+                             :
-:   | capacity  |     2 |   :     :     | |   | |   +---------------+           :
-:   +-----------+-------+   :     :     | |   | '-->| name: "Fido"  |           :
-:                           :     :     | |   |     +---------------+           :
+:   | capacity  |     2 |   :     :     | |   | |   +-------------+             :
+:   +-----------+-------+   :     :     | |   | '-->| name: Fido  |             :
```
djmitche, line 43: Maybe `{name: "Fido"}` just to suggest struct-ness and string-ness a bit more?

djmitche (approved):
Looks good, just a few ideas for the diagram

njr0:
I think it would be good to include a mention of vtable...particularly if there were a sentence explaining what it is in the notes.

src/traits/trait-objects.md
mgeisler: I see what you mean — though the style on the right is consistent with how the `Vec` fields are shown on the left. I've used this "table style" throughout.

Now, I did try to make Fido be `"Fido"`, but I didn't manage to escape the `"` for the svgbob plugin. The double quote is used to quote special characters for the plugin.

src/traits/trait-objects.md
```diff
@@ -47,32 +39,40 @@ Memory layout after allocating `pets`:
...
-* In the example, `pets` holds *fat pointers* to objects that implement `Pet`. The fat pointer consists of two components, a pointer to the actual object and a pointer to the virtual method table for the `Pet` implementation of that particular object.
-* Compare these outputs in the above example:
+- Types that implement a given trait may be of different sizes. This makes it
+  impossible to have things like `Vec<dyn Pet>` in the example above.
+- `dyn Pet` is a way to tell the compiler about a dynamically sized type that
+  implements `Pet`.
+- In the example, `pets` is allocated on the stack and the vector data is on the
+  heap. The two vector elements are *fat pointers*:
+  - A fat pointer is a double-width pointer. It has two components: a pointer to
+    the actual object and a pointer to the virtual method table (vtable) for the
```
mgeisler, line 70: @njr0, is this enough about vtables?

src/traits/trait-objects.md
mgeisler: Yeah, let me try that!

njr0:
Yeah, that's great...the fact that `vtable` is a short for virtual *method* table is super-helpful in explaining why you have `::name` in it. Obviously, having the field and the method have the same name was confusing, but even after @djmitche said `vtable` I assumed it stood for `variable`.

Anyway, the changes altogether seem like a huge upgrade in comprehensibility to me...especially for people who didn't understand it before coming to this section!

src/traits/trait-objects.md
mgeisler: I think I have a nicer and more correct solution: I now show the `"Fido"` string as a separate object. This is consistent with that we do in the diagrams for moved strings.

mgeisler:
> I think it would be good to include a mention of vtable...particularly if there were a sentence explaining what it is in the notes.

I added a link to Wikipedia for the vtable.

src/traits/trait-objects.md
mgeisler: I added small labels to the boxes, I hope you like it.

mgeisler:
The new diagram should show up on https://google.github.io/comprehensive-rust/traits/trait-objects.html in a few seconds.

njr0:
Yes, that's great. 9 lives 😄.

djmitche:
That's a fantastic result :)

PR #1296, opened 2023-10-03
Simplify solution to Luhn exercise

3mber666:
Maybe we can use bool double and make the code easier to read and comprehend

### DIFF
--- a/src/exercises/day-1/luhn.rs
+++ b/src/exercises/day-1/luhn.rs
@@ -15,28 +15,26 @@
 // ANCHOR: solution
 // ANCHOR: luhn
 pub fn luhn(cc_number: &str) -> bool {
-    // ANCHOR_END: luhn
-    let mut digits_seen = 0;
     let mut sum = 0;
-    for (i, ch) in cc_number.chars().rev().filter(|&ch| ch != ' ').enumerate() {
-        match ch.to_digit(10) {
-            Some(d) => {
-                sum += if i % 2 == 1 {
-                    let dd = d * 2;
-                    dd / 10 + dd % 10
+    let mut double = false;
+
+    for (i, c) in cc_number.chars().filter(|&f| f != ' ').rev().enumerate() {
+        if let Some(digit) = c.to_digit(10) {
+            if double {
+                let double_digit = digit * 2;
+                sum += if double_digit > 9 {
+                    double_digit - 9
                 } else {
-                    d
+                    double_digit
                 };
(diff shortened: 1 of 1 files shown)


mgeisler:
Hi @3mber666, could you try fixing the failing test? The [test run](https://github.com/google/comprehensive-rust/actions/runs/6396127620/job/17378063511) says:

```
running 6 tests
test test_invalid_cc_number ... ok
test test_empty_cc_number ... FAILED
test test_single_digit_cc_number ... FAILED
test test_two_digit_cc_number ... ok
test test_valid_cc_number ... ok
test test_non_digit_cc_number ... ok

failures:

---- test_empty_cc_number stdout ----
thread 'test_empty_cc_number' panicked at 'assertion failed: !luhn(\"\")', src/exercises/day-1/luhn.rs:58:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- test_single_digit_cc_number stdout ----
thread 'test_single_digit_cc_number' panicked at 'assertion failed: !luhn(\"0\")', src/exercises/day-1/luhn.rs:66:5


failures:
    test_empty_cc_number
    test_single_digit_cc_number
```

src/exercises/day-1/luhn.rs
```diff
@@ -15,25 +15,29 @@
 // ANCHOR: solution
 // ANCHOR: luhn
 pub fn luhn(cc_number: &str) -> bool {
-    // ANCHOR_END: luhn
-    let mut digits_seen = 0;
+    let cc_number = cc_number.trim().replace(" ", "");
```
mgeisler, line 18: Is this necessary? I would expect the `filter` call below to take of skipping the spaces?

I think it's more elegant to filter below since it avoids an allocation: `str::replace` will heap-alloacte space for the result, but filtering the existing string is an in-place operation.

src/exercises/day-1/luhn.rs
```diff
@@ -15,25 +15,30 @@
 // ANCHOR: solution
 // ANCHOR: luhn
 pub fn luhn(cc_number: &str) -> bool {
-    // ANCHOR_END: luhn
```
mgeisler, line 18: Please don't remove the comment, it's necessary to avoid showing the solution on [the exercise page](https://google.github.io/comprehensive-rust/exercises/day-1/luhn.html) :sweat_smile:

mgeisler (approved):
Thanks for all the work!

PR #1302, opened 2023-10-05
Update `mdbook` to version 0.4.35

mgeisler:
This also ports over the latest changes to `book.js` and `index.hbs`.

We probably need to come up with a more systematic way of doing this, e.g., we could store our changes in a patch and automatically attempt to apply it onto the latest upstream files.


djmitche (approved):
Sorry for the review delay!

PR #1307, opened 2023-10-05
Format the glossary using a Markdown list

mgeisler:
This format has two advantages:

- is easier to maintain in the Markdown file. Tables cannot be wrapped in Markdown, which would severely limit what we can write for each term.
- it can be styled better. The layout now looks more like what you find in a dictionary.

Before:

image

After:

image

Followup to #1277.

### DIFF
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -4,73 +4,101 @@ The following is a glossary which aims to give a short definition of many Rust
 terms. For translations, this also serves to connect the term back to the
 English original.
 
-<!-- Translators: please list the English term in italic after your translation.
-For Danish, this would look like this: `allokere (_allocate_)`. -->
+<style>
+h1#glossary ~ ul {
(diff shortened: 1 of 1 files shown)


mgeisler:
We will eventually run into https://github.com/dprint/dprint-plugin-markdown/issues/87 when we start auto-formatting the Markdown (but we can add `dprint` tags to ignore the formatting here until the bug is fixed).

victorhsieh (approved):
Just to throw out a wild idea, though not very important: It may be possible to use CSS to format it as a table...

mgeisler:
> Just to throw out a wild idea, though not very important: It may be possible to use CSS to format it as a table...

I don't see an easy way for this — at least not without adding a bunch of HTML to the Markdown. But if someone comes up with an approach, then we can apply that later :smile:

Issue #1328, opened 2023-10-09
es: Formatting in TOC


mgeisler:
There are several words which are intended to be italic in the TOC in the Spanish translation:

![image](https://github.com/google/comprehensive-rust/assets/89623/a5339acf-722c-437f-a603-499179b3dab6)

Unfortunately, formatting is not supported in the `SUMMARY.md` file right now (https://github.com/google/mdbook-i18n-helpers/issues/66). Until that bug is fixed, we should remove the `_` from those words.

mgeisler:
This is no longer necessary: mdbook-i18n-helpers 0.3.0 can handle the formatting in the TOC.

Issue #1329, opened 2023-10-09
es: Formatting errors in svgbob diagrams


mgeisler:
There are a few formatting problems in the svgbob diagrams in the Spanish translation. As an example, [16.2. Cadenas movidas en Rust](https://google.github.io/comprehensive-rust/es/ownership/moved-strings-rust.html) looks like this right now:

![image](https://github.com/google/comprehensive-rust/assets/89623/38cb335b-c0c0-4441-853c-e4284a84a034)

The problem is that [svgbob](https://ivanceras.github.io/content/Svgbob/Specification.html) isn't Markdown, so we cannot use emphasis here.

mgeisler:
@deavid and @vzz1x2, could one of you take a quick look at this?

carreter:
This still seems to be an issue:
![image](https://github.com/google/comprehensive-rust/assets/5355463/18ae4574-9c1f-487e-80e2-f3fc78d166f9)

I will take a crack at it as part of figuring out how the translation system works!

PR #1333, opened 2023-10-10
es: fix some problems with the Spanish translation

Phosphorus-M:
Hi! I was checking the book and I found little things to fix. 
I hope you review it.

### DIFF
--- a/po/es.po
+++ b/po/es.po
@@ -2111,7 +2111,6 @@ msgid "Welcome to Day 1"
 msgstr "Te damos la bienvenida al Día 1"
 
 #: src/welcome-day-1.md:3
-#, fuzzy
 msgid ""
 "This is the first day of Rust Fundamentals. We will cover a lot of ground "
 "today:"
@@ -2137,6 +2136,7 @@ msgstr ""
 #: src/welcome-day-1.md:12
 msgid "Pattern matching: destructuring enums, structs, and arrays."
 msgstr ""
+"Emparejamiento de Patrones: desestructuración de enums, structs y arrays."
 
 #: src/welcome-day-1.md:16
 msgid "Please remind the students that:"
@@ -4001,8 +4001,8 @@ msgid ""
 msgstr ""
 "¿Se podrían usar slices `&[i32]` en lugar de matrices de 3 × 3 predefinidas "
 "en el código para tus argumentos y tipos de resultados devueltos? Por "
-"ejemplo, \\``&\\[&\\[i32\\]\\]\\` para un slice de slices de dos "
-"dimensiones. ¿Por qué? ¿Por qué no?"
+"ejemplo, `&[&[i32]]` para un slice de slices de dos dimensiones. ¿Por qué? "
+"¿Por qué no?"
 
 #: src/exercises/day-1/for-loops.md:87
 msgid ""
@@ -4214,10 +4214,10 @@ msgid ""
 "related to the [`while let` loop](while-let-expressions.md). It will "
 "automatically call `into_iter()` on the expression and then iterate over it:"
 msgstr ""
-"El [bucle `for`][`for` loop](https://doc.rust-lang.org/std/keyword.for.html) "
-"está estrechamente relacionado con el [bucle `while let`](while-let-"
-"expressions.md). Llamará automáticamente a `into_iter()` en la expresión y "
-"después iterará sobre ella:"
+"El [bucle `for`](https://doc.rust-lang.org/std/keyword.for.html) está "
+"estrechamente relacionado con el [bucle `while let`](while-let-expressions."
+"md). Llamará automáticamente a `into_iter()` en la expresión y después "
+"iterará sobre ella:"
 
 #: src/control-flow/for-expressions.md:7
 msgid ""
--- a/po/es.po
+++ b/po/es.po
@@ -10,7 +10,7 @@ msgstr ""
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=1; plural=0;\n"
-"X-Generator: Poedit 3.3.2\n"
+"X-Generator: Poedit 3.4\n"
 
 #: src/SUMMARY.md:4 src/index.md:1
 msgid "Welcome to Comprehensive Rust 🦀"
@@ -289,7 +289,7 @@ msgstr "Manejo de la Memoria en Rust"
 
 #: src/SUMMARY.md:88 src/ownership.md:1
 msgid "Ownership"
-msgstr "Propiedad (_Ownership_)"
+msgstr "Ownership"
 
 #: src/SUMMARY.md:89 src/ownership/move-semantics.md:1
 msgid "Move Semantics"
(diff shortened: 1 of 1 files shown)


mgeisler:
Awesome, thanks @Phosphorus-M! I think this fixes #1328? Perhaps also #1329?

po/es.po
```diff
@@ -289,7 +289,7 @@ msgstr "Manejo de la Memoria en Rust"
 
 #: src/SUMMARY.md:88 src/ownership.md:1
 msgid "Ownership"
-msgstr "Propiedad (_Ownership_)"
+msgstr "Ownership"
```
JaviSorribes, line 292: We had a very in depth discussion at https://github.com/google/comprehensive-rust/discussions/1184 , and not sure that we had full consensus, but I think we pretty much came up with `translated_msg (_english_msg_)`

If this is a formatting concern due to the underscore, should we just use `translated_msg (english_msg)`  instead?

po/es.po
```diff
@@ -467,7 +467,7 @@ msgstr "Métodos por Default"
 
 #: src/SUMMARY.md:143 src/traits/trait-bounds.md:1
 msgid "Trait Bounds"
-msgstr "Límites Trait (Bounds)"
+msgstr "Trait Bounds"
```
JaviSorribes, line 470: Límites de Traits (Trait Bounds)

po/es.po
```diff
@@ -872,11 +872,11 @@ msgstr "Send y Sync"
 
 #: src/SUMMARY.md:274
 msgid "Send"
-msgstr "Enviar"
+msgstr "Send"
```
JaviSorribes, line 875: I restarted a conversation in https://github.com/google/comprehensive-rust/discussions/1184#discussioncomment-7243894 , because I'm not sure what the guidance is, whether to use English or Spanish for these.

po/es.po
```diff
@@ -3275,7 +3275,7 @@ msgid ""
 "starting index, meaning that `&a[0..a.len()]` and `&a[..a.len()]` are "
 "identical."
 msgstr ""
-"Si el slice comienza en el índice 0, la sintaxis de intervalo de Rust nos "
+"Si el slice comienza en el índice 0, la sintaxis de rango de Rust nos "
```
JaviSorribes, line 3278: I don't agree with this suggestion. "rango" can mean rank, which is very misleading here. "intervalo" is unambiguous

image

po/es.po
Phosphorus-M: I don't oppose the idea of translating certain terms, as it's a topic of discussion in some Discord and Telegram channels. Many individuals support the notion of retaining the original English term, but it doesn't hurt to include a translated alias.

I concur with this perspective because it provides an additional means for users to grasp the terminology. For those who are not well-versed in English, comprehending the terms can be challenging.

In this case:
> [¿Qué es el Ownership?](https://rustlanges.github.io/rust-book-es/ch04-01-what-is-ownership.html)

We employ the term "Ownership" but provide a note stating that "Propiedad" is an acceptable alternative for conveying the same concept.

I intend to revert this change in order to maintain both options.

po/es.po
Phosphorus-M: Yesterday, @ph4un00b and I discussed this. "Intervalo" isn't a poor translation, it's akin to "time interval" and offers a more precise definition, especially for those with a mathematical background. However, for the majority of people, using terms like "rangos" when referring to range types is more commonplace.

image

Some examples of that:
[Python](https://elpythonista.com/range-enumerate#:~:text=valor%20de%20parada.-,%C2%BFC%C3%B3mo%20crear%20rangos%20en%20Python%3F,-La%20sint%C3%A1xis%20para)
[Ruby](http://rubytutorial.wikidot.com/rangos)
[Swift](https://www.lafactoriaapple.com/swift/bucles/rangos.php)

Even when we're discussing Rust, we use the term "rango"
[FreecodeCamp Tutorial](https://www.freecodecamp.org/espanol/news/rust-en-replit/#:~:text=la%20sintaxis%20de%20rango%20de%20Rust)

po/es.po
mgeisler: In this particular case, "Send" (or rather `Send`) is correct since it's the name of a trait in the standard library. So it's like `String` or `HashMap`.

I think that's also what you concluded in #1184.

po/es.po
deavid: I like it either way. Intervalo and rango works in this context for me.

po/es.po
```diff
@@ -5865,13 +5865,13 @@ msgstr ""
...
 #: src/exercises/day-1/luhn.md:12
-#, fuzzy
 msgid ""
 "After doubling a digit, sum the digits if the result is greater than 9. So "
 "doubling `7` becomes `14` which becomes `1 + 4 = 5`."
 msgstr ""
-"Después de duplicar un dígito, se suman los dígitos que contiene. Por tanto, "
-"si duplicas `7`, pasará a ser `14`, lo cual pasará a ser `5`."
+"Después de duplicar un dígito, se suman los dígitos si el resultado es mayor "
+"a 9. Por tanto, si duplicas `7`, pasará a ser `14`, lo cual pasará a ser `1 "
```
deavid, line 5873:
```suggestion
"que 9. Por tanto, si duplicas `7`, pasará a ser `14`, lo cual pasará a ser `1 "
```

deavid (commented):
Overall I have no issue with the changes in this PR. I would approve but I'd like to discuss with @JaviSorribes before doing so.

po/es.po
mgeisler: The underscores are a problem here, as mentioned in #1328. So they should go for sure.

po/es.po
mgeisler: Let me resolve this conversation since using `Send` is correct here.

po/es.po
mgeisler: @Phosphorus-M, can you take out the translation here for now? Then try to reach a consensus with @JaviSorribes and the other reviewers and members of the Spanish Rust community.

Basically, I would love to merge this since it removes the weird-looking `_foo_` words from the TOC. So if you could take out the translation here, then we can make forward progress. The translation can still go in, it should just be discussed in a smaller and more targeted PR.

po/es.po
mgeisler: Actually, you can keep the formatting! I'm working on making it look correct in https://github.com/google/mdbook-i18n-helpers/pull/100. It should be merged in a few days and then formatting becomes file in the summary file.

po/es.po
mgeisler: Hi all, please come to a conclusion here so we can merge this — I know it's hard to edit a large body of text like this, so I suggest taking out the change for now.

JaviSorribes:
Hi team, I'm ok with moving ahead with this PR if everyone else is in favor. Also, I'm going to step down from code reviewers. I'm sorry but I don't have the bandwidth and I feel like I'm slowing everyone else down.

mgeisler:
> Hi team, I'm ok with moving ahead with this PR if everyone else is in favor. Also, I'm going to step down from code reviewers. I'm sorry but I don't have the bandwidth and I feel like I'm slowing everyone else down.

Thanks @JaviSorribes for your help! It's completely fine to leave comments like you did, it has to be a collaborative work to write a large translation like this.

@deavid and @vzz1x2, can either of you review the PR and work on getting it merged?

deavid (approved):

deavid:
@mgeisler done. I reviewed this a week ago or so and I was already okay with it. I re-reviewed it again. I don't see anything wrong, so I merged it directly. I'm sure this PR improves way more than any possible mistake it could have, and if it does, we can always send another PR to fix any minor issues.

vzz1x2:
@deavid, Im coming late to the party, but I agree with you, after reviewing it myself, it is much better having it than dont.

PR #1336, opened 2023-10-10
[54.1 join] add cargo dependencies in speaker notes

### LINKED ISSUE #1330: 54.1 Join (dependencies)
### Discussed in https://github.com/google/comprehensive-rust/discussions/1317

<div type='discussions-op-text'>

<sup>Originally posted by **njr0** October  6, 2023</sup>
I had to add these dependencies to may `Cargo.toml` to make it work. As with the last few suggestions, I think it will help to includes these in the speaker notes or main text, especially since this can't be run in the playground. (And who wants to be in a playground anyway?)

```
[dependencies]
tokio = { version = "1.32.0", features = ["full"] }
reqwest = { version = "0.11.22", features = ["blocking", "rustls-tls"] }
futures = "0.3"
anyhow = "1.0"
```
</div>

0scvr:
The example in 54.1 Join requires additional dependencies which need to be installed prior to running. This PR adds the required dependencies in the speaker notes of the page.

Closes #1330.

Preview:
join

### DIFF
--- a/src/async/control-flow/join.md
+++ b/src/async/control-flow/join.md
@@ -33,6 +33,14 @@ async fn main() {
 
 <details>
 
+> Before running this example you should install the following dependencies:
+> * [anyhow](https://crates.io/crates/anyhow)
+> * [futures](https://crates.io/crates/futures)
+> * [reqwest](https://crates.io/crates/reqwest)
+> * [tokio](https://crates.io/crates/tokio)
+>
+> You can install them by running the following command in your terminal: `cargo add <package_name>`
+
 Copy this example into your prepared `src/main.rs` and run it from there.
 
 * For multiple futures of disjoint types, you can use `std::future::join!` but


djmitche:
This is already written down, as noted in #1330. I think it's too late for a speaker to run `cargo add` when they are presenting a slide, so I'd like to keep those instructions in the preparation for the class. I think the best way to help a reader of the course (who does not have an instructor) is to link back to those instructions in each set of speaker notes where they are required.

mgeisler:
Hi @0scvr, should we close this PR and instead make a new one to link back to the setup instructions?

mgeisler:
> > Hi @0scvr, should we close this PR and instead make a new one to link back to the setup instructions?
> 
> Please assign me the issue when you do, if possible.

I think we can keep #1330 for this, where you are already assigned :slightly_smiling_face: Thanks!

Issue #1327, opened 2023-10-09
Skip large code blocks in Bare-metal Rust


mgeisler:
As part of #1257, we should skip the large and mostly untranslatable code blocks found in Bare-metal Rust.

To do this, simply add

```
<!-- mdbook-xgettext: skip -->
```

before the code block. Notice how the code disappears from `messages.pot` when you run

```
MDBOOK_OUTPUT='{"xgettext": {"pot-file": "messages.pot"}}' mdbook build -d po
```

@qwandor, fyi.

mgeisler:
Hi @0scvr, that's amazing! Thanks!

mgeisler:
Should we close this now?

mgeisler:
> By replacing comments and strings inside the codeblocks we can probably improve this further but I don't think it's worth the effort, unless there's a real need. I think this can be closed for now.

I have #531 as a long-shot approach for this: if we could get `mdbook` to show `stderr`, we could replace almost all `println!` calls with `dbg!`. That would eliminate most of the literal strings in our code examples and automatically remove the code blocks from the translations. Which would be a nice way to lower the burden on our translators.

But I agree, let's close this one now that you've cleaned up most of the strings. Thanks a lot!

mgeisler:
> That would eliminate most of the literal strings in our code examples and automatically remove the code blocks from the translations. Which would be a nice way to lower the burden on our translators.

I just remembered that we also have https://github.com/google/mdbook-i18n-helpers/issues/95 which is a better long-term solution: we could only extract comments and literal strings from the code.

PR #1339, opened 2023-10-10
Add mdbook-xgettext skip for untranslatable codeblocks in Bare Metal

0scvr:
See #1327

Removes 280 lines from `messages.pot`.


mgeisler:
Nice, this removed 280 lines from the translations:

```
% diff po/messages-{before.pot,after.pot} | diffstat
 messages-after.pot |  286 -----------------------------------------------------------------------
 1 file changed, 3 insertions(+), 283 deletions(-)

```

PR #1344, opened 2023-10-11
Fix typo: missing n in unknown

noamzaks:


dyeroshenko (approved):

mgeisler:
Hey all! A small note: you should never edit the `msgid` fields of the PO files. It is only `msgmerge` that is allowed to update that field. This is [mentioned briefly in our documentation](https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md#creating-and-updating-translations).

The problem is that the translations are not using the current English text. Instead they're using the English text as it looked at the time mentioned in the POT-Creation-Date field at the top of the PO file. This was implemented in #1243.

We should revert the edits to the PO files to fix this.

mgeisler:
I created #1347 to remove the edits to the PO files.

Perhaps we can create a check somehow which catches this in the future? We've have a few PRs like this where people do the natural thing and fix a typo in all files where it occurs.

noamzaks:
> Perhaps we can create a check somehow which catches this in the future? We've have a few PRs like this where people do the natural thing and fix a typo in all files where it occurs.

Do we want to search for any edits of PO files, or specific changes of anything inbetween the `msgid` and `msgstr`?

```
#: src/exercises/day-3/safe-ffi-wrapper.md:26
msgid ""
"`*const i8` to `&CStr`: you need something which can find the trailing `\\0` "
"character"
msgstr ""
```

I'm assuming only everything between `msgid` and `msgstr`?

mgeisler:
> > Perhaps we can create a check somehow which catches this in the future? We've have a few PRs like this where people do the natural thing and fix a typo in all files where it occurs.
> 
> Do we want to search for any edits of PO files, or specific changes of anything inbetween the `msgid` and `msgstr`?

Good question... The tricky thing is that we do change the `msgid` fields from time to time — as mentioned in the TRANSLATION file.

> ```
> #: src/exercises/day-3/safe-ffi-wrapper.md:26
> msgid ""
> "`*const i8` to `&CStr`: you need something which can find the trailing `\\0` "
> "character"
> msgstr ""
> ```
> 
> I'm assuming only everything between `msgid` and `msgstr`?

Yes, correct: the text there is the original English text. It has to match the text in the Markdown for us to translate it correctly (via [mdbook-i18n-helpers](https://github.com/google/mdbook-i18n-helpers)).

Saying this, I get an idea : I think we can spot correct edits of the `msgid` fields by looking for a chance in the POT-Creation-Date field. When someone runs `msgmerge` to add in new English strings, the field will change (it is copied from the `messages.pot` file).

PR #1346, opened 2023-10-11
Update dining-philosophers-async.rs

njr0:
Also replace Thales with Hypatia in the async version of Dining Philosophers.


mgeisler (approved):
Thanks!

mgeisler:
Corresponds to #1345.

PR #1347, opened 2023-10-11
Revert "Fix typo: missing n in unknown"

mgeisler:
Reverts most of google/comprehensive-rust#1344.

The edit to the Markdown file has been kept since we _do_ want to fix the typo!

### DIFF
--- a/po/da.po
+++ b/po/da.po
@@ -11109,7 +11109,7 @@ msgstr ""
 #: src/exercises/day-3/safe-ffi-wrapper.md:27
 msgid ""
 "`&CStr` to `&[u8]`: a slice of bytes is the universal interface for \"some "
-"unknown data\","
+"unknow data\","
 msgstr ""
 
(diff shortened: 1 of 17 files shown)


sakex (approved):

deavid (approved):

mgeisler:
Thanks @sakex for the super-quick review. However, please don't enable auto-merge for other people's PRs:

image

In this case I was quick enough to force-push a change which un-reverted the change to the Markdown file. But if I had been slower, I would have accidentally reverted all of #1344, which I didn't want.

Sorry for all the complexity here... :grimacing: There are more corner-cases than I thought, but I hope we can figure it out as we learn more.

sakex:
Ok sorry for that. I usually do that for people who don't have merge rights. I won't do it for others then

mgeisler:
> Ok sorry for that. I usually do that for people who don't have merge rights. I won't do it for others then

I think you should continue doing this for people who don't have write access — but I do have access 😄

PR #1351, opened 2023-10-11
Added action to prevent changing msgid

noamzaks:
See discussion in #1344.

This checks that all `.po` files changed in pull requests either contain changes in POT-Creation-Date (i.e. are run by the specific tool which should change msgid's), or don't change msgid's (by checking that each line changed is NOT between msgid and msgstr)

I have tested this on my fork, see [valid changes](https://github.com/noamzaks/comprehensive-rust/pull/1/files) where the action passes, and [invalid changes](https://github.com/noamzaks/comprehensive-rust/pull/2/files) where it doesn't.

The action also has output for failing/passing:
[passing](https://github.com/noamzaks/comprehensive-rust/actions/runs/6488171290/job/17620037473?pr=1)
[failing](https://github.com/noamzaks/comprehensive-rust/actions/runs/6488175171/job/17620050716?pr=2)

### DIFF
--- a/.github/workflows/prevent-msgid.yml
+++ b/.github/workflows/prevent-msgid.yml
@@ -8,34 +8,51 @@ on:
 jobs:
   prevent-msgid-changes:
     runs-on: ubuntu-latest
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
+      steps:
+        - name: Checkout
+          uses: actions/checkout@v3
+          with:
+            fetch-depth: 0
 
-      - name: Check for changes between msgid and msgstr
-        run: |
-          git reset origin/main
-          for filename in $(git diff --name-only | grep '.po')
-          do
-            echo Checking $filename changes
-            if git diff $filename | grep -q POT-Creation-Date; then
-              echo Assuming \"$filename\" was changed automatically, skipping msgid check
-              continue
-            fi
+        - name: Reset git
+          run: git reset origin/main
 
-            changed_lines=$(git blame $filename | grep -n '^0\{8\} ' | cut -f1 -d:)
-            for changed_line in $changed_lines
-            do
-              if sed -n ${changed_line}p $filename | grep -q msgstr; then
-                continue
-              fi
+        - name: Check po file changes
+          run: |
+              import os
 
-              if sed -n "${changed_line},\$p" $filename | awk '!/^msgid/ {print} /^msgid/ {exit}' | grep -q msgstr; then
-                echo Changed msgid in file \"$filename\" \(line $changed_line\)!
-                echo Please read https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md#creating-and-updating-translations.
-                exit 1
-              fi
-            done
-          done
+              for filename in os.popen("git diff --name-only").read().split():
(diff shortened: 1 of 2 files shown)


.github/workflows/prevent-msgid.yml
```diff
@@ -0,0 +1,41 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Check for changes between msgid and msgstr
+        run: |
+          git reset origin/main
+          for filename in $(git diff --exit-code --name-only | grep '.po')
```
mgeisler, line 20: Small question: I wonder why `--exit-code` is here? The man page says

> Make the program exit with codes similar to diff(1). That is, it exits with 1 if there were differences and 0 means no differences.

but it's not clear to me why we want the exit code here?

.github/workflows/prevent-msgid.yml
```diff
@@ -0,0 +1,41 @@
...
+          do
+            echo Checking $filename changes
+            if git diff --exit-code $filename | grep -qE POT-Creation-Date; then
```
mgeisler, line 23: I think we don't need `-E` for `grep` here since the pattern doesn't use anything special?

mgeisler:
Wow, this is great! Thank you for putting this together!

I can see what this does, but I wonder if it would be easier to understand if it was a Python script? They can be embedded directly into the GH action:

https://github.com/mgeisler/textwrap/blob/04d36f18155a51e3346502312cf567f2ddf764a6/.github/workflows/prepare-release.yml#L69-L80

Though saying this, it would be better to just create a `.py` file inside the `.github/workflows/` folder.

mgeisler:
I think we can merge this as a shell script already, but if you find time, I would be interested in seeing it as a Python script instead. That would hopefully help with some of the more obscure (to me) lines such as

```sh
             if sed -n ${changed_line}p $filename | grep -q msgstr; then
```

noamzaks:
Definitely easier to understand (and write) as a Python script. ~Doesn't it "cost more minutes"? You need a setup-python action (I think).~ Didn't know there is Python built in to Actions workers.

I can turn this into a `.py` script inside `.github/workflows`, but then how will it run?

mgeisler:
> Definitely easier to understand (and write) as a Python script.

Awesome!

> ~Doesn't it "cost more minutes"? You need a setup-python action (I think).~ Didn't know there is Python built in to Actions workers.

Yeah, it seems to be available right out of the box.

> I can turn this into a `.py` script inside `.github/workflows`, but then how will it run?

You should be able to do

```yaml
      - name: Configure Git user
        run: python3 .github/workflows/prevent-changing-msgid.py
```

(not tested... I think the path will be relative to the repository root, but perhaps it's relative to `.github/workflows/`?)

noamzaks:
Switched to Python and re-tested (links in the first post updated accordingly)

mgeisler (approved):
Thanks a lot!

mgeisler:
Hi @noamzaks, I created #1364 to add a few comments to the script. You might want to add a few more to explain future maintainers what it does.

I also notes that we should

- make the script read-only (remove the `git reset`)
- make it use more Git plumbing commands instead of the high-level porcelain commands used today

It's not high-priority, but it would be nice to have.

PR #1355, opened 2023-10-12
fix: typo in mark_visited fn docstring

spaydar:
Fixes a typo in the `mark_visited` function docstring.

See [`HashSet.insert` documentation](https://doc.rust-lang.org/stable/std/collections/struct.HashSet.html#method.insert)


mgeisler (approved):
Thanks, nice catch!

PR #1358, opened 2023-10-12
Add trait bound to example for MutexGuard that is !Send + Sync

fbornhofen:
This example only makes sense (and is therefore easier to understand logically :) ) if T is Sync. See https://github.com/rust-lang/rust/issues/41622 - this used to be a bug initially.


sakex (approved):
Thanks Fabian :)

PR #1359, opened 2023-10-12
Format all PO files with `dprint`

mgeisler:
This enforces a consistent formatting for the PO files. The goal of this is to avoid large diffs due to random and unnecessary reformatting.

We use the format of `msgcat`: this is also waht `msgmerge` produces and it’s easy to replicate for people by installing Gettext and running `dprint fmt`.

This is a follow-up to #1351 which started enforcing that the `msgid` fields don’t change due to reformatting in a PR.

If this turns out to be cumbersome, then we can disable it again.


dprint.json
```diff
@@ -8,6 +8,9 @@
     "commands": [{
       "command": "rustfmt --edition 2021",
       "exts": ["rs"]
+    }, {
+      "command": "msgcat -",
```
sakex, line 12: How does this work? Is it piping rustfmt output to msgcat ?

sakex (approved):

dprint.json
mgeisler: Almost, it's piping all `.po` files (from `exts` below) through `msgcat -`. This normalizes the PO files.

mgeisler:
Note that this will require translators to install the Gettext tools and `dprint` so they can run `dprint fmt` before sending out a file for review. I've tried to avoid this until now since I wanted to make it as easy as possible to contribute to a translation. But when the Poeditors don't format things consistently, we end up with a mess.

henrif75:
> Note that this will require translators to install the Gettext tools and `dprint` so they can run `dprint fmt` before sending out a file for review. I've tried to avoid this until now since I wanted to make it as easy as possible to contribute to a translation. But when the Poeditors don't format things consistently, we end up with a mess.

We'll probably need to update the instructions in TRANSLATE.md, right?

mgeisler:
> We'll probably need to update the instructions in TRANSLATE.md, right?

Yes! Thanks, that's a great point!

mgeisler:
@henrif75 what do you think of the update to TRANSLATIONS?

henrif75:
> @henrif75 what do you think of the update to TRANSLATIONS?

Sure, I'll do it.

henrif75:
> @henrif75 what do you think of the update to TRANSLATIONS?

They're great, thanks!

PR #1361, opened 2023-10-12
ja: refresh po file

HidenoriKobayashi:
Refresh the po file for Japanese translation. Part of google#652.

Steps taken:
1. MDBOOK_OUTPUT='{"xgettext": {"pot-file": "messages.pot"}}' \
  mdbook build -d po
1. msgmerge --update po/ja.po po/messages.pot

### DIFF
--- a/po/ja.po
+++ b/po/ja.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: Comprehensive Rust 🦀\n"
-"POT-Creation-Date: 2023-09-24\n"
+"POT-Creation-Date: 2023-10-15T23:41:49Z\n"
 "PO-Revision-Date: 2023-06-06 13:18+0900\n"
 "Last-Translator: Kenta Aratani <kentaaratani@coinez.jp>\n"
 "Language-Team: Japanese <translation-team-ja@lists.sourceforge.net>\n"
@@ -52,8 +52,8 @@ msgstr "ローカル環境での実行"
 msgid "Day 1: Morning"
 msgstr "Day 1： AM"
 
-#: src/SUMMARY.md:19 src/SUMMARY.md:79 src/SUMMARY.md:137 src/SUMMARY.md:195
-#: src/SUMMARY.md:221 src/SUMMARY.md:271
+#: src/SUMMARY.md:19 src/SUMMARY.md:80 src/SUMMARY.md:135 src/SUMMARY.md:193
+#: src/SUMMARY.md:219 src/SUMMARY.md:269
 msgid "Welcome"
 msgstr "Welcome"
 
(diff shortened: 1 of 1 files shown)


po/ja.po
```diff
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: Comprehensive Rust 🦀\n"
-"POT-Creation-Date: 2023-09-24\n"
+"POT-Creation-Date: \n"
```
mgeisler, line 4: Thanks for refreshing the file. Could you do this again after installing the latest version of `mdbook-i18n-helpers`? The new version will insert the current date here, which is important to ensure a consistent snapshot build of the translation (the translation becomes frozen in time using the timestamp here).

mgeisler (changes requested):

po/ja.po
kantasv: TL;DR I could reproduce the same blank POT-Creation-Date with 0.2.4 helper. It might be a bug on the helper side, or other dependency issue? 

### Context

- Confirmed that the lastes helper is installed:
```
➜  comprehensive-rust git:(main) ✗ cargo search mdbook-i18n-helpers
mdbook-i18n-helpers = "0.2.4"    # Plugins for a mdbook translation workflow based on Gettext.
```

- Updated Rust with `rustup update` (prev `cargo 1.72.0`-> new `cargo 1.73.0`) and tried refreshing POT creation date, but still an empty entry was generated for it.

kantasv: The new change to add POT-Creation-Date seems to come from this MR on the helper repo:
https://github.com/google/mdbook-i18n-helpers/pull/87/files

### The feature for POT-Creation-Date is included to v.0.2.4 helper release
I confirmed that the commit [b19c24f](https://github.com/google/mdbook-i18n-helpers/commit/b19c24f478c0368ea0c65b75d1e6b9c2a331278c) to the helper repo ( corresponding MR: **Set POT-Creation-Date in newly generated POT files**) was properly included to the release for v0.2.4., as in the release diff from [v0.2.3->0.2.4](https://github.com/google/mdbook-i18n-helpers/compare/0.2.3...0.2.4).

### Is this JA lang specific problem? - No
To check if this is JA specific issue or not, I tried refreshing po files with a few languages. FR and DE might have the same empty blank when I use `msgmerge --update po/ja.po po/messages.pot`. The problem is at least reproducible in my enviroment. 

```
"POT-Creation-Date: 2023-09-18\n" -> "POT-Creation-Date: \n"
"POT-Creation-Date: 2023-09-19\n" -> "POT-Creation-Date: \n"
```

kantasv: This line seems to add `pot_creation_date` to catalog metadata on the helper repo. But I'm not sure how the final ja.po file yields "POT-Creation-Date" key. 

https://github.com/google/mdbook-i18n-helpers/blob/0cdc55b16deefff768811a587001cbc2694338e2/src/bin/mdbook-xgettext.rs#L52

mgeisler: Hey @kantasv! Yeah, you've found the correct line! The data flow is the following:

- `MDBOOK_OUTPUT='{"xgettext": {"pot-file": "messages.pot"}}' mdbook build -d po` invokes `mdbook-xgettext`
- `mdbook-xgettext` is a binary from `mdbook-i18n-helpers` and it writes the current date and time to POT-Creation-Date.
- `msgmerge --update po/ja.po po/messages.pot` updates `ja.po` based on the `messages.pot`. This will copy over the POT-Creation-Date field from `messages.pot`.

I believe you simply need to run

```
cargo install mdbook-i18n-helpers
```

to install version 0.2.4 — I don't think the output of `cargo search` is showing you the version you have installed locally.

After installing verison 0.2.4, you need to regenerate the `messages.pot` file and then run `msgmerge` as well.

HidenoriKobayashi: Done. It now contains a date.

mgeisler: Great! The whole system is very simple and quite flexible... if there is no date, the translation is built as of "now" and it will always lag behind the English source texts. With a date, the translation becomes frozen at that date.

The most recent [publication run shows this](https://github.com/google/comprehensive-rust/actions/runs/6517213745/job/17701480597):

image

Right now, all PO files have a date and if not, well then they'll get a date the next time someone uses a sufficiently new version of `mdbook-i18n-helpers`.

po/ja.po
```diff
@@ -21516,6 +20297,122 @@ msgid ""
...
+#~ "     Running `target/debug/exercise`\n"
+#~ "Hello, world!\n"
+#~ "```"
+
+#~ msgid ""
+#~ "```shell\n"
+#~ "$ cargo run\n"
+#~ "   Compiling exercise v0.1.0 (/home/mgeisler/tmp/exercise)\n"
+#~ "    Finished dev [unoptimized + debuginfo] target(s) in 0.24s\n"
+#~ "     Running `target/debug/exercise`\n"
```
mgeisler, line 20335: Feel free to delete these old translations. They're put in a kind of "paper bin" by `msgmerge`, but they should be deleted when they don't look useful.

mgeisler (approved):
Looks good to me, thanks!

PR #1364, opened 2023-10-14
Add comments and license text to `prevent-msgid.py`

mgeisler:


.github/workflows/prevent-msgid.py
```diff
@@ -1,19 +1,72 @@
...
         continue
 
-    if "POT-Creation-Date" in os.popen(f"git diff --unified=0 {filename}").read():
-        print(f"Assuming {filename} was changed automatically, skipping msgid check")
+    # If POT-Creation-Date has changed, then we assume that the commit
+    # is the result of `msgmerge --update`. It is expected that the
+    # `msgid` fields change when `msgmerge` is run, so there is
+    # nothing to check.
+    if "POT-Creation-Date" in os.popen(
+            f"git diff --unified=0 {filename}").read():
```
noamzaks, line 56: I think `git diff` can be changed to `git diff-files` cleanly here.

Didn't find some plumbing replacement for git blame

mgeisler: Hey @noamzaks! Yeah, sounds good with `git diff-files`! It would be great if you could test it and make a PR.

noamzaks (commented):
lgtm

mgeisler:
@noamzaks, thanks for looking over the comments!

djmitche (approved):

PR #1376, opened 2023-10-17
Update luhn.rs

justanotheranonymoususer:
Fix semicolon

I'm only learning Rust but I think it makes more sense


djmitche (approved):

djmitche:
Thank you! That's definitely a good fix.

Issue #1349, opened 2023-10-11
Include instructions for connecting to server in Tasks example


djmitche:
### Discussed in https://github.com/google/comprehensive-rust/discussions/1316

<div type='discussions-op-text'>

<sup>Originally posted by **njr0** October  6, 2023</sup>
..
I made it run OK (by just adding `tokio = { version = "1.32.0", features = ["full"] }` as a dependency in `Cargo.toml`, and I could  get it to say `connection from 127.0.0.1:64382` etc. by using `curl`, but I guess I'd have to write code to connect to the socket and receive the question and send a reply and stuff. Presumably, that's what's in the "prepared `main.rs`.
..
</div>

---

`nc` or `telnet` are good tools for this. It's not HTTPS, so curl's not a good choice.

This should be mentioned in the speaker notes.

Gourav2609:
Please assign this issue to me @djmitche

djmitche:
done!

Gourav2609:
Thanks @djmitche Will open a PR soon

Gourav2609:
@djmitche 
Just to make sure i am doing the right thing..

the changes are to be made in [src\running-the-course\course-structure.md
]( https://github.com/google/comprehensive-rust/blob/main/src/running-the-course/course-structure.md#concurrency-in-rust.)
do i just have to add a line ?

`nc or telnet are good tools for this. It's not HTTPS, so curl's not a good choice.`

please let me know🙂

djmitche:
I think it would be in the speaker notes (`<details>`) in https://github.com/google/comprehensive-rust/blob/c4ce2ab8a97dc1c0c088d75e32483d8adac79059/src/async/tasks.md. I don't think it's necessary to include the "not curl" part. And, a link for `nc` and `telnet` is probably useful, too.

Gourav2609:
@djmitche Please review the PR #1382 
Thanks

PR #1382, opened 2023-10-18
Updated about `nc` and `telnet` in task.md

Gourav2609:
Updated about `nc` and `telnet` with their documentation link provided as well
Closes issue #1349

### DIFF
--- a/src/async/tasks.md
+++ b/src/async/tasks.md
@@ -60,4 +60,10 @@ Copy this example into your prepared `src/main.rs` and run it from there.
 
 * Refactor the async block into a function, and improve the error handling using `?`.
 
+* `nc` or `telnet` are good tools for this. It's not HTTPS
+
+     - [Netcat (nc)](https://docs.rs/nc/latest/nc/): Netcat is a          versatile networking utility that can be used for various networking tasks, including connecting to network services.
+
+     - [Telnet](https://docs.rs/telnet/latest/telnet/): Telnet is a network protocol used for remote terminal connections to devices over a network. It's a simple way to test network connectivity.
+
 </details>


Gourav2609:
@djmitche I have just filled the CLA after submitting the PR.
is that fine?

src/async/tasks.md
```diff
@@ -60,4 +60,10 @@ Copy this example into your prepared `src/main.rs` and run it from there.
 
 * Refactor the async block into a function, and improve the error handling using `?`.
 
+* `nc` or `telnet` are good tools for this. It's not HTTPS
+
+     - [Netcat (nc)](https://docs.rs/nc/latest/nc/): Netcat is a          versatile networking utility that can be used for various networking tasks, including connecting to network services.
+
+     - [Telnet](https://docs.rs/telnet/latest/telnet/): Telnet is a network protocol used for remote terminal connections to devices over a network. It's a simple way to test network connectivity.
```
djmitche, line 67: I think both of these links are different from what the descriptions are referring to! But, going to look for links, there's not much! Maybe these links to the manpage are best:
 * https://www.unix.com/man-page/linux/1/nc/
 * https://www.unix.com/man-page/linux/1/telnet/

src/async/tasks.md
```diff
@@ -60,4 +60,10 @@ Copy this example into your prepared `src/main.rs` and run it from there.
 
 * Refactor the async block into a function, and improve the error handling using `?`.
 
+* `nc` or `telnet` are good tools for this. It's not HTTPS
```
djmitche, line 63: Let's move this line up to line 53. "..run it from there. Try connecting to it with a TCP connection tool like `nc` or `telnet`".

djmitche (changes requested):

src/async/tasks.md
Gourav2609: i have shifted the content to line 53...
Do i need do shift the basic definitions of each `nc` and `telnet` to somewhere else as currently they just below this line

src/async/tasks.md
Gourav2609: Thanks for the links @djmitche 
i have updated the link in my latest commit [Updated links](https://github.com/google/comprehensive-rust/pull/1382/commits/652b5945f61556f9f873c74baa9b00babd2ca08b)

src/async/tasks.md
```diff
@@ -51,13 +51,19 @@ async fn main() -> io::Result<()> {
 
 Copy this example into your prepared `src/main.rs` and run it from there.
 
+Try connecting to it with a TCP connection tool like `nc` or `telnet`.
+
+* [Netcat (nc)](https://docs.rs/nc/latest/nc/): Netcat is a versatile networking utility that can be used for various networking tasks, including connecting to network services. 
+
+* [Telnet](https://docs.rs/telnet/latest/telnet/): Telnet is a network protocol used for remote terminal connections to devices over a network. It's a simple way to test network connectivity.
+
```
djmitche, line 59: These docs links are still wrong! The linked `nc` crate does not do what the text says it does..

djmitche (commented):
As for whether it's OK to lay out this way -- I think this is a little confusing. It has one sentence in a paragraph suggesting that the instructor do something, followed by two list items for things in that pargraph (nc and telnet), followed immediately by more list items suggesting that the instructor do something.

What might be a clearer way to express that for the reader?

Gourav2609:
> What might be a clearer way to express that for the reader?

i think those definitions would not be necessary since the users would read about them anyway by clicking on  `nc` and `telnet`.

or we can use something similar like `detail` dropdown to display their definitions  at the end of the page

src/async/tasks.md
Gourav2609: sry . i dont know how did i miss that
Fixed them in my latest commit `Fixed links`

djmitche:
I lean toward your first idea:

> i think those definitions would not be necessary since the users would read about them anyway by clicking on nc and telnet.

djmitche (changes requested):
(marking changes-requested based on my last comment)

Gourav2609:
> I lean toward your first idea:

@djmitche 
sry but i did not get it...

should we include the text of nc and telnet or just the links
or we can create something like the already existing detail dropdown for the definition of the following (nc , telnet) at the bottom

djmitche:
Just the links, please :)

Gourav2609:
> Just the links, please :)
 

@djmitche 

Done 😊.. Please refer the latest commit [[Just the links](https://github.com/google/comprehensive-rust/pull/1382/commits/bc481810c3c36319d20b045e3e6795207434cf99)]

PR #1388, opened 2023-10-19
zh-CN: Update Index page

liruifengv:


SketchK (approved):

mgeisler:
Thanks so much @liruifengv for fixing the fuzzy entries on the front page! I cannot read Chinese, but it's beautiful to see https://google.github.io/comprehensive-rust/zh-CN/ fully translated :tada: 

The "Speaker Notes" part is injected by some JavaScript and I would like to find a way to translate it as well. I'm just not sure how yet.

PR #1391, opened 2023-10-19
Refreshed existing IT translations, and added a few more

detro:
Still a long way to go.

Also added a small QoL improvement to the translators instructions.

### DIFF
--- a/TRANSLATIONS.md
+++ b/TRANSLATIONS.md
@@ -180,8 +180,8 @@ output.
 
 ### Building a Translation
 
-Make sure you have gone through the [build setup](./README.md#building)
-at least once.
+Make sure you have gone through the [build setup](./README.md#building) at least
+once.
 
 To use the `po/xx.po` file for your output, run the following command:
 
--- a/TRANSLATIONS.md
+++ b/TRANSLATIONS.md
@@ -49,9 +49,16 @@ brew install gettext
 
 ### `dprint`
 
-Install [`dprint`] using their installation instructions.
+Install [`dprint`] using their
+[installation instructions](https://dprint.dev/install/).
 
 [`dprint`]: https://dprint.dev/
+
+Alternatively, on MacOS with [Homebrew](https://brew.sh/), you can install with:
+
+```shell
+brew install dprint
+```
 
 Ensure you can build the book, and that `mdbook serve` works. For this, follow
 the instructions in the [README](README.md).
@@ -172,6 +179,9 @@ output.
 > translation as all entries will be marked as "fuzzy".
 
 ### Building a Translation
+
+Make sure you have gone through the [build setup](./README.md#building)
+at least once.
 
 To use the `po/xx.po` file for your output, run the following command:
 
--- a/TRANSLATIONS.md
+++ b/TRANSLATIONS.md
@@ -49,7 +49,7 @@ brew install gettext
 
 ### `dprint`
 
-Install [`dprint`] using their
+Install `dprint` using their
 [installation instructions](https://dprint.dev/install/).
 
 [`dprint`]: https://dprint.dev/
--- a/po/it.po
+++ b/po/it.po
@@ -1,8 +1,8 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: Comprehensive Rust 🦀\n"
-"POT-Creation-Date: 2023-10-19\n"
-"PO-Revision-Date: 2023-10-19 19:23+0100\n"
+"POT-Creation-Date: 2023-10-24T09:24:14+01:00\n"
+"PO-Revision-Date: 2023-10-24 09:36+0100\n"
 "Last-Translator: Ivan De Marino <ivan.de.marino@gmail.com>\n"
 "Language-Team: \n"
 "Language: it\n"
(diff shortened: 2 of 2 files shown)


po/it.po
```diff
@@ -1,16 +1,16 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: Comprehensive Rust 🦀\n"
-"POT-Creation-Date: 2023-09-19\n"
-"PO-Revision-Date: 2023-09-19 09:35+0100\n"
+"POT-Creation-Date: 2023-10-19\n"
```
mgeisler, line 4: Hi @detro, updating this by hand will work, but it's not the intended workflow (I can tell that you did it by hand since it has only a date without a time).

You need to run `cargo install mdbook-i18n-helpers` to get version 0.2.4 of our helper tools. This will ensure that your `messages.pot` file gets a POT-Creation-Date field set to now, including hours, minutes, and seconds.

When you run `msgmerge`, this field is copied to the `it.po` file. This way we know when the English text was last extracted and we use this when publishing the course:

https://github.com/google/comprehensive-rust/blob/9637e3c6079c00f512f54bd923e8f71e37fd3e4a/.github/workflows/publish.yml#L52-L58

TRANSLATIONS.md
```diff
@@ -49,10 +49,17 @@ brew install gettext
 
 ### `dprint`
 
-Install [`dprint`] using their installation instructions.
+Install [`dprint`] using their
```
mgeisler, line 52: Let's drop the double link, we already link to `dprint` above:

```suggestion
Install `dprint` using their
```

mgeisler:
> Still a long way to go.

Have you experiemented with the [cloud-translate](https://github.com/mgeisler/cloud-translate) tool?

detro:
> > Still a long way to go.
> 
> Have you experiemented with the [cloud-translate](https://github.com/mgeisler/cloud-translate) tool?

Yep. We have most of it translated via it, but validating each takes a while. Especially for long, complex sentences, where Italian gets pretty weird from Google translate.

po/it.po
detro: Mmm, I knew that what I did I wasn't supposed to do.

I did look at your previous CL about this, and probably misunderstood what I was supposed to do.

Let me fix it.

mgeisler: That's very fair! I was the one who started by "cheating" here :smile:

When I introduced the new "snapshot publishing" system, I needed to add a POT-Creation-Date field to each file. This ensure that the PO files stop getting more out of date. So I just used the date of the last edit as a first approximation — knowing that it the field would be made correct the next time someone creates a `messages.pot` file and runs `msgmerge`.

mgeisler:
> > > Still a long way to go.
> > 
> > 
> > Have you experiemented with the [cloud-translate](https://github.com/mgeisler/cloud-translate) tool?
> 
> Yep. We have most of it translated via it, but validating each takes a while. Especially for long, complex sentences, where Italian gets pretty weird from Google translate.

Okay, great that you know about it — and I agree that the results can be very confusing and hilarious!

detro:
@mgeisler let me know if this is how it should be now.

I suppose I need to remember to run the `msgmerge --update po/XX.po po/messages.pot` every time just before we push a branch for review, right?

mgeisler:
> I suppose I need to remember to run the `msgmerge --update po/XX.po po/messages.pot` every time just before we push a branch for review, right?

Not quite: you should only run `msgmerge` when you want to _incorporate new English text into your translation_. That's what I call [Refreshing an Existing Translation](https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md#refreshing-an-existing-translation) here.

Most of the PRs should only add missing translations: that's how you will slowly work towards a full translation. The POT-Creation-Date header ensures that your translation doesn't deteriorate over time: it will remain fixed at the timestamp stored there.

So how do you update the POT-Creation-Date timestamp? By generating a new `messages.pot` file and running `msgmerge`. Since this imports all the new English strings, messages in the translation will be marked fuzzy and there will be untranslated messages as well. Fixing those should be done in followup PRs to make it easy to review.

mgeisler:
@nicomazz and @bznein, we'll need your help to actually review the updated Italian translation here — non parlo Italiano :slightly_smiling_face:

po/it.po
```diff
@@ -1356,9 +1353,8 @@ msgstr ""
 "argomenti specialistici:"
 
 #: src/running-the-course/course-structure.md:19
-#, fuzzy
 msgid "Rust in Android"
-msgstr "Benvenuto in Rust su Android"
+msgstr "Rust su Android"
```
nicomazz, line 1357: optional: I think "Rust in Android" sounds better, but I'm fine also with the current version

po/it.po
```diff
@@ -2470,6 +2461,196 @@ msgstr ""
 "prevedibili come C e C++ (nessun Garbage Collector) così come l'accesso "
 "all'hardware di basso livello (quando serve)"
 
+#: src/why-rust/an-example-in-c.md:4
+msgid "Let's consider the following \"minimum wrong example\" program in C:"
+msgstr ""
```
nicomazz, line 2466: Any translation here?

po/it.po
```diff
@@ -2470,6 +2461,196 @@ msgstr ""
...
+"\t}\n"
+"\n"
+"\treturn 0;\n"
+"}\n"
+"```"
+msgstr ""
+
+#: src/why-rust/an-example-in-c.md:48
+msgid "How many bugs do you spot?"
+msgstr ""
```
nicomazz, line 2515: No translation for this and the follwoing?

nicomazz (approved):
I did my best, towards the end I just skimmed through though: in the middle of the cl many things are left untranslated; I'm assuming this is expected

po/it.po
```diff
@@ -73,958 +73,962 @@ msgstr "Piccolo esempio"
...
 
-#: src/SUMMARY.md:312 src/thanks.md:1
+#: src/SUMMARY.md:313 src/thanks.md:1
 msgid "Thanks!"
 msgstr "Grazie!"
 
-#: src/SUMMARY.md:313
+#: src/SUMMARY.md:314 src/glossary.md:1
+msgid "Glossary"
+msgstr ""
```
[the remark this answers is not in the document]
detro: fixed.

po/it.po
detro: E' una nuova entry, non ancora tradotta

po/it.po
detro: Nuova entry, non ancora tradotta

detro:
OK, aside from the new entries not translated yet (that's why they are empty), I think I have addressed all comments and will proceed to merge now.

PR #1392, opened 2023-10-19
Add CXX tutorial

### LINKED ISSUE #823: Add examples of how to use CXX

randomPoison:
Add a number of slides that cover most of CXX's functionality and demonstrate how it can be used.

Fixes #823.


src/android/interoperability/cpp/blobstore/src/main.rs
mgeisler, line 1: This looks like code from https://cxx.rs/tutorial.html? I do love the tutorial there, which is why I'm basically going through it infront of the class when teaching :smile: 

We must put third-party code into a suitable folder in `/third_party` to mark it as such.

mgeisler:
Thanks @randomPoison for putting this up! I'm really excited about us getting small bite-sized slides on this since it's been a big hole in our coverage.

Before we merge this, we should think about how this part will work. First, `cxx` is not included in the playground, so `mdbook test` fails. This is similar to the Android, bare-metal, and concurrency chapters where the instructor has to copy-paste the relevant examples into a separate editor and run them there.

Next, will you be able to show the features of CXX using a running example? Or will we use different examples? On one hand, I like the realistic nature of the CXX tutorial with the blob store client, but I also feel that it's a bit too big of a project for this course. So I actually like the examples you came up with: the `PlayingCard` struct is small and easy to understand right away.

What I'm missing is more commentary and contrasting: can we (auto-)generate and show the corresponding C++ code for this? For `Suit`, you mention in the notes that CXX will generate a Rust struct, so I would love to see that as a student :slightly_smiling_face: I suggest including instructions on how to view this struct — both for the instructor and for people reading the course by themselves.

I would expand the page on the type mapping with notes on _why_ the mapping is necessary. It's probably not immediately clear why a Rust `String` doesn't map 1:1 to `std::string` in C++.

Overall, I think this new chapter will be great, thanks for getting it started!

src/android/interoperability/cpp/blobstore/src/main.rs
randomPoison: Pretty much all of the code examples are taken from the CXX tutorial, with some modifications. Should I move all of the snippets into files in `/third-party`?

mgeisler: Yes, they all have to be separate if they are not our own creation.

randomPoison:
@mgeisler 

> Before we merge this, we should think about how this part will work. First, `cxx` is not included in the playground, so `mdbook test` fails. This is similar to the Android, bare-metal, and concurrency chapters where the instructor has to copy-paste the relevant examples into a separate editor and run them there.

I wasn't really designing the code snippets to be run directly. They're mostly just partial demonstrations of the CXX syntax or the generated code, so there's not usually anything to run.

> Next, will you be able to show the features of CXX using a running example?

I wasn't initially planning to create an example project, but Stephen expressed interest in adding one. It would also address the first point you brought up, since it would be an actually runnable example that I could pull up in class. I gave the blobstore example in the CXX repo a look, and I think (with some minor modification) it would be a good one to use as an example. I'll work on getting that integrated.

> What I'm missing is more commentary and contrasting: can we (auto-)generate and show the corresponding C++ code for this?

Ah, I hadn't included more of the generated code because it's not really shown in the CXX tutorial, but I think I should be able to add some.

> For `Suit`, you mention in the notes that CXX will generate a Rust struct, so I would love to see that as a student 🙂 I suggest including instructions on how to view this struct — both for the instructor and for people reading the course by themselves.

I've added a slide to demonstrate the code generated for a shared enum, and I've also added speakers notes to one of the earlier slides noting how to see the generated Rust and C++ code.

> I would expand the page on the type mapping with notes on _why_ the mapping is necessary. It's probably not immediately clear why a Rust `String` doesn't map 1:1 to `std::string` in C++.

There isn't really much to explain about why the type mappings are necessary: They're largely just defining a type in one language that corresponds to an existing type in the other language. Really the only confusing one is `String` vs `std::string`. I've added a speaker note elaborating on why these two can't be mapped to each other directly.

randomPoison:
Oh, one other thing that I realized we need is a slide or two about using CXX in Android. There's CXX support in the Android build system, so we should include slides showing what that setup looks like, and make sure we also have an `Android.bp` for the runnable example project so that we can demonstrate building/running for Android.

src/android/interoperability/cpp/blobstore/src/main.rs
randomPoison: How should I treat snippets that were initially copied from the CXX tutorial, but were then heavily modified?

randomPoison: I moved all of the snippets into `/third-party`, let me know if this works!

randomPoison:
@mgeisler I think I've addressed all outstanding feedback. Could you re-review now and let me know if I should make any further changes?

@fw-immunant This might also be a good point for you to come in and review it, now that I've gotten a pass of cleanup done.

src/android/interoperability/cpp/cpp-bridge.md
```diff
@@ -0,0 +1,52 @@
...
+                parts: *mut ::cxx::core::ffi::c_void,
+            ) -> u64;
+        }
+        unsafe {
+            __put(self, parts as *mut MultiBuf as *mut ::cxx::core::ffi::c_void)
+        }
+    }
+}
+
+// And so on...
```
mgeisler, line 41: The code snippets are extracted for translation — but if you remove the English text, the translator will know that they can just reuse the original text:

```suggestion
// ...
```

src/android/interoperability/cpp/cpp-bridge.md
```diff
@@ -0,0 +1,52 @@
+# C++ Bridge Declarations
+
+```rust,ignore
```
mgeisler, line 3: As a more general tool, let us mark the untranslatable code snippets with

````suggestion

<!-- mdbook-xgettext: skip -->
```rust,ignore
````

This will completely avoid putting the code into the PO files and thus greatly reduce the workload of the translators.

Comment (mgeisler) [reply src/android/interoperability/cpp/cpp-bridge.md:3]: Please add such a comment before every code block which cannot be translated, including the ones that include their content from another file.

You can see the effect of this with

```
cargo install mdbook-i18n-helpers
MDBOOK_OUTPUT='{"xgettext": {"pot-file": "before.pot"}}' mdbook build -d po
MDBOOK_OUTPUT='{"xgettext": {"pot-file": "after.pot"}}' mdbook build -d po
diff po/before.pot po/after.pot
```

You can then regenerate the POT file periodically to check that it only contains text that can be meaningfully translated.

File: src/android/interoperability/cpp/cpp-bridge.md [hunk 3]
```diff
@@ -0,0 +1,52 @@
...
+#[cxx::bridge]
+mod ffi {
+{{#include ../../../../third_party/cxx/blobstore/src/main.rs:cpp_bridge}}
+}
+```
+
+Results in (roughly) the following Rust:
+
+```rust,ignore
+#[repr(C)]
```
mgeisler, line 13: Very nice! During the class, I would want to generate this code and show the students where it is using [`bat`](https://github.com/sharkdp/bat) or similar...

How do I do this?

src/android/interoperability/cpp/type-mapping.md
```diff
@@ -0,0 +1,26 @@
...
+| `Vec<T>`          | `rust::Vec<T>`       |
+| `CxxVector<T>`    | `std::vector<T>`     |
+
+<details>
+
+* These types can be used in the fields of shared structs and the arguments and
+  returns of extern functions.
+* Note that Rust's `String` does not map directly to `std::string`. There are a
+  few reasons for this:
+  * `std::string` does not uphold the utf-8 invariant that `String` requires.
```
mgeisler, line 20:
```suggestion
  * `std::string` does not uphold the UTF-8 invariant that `String` requires.
```

mgeisler: Thanks for including the explanation here! I think it's important so people get a nice mental model of this.

src/android/interoperability/cpp/android-build-rust.md
```diff
@@ -0,0 +1,12 @@
+# Building in Android
+
+Create a `rust_binary` that depends on `libcxx` and your `cc_library_static`.
+
+```json
```
mgeisler, line 5: I used `javascript` as the language in the Android part (should be mostly the same, so I'm just mentioning it for consistency).

I also used a particular format for referring to external files, see [Rust Binaries](https://google.github.io/comprehensive-rust/android/build-rules/binary.html). 
I did that consistently in those slides: this was important since I hope people can copy-paste the code into their own slides and thus follow-along at home.  To make my life easy, all the files are then actual live files so that I can navigate to the right directory and do `m hello_rust_with_dep`.

See [build_all.sh](https://github.com/google/comprehensive-rust/blob/main/src/android/build_all.sh) for the script that I use to build everything once in a while. You should try to hook up your `Android.bp` files to the same script so that we can prevent bit-rot.

mgeisler: I just created #1434 to explain this process a bit — it's a little hacky, but I think it's better than nothing :grimacing:

third_party/cxx/blobstore/BUCK
mgeisler, line 1: We might not need to copy this file?

mgeisler:
Hi @randomPoison, thanks for all the updates! I think it looks like a great improvement already!

I will let @fw-immunant read it through as well and then I think we should merge this soon.

> > Before we merge this, we should think about how this part will work. First, `cxx` is not included in the playground, so `mdbook test` fails. This is similar to the Android, bare-metal, and concurrency chapters where the instructor has to copy-paste the relevant examples into a separate editor and run them there.
> 
> I wasn't really designing the code snippets to be run directly. They're mostly just partial demonstrations of the CXX syntax or the generated code, so there's not usually anything to run.

When I'm teaching, I find it important to copy-paste the examples into a file and show the students that it can actually run. Doing this live is important since it encourages experimentation: someone will inevitably ask "How does `X` translate?" and then you can quickly try it out with the class.

The first question I have when I see a code generator like CXX is: where is the output? Can I see it? Can I understand it? Showing the students the actual code and going over the high-level structure would be important. I do the same when talking about AIDL: I show how you can find the output files and how they look.

Yes, people could do this copy-pasting themselves, but it helps anchor the information when you see it actually done. Otherwise it's too each to just skim over the slides.

I think we could spend 30-45 minutes on this part since it's important with all the interop that we need in Android. We could even spend longer if we could find a project that people could dive into here — the Android day is currently set to be half a day, but I would be happy if it was a full day course.
 
> > Next, will you be able to show the features of CXX using a running example?
> 
> I wasn't initially planning to create an example project, but Stephen expressed interest in adding one. It would also address the first point you brought up, since it would be an actually runnable example that I could pull up in class. I gave the blobstore example in the CXX repo a look, and I think (with some minor modification) it would be a good one to use as an example. I'll work on getting that integrated.

Thank you! I personally really like the blobstore tutorial: it's really well done because it shows integration in both directions. I also love that it's a real world example.

> > What I'm missing is more commentary and contrasting: can we (auto-)generate and show the corresponding C++ code for this?
> 
> Ah, I hadn't included more of the generated code because it's not really shown in the CXX tutorial, but I think I should be able to add some.

I think the examples you've added are great, they're bound to stir up a good discussion in class :smile:

src/android/interoperability/cpp/cpp-bridge.md
randomPoison: Okay, can do! Which code blocks can't be translated, though? Do the translators usually translate any comments in code snippets?

src/android/interoperability/cpp/cpp-bridge.md
randomPoison: Probably the easiest way to view the generated code is to follow the speaker notes from the Bridge Module slide, i.e.:

* To view the generated Rust code, use [cargo-expand] to view the expanded proc
  macro. For most of the examples you would use `cargo expand ::ffi` to expand
  just the `ffi` module (though this doesn't apply for Android projects).
* To view the generated C++ code, look in `target/cxxbridge`.

You won't easily be able to do this directly with the code snippet in the slide, but you could pull up the blobstore example and build with Cargo to show this. The blobstore example is setup to build with both Cargo and Soong, so the easiest way to demo stuff will probably be with the Cargo version of the project.

[cargo-expand]: https://github.com/dtolnay/cargo-expand

src/android/interoperability/cpp/android-build-rust.md
randomPoison: > I also used a particular format for referring to external files, see [Rust Binaries](https://google.github.io/comprehensive-rust/android/build-rules/binary.html).

I'm not sure what you're referring to here? What "format" are you referring to? Do you mean the way that you're `#include`ing the snippets in from a larger example, whereas I'm writing most of them inline?

randomPoison:
> When I'm teaching, I find it important to copy-paste the examples into a file and show the students that it can actually run. Doing this live is important since it encourages experimentation: someone will inevitably ask "How does `X` translate?" and then you can quickly try it out with the class.
> 
> The first question I have when I see a code generator like CXX is: where is the output? Can I see it? Can I understand it? Showing the students the actual code and going over the high-level structure would be important. I do the same when talking about AIDL: I show how you can find the output files and how they look.
> 
> Yes, people could do this copy-pasting themselves, but it helps anchor the information when you see it actually done. Otherwise it's too each to just skim over the slides.
> 
> I think we could spend 30-45 minutes on this part since it's important with all the interop that we need in Android. We could even spend longer if we could find a project that people could dive into here — the Android day is currently set to be half a day, but I would be happy if it was a full day course.

This all sounds good, but I'm not sure what the actionable item is for this PR? I've tried to show the generated code as much as possible, and we now have a working example that we can demo to the class. Could you elaborate on what other changes you'd want to see so that it's easier to copy the sample code into a working example?

src/android/interoperability/cpp/cpp-bridge.md
mgeisler: Things are moving fast :smile: We've just gotten a PR which will help a lot with the translation effort by only extracting comments and strings: https://github.com/google/mdbook-i18n-helpers/pull/109. This makes the large code blocks much less annoying for the translators.

To answer your questions:

> Okay, can do! Which code blocks can't be translated, though?

Code where all static strings are things like `"a: {a}"` from a format string.

> Do the translators usually translate any comments in code snippets?

It's a bit uneven today. The Simplified Chinese translation actually translated some comments: https://google.github.io/comprehensive-rust/zh-CN/hello-world/small-example.html (but not all, as far as I know).

mgeisler: > This makes the large code blocks much less annoying for the translators.

Meaning: don't worry too much about this now.

src/android/interoperability/cpp/android-build-rust.md
mgeisler: > I'm not sure what you're referring to here? What "format" are you referring to?

Sorry, I was thinking of the 

````markdown

_filename.rs_:

```javascript

```
````

convention with the italic filename. If the precise name doesn't matter, then don't worry about following it (I found that the names matter a lot for the care of the AIDL code).

**Comment by mgeisler** [comment]
> This all sounds good, but I'm not sure what the actionable item is for this PR? I've tried to show the generated code as much as possible, and we now have a working example that we can demo to the class. Could you elaborate on what other changes you'd want to see so that it's easier to copy the sample code into a working example?

You're right, my rambling was not super actionable!

I think we should merge what we have now — it's a huge improvement over the current course! You'll test it out in class in about two weeks and then we'll hopefully know more what works and what doesn't.

Thanks for landing this!

**Review by mgeisler — APPROVED** [review]
Thanks!

**Comment by mgeisler** [comment]
@randomPoison and @fw-immunant, please coordinate with each other to see if Frances should also review this.

File: src/android/interoperability/cpp/cpp-bridge.md [hunk 6]
```diff
@@ -0,0 +1,52 @@
...
+    }
+}
+
+// ...
+```
+
+<details>
+
+* The programmer does not need to promise that the signatures they have typed in
+  are accurate; that would be unreasonable. CXX performs static assertions that
```
fw-immunant, line 47: More specifically it seems like it would make API changes down the road require remembering to manually recheck signatures, which could be a significant source of bugs.

src/android/interoperability/cpp/rust-result.md
```diff
@@ -0,0 +1,16 @@
...
+```rust,ignore
+{{#include ../../../../third_party/cxx/book/snippets.rs:rust_result}}
+```
+
+<details>
+
+* Rust functions that return `Result` are translated to exceptions on the C++
+  side.
+* The exception thrown will always be of type `rust::Error`, which primarily
+  exposes a way to get the error message string.
```
fw-immunant, line 12: Is this via the Rust [`Error`](https://doc.rust-lang.org/stable/std/error/trait.Error.html) trait?

fw-immunant (approved):
LGTM, a couple minor comments inline.

src/android/interoperability/cpp/cpp-bridge.md
randomPoison: I'm not sure what you're referring to here?

src/android/interoperability/cpp/rust-result.md
randomPoison: Not sure, more likely it's using the `Display` impl though, since that's what you're supposed to use with the `std::Error` trait anyway.

mgeisler:
Hi @randomPoison, please resolve the last discussions with @fw-immunant and merge this.

src/android/interoperability/cpp/cpp-bridge.md
fw-immunant: I mean that the justification is not just that it would be "unreasonable" but more specifically that it would be a source of bugs to trust the programmer's signatures unconditionally.

src/android/interoperability/cpp/rust-result.md
fw-immunant: https://cxx.rs/binding/result.html#returning-result-from-rust-to-c indicates that this does use `Display`; we might note that here as well.

src/android/interoperability/cpp/cpp-bridge.md
randomPoison: Ah sure. Removing the weird wording.

src/android/interoperability/cpp/rust-result.md
randomPoison: Added a note about `Display`.

PR #1401, opened 2023-10-21
zh-CN: translate android/interoperability/with-c/rust.md

nodmp:
Part of #324

**The translation of "mangling" refers to the Chinese translation on Wikipedia (https://en.wikipedia.org/wiki/Name_mangling#Rust).**


po/zh-CN.po
```diff
@@ -12831,13 +12881,21 @@ msgid ""
 "adb shell /data/local/tmp/analyze_numbers\n"
 "```"
 msgstr ""
+"```shell\n"
+"m analyze_numbers\n"
+"adb push \"$ANDROID_PRODUCT_OUT/system/bin/analyze_numbers /data/local/tmp"
```
nodmp, line 12886:
```suggestion
"adb push \"$ANDROID_PRODUCT_OUT/system/bin/analyze_numbers /data/local/"
```

po/zh-CN.po
```diff
@@ -12831,13 +12881,21 @@ msgid ""
...
+"\"\n"
```
nodmp, line 12887:
```suggestion
"tmp\"\n"
```

SketchK (approved):

po/zh-CN.po
```diff
@@ -12751,10 +12751,26 @@ msgid ""
 "}\n"
 "```"
 msgstr ""
+"```rust,editable\n"
```
mgeisler, line 12754: It's fine to decide not to translate the code — if you think this would make sense in general, then we can add a special `<!-- mdbook-xgettext: skip -->` comment right before the code block. That will prevent is from being extracted into the `messages.pot` file the next time someone refreshes the translation. See #1327 and related issues for details.

nodmp: I'll be submitting a patch to fix all this later, but for now, if I refresh the translation, it's going to break my other PR's!

PR #1405, opened 2023-10-21
Update std.md

justanotheranonymoususer:
Fix typo


mgeisler (approved):
Thanks for catching this!

PR #1421, opened 2023-10-24
Be more precise about printing in compound-types.md

### LINKED ISSUE #1412: Wrong speaker notes in "Compound types" chapter
In the "Compound types" chapter, the speaker notes suggest:
`We could also have used {a} and {a:?} without specifying the value after the format string.`

This does not work on this code:
```
error[E0277]: `[i8; 10]` doesn't implement `std::fmt::Display`
 --> src/main.rs:4:18
  |
4 |     println!("a: {a}");
  |                  ^^^ `[i8; 10]` cannot be formatted with the default formatter
  |
  = help: the trait `std::fmt::Display` is not implemented for `[i8; 10]`
  = note: in format strings you may be able to use `{:?}` (or {:#?} for pretty-print) instead
  = note: this error originates in the macro `$crate::format_args_nl` which comes from the expansion of the macro `println` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0277`.
error: could not compile `playground` (bin "playground") due to previous error
```

mgeisler:
Fixes #1412.

### DIFF
--- a/src/basic-syntax/compound-types.md
+++ b/src/basic-syntax/compound-types.md
@@ -12,7 +12,7 @@ Array assignment and access:
 fn main() {
     let mut a: [i8; 10] = [42; 10];
     a[5] = 0;
-    println!("a: {:?}", a);
+    println!("a: {a:?}");
 }
 ```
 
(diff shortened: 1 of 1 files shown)


src/basic-syntax/compound-types.md
```diff
@@ -39,10 +39,10 @@ Arrays:
 
 * We can use literals to assign values to arrays.
 
-* In the main function, the print statement asks for the debug implementation with the `?` format
-  parameter: `{}` gives the default output, `{:?}` gives the debug output. We
-  could also have used `{a}` and `{a:?}` without specifying the value after the
-  format string.
+* The `println!` macro asks for the debug implementation with the `?` format
```
sakex, line 42: Maybe add a comment on what  `Debug` output means (ie we can get it by adding `#[derive(Debug)]` to the struct definition)

sakex (approved):

src/basic-syntax/compound-types.md
mgeisler: I wanted to keep it light here, we don't know anything about traits or so at this point and so I think `#[derive(Debug)]` is a little early. It will appear later in the afternoon.

PR #1425, opened 2023-10-25
Updating the uk.po file to speed up translation work on the Ukrainian translation of the book

thefirans:
Updated translated version of `uk.po` file using `cloud-translate`, for further editing


mgeisler:
Hey @thefirans,

You'll see that the format check gives an error. The problem is that `cloud-translate` doesn't format the file the same way as `msgcat` or `msgmerge` does. If you're on a Linux machine, you should follow the [contribution steps (they were updated recently)](https://github.com/google/comprehensive-rust/blob/main/CONTRIBUTING.md) to install `dprint` and then use that to auto-format the file.

If you have the Gettext tools installed, you can also just do

```sh
msgcat -o po/uk.po po/uk.po
```

since that's what `dprint` runs behind the scenes.

By the way, I checked the diff locally using

```sh
gh pr diff 1425 | bat -l patch
```

and it shows me that a few `msgid` entries have changed (the wrapping is different). That is what makes the format check fail.

thefirans:
@mgeisler - Thanks for the suggestions! Indeed, I formatted the `uk.po` file correctly, after that the format test passed successfully. However some errors I still get, don't know if it is directly related to my file or not. 

Setup pages error:
```
Warning: Get Pages site failed
Error: Create Pages site failed
Error: AxiosError: Request failed with status code 403
```

mgeisler:
> @mgeisler - Thanks for the suggestions! Indeed, I formatted the `uk.po` file correctly, after that the format test passed successfully.

Great to hear!

> However some errors I still get, don't know if it is directly related to my file or not.
> 
> Setup pages error:
> 
> ```
> Warning: Get Pages site failed
> Error: Create Pages site failed
> Error: AxiosError: Request failed with status code 403
> ```

This looks like a transient error to me and I don't think we use anything called "Axios" in our software stack. So I'm not sure where it comes from :slightly_smiling_face:

thefirans:
> This looks like a transient error to me and I don't think we use anything called "Axios" in our software stack. So I'm not sure where it comes from 🙂

Alright then, looks like all the tests are passing :)

mgeisler:
> > This looks like a transient error to me and I don't think we use anything called "Axios" in our software stack. So I'm not sure where it comes from 🙂
> 
> Alright then, looks like all the tests are passing :)

Yep, now we "just" need @dyeroshenko to look over the translation.

Am I correct in thinking that you're run `cloud-translate` and also removed all fuzzy markers? Normally, I would suggest to only remove the fuzzy markers when a human has proofread the machine translations — but if you prefer to do it this way, then I'm also okay with it.

dyeroshenko:
Machine translation can be funny sometimes, but LGTM!

dyeroshenko:
@mgeisler LMK, if we are happy to merge? Tests are passing and LGTM for translation quality.

mgeisler:
> @mgeisler LMK, if we are happy to merge? Tests are passing and LGTM for translation quality.

Yes, please!

mgeisler:
@dyeroshenko, you get to do the merge — I'm not supposed to be involved in the translations. You have to self-organize, otherwise it will never scale :smile:

dyeroshenko:
> @dyeroshenko, you get to do the merge — I'm not supposed to be involved in the translations. You have to self-organize, otherwise it will never scale 😄

Just wanted to double-check because of the change size and migrating to `cloud-translate` :) 

Merging!

mgeisler:
Thanks!

Issue #1417, opened 2023-10-23
Detect deletion of slides without corresponding book.toml entry


djmitche:
We have occasionally deleted a file without adding an entry to `book.toml`, leaving existing links to that file with a 404 error page.

It shouldn't be too hard to detect, in the GitHub actions, that a file has been deleted, and verify that there is an entry in `book.toml` for the `.html` filename that it would have produced.

Gourav2609:
I have opened a PR regarding this .. Please check this out

bridadan:
If this is still relevant I'd be happy to pick up the work on this again

djmitche:
We've had less "churn" in book pages, but this would still be useful!

bridadan:
Great, in that case I'll take it!

PR #1430, opened 2023-10-25
Detect deletion of slides 

Gourav2609:
closes issue #1417

### DIFF
--- a/.github/workflows/verify-deleted-file-entries.yml
+++ b/.github/workflows/verify-deleted-file-entries.yml
@@ -22,39 +22,35 @@ jobs:
           # Install your dependencies here.
 
       - name: Check if it's a pull request event
-        id: is-pull-request
+        id: check-pr-event
         run: |
           if [ -n "$GITHUB_EVENT_NAME" ] && [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then
-            echo "::set-output name=is_pull_request::true"
+            echo "is_pull_request=true" >> $GITHUB_ENV
           else
-            echo "::set-output name=is_pull_request::false"
+            echo "is_pull_request=false" >> $GITHUB_ENV
           fi
 
       - name: Get base branch
         id: base-branch
         run: |
-          is_pull_request="${{ steps.is-pull-request.outputs.is_pull_request }}"
+          is_pull_request="${{ env.is_pull_request }}"
           if [ "$is_pull_request" == "true" ]; then
             base_branch=$(jq -r .pull_request.base.ref "$GITHUB_EVENT_PATH")
-            echo "::set-output name=base_branch::$base_branch"
+            echo "base_branch=$base_branch" >> $GITHUB_ENV
           else
-            echo "::set-output name=base_branch::"
+            echo "base_branch=" >> $GITHUB_ENV
           fi
 
-      - name: List deleted slides
-        id: deleted-slides
+      - name: Set deleted slides in an environment file
         run: |
-          base_branch="${{ steps.base-branch.outputs.base_branch }}"
+          base_branch="${{ env.base_branch }}"
           if [ -n "$base_branch" ]; then
-            deleted_slides=$(git diff --name-only $base_branch...HEAD --diff-filter=D)
-            echo "::set-output name=deleted_slides::$deleted_slides"
-          else
-            echo "::set-output name=deleted_slides::"
+            echo "DELETED_SLIDES=$(git diff --name-only \"$base_branch\"...HEAD --diff-filter=D)" >> $GITHUB_ENV
           fi
 
       - name: Check for entries in book.toml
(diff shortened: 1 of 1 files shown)


.github/workflows/verify-deleted-file-entries.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: actions/checkout@v2
+
+      - name: Install dependencies
+        run: |
+          # Install your dependencies here.
+
+      - name: List deleted slides
+        id: deleted-slides
+        run: |
+          deleted_slides=$(git diff --name-only HEAD^ HEAD)
```
djmitche, line 23: I think this will include all _modified_ slides, right?

Also, this only compares HEAD^ to HEAD -- what if the PR has multiple commits in it?

.github/workflows/verify-deleted-file-entries.yml
```diff
@@ -0,0 +1,38 @@
...
+          echo "::set-output name=deleted_slides::$deleted_slides"
+
+      - name: Check for entries in book.toml
+        run: |
+          deleted_slides="${{ steps.deleted-slides.outputs.deleted_slides }}"
+
+          for slide in $deleted_slides; do
+            slide_filename=$(basename -- "$slide")
+            slide_name="${slide_filename%.*}"
+            entry_exists=$(grep -c "\[slide.$slide_name\]" book.toml)
```
djmitche, line 33: I'm not sure what this pattern is intended to match. It looks like a character class?

Gourav2609:
any suggestions🙂  ?

djmitche (changes requested):
It would be great to be able to test this out. We could do so in this PR, but maybe a better approach would be to put this functionality in a shell script that could also be run locally, for testing purposes.

djmitche:
Oh, and please update the title of the PR to reflect what it does.

.github/workflows/verify-deleted-file-entries.yml
Gourav2609: updated it in my new commit .

.github/workflows/verify-deleted-file-entries.yml
Gourav2609: The pattern `"\[slide.$slide_name\]"` is not intended to create a character class. Instead, it is meant to match a string in the format of [slide.slide_name] where slide_name is a placeholder for the actual name of the slide.

djmitche: Ah -- and where does that appear in `book.toml`?

.github/workflows/verify-deleted-file-entries.yml
```diff
@@ -0,0 +1,63 @@
...
+
+      - name: Get base branch
+        id: base-branch
+        run: |
+          is_pull_request="${{ env.is_pull_request }}"
+          if [ "$is_pull_request" == "true" ]; then
+            base_branch=$(jq -r .pull_request.base.ref "$GITHUB_EVENT_PATH")
+            echo "base_branch=$base_branch" >> $GITHUB_ENV
+          else
+            echo "base_branch=" >> $GITHUB_ENV
```
djmitche, line 41: This will end up running `git diff ...HEAD` on anything that is not a pull request, which does nothing. If that is the intent, perhaps this action should only run on pull requests?

.github/workflows/verify-deleted-file-entries.yml
```diff
@@ -0,0 +1,63 @@
...
+          fi
+
+      - name: Set deleted slides in an environment file
+        run: |
+          base_branch="${{ env.base_branch }}"
```
djmitche, line 46: What is the reason to spread this processing over multiple workflow steps? Couldn't this all be accomplished in a single step?

djmitche (changes requested):
I'd really like to have a way to test this! I suspect this won't work as written (at least since it's looking for a pattern that doesn't exist in the file), but it's hard to tell for sure.

Gourav2609:
sry @djmitche for late reply ... will soon come up with changes

mgeisler:
Hi @Gourav2609, thanks for looking at this issue!

As a high-level comment, I would like code such as this to be written in Python. See https://github.com/google/comprehensive-rust/blob/main/.github/workflows/check-msgid-changes.yml which calls directly out to https://github.com/google/comprehensive-rust/blob/main/.github/workflows/check-msgid-changes.py.

We cannot maintain large amounts of Bash code, so that is why I think we should write this in a high-level language instead.

djmitche:
@Gourav2609 are you able to continue working on this?

Gourav2609:
@djmitche sry.. Got a little busy with college hackathon.. Will soon start working on it

djmitche:
Closing this as stale..

PR #1433, opened 2023-10-27
Fixture code and footer content tags in rtl content

moaminsharifi:
Part of #671 
and #1413
## Issue:

In the code part of content which always is in english and must be `text-align: left` but with `<html ... dir=rtl >` cuz conflict.
### For example:
استفاده از cargo - Comprehensive Rust 🦀
اجرای cargo روی ماشین local - Comprehensive Rust 🦀
other one in footer:
footer-ارسال و همگام سازی - Comprehensive Rust 🦀

## **how must be:**
FIX-اجرای cargo روی ماشین local - Comprehensive Rust 🦀
FIX-استفاده از cargo - Comprehensive Rust 🦀
FIX-footer-ارسال و همگام سازی - Comprehensive Rust 🦀

### Suggestion:
Due to the fact that the other way @mgeisler  *at [issue 1413](https://github.com/google/comprehensive-rust/issues/1413#issuecomment-1780763297)* suggested doesn't work at all, we need to modify the CSS

Adding rtl.css file with those content:
```
[dir="rtl"] .hljs,
[dir="rtl"] pre > code {
  text-align: left;
}
[dir="rtl"] #cookieBar {
  direction: ltr;
}
```


hamidrezakp:
@moaminsharifi Thank you.
But your branch contains unrelated commits too. can you do a rebase on `google/comprehensive-rust/main` and force-push again?

moaminsharifi:
> @moaminsharifi Thank you. But your branch contains unrelated commits too. can you do a rebase on `google/comprehensive-rust/main` and force-push again?

I will do it later,

> Also, have you tried adding `dir="auto"` to element? Browser itself will choose the right direction for text. if it starts with a RTL language it will be RTL and if LTR, it will be LTR. That's way better and it makes text BIDI. But we have to have it on every text element on the page, and it must be generated from mdbook.

Those specific parts like <pre> and <code> tag as I checked only we use for Rust code, Bash and any other english texts.

Plus adding `<tag ... dir=auto>` with js may cuz issue when:
1- user not enabled js
2- conflict with js render of code blocks
https://github.com/google/comprehensive-rust/issues/1413#issuecomment-1781428874

hamidrezakp:
> Plus adding `<tag ... dir=auto>` with js may cuz issue when: 1- user not enabled js 2- conflict with js render of code blocks [#1413 (comment)](https://github.com/google/comprehensive-rust/issues/1413#issuecomment-1781428874)

Not with js, but with be generated with mdbook.

mgeisler:
Thanks for looking at this!

> In the code part of content which always is in english and must be `text-align: left` but with `<html ... dir=rtl >` cuz conflict.

This is not technically true: the Simplified Chinese translation actually [translated some of the Rust comments](https://google.github.io/comprehensive-rust/zh-CN/hello-world/small-example.html):

image

I don't know how this would be expected to look in Persian, but I just wanted to mention that it is possible to translate this (that's actually why the code snippets are in the PO files). Please also see https://github.com/google/mdbook-i18n-helpers/issues/95 and let us know there if this is a terrible idea from the point of view of right-to-left text. I suspect that it is?

moaminsharifi:
> Thanks for looking at this!
> 
> > In the code part of content which always is in english and must be `text-align: left` but with `<html ... dir=rtl >` cuz conflict.
> 
> This is not technically true: the Simplified Chinese translation actually [translated some of the Rust comments](https://google.github.io/comprehensive-rust/zh-CN/hello-world/small-example.html):
> 
> image
> 
> I don't know how this would be expected to look in Persian, but I just wanted to mention that it is possible to translate this (that's actually why the code snippets are in the PO files). Please also see [google/mdbook-i18n-helpers#95](https://github.com/google/mdbook-i18n-helpers/issues/95) and let us know there if this is a terrible idea from the point of view of right-to-left text. I suspect that it is?

as I mentioned in the https://github.com/google/comprehensive-rust/issues/1413#issuecomment-1781428874
The issue is mostly about 
normal  part like:
 \`\`\`
fn main() {
println!("Edit me!");
}
 \`\`\`
or
 \`\`\`rust| bash
fn main() {
println!("Edit me!");
}
 \`\`\`

which use web-kit value
image
when we add `editable` to code block then will be fixed.


About rendering this part:

 \`\`\`rust,editable
fn main() {              // Program entry point
    let mut x: i32 = 6;  // Mutable variable binding
    print!("{x}");       // Macro for printing, like printf
    while x != 1 {       // No parenthesis around expression
        if x % 2 == 0 {  // Math like in other languages
            x = x / 2;
        } else {
            x = 3 * x + 1;
        }
        print!(" -> {x}");
    }
    println!();
}
 \`\`\`

یک مثال ساده - Comprehensive Rust 🦀
It's seems totally ok for me. (again, because we use `rust,editable` then it is always text algin left)


About [google/mdbook-i18n-helpers#95](https://github.com/google/mdbook-i18n-helpers/issues/95), how we must convert this part in to .po file?
 \`\`\`rust,editable
fn main() {              // Program entry point
    let mut x: i32 = 6;  // Mutable variable binding
    print!("{x}");       // Macro for printing, like printf
    while x != 1 {       // No parenthesis around expression
        if x % 2 == 0 {  // Math like in other languages
            x = x / 2;
        } else {
            x = 3 * x + 1;
        }
        print!(" -> {x}");
    }
    println!();
}
 \`\`\`
Now We just take whole part as msgid. if we separate it like as we know `//` is a way to comment in the line after the code in Rust.
How We must handle msgid and msgstr? can You @mgeisler give me an example?

@hamidrezakp  for using mbBook I am not sure how we can change rendering.

mgeisler:
> How We must handle msgid and msgstr? can You @mgeisler give me an example?

If you're asking what the `msgid` and `msgstr` fields do in the PO file, then please read [TRANSLATIONS.md](https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md) and let me know if you still have questions. In short, `msgid` is populated with the English text and `msgstr` is the corresponding translation.

mgeisler (approved):
Thanks for the code, @moaminsharifi and thanks for the review, @Manishearth!

mgeisler:
> Now We just take whole part as msgid. if we separate it like as we know `//` is a way to comment in the line after the code in Rust. How We must handle msgid and msgstr? can You @mgeisler give me an example?

Luckily @dalance gave an example in the PR: https://github.com/google/mdbook-i18n-helpers/pull/109#issuecomment-1784795489.

I expect to merge the PR soon and then we'll create a new release of mdbook-i18n-helpers. Then we'll see how it impacts RTL text (or you can already try it out with the code in the PR).

Issue #182, opened 2023-01-18
Impossible to open speaker notes again after closing them


qwandor:
I can't find any way to open the speaker notes again after closing them once. Steps to reproduce:

1. Go to https://google.github.io/comprehensive-rust/.
2. Click the link to open speaker notes in a new window. The new window opens, and the link disappears from the main window.
3. Close the speaker notes window. I would expect the link in the main window to reappear at this point, but it does not.
4. Go to a different page in the main window. Still no link to reopen speaker notes.
5. Refresh the main window. There is still no link to reopen speaker notes.
6. Close the main window, and open https://google.github.io/comprehensive-rust/ again. There's still no link to reopen speaker notes again, they are gone forever.

mgeisler:
Hey @qwandor,

Yes, the state is set to `popup` when you first open them in an external popup. You then get a close button in the top-right corner of the screen:

![image](https://user-images.githubusercontent.com/89623/213187353-7db04180-c287-4af2-876e-24c3025aa8b6.png)

Clicking this button will change the state back so that the notes appear inline again.

This is not ideal and I was afraid that it would be hard to notice the new button... I played with making it flash to red before fading back to grey. However, that was a bit too much CSS animation coding so I gave it up.

It used to be the case that closing the popup would make the notes appear inline again. However, this lead to another problem: if you paged quickly through the notes, you would end up in a broken state. I fixed that via #161.

In short, navigating to another page in the speaker looks the same to us as you closing the window. So it's hard to reliably react on you closing the window without also messing up the state when you navigate naturally between pages.

mgeisler:
I'm slowly working towards a redesign of the course theme (#1089). There, the idea is to combine the "open popup" button and the "close popup" button into a single toggle button. That should make it much clearer where to click to bring the notes back into the main window.

fw-immunant:
Seems to be fixed.

Issue #1232, opened 2023-09-22
Rewrite speaker note system


mgeisler:
The [`speaker-notes.js`](https://github.com/google/comprehensive-rust/blob/main/theme/speaker-notes.js) file is a little home-grown system for injecting speaker notes to the pages.

I would like to rewrite this to use a more modern development style:

- we should minimize the code when publishing the course
- we could perhaps use a library for some of the state logic?
- we could take advantage of https://github.com/google/mdbook-i18n-helpers/issues/70 when it's ready to do less on the fly DOM manipulation

mgeisler:
Making the labels translatable is also being looked at in #1390.

PR #1441, opened 2023-10-31
Reset state to `inline-open` when speaker note popup is closed

elliotbrandwein:
Partially addresses #182.

### DIFF
--- a/theme/speaker-notes.js
+++ b/theme/speaker-notes.js
@@ -118,7 +118,7 @@
         // bind the popup to reset the speaker note state on close of the popup
         popup.onload = () => {
           popup.onbeforeunload = () => {
-            setState('inline-open');
+            setState("inline-open");
             applyState();
           };
         };
--- a/theme/speaker-notes.js
+++ b/theme/speaker-notes.js
@@ -116,8 +116,11 @@
         setState("popup");
         applyState();
         // bind the popup to reset the speaker note state on close of the popup
-        popup.onload = ()=>{
-          popup.onbeforeunload = ()=>{setState('inline-open'); applyState();};
+        popup.onload = () => {
+          popup.onbeforeunload = () => {
+            setState('inline-open');
+            applyState();
+          };
         };
       } else {
         window.alert(
--- a/theme/speaker-notes.js
+++ b/theme/speaker-notes.js
@@ -115,6 +115,10 @@
       if (popup) {
         setState("popup");
         applyState();
+        // bind the popup to reset the speaker note state on close of the popup
+        popup.onload = ()=>{
+          popup.onbeforeunload = ()=>{setState('inline-open'); applyState();};
+        };
       } else {
         window.alert(
           "Could not open popup, please check your popup blocker settings."


mgeisler:
Hi @elliotbrandwein, thanks for the PR!

I tried a few things to get this working back when I implemented it: and I ran into a lot of weird restrictions on when one can use the `onbeforeunload` event...

Your solution works okay _until_ you navigate away from the page that opened the popup window. The two windows are synchronized, but after navigating forward or backward in either window, the unload event no longer triggers and so the state isn't reset.

My understanding is that browsers basically refuse to implement this according to spec because of spammy popup windows.

elliotbrandwein:
> Hi @elliotbrandwein, thanks for the PR!
> 
> I tried a few things to get this working back when I implemented it: and I ran into a lot of weird restrictions on when one can use the `onbeforeunload` event...
> 
> Your solution works okay _until_ you navigate away from the page that opened the popup window. The two windows are synchronized, but after navigating forward or backward in either window, the unload event no longer triggers and so the state isn't reset.
> 
> My understanding is that browsers basically refuse to implement this according to spec because of spammy popup windows.

I see. at this point i might be dipping into https://github.com/google/comprehensive-rust/issues/1232 but i can see the fix for this issue being that i build a speaker notes state system that keeps track of the page of the speaker notes & if the popup is active or not. If you dont mind feature creep Implement that into this pr

mgeisler:
> If you dont mind feature creep I can implement that into this pr

That would be wonderful actually! I've been waiting for someone with enough JavaScript skills to come and help out :sweat_smile:

moaminsharifi:
Do you need help on this?
@elliotbrandwein @mgeisler

mgeisler:
Hey @moaminsharifi, I would be happy to get help here :smile:

moaminsharifi:
After checking JS window documentation, My idea due to current system which is mostly based on pure js and not any webframe work is make it something like SPA-ish:

1- Disable navigation links' default behavior (using the css selector of `a.mobile-nav-chapters`) 

2- when user clicks on navigation links , we use fetch module of browser to get all html content and render it 
- We need to render the sidebar
- We need to render the content part
- We need to check if popup open then render that part ( and if user click in popup then update main window)

> https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API
> https://www.w3schools.com/jsref/met_doc_write.asp

3- check URL in browser with JS.

> https://stackoverflow.com/questions/824349/how-do-i-modify-the-url-without-reloading-the-page

I want to know I much I can modify in js part? like adding new js lib is ok?  @mgeisler

elliotbrandwein:
> After checking JS window documentation, My idea due to current system which is mostly based on pure js and not any webframe work is make it something like SPA-ish:
> 
> 1- Disable navigation links' default behavior (using the css selector of `a.mobile-nav-chapters`)
> 
> 2- when user clicks on navigation links , we use fetch module of browser to get all html content and render it
> 
> * We need to render the sidebar
> * We need to render the content part
> * We need to check if popup open then render that part ( and if user click in popup then update main window)
> 
> > https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API
> > https://www.w3schools.com/jsref/met_doc_write.asp
> 
> 3- check URL in browser with JS.
> 
> > https://stackoverflow.com/questions/824349/how-do-i-modify-the-url-without-reloading-the-page
> 
> I want to know I much I can modify in js part? like adding new js lib is ok? @mgeisler

That definitely works to solve the ticket [here](https://github.com/google/comprehensive-rust/issues/1232) but for the issue at hand of the main tab not getting the signal that the notes tab closed we still need to solve for. 

My thinking is that since `onbeforeunload` isn't reliable we use localStorage and a storage listener to trigger the logic for when the popup tab is open/closed.

That being this fix would have to wait until after you create the new tab lib, or it would need to be redone as part of your cl.

mgeisler:
> I want to know I much I can modify in js part? like adding new js lib is ok? @mgeisler

Hi @moaminsharifi, I would be fine with that — we don't have any particular limitations for for this, apart that the result should be maintainable and load fast, etc :slightly_smiling_face: 

> My thinking is that since `onbeforeunload` isn't reliable we use localStorage and a storage listener to trigger the logic for when the popup tab is open/closed.

I might have tried this, but I'm not sure! It's nearly a year since I played around with this, so I'm really appreciating someone new taking a look at this with fresh eyes!

mgeisler:
@elliotbrandwein, since I think the change here does improve the situation in some cases, we can merge the PR if you like? Then you and @moaminsharifi could discuss a better long-term solution (if one exists :smile:). That way we can stop discussing on the PR and collect our thoughts in concrete issues instead.

mgeisler (approved):

elliotbrandwein:
updated the format of the pr so it should (hopefully) pass the pre submits and get merged

mgeisler:
> updated the format of the pr so it should (hopefully) pass the pre submits and get merged

Thanks @elliotbrandwein! I hope you and @moaminsharifi can come up with some good solution for the general problem of syncing the state.

PR #1443, opened 2023-11-03
zh-CN: Fix miss htts for https

lispking:
htts => https


mgeisler (approved):
Thanks for catching this!

mgeisler:
Another place where #957 would have been helpful!

PR #1445, opened 2023-11-04
ja: added translation for control-flow

ternbusty:
Hello, JA translation team! (#652) I've just included some translations for the "control-flow" section (chapter 8). I'm open to any feedback or suggestions. Thank you in advance!

### DIFF
--- a/po/ja.po
+++ b/po/ja.po
@@ -3758,58 +3758,75 @@ msgid ""
 "becomes the value of the `if` expression. Other control flow expressions "
 "work similarly in Rust."
 msgstr ""
+"今まで見てきたように、Rust において `if` は式です。`if` 式のブロックは値を"
+"持っており、条件判定の結果に応じて評価されたブロックの値が `if` 式の値となり"
+"ます。Rust では他の制御フローの式も同様の動作をします。"
 
 #: src/control-flow/blocks.md:3
 msgid ""
 "A block in Rust contains a sequence of expressions. Each block has a value "
 "and a type, which are those of the last expression of the block:"
 msgstr ""
+"Rust におけるブロックの中にはいくつかの式が存在します。それぞれのブロックは値"
+"と型を持っており、ブロックの値や型はそのブロック内にある最後の式の値や型と一"
+"致します。"
 
 #: src/control-flow/blocks.md:27
 msgid ""
 "If the last expression ends with `;`, then the resulting value and type is "
 "`()`."
 msgstr ""
+"最後の式が `;` で終了した場合、ブロック全体の値と型は `()` になります。"
 
 #: src/control-flow/blocks.md:29
 msgid ""
 "The same rule is used for functions: the value of the function body is the "
 "return value:"
 msgstr ""
+"同じルールが関数についても適用されます。関数の body ブロックの値が、その関数"
+"の返り値となります。"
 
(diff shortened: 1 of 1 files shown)


po/ja.po
```diff
@@ -3758,162 +3758,214 @@ msgid ""
...
 msgid "Key Points:"
-msgstr ""
+msgstr "キーポイント: "
 
 #: src/control-flow/blocks.md:46
 msgid ""
 "The point of this slide is to show that blocks have a type and value in "
 "Rust. "
 msgstr ""
+"このスライドのポイントは、Rust における block は型と値を持つということです。"
```
keiichiw, line 3800:
```suggestion
"このスライドのポイントは、Rust における ブロック は型と値を持つということです。"
```

We translate "block" to 「ブロック」in other places

po/ja.po
```diff
@@ -3758,162 +3758,214 @@ msgid ""
...
+"`step_by` は、元のイテレータとは別の、各要素をスキップする `Iterator` を返す"
+"メソッドです。"
 
 #: src/control-flow/for-expressions.md:29
 msgid ""
 "Modify the elements in the vector and explain the compiler errors. Change "
 "vector `v` to be mutable and the for loop to `for x in v.iter_mut()`."
 msgstr ""
+"ベクタ内の要素を変更してみて、その結果生じるコンパイルエラーについて説明して"
+"ください。また、vector `v` をミュータブルに、for ループを `for x in v."
```
keiichiw, line 3884:
```suggestion
"ください。また、ベクタ `v` をミュータブルに、for ループを `for x in v."
```

keiichiw (approved):
Welcome and thank you so much for your contribution!

I only have a few comments. LGTM!

po/ja.po
```diff
@@ -3758,162 +3758,214 @@ msgid ""
...
 
 #: src/control-flow/for-expressions.md:26
 msgid "Index iteration is not a special syntax in Rust for just that case."
 msgstr ""
+"Rust では、インデックスによる反復処理のために特別な構文は提供されていません。"
 
 #: src/control-flow/for-expressions.md:27
 msgid "`(0..10)` is a range that implements an `Iterator` trait. "
-msgstr ""
+msgstr "`(0..10)` は `Iterator` トレイトを実装した range です。"
```
chikoski, line 3868: Perhaps we need to create another PR to the original text, but we could add more clarity by updating "range" with "Range". By this modification, the audiences could understand that (0..10) defines a Range type data which implements Iterator trait as they can find the Range data type, not range.

chikoski (commented):
Overall, the translated texts look good to me, while I left a minor comment.

po/ja.po
ternbusty: Thank you for the review. I have changed "block" to「ブロック」.

po/ja.po
ternbusty: I've corrected "vector" to 「ベクタ」. I will be more careful about inconsistent notation in future pull requests!

po/ja.po
ternbusty: Thank you for your review. As you pointed out, "Range" is a type, so I have promptly changed the Japanese translation to "Range" and made adjustments so that it's clear it's a type.

po/ja.po
keiichiw: Thanks! You can close the conversation by yourself once you address the comment so we can use open threads as a TODO list:)

PR #1448, opened 2023-11-06
Update zh-CN.po （修正25.5 impl Trait中形参实参使用错误）

JChaooo:
25.5 impl Trait中，
“与特征边界类似，impl Trait 语法可以在函数实参 和返回值中使用”
更正为
“与特征边界类似，impl Trait 语法可以在函数形参 和返回值中使用”

### DIFF
--- a/po/zh-CN.po
+++ b/po/zh-CN.po
@@ -9161,7 +9161,7 @@ msgstr "`impl Trait`"
 msgid ""
 "Similar to trait bounds, an `impl Trait` syntax can be used in function "
 "arguments and return values:"
-msgstr "与特征边界类似，`impl Trait` 语法可以在函数实参 和返回值中使用："
+msgstr "与特征边界类似，`impl Trait` 语法可以在函数形参 和返回值中使用："
 
 #: src/traits/impl-trait.md:6
 msgid ""


po/zh-CN.po
```diff
@@ -9161,7 +9161,7 @@ msgstr "`impl Trait`"
 msgid ""
 "Similar to trait bounds, an `impl Trait` syntax can be used in function "
 "arguments and return values:"
-msgstr "与特征边界类似，`impl Trait` 语法可以在函数实参 和返回值中使用："
+msgstr "与特征边界类似，`impl Trait` 语法可以在函数形参 和返回值中使用："
```
superwhd, line 9164: In my impression, "arguments" should be translated as "实参". Could you elaborate why you'd like to make this change?

JChaooo: If I haven't misunderstood this document, here, "impl Trait" should be used in the parameter declaration of a function.  Therefore, during parameter declaration, it should be referred to as "formal parameter," and it becomes the "actual argument" when used in the function.  In the English version of the document, the term "arguments" is used, which means "参数" in Chinese (without specifying what kind of parameter).  However, in the current Chinese translation, the term used is "实参" (actual argument).  According to the demo example, its actual meaning should be "形参" (formal parameter).

image

SketchK (approved):

po/zh-CN.po
superwhd: Thanks for sharing the context! I agree with using "形参" here. Also IMHO the original English text should've used 'parameters' instead of 'arguments'.

@mgeisler What do you think?

superwhd (approved):

NoahDragon (approved):

PR #1457, opened 2023-11-09
Merge two build steps

mgeisler:
We can simplify the conditional steps by installing and running `msgfmt` in the same step.


mgeisler:
Actually, I found a better way for this: we can use `msgfmt` to show statistics about the course in the English case.

PR #1429, opened 2023-10-25
Re-apply formatting to SUMMARY.md

mgeisler:
With the latest version of mdbook-i18n-helpers, we now have support for formatting in the `SUMMARY.md` file. This allows us to revert the changes made in #130 when we first introduced the translation support.

This also aligns a few headings to match the heading in the outline. While this isn’t strictly necessary, it removes a bit of work for the translators.

PR #1458, opened 2023-11-09
Test that we can always extract strings

mgeisler:
This adds a check that `mdbook-xgettext` can always extract the strings of the course. Without this, it’s possible to merge a change which will make `mdbook-xgettext` error out the next time a translator tries to refresh their translation.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -122,7 +122,6 @@ jobs:
         uses: ./.github/workflows/setup-rust-cache
 
       - name: Install Gettext
-        if: matrix.language != 'en'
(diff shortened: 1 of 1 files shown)


mgeisler:
I noticed this problem in #1073 which introduces formatting in the `SUMMARY.md` file. This currently doesn't work, but the next version of mdbook-i18n-helpers will fix that (https://github.com/google/mdbook-i18n-helpers/issues/66). I plan on re-introducing formatting in #1429, but that might of course become obsolete by #1073.

djmitche (approved):

PR #1468, opened 2023-11-11
Fill the remaining glossary definitions

### LINKED ISSUE #1466: Fill out the English glossary
We have had a glossary for a few weeks now: https://google.github.io/comprehensive-rust/glossary.html

It is mostly incomplete and we could use help with filling out the definitions.

ANURAGSISODIYAA:
fixes #1466

### DIFF
--- a/po/fa.po
+++ b/po/fa.po
@@ -17504,313 +17504,305 @@ msgid ""
 msgstr ""
 
 #: src/glossary.md:30
-msgid ""
-"allocate:  \n"
-"Dynamic memory allocation on [the heap](memory-management/stack-vs-heap.md)."
-msgstr ""
+msgid "allocate:"
+msgstr "allocate:  \nDynamic memory allocation on [the heap](memory-management/stack-vs-heap.md)."
 
 #: src/glossary.md:32
 msgid "argument:"
-msgstr "Argument: Information passed to a function or method when it is called."
+msgstr "argument:  \nInformation passed to a function or method when it is called."
 
 #: src/glossary.md:33
-msgid ""
-"Bare-metal Rust:  \n"
-"Low-level Rust development, often deployed to a system without an operating "
-"system. See [Bare-metal Rust](bare-metal.md)."
-msgstr "Bare-metal Rust: Low-level Rust development, often deployed to a system without an operating system. See [Bare-metal Rust](bare-metal.md)."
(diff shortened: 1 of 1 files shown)


hamidrezakp:
Oh thanks man,
@moaminsharifi, Can you update and translate this parts after this PR lands?

moaminsharifi:
> Oh thanks man, @moaminsharifi, Can you update and translate this parts after this PR lands?

I can translate those, but it will take me some time since my PR queue is busy at the moment.

po/fa.po
```diff
@@ -17511,58 +17511,58 @@ msgstr ""
 
 #: src/glossary.md:32
 msgid "argument:"
-msgstr ""
+msgstr "Argument: Information passed to a function or method when it is called."
```
mgeisler, line 17514: Wow, thanks for filling out all of these!

Please note that the process is a bit different: you need to modify this Markdown file to fill out the missing English definitions: https://github.com/google/comprehensive-rust/blob/main/src/glossary.md.

Make sure to follow the special style I used there:

```markdown
- term:\
  Explanation of the term.
```

That will ensure that the `term:` part is highlighted in bold and that the explanation comes on the next line. It will also ensure that each term is separated out into its own individually translatable message here!

After that is merged, the English `msgid` fields you see here will update naturally as part of the next [translation refresh](https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md#refreshing-an-existing-translation).

mgeisler (changes requested):
I love the explanations here, they're top-notch! Thanks a ton for adding all the links to the individual pages, that's a lot of work!

We just need to incorporate the changes using the right process, as explained in my other comments.

ANURAGSISODIYAA:
Oh Soory ! I would correct the formatig.

po/fa.po
ANURAGSISODIYAA: sure

ANURAGSISODIYAA (commented):
Updated glossary terms in correct format and also filled some missing terms

mgeisler:
I think we can close this now that we have #1466 merged!

The glossary terms will flow into the Persian translation on the next refresh.

mgeisler:
Thanks for #1466, which makes this obsolete.

PR #1460, opened 2023-11-09
Update to mdbook-i18n-helpers 0.3.0

mgeisler:
This version has much improved support for the translation of code blocks. See https://github.com/google/mdbook-i18n-helpers/issues/95 for details.

Most PO files won’t need any update since most of them don’t translate the comments in the code blocks. Those that do, can run `mdbook-i18n-normalize` when they feel like it.

PR #1469, opened 2023-11-11
File normalized with `mdbook-i18n-helpers` 0.3.0

moaminsharifi:
#317 #1460


mgeisler:
Because of #1471 and https://github.com/google/mdbook-i18n-helpers/issues/125, I suggest closing this PR: normalization is not supported when the translation has "more" content than the original string (the `foo <span>bar</span>` parts counts as more content in this case).

PR #1474, opened 2023-11-12
Add kantasv to CODEOWNERS

kantasv:
This adds myself to CODEOWNERS. My main focus is translation, but I'd like to add me for the future maintenance within the projcect, such as github actions and tools. Thanks!


mgeisler (approved):
Awesome to have you here!

PR #1452, opened 2023-11-07
Update other-resources.md

VENCYR:
Added Rust Primer and CodeCrafters

PR #1477, opened 2023-11-13
Revert "Update other-resources.md"

randomPoison:
Reverts google/comprehensive-rust#1452, as the learning resource added there is not fully free content.


mgeisler (approved):
Let's take these out again: the other resources are (fully) free and cover Rust well.

PR #1482, opened 2023-11-17
ja: Normalize by running through mdbook-i18n-normalize 0.3.0

ternbusty:
Hello, JA translation team! (https://github.com/google/comprehensive-rust/issues/652) 

As part of #1460, I normalized ja.po using mdbook-i18n-normalize 0.3.0.

Steps taken:

1. Refreshed `po` file using the following command:

```
MDBOOK_OUTPUT='{"xgettext": {"pot-file": "messages.pot"}}' \
  mdbook build -d po
msgmerge --update po/ja.po po/messages.pot
```

2. Ran `mdbook-i18n-normalize`

I'm open to any feedback or suggestions. Thank you in advance!

### DIFF
--- a/po/ja.po
+++ b/po/ja.po
@@ -1,15 +1,16 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: Comprehensive Rust 🦀\n"
-"POT-Creation-Date: 2023-11-17T12:51:38+09:00\n"
+"POT-Creation-Date: 2023-10-15T23:41:49Z\n"
 "PO-Revision-Date: 2023-06-06 13:18+0900\n"
 "Last-Translator: Kenta Aratani <kentaaratani@coinez.jp>\n"
 "Language-Team: Japanese <translation-team-ja@lists.sourceforge.net>\n"
+"Language: ja\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"Language: ja\n"
 "Plural-Forms: nplurals=1; plural=0;\n"
+"X-Generator: Poedit 3.3.2\n"
 
 #: src/SUMMARY.md:4 src/index.md:1
 msgid "Welcome to Comprehensive Rust 🦀"
@@ -52,7 +53,7 @@ msgid "Day 1: Morning"
 msgstr "Day 1： AM"
 
 #: src/SUMMARY.md:19 src/SUMMARY.md:80 src/SUMMARY.md:135 src/SUMMARY.md:193
-#: src/SUMMARY.md:231 src/SUMMARY.md:281
+#: src/SUMMARY.md:219 src/SUMMARY.md:269
 msgid "Welcome"
 msgstr "Welcome"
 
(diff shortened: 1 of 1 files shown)


mgeisler:
> Steps taken:
> 
> 1. Refreshed `po` file using the following command:
> 
> ```
> MDBOOK_OUTPUT='{"xgettext": {"pot-file": "messages.pot"}}' \
>   mdbook build -d po
> msgmerge --update po/ja.po po/messages.pot
> ```
> 
> 2. Ran `mdbook-i18n-normalize`

Hi @ternbusty, thanks for working on this!

You could not really know this from our documentation, but the two steps should be done in the opposite order. The reason is that `mdbook-i18n-normalize` known how to turn

```po
msgid ""
"fn main() {\n"
"  let x = 10; // foo\n"
"  let y = 20; // bar\n"
"}"
msgstr ""
"fn main() {\n"
"  let x = 10; // Japanese foo\n"
"  let y = 20; // Japanese bar\n"
"}"
```

into

```po
msgid "// foo\n"
msgstr "// Japanese foo\n"

msgid "// bar\n"
msgstr "// Japanese bar\n"
```

That is, it knows how to find the matching translations and thus how to _preserve_ the translations.

When you use `msgmerge` on the file, no such matching is done. The result is likely that the comments are un-translated in the `ja.po` file. That's a valid state, but some work is lost if the translations were translated before the `msgmerge`. It's quite subtle and you will need to find a code snippet with translated comments to see the difference.

In general, I would recommend against running `msgmerge` at the same time as you run `mdbook-i18n-helpers`. When you run them together, you end up with an extra-large PR, but if you just run the normalization tool by itself, then the PR diff should be more understandable.

ternbusty:
Hi @mgeisler, Thank you for the detailed explanation!
I have reverted the above two commits and only performed normalization using mdbook-i18n-helpers without executing msgmerge.

mgeisler:
> Hi @mgeisler, Thank you for the detailed explanation! I have reverted the above two commits and only performed normalization using mdbook-i18n-helpers without executing msgmerge.

Thanks, it now looks correct from a quick skim (I don't speak Japanese, but the diff looks like I would expect it to look).

One last thing... because of a small bug (https://github.com/google/mdbook-i18n-helpers/issues/117), you had to modify the POT-Creation-Date. It's great that you figured it out, but you should set the date to the original date +1 second instead.

The timestamp found there is used when we publish the translation. We use `git restore` to bring the Markdown files back to how they looked at the time in the POT-Creation-Date. This allows us to "freeze" the translation in time and give people a stable ground to work on.

This also means that when you advance the time to "now", the next publication will have lots of untranslated strings. You should be able to see this if you download the [comprehensive-rust-ja](https://github.com/google/comprehensive-rust/suites/18300736434/artifacts/1058503709) artifact from the bottom of the [Test](https://github.com/google/comprehensive-rust/actions/runs/6911382975) page.

Until we fix https://github.com/google/mdbook-i18n-helpers/issues/117 and make a new release, we will have this little issue.

ternbusty:
Thank you for your guidance. I've adjusted the POT-Creation-Date to be the original date +1 second.

mgeisler:
@kantasv, @keiichiw, and @chikoski, I think this looks good from a technical point of view!

PR #1484, opened 2023-11-19
Cleaner implementation luhn.rs

AadeeWasTaken:
I think that this implementation gets rid of most of the nesting making the code much easier to understand.

### DIFF
--- a/src/exercises/day-1/luhn.rs
+++ b/src/exercises/day-1/luhn.rs
@@ -15,32 +15,36 @@
 // ANCHOR: solution
 // ANCHOR: luhn
 pub fn luhn(cc_number: &str) -> bool {
-    // ANCHOR_END: luhn
+    let cc_number = cc_number.replace(" ", "");
     let mut sum = 0;
-    let mut double = false;
-    let mut digit_seen = 0;
+    let mut even = false;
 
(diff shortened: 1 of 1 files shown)


src/exercises/day-1/luhn.rs
```diff
@@ -15,34 +15,38 @@
...
         return false;
     }
 
+    for i in (0..cc_number.len()).rev() {
+        let mut num = cc_number
+            .chars() //for new line
+            .nth(i)
+            .unwrap()
+            .to_digit(10)
+            .unwrap();
```
mgeisler, line 36: The use of `unwrap` here is a step back: the code here now relies on the `parse` call above to not error out. It is better to handle this via the code structure (e.g., `match` or similar).

src/exercises/day-1/luhn.rs
```diff
@@ -15,34 +15,38 @@
 // ANCHOR: solution
 // ANCHOR: luhn
 pub fn luhn(cc_number: &str) -> bool {
-    // ANCHOR_END: luhn
```
mgeisler, line 18: The comment here is important: it is used to include just the function signature into the exercise slide.

mgeisler:
Hi @AadeeWasTaken, let us close this one in preference of your other PR (#1486).

PR #1486, opened 2023-11-20
Update luhn.rs

AadeeWasTaken:
Made it easier to read and also got rid of unwrap

### DIFF
--- a/src/exercises/day-1/luhn.rs
+++ b/src/exercises/day-1/luhn.rs
@@ -16,31 +16,32 @@
 // ANCHOR: luhn
 pub fn luhn(cc_number: &str) -> bool {
     // ANCHOR_END: luhn
+    let cc_number = cc_number.replace(" ", "");
     let mut sum = 0;
-    let mut double = false;
-    let mut digit_seen = 0;
+    let mut even = false;
 
-    for c in cc_number.chars().filter(|&f| f != ' ').rev() {
-        if let Some(digit) = c.to_digit(10) {
-            if double {
-                let double_digit = digit * 2;
-                sum += if double_digit > 9 {
-                    double_digit - 9
-                } else {
-                    double_digit
-                };
-            } else {
-                sum += digit;
-            }
-            double = !double;
-            digit_seen += 1;
-        } else {
-            return false;
-        }
+    if cc_number.len() < 2 {
+        return false;
     }
 
-    if digit_seen < 2 {
-        return false;
+    for c in cc_number.chars().rev() {
+        let mut num;
+
+        match c.to_digit(10) {
+            Some(n) => num = n,
+            None => return false,
+        }
+
+        if even {
+            num = num * 2;
+
+            if num > 9 {
+                num = (num / 10) + (num % 10);
+            }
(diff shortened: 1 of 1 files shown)


src/exercises/day-1/luhn.rs
```diff
@@ -16,31 +16,32 @@
 // ANCHOR: luhn
 pub fn luhn(cc_number: &str) -> bool {
     // ANCHOR_END: luhn
+    let cc_number = cc_number.replace(" ", "");
     let mut sum = 0;
-    let mut double = false;
-    let mut digit_seen = 0;
+    let mut even = false;
```
mgeisler, line 21: I think what you're essentially doing here is trading an allocation (`cc_number.replace(" ", "")` will typically have to allocate) for the `digit_seen` variable which tracks the number of characters seen.

I would prefer to avoid the allocation and thus remove the call to `replace` — allocation is not necessary when we can just filter out the spaces.

Your PR is useful since it can refactor the nested matching and if-statements into something that uses an early return.

src/exercises/day-1/luhn.rs
```diff
@@ -16,31 +16,32 @@
...
+        match c.to_digit(10) {
+            Some(n) => num = n,
+            None => return false,
+        }
+
+        if even {
+            num = num * 2;
+
+            if num > 9 {
+                num = (num / 10) + (num % 10);
```
mgeisler, line 39: We have this code in the past :slightly_smiling_face:  See [the history for this exercise](https://github.com/google/comprehensive-rust/commits/main/src/exercises/day-1/luhn.rs).

People said in the past that the `if sum > 9  { sum = sum - 9 }` logic is easier to understand.

src/exercises/day-1/luhn.rs
```diff
@@ -16,31 +16,32 @@
...
             }
-            double = !double;
-            digit_seen += 1;
-        } else {
-            return false;
```
mgeisler, line 38: You should be able to turn this return into an early return.

src/exercises/day-1/luhn.rs
```diff
@@ -16,31 +16,32 @@
...
+        return false;
+    }
+
+    for c in cc_number.chars().rev() {
+        let mut num;
+
+        match c.to_digit(10) {
+            Some(n) => num = n,
+            None => return false,
+        }
```
mgeisler, line 33: Please use the `if let Some(num) = c.to_digit(10) else return false` pattern here.

src/exercises/day-1/luhn.rs
mgeisler: Can you take out this part since the allocation is ultimately unnecessary?

mgeisler (changes requested):
Hi @AadeeWasTaken, could you merge in the latest `main` and see if there are still things to improve here?

Please also take out the parts that I commented on. Thanks!

mgeisler:
The exercise has been updated a little in the meantime, so I suggest we close this PR. @aadeexyz, please open another PR if you still think this can be made clearer!

PR #1490, opened 2023-11-20
Skip calling “apt update” in GH actions

mgeisler:
We are currently getting an error when we run “apt update”:

    Reading package lists...
    E: Repository 'https://packages.microsoft.com/ubuntu/22.04/prod jammy InRelease' changed its 'Origin' value from 'microsoft-ubuntu-jammy-prod jammy' to 'Pulp 3'
    E: Repository 'https://packages.microsoft.com/ubuntu/22.04/prod jammy InRelease' changed its 'Label' value from 'microsoft-ubuntu-jammy-prod jammy' to ''
    Error: Process completed with exit code 100.

I’m not sure why we get this, but updating should not be necessary in the first place: the Ubuntu images already come with an up-to-date list of packages.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,9 +76,7 @@ jobs:
         uses: ./.github/workflows/setup-rust-cache
 
       - name: Install dependencies
-        run: |
-          sudo apt update
-          sudo apt install gcc-aarch64-linux-gnu
+        run: sudo apt install gcc-aarch64-linux-gnu
 
       - name: Install toolchain
         run: rustup target add ${{ matrix.target }}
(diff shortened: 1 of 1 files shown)


mgeisler:
This seems to fix the (temporary?) build errors we [see on the `main` branch](https://github.com/google/comprehensive-rust/actions/runs/6931470806/job/18856278197).

@djmitche and @randomPoison, what do yo think of it? I originally wanted to take this out in #634 — I've never bothered to update the caches in any of my other actions. The documentation does actually recommend updating the cache, but I'm not sure why when it comes to system package caches which are demonstratively already up to date (you _do_ need it if you add new package archives, but we don't do that).

mgeisler:
I'm seeing other repositories with the same errors, e.g., https://github.com/microsoft/vscode/issues/198658.

mgeisler:
> I think this is fine. Most likely something has gotten mixed up in the apt repo, and it will likely happen again. Avoiding getting mixed up in that is worthwhile.

Yeah, that's a nice way to put it :slightly_smiling_face: I guess we'll see if something starts breaking later down the road.

PR #1492, opened 2023-11-21
Test translations using same source as for publish

mgeisler:
Before we would run `mdbook test` using the current Markdown sources.
This is subtly wrong: we publish the course using back-dated sources,
so we should therefore also run the tests using the same sources (this
ensures that the code snippets actually work).

After this commit, all translatable content lives in exactly two
directories:

- `src/`
- `third_party/`

We need to restore both directories when testing and when publishing.
This ensures consistency in the Markdown text and in the included
source code.

A new `.github/workflows/build.sh` script takes care of preparing the
two directories according to the date in the PO file (if any).

To ensure we can restore all of `third_party/` to an old commit, the
non-changing `third_party/mdbook/book.js` file has been moved to
`theme/book.js`. The file is generated by `mdbook init --theme`,
making it suitable for modification by the user (us). Symlinks have
been added to `third_party/mdbook/` to indicate that the files
ultimately came from upstream.


henrif75 (approved):

djmitche (approved):
Looks good to me! This brings `publish.yml` and `build.yml` closer, which should help eliminate problems that occur only on publishing (merging to main) and not in PR checks.

mgeisler:
Thanks both for the review! I'll test it out on https://github.com/mgeisler/comprehensive-rust/actions/runs/7411755259 and then merge the PR.

mgeisler:
Seems to work fine!

PR #1442, opened 2023-11-01
ja: translate Ch. 41 and 42

HidenoriKobayashi:
Japanese translation for Ch. 41 & 42.

Part of #652.

PR #1497, opened 2023-11-28
ja: translate Ch. 43 and 44 (#1442)

HidenoriKobayashi:
Part of #652.


HidenoriKobayashi:
I didn't realize that I was going into the next chapter, again. Sorry.

po/ja.po
```diff
@@ -13016,6 +13035,10 @@ msgid ""
 "generating unaligned accesses, so it should be fine in this case, but this "
 "is not necessarily the case in general."
 msgstr ""
+"アンアラインド・アクセスがフォールトになります。我々はコンパイラがアンアライ"
```
keiichiw, line 13038: How about「アラインされていないアクセスが失敗します」?
I don't think 「アンアラインド・アクセス」 is a common phrase.

po/ja.po
```diff
@@ -13035,30 +13066,41 @@ msgid ""
 "for DRAM, and another 1 GiB higher up for more devices. This matches the "
 "memory layout that QEMU uses."
 msgstr ""
+"単純化のために、ハードコードしたページテーブル（`idmap.S`参照）を利用します。"
+"このページテーブルは最初の1GiBをデバイス用に、次の1GiBをDRAM用に、次の1GiBを"
+"さらなるデバイス用に透過的にマップします。これはQEMUのメモリマップに合致しま"
```
keiichiw, line 13071: nit
```suggestion
"さらなるデバイス用に透過的にマップします。これはQEMUのメモリレイアウトに合致しま"
```

po/ja.po
```diff
@@ -14077,34 +14284,48 @@ msgid ""
 "are not available in `core` or `alloc`. How can we manage synchronisation or "
 "interior mutability, such as for sharing state between different CPUs?"
 msgstr ""
+"`std::sync`が提供する`std::sync::Mutex` とその他の同期プリミティブはcore`また"
```
keiichiw, line 14287:
```suggestion
"`std::sync`が提供する`std::sync::Mutex` とその他の同期プリミティブは`core`また"
```

po/ja.po
```diff
@@ -13132,66 +13186,88 @@ msgid ""
...
 msgstr ""
+"`addr_of!`を用いると、中間的な参照を作らずに構造体のフィールドにアクセスする"
+"ことができます。"
 
 #: src/bare-metal/aps/mmio.md:9
 msgid ""
 "Volatile access: read or write operations may have side-effects, so prevent "
 "the compiler or hardware from reordering, duplicating or eliding them."
 msgstr ""
+"Volatileアクセス：（MMIO領域に対する）リードやライトは副作用があることがある"
```
keiichiw, line 13228:
```suggestion
"Volatileアクセス：MMIO領域に対する読み書きは副作用があることがある"
```

* I think you translate "read" and "write" to "読み" and "書き" at places, respectively. Either set is fine but it should be consistent
* I don't think `()` is needed to add phrases that are omitted in the English version for most cases because normal readers don't  care about the correspondence between Japanese version and English version.

keiichiw: Or, I think we can use `read` and `write` as is like "MMIO領域に対するreadやwriteは副作用があることがある". I personally feel that "read" is more natural than "リード", as it's clear that it's not "lead"

po/ja.po
```diff
@@ -13257,30 +13333,43 @@ msgid ""
...
+"ことができるようになることが理由です。"
 
 #: src/bare-metal/aps/uart.md:60
 msgid ""
 "We could have done it the other way around (making `new` safe but "
 "`write_byte` unsafe), but that would be much less convenient to use as every "
 "place that calls `write_byte` would need to reason about the safety"
 msgstr ""
+"逆に（`new`をセーフにして、`write_byte` をアンセーフに）することもできました"
+"が、そうすると`write_byte`の全呼び出し箇所において安全性について推論しなけれ"
```
keiichiw, line 13350: 「推論」sounds like that the compiler infer the safety, but that's not case. Here, we developers would need to reason. So,  I guess「安全性を考慮する」or 「安全性を確認する」would be better. It's not a direct translation, though

po/ja.po
```diff
@@ -13326,18 +13419,23 @@ msgid ""
 "read. Plus, some of them are bit fields which would be nice to access in a "
 "structured way."
 msgstr ""
+"実際のところPL011には[もっと多くのレジスタ](https://developer.arm.com/"
+"documentation/ddi0183/g/programmers-model/summary-of-registers)があり、それら"
+"にアクセスするためにオフセットを足してポインタを得ることは間違えになりやす"
+"く、可読性を低下させます。さらに、いくつかはビットフィールドなので、体系的に"
```
keiichiw, line 13425: nit
```suggestion
"く、可読性を低下させます。さらに、いくつかはビットフィールドなので、構造化された方法で"
```

keiichiw (approved):
Thanks! I added several comments, which are not critical at all.

HidenoriKobayashi:
Thanks for the review and all the suggestions! I fixed the wording as suggested.

HidenoriKobayashi:
I forgot to mention you in my previous comment. Please take a look at the fixes. @keiichiw

keiichiw:
Sorry, I missed your reply in the mail box. LGTM!

PR #1518, opened 2023-11-29
Use `std::string_view` (not just `const char*`) as an analogy of `&str`.

anforowicz:


anforowicz:
I think that `std::string_view` (pointer + length) is a better analogy of `&str` (also pointer + length) than `const char*` (only a pointer).  WDYT?

djmitche (approved):
I think this makes a lot more sense, thanks!

PR #1519, opened 2023-11-29
Update setup.md

ningcng:
```
$ lunch aosp_cf_x86_64_phone-userdebug

Invalid lunch combo: aosp_cf_x86_64_phone-userdebug
Valid combos must be of the form <product>-<release>-<variant>
```


mgeisler (approved):
Wow, thanks a lot for trying this out! I suspect we have the same bug in the `src/android/build-all.sh` script?

PR #1521, opened 2023-11-29
Add course and session markers for deep-dives

### LINKED ISSUE #1510: Add course metadata to deep-dives
Currently, "Fundamentals" has Markdown frontmatter that specifies its structure and timing. Let's add a similar thing to the deep-dives.

Let's do this in such a way that I don't have to introduce timing information, just structure.

djmitche:
This doesn't add any timing information. The authors of each deep-dive can do that separately.

Fixes #1510.


mgeisler (approved):
Thanks!

PR #1528, opened 2023-11-30
Add slide on GoogleTest

mgeisler:
We expect Android engineers to start using this crate going forward.

### DIFF
--- a/src/testing/Cargo.toml
+++ b/src/testing/Cargo.toml
@@ -4,7 +4,15 @@ version = "0.1.0"
 edition = "2021"
 publish = false
 
+[[example]]
+name = "googletest-example"
+crate-type = ["staticlib"]
+path = "googletest.rs"
+test = true
+
 [[bin]]
 name = "luhn"
 path = "exercise.rs"
 
+[dependencies]
+googletest = "0.10.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -749,6 +749,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "googletest"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "09213705c85aa0e4b4fff44a3a826a556979a34a266df6bcda703a49c69fb61e"
+dependencies = [
+ "googletest_macro",
+ "num-traits",
+ "regex",
+ "rustversion",
+]
+
+[[package]]
+name = "googletest_macro"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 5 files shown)


mgeisler:
The 12 minutes is a rough guess: we'll have to try this out by actually teaching the class.

mgeisler:
@marshallpierce, this is a bit of new material — don't worry if you don't have time to play with it for the class next week.

src/testing/useful-crates.md
```diff
@@ -1,13 +0,0 @@
...
-
-# Useful Crates
-
-Rust comes with only basic support for writing tests.
-
-Here are some additional crates which we recommend for writing tests:
-
-* [googletest](https://docs.rs/googletest): Comprehensive test assertion library in the tradition of GoogleTest for C++.
-* [proptest](https://docs.rs/proptest): Property-based testing for Rust.
-* [rstest](https://docs.rs/rstest): Support for fixtures and parameterised tests.
```
qwandor, line 13: Do we no longer recommend proptest or rstest?

src/testing/mocking.md
```diff
@@ -0,0 +1,44 @@
+---
+minutes: 10
+---
+
+# Mocking
+
+If you want to do mocking, we recommend [Mockall]. You need to refactor your
+code to use traits, which you can then quickly mock:
```
qwandor, line 8: We should first mention that we generally recommend against mocking, as per go/choose-test-double and go/mocking.

src/testing/Cargo.toml
```diff
@@ -4,7 +4,22 @@ version = "0.1.0"
 edition = "2021"
 publish = false
 
+[[example]]
+name = "googletest-example"
+crate-type = ["staticlib"]
```
qwandor, line 9: Why a staticlib? Are you planning to link it into C code?

src/testing/mocking.md
mgeisler: Yeah, but not in the public documentation... I added the "if" because of that, but let me make it stronger by adding a note to the speaker notes! Thanks for the idea :smile:

src/testing/useful-crates.md
mgeisler: I think we (meaning Android) never really recommended them. Personally, I've never used them and I've also not seem them being used in the libraries I've looked at.

src/testing/Cargo.toml
mgeisler: This was the type used as an example in [Examples](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#examples). It allows me to avoid an empty (and unwanted) `main` function.

PR #1529, opened 2023-11-30
Fix Fibonacci numbers example

jlunz:
The exercise has n <= 2 for the base case, but it is n < 2 in the solution. 
You currently get fib(n) = 10946 for n = 20, but it should be 6765.

### DIFF
--- a/src/types-and-values/exercise.rs
+++ b/src/types-and-values/exercise.rs
@@ -16,7 +16,7 @@
 // ANCHOR: fib
 fn fib(n: u32) -> u32 {
     // ANCHOR_END: fib
-    if n < 2 {
+    if n <= 2 {
         return 1;
     } else {
         return fib(n - 1) + fib(n - 2);


mgeisler:
Hi @jlunz, on my, that's a good catch!

I got curious now and looked up [Fibonacci sequence](https://en.wikipedia.org/wiki/Fibonacci_sequence), which leads to [Sequence A000045](https://oeis.org/A000045), which goes

```
0, 1, 1, 2, 3, 5, 8
```

The function we use don't do that since `fib(0) == 1`... Could you fix this and add a test for it? That would be awesome!

mgeisler:
> The example text specifies that the first and second numbers are both 1, so we're starting with 1,1 and 1-indexing. Which means that the `n <= 2` is correct. I don't think students will have great difficulty adjusting if they learned some slightly different formulation of the sequence.

Fair enough! I was just keen to follow the sequence given in the reference, but it's of course a very minor point!

@jlunz, if you feel like it, you're still welcome to add a test for this :smile:

jlunz:
@mgeisler Sure, I can give it a try. I assume you mean an inline test like e.g. here: https://github.com/google/comprehensive-rust/blob/7d10518b244fcb0907b5c1e415dab6f70a2e28be/src/control-flow-basics/exercise.rs#L28 ?

PR #1532, opened 2023-11-30
Add slide to show the Mockall crate in action

mgeisler:
Split off from #1528 so we can talk about things in isolation.

### DIFF
--- a/src/testing/mocking.md
+++ b/src/testing/mocking.md
@@ -1,0 +1,58 @@
+---
+minutes: 10
+---
+
+# Mocking
+
+If you want to do mocking, we recommend [Mockall]. You need to refactor your
+code to use traits, which you can then quickly mock:
+
+```rust,ignore
+{{#include mockall.rs}}
+```
+
+[Mockall]: https://docs.rs/mockall/
+
+<details>
+
+- Note that *mocking is controversial*: mocks allow you to completely isolate a
+  test from its dependencies. The immediate result is faster and more stable
+  test execution. On the other hand, the mocks can be configured wrongly and
+  return output different from what the real dependencies would do.
+
+  If at all possible, it is recommended that you use the real dependencies. As
+  an example, many databases allow you to configure an in-memory backend. This
+  means that you get the correct behavior in your tests, plus they are fast and
+  will automatically clean up after themselves.
+
+  Similarly, many web frameworks allow you to start an in-process server which
+  binds to a random port on `localhost`. Always prefer this over mocking away
+  the framework since it helps you test your code in the real environment.
+
+- Mockall is not part of the Rust Playground, so you need to run this example in
+  a local environment. Use `cargo add mockall` to quickly add Mockall to an
+  existing Cargo project.
+
+- Mockall has a lot more functionality. In particular, you can set up
+  expectations which depend on the arguments passed:
+
+    ```rust,ignore
+    let mut mock_cat = MockPet::new();
+    mock_cat
+        .expect_is_hungry()
+        .with(mockall::predicate::gt(Duration::from_secs(3 * 3600)))
+        .return_const(true);
+    mock_cat
+        .expect_is_hungry()
+        .return_const(true);
+    assert_eq!(mock_cat.is_hungry(Duration::from_secs(1 * 3600)), false);
+    assert_eq!(mock_cat.is_hungry(Duration::from_secs(5 * 3600)), true);
+    ```
+
+- You can use `.times(n)` to limit the number of times a mock method can be
+  called to `n` --- the mock will automatically panic when dropped if this isn't
+  satisfied.
+
+- Mockall is available for use in AOSP.
+
+</details>
--- a/src/testing/mocking.md
+++ b/src/testing/mocking.md
@@ -1,5 +1,5 @@
 ---
-minutes: 10
+minutes: 5
 ---
 
 # Mocking
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -469,6 +469,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "difflib"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6184e33543162437515c2e2b48714794e37845ec9851711914eec9d308f6ebe8"
+
+[[package]]
 name = "digest"
 version = "0.10.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -477,6 +483,12 @@ dependencies = [
  "block-buffer",
  "crypto-common",
 ]
+
+[[package]]
+name = "downcast"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1435fa1053d8b2fbbe9be7e97eca7f33d37b28409959813daefc1446a14247f1"
 
 [[package]]
 name = "dtoa"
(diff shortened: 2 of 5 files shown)


src/testing/mocking.md
```diff
@@ -0,0 +1,44 @@
...
+        .return_const(true);
+    assert_eq!(mock_cat.is_hungry(Duration::from_secs(1 * 3600)), false);
+    assert_eq!(mock_cat.is_hungry(Duration::from_secs(5 * 3600)), true);
+    ```
+
+- You can use `.times(n)` to limit the number of times a mock method can be
+  called to `n` --- the mock will automatically panic when dropped if this isn't
+  satisfied.
+
+- Mockall is available for use in AOSP.
```
mgeisler, line 42: This is not strictly true today, but it should be true by ~next week.

src/testing/mocking.md
```diff
@@ -0,0 +1,58 @@
+---
+minutes: 10
```
djmitche, line 2: I don't want to litigate down to the minute, IMHO this slide should be short: "If it's mocking you need, there's a great crate for that which is used in Android, called mockall. Here's an example, which creates a MockPet type. Find more at docs.rs when you need it"

mgeisler:
Hey @djmitche, sorry that I didn't motivate this better!

> I mean, `serde` is fantastic too, as is `ring`, `log`, `pretty_assertions`, .. where do we draw the line and assume that students can find the relevant crate when they need it?

The guide for me (and the Android team) has been that we've gotten requests from teams that ask us "How do I do mocking in my unit tests?" This slide is one part of the answer — I will also write some longer internal documentation about mocking to help those teams (search for "rust-mocking-recommendations" internally to find some background).

Basically, a picture is forming about some topics that we get more questions about than others... Mocking is one of them and right now we don't have a good answer for those teams.

I would also be happy to write about Serde, but that would be in a different part. @marshallpierce has talked about adding content about parsing and it would presumably fit there.

src/testing/mocking.md
mgeisler: Yeah, maybe 10 minutes is overestimating it here. I really have to schedule a class myself to understand how much time to spend on each slide.

Let change it to 5 minutes as a start?

```suggestion
minutes: 5
```

src/testing/mocking.md
```diff
@@ -0,0 +1,58 @@
+---
+minutes: 5
+---
+
+# Mocking
+
+If you want to do mocking, we recommend [Mockall]. You need to refactor your
```
djmitche, line 7:
```suggestion
If you want to do mocking, AOSP recommends [Mockall], although there are several other crates available. You need to refactor your
```
..or something to that effect. I think the existing text implies "mockall is how people do test mocks in Rust", and that's not accurate.

src/testing/mocking.md
```diff
@@ -0,0 +1,58 @@
...
+  return output different from what the real dependencies would do.
+
+  If at all possible, it is recommended that you use the real dependencies. As
+  an example, many databases allow you to configure an in-memory backend. This
+  means that you get the correct behavior in your tests, plus they are fast and
+  will automatically clean up after themselves.
+
+  Similarly, many web frameworks allow you to start an in-process server which
+  binds to a random port on `localhost`. Always prefer this over mocking away
+  the framework since it helps you test your code in the real environment.
```
djmitche, line 30: This seems out of context here -- if students are asking how to use mocks in Rust, they are probably more focused on the "in Rust" part and do not want to be told that they shouldn't be using mocks.

src/testing/mocking.md
```diff
@@ -0,0 +1,58 @@
...
+    mock_cat
+        .expect_is_hungry()
+        .return_const(true);
+    assert_eq!(mock_cat.is_hungry(Duration::from_secs(1 * 3600)), false);
+    assert_eq!(mock_cat.is_hungry(Duration::from_secs(5 * 3600)), true);
+    ```
+
+- You can use `.times(n)` to limit the number of times a mock method can be
+  called to `n` --- the mock will automatically panic when dropped if this isn't
+  satisfied.
```
djmitche, line 54:
```suggestion
- Mockall has a lot more functionality, and has extensive documentation at https://docs.rs/mockall.
```

djmitche (changes requested):
Sorry to be a bit of a pest here. I'm trying to be conscious of the fact that lots of people use this course and we should be careful to not make broad statements that are in fact only true in specific contexts (like AOSP or Chromium).

src/testing/mocking.md
mgeisler: I think it's important to mention just a few features (such as the one in the example). It's in the notes, so it should take up very little space in the course — but it will be useful for the participants.

src/testing/mocking.md
mgeisler: I did a survey and from what I can tell, Mockall is vastly more used than any other mocking library right now:

|                                               | Released | Updated | Dependants | Downloads/month |
|-----------------------------------------------|---------:|--------:|-----------:|----------------:|
| [Mockall](http://lib.rs/crates/mockall)       |     2019 |  _2023_ |        295 |              1M |
| [Mockers](http://lib.rs/crates/mockers)       |     2016 |  _2022_ |          3 |             560 |
| [Simulacrum](http://lib.rs/crates/simulacrum) |     2017 |    2018 |          1 |              6k |
| [Pseudo](http://lib.rs/crates/pseudo)         |     2017 |    2018 |          4 |             653 |
| [Double](http://lib.rs/crates/double)         |     2017 |    2019 |          1 |             208 |
| [Mockiato](http://lib.rs/crates/mockiato)     |     2019 |    2019 |          1 |              2k |
| [Mock-it](http://lib.rs/crates/mock-it)       |     2018 |  _2022_ |          2 |             406 |

Most libraries haven't been updated in a few years and most of them have very few reverse dependencies.

Of course Mockall isn't the end all of mocking, so a new library could be made which is even better. But right it's the one we will recommend to our engineers in Android.

At the end of the day, the course here is written for Android engineers. I've tried my best to take out the Android-specific parts and make as much as it generic. I think the advise here is also generic enough for this section.

However, let us add a note on the slide (and perhaps somewhere in the beginning) to say that the material here is an _opinionated_ guide to Rust. It tried to be fair and generic, but it is sponsored by Google, and it will recommend tools useful today to Android and Google engineers. Does that sound fair?

djmitche: Ah, thanks for that analysis of the libraries! A few search results suggested a much wider range of options. Given that information, this slide makes sense.

src/testing/mocking.md
djmitche: I still want to keep it short, though, just in the interest of overall course length. I think this lengthy speaker note encourages a digression into the relative merits of various testing strategies, which will probably generate questions and comments from students, and suddenly we've been on this slide for 15 minutes.

I agree that an article would be great, but maybe this could be moved to "More to Explore"?

At any rate, LGTM.

src/testing/mocking.md
mgeisler: Thanks! I will add a note explaining that the recommendations on the slide is current at of now and apply directly to Android. They might also apply to others, but as you point out, there _are_ other options out there!

src/testing/mocking.md
mgeisler: > I still want to keep it short, though, just in the interest of overall course length.

Fair point! I was not thinking that the instructor would go through all the pros and cons — I guess I was writing this bullet point more for the people who will read the slide by themselves.

The problem of multiple personas reading the slides is tricky... as we've discussed before :smile: It just occurred to me that we could create more sections in the notes (and first rename them to "Notes"):

```markdown
<details>

### Speaker

- Make sure to mention X and Y.
- Try removing `x`, note the compilation error.

### Participants

- The concept of Foo was originally introduced in C++ as Bar. See Wikipedia.
- If you want to explore more, then try ...

</details>
```

Basically an extension of the "More to Explore" headings we've added a few places.

PR #1533, opened 2023-11-30
Move Mockall and GoogleTest slides to Android section

mgeisler:
After #1528 and #1532, we now have actual slides which showcase the crates in action. So we can reclaim a few minutes by removing the slide which mentions Mockall and GoogleTest slide.

The slide mentioned [proptest](https://docs.rs/proptest) and [rstest](https://docs.rs/rstest) as well. While I'm sure the libraries are useful, we don't have them imported into AOSP and I've never personally used them. We should therefore not advertise them yet at this point since they won't be useful to Android engineers.

Of course we can mention things that are not in AOSP (or in Chromium), but I think we should do it in the speaker notes at most.


qwandor (approved):

djmitche (changes requested):
I think this is a good argument for not advocating specific crates, or at least not presenting crates about which there's some disagreement as standard.

Speaking for myself, I've never seen googletest used outside of google, but I've seen quite a bit of proptest and rstest. In a general "Rust course" this seems a strange priority, but in a Rust course for Android devs written by Googlers, it makes more sense. But even within Google this course is not specifically for Android devs.

I wonder if we should present this as
 * In addition to the language's testing support, there are lots of great testing crates out there, such as .... (so: keep this "Useful Crates" slide)
 * Android uses `googletest` and `mockall` so we'll spend a few minutes looking at those crates specifically (so: include that context in the slides for those crates)

randomPoison:
I'm in favor of getting rid of the "useful crates" slide, though I don't feel strongly whether we keep the slides on googletest and mockall. Personally I think it's helpful to highlight a couple of ecosystem crates that provide more advanced testing functionality since the built-in test infrastructure is so basic. googletest and mockall seem like valid options, I think the more important thing to emphasize is "there is limited built-in functionality, but there are ecosystem libraries that fill in the gaps".

Alternately, if we want to still talk about googletest and mockall since they're in AOSP, but we don't feel like these slides are a good fit for the Fundamentals course, we could move these to the Android course. It wouldn't hurt to have a small section on testing within AOSP, since AOSP tests are likely required to use googletest as the standard testing crate.

mgeisler:
> Speaking for myself, I've never seen googletest used outside of google, but I've seen quite a bit of proptest and rstest. In a general "Rust course" this seems a strange priority, but in a Rust course for Android devs written by Googlers, it makes more sense.

Yeah, you're right... I've never seen googletest used outside of Google either :smile: That can of course change with time, but right now it doesn't have a lot of usage: https://lib.rs/crates/googletest/rev. However, it's the one we recommend for Android, so that is why I added it here.

> Alternately, if we want to still talk about googletest and mockall since they're in AOSP, but we don't feel like these slides are a good fit for the Fundamentals course, we could move these to the Android course. It wouldn't hurt to have a small section on testing within AOSP, since AOSP tests are likely required to use googletest as the standard testing crate.

I think your suggestion here is the best way forward: testing is a must for AOSP engineers, so I feel we must mention these crates there.

Let me move the pages to the Android section and then I hope we can merge this.

mgeisler:
I've updated the PR:

- the "Useful Crates" slide is still gone since I don't want to recommend crates I don't use myself
- the Mockall and GoogleTest crates are now featured in the Android class

I hope this feels better for everybody?

djmitche (approved):
Nice!

src/android/testing.md
```diff
@@ -0,0 +1,37 @@
+# Testing in Android
+
+Buildings on [Testing](../testing.md), we will now look at how unit tests work
```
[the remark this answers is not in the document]
mgeisler: Ah, hilarious typo :rofl:

src/android/testing.md
```diff
@@ -0,0 +1,37 @@
+# Testing in Android
+
+Building on [Testing](../testing.md), we will now look at how unit tests work
+in AOSP. Use the `rust_test` module for your unit tests:
```
mgeisler, line 4: Fix formatting:

```suggestion
Building on [Testing](../testing.md), we will now look at how unit tests work in
AOSP. Use the `rust_test` module for your unit tests:
```

mdbook-course/src/bin/course-schedule.rs
```diff
@@ -45,7 +45,7 @@ fn timediff(actual: u64, target: u64, slop: u64) -> String {
             duration(actual),
             duration(actual - target),
         )
-    } else if actual < target - slop {
+    } else if actual + slop < target {
```
mgeisler, line 48: @djmitche, are you happy with this change? I think I might have done the timing information wrong in the moved slides to trigger an integer underflow here?

djmitche: Indeed, good point!

djmitche (approved):

PR #1538, opened 2023-11-30
Do not include timing info in sub-chapters of slides

### LINKED ISSUE #1522: Sub-slides should not have time included in the speaker notes
Currently, a sub-slide will have the same timing information substituted into the speaker notes as the parent slide. But that parent slide already sums the times in all of the sub-slides!

### LINKED ISSUE #1552: Need to consistently use title casing for section titles
@mgeisler points out in https://github.com/google/comprehensive-rust/pull/1539#discussion_r1411900480 that we should consistently use title casing for section titles, but it seems that Chromium sections are a mixed bag (e.g. in `SUMMARY.md` there is the "Build Rules"  section, but also "Adding third party crates" and "Unsafe code" sections).

Given above let me open an issue to track this, and to help up remember to fix this.  This may need a bit for the dust to settle first, since we are still shuffling and/or adding Chromium-related sections (e.g. in https://github.com/google/comprehensive-rust/pull/1551).

djmitche:
Timing information is for a slide as a whole, so we use a sentence like "This slide and its sub-slides should take about 15 minutes" in the first chapter in the slide. Subsequent chapters in the same slide do not need timing information.

Fixes #1522.

### DIFF
--- a/mdbook-course/src/course.rs
+++ b/mdbook-course/src/course.rs
@@ -447,6 +447,13 @@ impl Slide {
         Ok(())
     }
 
+    /// Determine whether the given chapter is a sub-chapter of this slide.
+    pub fn is_sub_chapter(&self, chapter: &Chapter) -> bool {
+        // The first souce_path in the slide is the "parent" chapter, so anything
+        // else is a sub-chapter.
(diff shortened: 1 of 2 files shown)


mgeisler (approved):
Sounds good!

PR #678, opened 2023-05-24
Update build_all.sh

freeaion:
The path, $ANDROID_BUILD_TOP and/or $ANDROID_PRODUCT_OUT, can contain spaces, which results in error. As long as shell var is not number, it is safe to enclose it with ""

PR #1542, opened 2023-12-01
Fix adb push command

ningcng:
```
$ adb push "$ANDROID_PRODUCT_OUT/system/bin/hello_rust /data/local/tmp"
adb: push requires <source> and <destination> arguments
```


mgeisler (commented):
Well spotted! This was "fixed" in #678, but I see now that the fix was too eager. Let me fix it better.

mgeisler (approved):

PR #895, opened 2023-07-02
Update Android.bp files

spoloxs:
Solves issue #371

Issue #371, opened 2023-02-09
Remove `prefer_rlib: true` from `Android.bp` files


mgeisler:
The line makes Soong statically compile all Rust dependencies. This in turn makes it easy to copy examples to the emulator. However, this mode is not the default and it can cause problems with doubly-defined symbols.

spoloxs:
I removed `prefer_rlib: true` from `Android.bp` files and raised the PR please check and tell if something is wrong or missing.

mgeisler:
Done via #895.

PR #1544, opened 2023-12-01
Link statically to avoid dynamic link error

ningcng:
```
$ adb shell /data/local/tmp/hello_rust_with_dep
CANNOT LINK EXECUTABLE "/data/local/tmp/hello_rust_with_dep": library "libtextwrap.dylib.so" not found: needed by main executable
```

### DIFF
--- a/src/android/aidl/birthday_service/Android.bp
+++ b/src/android/aidl/birthday_service/Android.bp
@@ -20,6 +20,7 @@ rust_binary {
         "libbinder_rs",
         "libbirthdayservice",
     ],
+    prefer_rlib: true, // To avoid dynamic link error.
 }
 // ANCHOR_END: birthday_server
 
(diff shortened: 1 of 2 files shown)


mgeisler:
This was removed in #895 as part of #371. @randomPoison, how did you test the change in #895?

@ningcng, could you add comment to document what this does? Otherwise it'll likely be removed again :smile:

mgeisler (approved):
Thanks!

mgeisler:
> ```
> $ adb shell /data/local/tmp/hello_rust_with_dep
> CANNOT LINK EXECUTABLE "/data/local/tmp/hello_rust_with_dep": library "libtextwrap.dylib.so" not found: needed by main executable
> ```

An alternative solution would be (I think) to use `adb sync` to copy over the necessary `libtextwrap.dylib.so` file.

@ningcng and @randomPoison, could one of you test if that does the right thing? If so, then I think it would be nicer to update the instructions to use `adb sync` to copy over everything needed.

PR #1550, opened 2023-12-01
Add adb root for service deployment

ningcng:
Otherwise
```
$ adb shell /data/local/tmp/birthday_server
thread 'main' panicked at birthday_service/src/server.rs:16:10:
Failed to register service: PERMISSION_DENIED
```

### DIFF
--- a/src/android/build_all.sh
+++ b/src/android/build_all.sh
@@ -56,7 +56,6 @@ source build/envsetup.sh
 lunch aosp_cf_x86_64_phone-userdebug
 #acloud reconnect --autoconnect adb
 
-adb root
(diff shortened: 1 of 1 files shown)


mgeisler (changes requested):
The `adb root` line now shows up on the slide, but if it's needed there, then it's needed on every slide. We should instead explain this once in the beginning of the Android class.

@randomPoison, can you check this for your January classes? @fw-immunant, you might also have a setup at hand for testing this since you taught the class in the past.

ningcng:
It's not necessary for non-service deployment.

mgeisler (approved):
Thanks for fixing this!

PR #1551, opened 2023-12-01
Chromium: new subsection about gtest tests

anforowicz:


anforowicz:
@adetaylor, can you PTAL?

It may be worth running through my examples + exercises once, to make sure that they really work (I think they do, but in a self-review I almost missed a discrepancy between the text and what I was testing in a Chromium repo - I used the old docs that I am fixing in https://chromium-review.googlesource.com/c/chromium/src/+/5079204/1/testing/rust_gtest_interop/README.md).

At any rate, even if things are not perfect - maybe we can merge what I have, and then follow-up with small fix PRs later?  This way we will have "something" about testing for the next week even if I am hit by a proverbial bus.

adetaylor (approved):

PR #1559, opened 2023-12-06
Use Self to create struct in method chapter

hurryabit:
If I'm not mistaken, it's idomatic to use `Self` in a struct constructor
rather than the name of the struct.


djmitche (approved):
Good catch!

mgeisler:
> If I'm not mistaken, it's idomatic to use `Self` in a struct constructor rather than the name of the struct.

Some people (like me), never want to use type aliases such as `Self`... I don't like having two names for the same thing :shrug: However, I think I'm in a minority here :smile: So thanks for the PR!

PR #1566, opened 2023-12-07
Simplify enums a little

mgeisler:
This takes out the huge example with lots of macro magic. I don't think we need it for an introductory course.

I also cleaned up the formatting a little and made sure to distinguish between types and values.

### DIFF
--- a/src/user-defined-types/enums.md
+++ b/src/user-defined-types/enums.md
@@ -22,7 +22,7 @@ enum PlayerMove {
 }
 
 fn main() {
-    let m = PlayerMove::Run(Direction::Left);
+    let m: PlayerMove = PlayerMove::Run(Direction::Left);
     println!("On this turn: {:?}", m);
 }
 ```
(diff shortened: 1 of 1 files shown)


src/user-defined-types/enums.md
```diff
@@ -31,10 +31,10 @@ fn main() {
 
 Key Points:
 
-* Enumerations allow you to collect a set of values under one type
-* Direction has two variants, `Left` and `Right`. These are referred to with the `Direction::..` namespace.
-* PlayerMove shows the three types of variants. Rust will also store a discriminant so that it can determine at runtime which variant is in a value.
-* This might be a good time to compare Structs and Enums:
+* Enumerations allow you to collect a set of values under one type.
+* `Direction` is a type with variants. There are two values of `Direction`: `Direction::Left` and `Direction::Right`.
+* `PlayerMove` is a type with three variants. In addition to the payloads, Rust will store a discriminant so that it known at runtime which variant is in a `PlayerMove` value.
```
djmitche, line 36:
```suggestion
* `PlayerMove` is a type with three variants. In addition to the payloads, Rust will store a discriminant so that it knows at runtime which variant is in a `PlayerMove` value.
```

djmitche (approved):

PR #1568, opened 2023-12-07
Expand "Docs" plus light copy-editing

mgeisler:
I don't like shortening words, so I prefer "Documentation" over "Docs". I find this less jargony and thus easier to read (and potentially also easier to translate).


djmitche (approved):
Needs a corresponding change in `SUMMARY.md`, but otherwise looks good.

mgeisler:
> Needs a corresponding change in `SUMMARY.md`, but otherwise looks good.

Thanks!

PR #1572, opened 2023-12-09
Render book as PDF in `publish.yml` workflow

### LINKED ISSUE #1543: Look into publishing PDFs with mdbook-typst
There is a new tool which can convert an mdbook to Typst format, and thus also to high-quality PDFs: https://github.com/LegNeato/mdbook-typst

max-heller:
Renders the book as a PDF and includes it in the published HTML bundle as `comprehensive-rust.pdf`.

Unfinished business:

- [ ] This doesn't add a link to the PDF from anywhere (see https://github.com/google/comprehensive-rust/issues/1543#issuecomment-1846238746).
- [x] This also may or may not do anything useful yet, thanks yaml scripting. I haven't tested rendering the translated books locally, so that may need extra work.
- [x] The book contains unicode characters that the default LaTeX font doesn't support (e.g. 🦀), so you may want to [swap out the fonts](https://pandoc.org/MANUAL.html#fonts).
    - [x] Use different fonts for translations (e.g., [Chinese](https://max-heller.com/comprehensive-rust/zh-CN/comprehensive-rust.pdf))
- [x] Slightly smaller margins
- [x] Color hyperlinks
- [ ] Do you know of a way to exclude/include content in the PDF? As an example, [1.2. Keyboard Shortcuts](https://google.github.io/comprehensive-rust/running-the-course/keyboard-shortcuts.html) should not be shown in the PDF. Similarly, the "In this segment" texts (as seen in [12. Pattern Matching](https://google.github.io/comprehensive-rust/pattern-matching.html)) should also go away.
- [x] ~It would also be great if we could apply the formatting in [67. Glossary](https://google.github.io/comprehensive-rust/glossary.html) — it's done via a stylesheet in the HTML, but that is of course lost in the PDF.~
    - Ignoring for now

Closes #1543

[Rendered](https://max-heller.com/comprehensive-rust/comprehensive-rust.pdf)


mgeisler:
Hi @max-heller, thanks for creating the PR!

> * [ ]  This also may or may not do anything useful yet, thanks yaml scripting. I haven't tested rendering the translated books locally, so that may need extra work.

Yeah, I hate that too :smile: Could I get you to enable running the workflows in your fork? That should allow us to see the effect of this PR in your fork (you probably need to merge the `main` branch into the `rendered-pdf` branch so that the publish workflow triggers).

> * The book contains unicode characters that the default LaTeX font doesn't support (e.g. 🦀), so you may want to [swap out the fonts](https://pandoc.org/MANUAL.html#fonts).

Good point! From looking through the PDF at

> [Rendered](https://github.com/google/comprehensive-rust/files/13622237/comprehensive-rust.pdf)

I don't immediately spot any bad effects of this — the little crab seems to have been ignored and disappeared. I'm thinking of removing the emoji from the course since it's tedious to type...

As for a font, it's a bit ironic that Computer Modern doesn't feel super modern to me any longer. I guess I should experiment a bit with finding a more crisp (sans-serif?) font... perhaps slightly smaller margins would be nice too as well as some color to the hyperlinks.

---

Do you know of a way to exclude/include content in the PDF? As an example, [1.2. Keyboard Shortcuts](https://google.github.io/comprehensive-rust/running-the-course/keyboard-shortcuts.html) should not be shown in the PDF. Similarly, the "In this segment" texts (as seen in [12. Pattern Matching](https://google.github.io/comprehensive-rust/pattern-matching.html)) should also go away.

It would also be great if we could apply the formatting in [67. Glossary](https://google.github.io/comprehensive-rust/glossary.html) — it's done via a stylesheet in the HTML, but that is of course lost in the PDF.

.github/workflows/publish.yml
```diff
@@ -40,8 +40,51 @@ jobs:
...
+      - name: Install mdbook-pandoc
+        run: curl --proto '=https' --tlsv1.2 -LsSf https://github.com/max-heller/mdbook-pandoc/releases/download/v0.2.0/mdbook-pandoc-installer.sh | sh
+      - name: Install pandoc
+        run: |
+          curl -LsSf https://github.com/jgm/pandoc/releases/download/3.1.9/pandoc-3.1.9-linux-amd64.tar.gz | tar zxf -
+          echo "$PWD/pandoc-3.1.9/bin" >> $GITHUB_PATH
+      - name: Install mdbook-pandoc dependencies
+        run: sudo apt-get install -y librsvg2-bin
+      - name: Install TeX Live
+        uses: teatimeguest/setup-texlive-action@v3
```
mgeisler, line 52: Would it be possible to simply this by simply installing Pandc and TeX Live using system packages?

Debian (and thus Ubuntu) has normally been good at including TeX packages — at least in my experience ~10 years ago where I always used the ones from the system instead of dealing with hand-installed stuff.

.github/workflows/publish.yml
```diff
@@ -40,8 +40,51 @@ jobs:
       - name: Install mdbook
         uses: ./.github/workflows/install-mdbook
 
+      - name: Install mdbook-pandoc
+        run: curl --proto '=https' --tlsv1.2 -LsSf https://github.com/max-heller/mdbook-pandoc/releases/download/v0.2.0/mdbook-pandoc-installer.sh | sh
```
mgeisler, line 44: This could be `cargo install mdbook-pandoc` instead: we cache the Rust binaries installed, so it'll be pretty quick after the first run.

(I try to avoid having long as "mysterious" command lines in the actions — I don't know what `-LsSf` means and I prefer not having to know about it).

.github/workflows/publish.yml
mgeisler: On my Debian Unstable machine, this worked fine to get a working PDF:

```
sudo apt install pandoc texlive texlive-luatex librsvg2-bin
```

mgeisler:
> I don't immediately spot any bad effects of this — the little crab seems to have been ignored and disappeared. I'm thinking of removing the emoji from the course since it's tedious to type...

Now that I've tried this locally, I see that there are a bunch of other places where emojis cause problems:

```
[WARNING] Missing character: There is no 🌍 (U+1F30D) (U+1F30D) in font [lmmono10-regular]:!
[WARNING] Missing character: There is no α (U+03B1) (U+03B1) in font [lmmono10-regular]:!
[WARNING] Missing character: There is no 🪐 (U+1FA90) (U+1FA90) in font [lmmono10-regular]:!
```

plus the line drawing characters in the notes of [26.2. Filesystem Hierarchy](https://google.github.io/comprehensive-rust/modules/filesystem.html).

.github/workflows/publish.yml
max-heller: This works well for TeX, but the repository for Ubuntu 22.04 (`ubuntu-latest` in Github Actions) includes a pretty old version of Pandoc that's missing extensions needed for optimal rendering

max-heller:
> Yeah, I hate that too 😄 Could I get you to enable running the workflows in your fork? That should allow us to see the effect of this PR in your fork (you probably need to merge the `main` branch into the `rendered-pdf` branch so that the publish workflow triggers).

[Running](https://github.com/max-heller/comprehensive-rust/actions/runs/7381017995/job/20079081292)

> Do you know of a way to exclude/include content in the PDF? As an example, [1.2. Keyboard Shortcuts](https://google.github.io/comprehensive-rust/running-the-course/keyboard-shortcuts.html) should not be shown in the PDF. Similarly, the "In this segment" texts (as seen in [12. Pattern Matching](https://google.github.io/comprehensive-rust/pattern-matching.html)) should also go away.

There's no builtin way to accomplish this that I can tell. Some options:
- [Run a Pandoc filter (AST transformer) to exclude marked sections](https://gist.github.com/tarleb/a0f41adfa7b0e5a9be441e945f843299) or whole files. This would be the most flexible and doesn't require changes to `mdbook-pandoc`
- Add functionality to `mdbook-pandoc` to exclude marked sections as in the previous option. This might be useful functionality to have built in, but adds complexity
- Add a configuration option to `mdbook-pandoc` to exclude certain files. Less fine-grained but doesn't require annotating the source

> It would also be great if we could apply the formatting in [67. Glossary](https://google.github.io/comprehensive-rust/glossary.html) — it's done via a stylesheet in the HTML, but that is of course lost in the PDF.

Not sure how to accomplish this with LaTeX unfortunately

.github/workflows/publish.yml
mgeisler: > This works well for TeX, but the repository for Ubuntu 22.04 (`ubuntu-latest` in Github Actions) includes a pretty old version of Pandoc that's missing extensions needed for optimal rendering

Okay, fair enough! I'm glad that the installation instructions are much simpler now!

.github/workflows/publish.yml
```diff
@@ -40,8 +40,19 @@ jobs:
       - name: Install mdbook
         uses: ./.github/workflows/install-mdbook
 
+      - name: Install mdbook-pandoc
+        run: cargo install mdbook-pandoc
```
mgeisler, line 44: Small thing: I started adding ` --locked --version x.y.z` to the `cargo install` lines after we ran into incompatibilities between a new version of `mdbook` and one of the plugins we use. I would suggest doing the same here for consistency.

.github/workflows/publish.yml
```diff
@@ -40,8 +40,19 @@ jobs:
...
+      - name: Install pandoc
+        run: |
+          curl -LsSf https://github.com/jgm/pandoc/releases/download/3.1.9/pandoc-3.1.9-linux-amd64.tar.gz | tar zxf -
+          echo "$PWD/pandoc-3.1.9/bin" >> $GITHUB_PATH
+      - name: Install mdbook-pandoc dependencies
+        run: sudo apt-get install -y texlive texlive-luatex librsvg2-bin
```
mgeisler, line 50: I think we could put all these steps into a single `Install mdbook-pandoc` step. Just to group things a bit.

mgeisler (changes requested):

max-heller:
> It would also be great if we could apply the formatting in [67. Glossary](https://google.github.io/comprehensive-rust/glossary.html) — it's done via a stylesheet in the HTML, but that is of course lost in the PDF.

One option might be to format the glossary with [`<dl>`, `<dt>`, and `<dd>` HTML elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl), which Pandoc converts to something reasonable:
```html
<dl>
  <dt>Coffee</dt>
  <dd>Black hot drink</dd>
  <dt>Milk</dt>
  <dd>White cold drink</dd>
</dl>
```
passed to `pandoc -f html -t pdf` produces
Screenshot 2024-01-02 at 7 18 10 PM
However, Pandoc mostly ignores HTML embedded in Markdown, so `mdbook-pandoc` would need to find a way to force Pandoc to process the embedded HTML like it does with `-f html`

mgeisler:
> [Running](https://github.com/max-heller/comprehensive-rust/actions/runs/7381017995/job/20079081292) ([rendered English book](https://max-heller.github.io/comprehensive-rust/comprehensive-rust.pdf))

Yay, great to see that it works!

> The translated books are built, but some are very broken because of [missing fonts](https://max-heller.com/comprehensive-rust/zh-CN/comprehensive-rust.pdf).

Uh, yeah that looks pretty rough :smile: 

I did a bunch of testing here and found that I can get nice output for the Chinese translations, Korean, Japanese and the Russian translation by

- Installing the Noto family of fonts (with `sudo apt install fonts-noto`)
- Setting `CJK*font`, but _only_ for the Chinese, Japanese, and Korean translations.

So my `book.toml` file ended up looking like this for Simplified Chinese (SC):

```toml
[output.pandoc.profile.pdf.variables]
mainfont = "Noto Serif"
sansfont = "Noto Sans"
monofont = "Noto Sans Mono"

CJKmainfont = "Noto Serif CJK SC"
CJKsansfont = "Noto Sans CJK SC"
CJKmonofont = "Noto Sans Mono CJK SC"
```

There are other fonts for the other [CJK languages](https://fonts.google.com/noto/use#faq:~:text=Which%20Noto%20fonts%20should%20I%20use%20for%20Chinese%2C%20Japanese%20or%20Korean%3F).

The `CJKmainfont` variable trigger the inclusion of `luatexja-fontspec`, which I installed with `sudo apt install texlive-lang-cjk`. This package is not just used for Japanese, but also for the Chinese translations. However, including this package gives weird output for other languages, so it's important that it's only set for the CJK languages. I wanted to set it via [an environment variable](https://rust-lang.github.io/mdBook/format/configuration/environment-variables.html), but it seems that they cannot express a mixed-case key like `CJKmainfont`.

I've played around quite a bit here, so I guess the packages to install can be trimmed down a lot, but this should be a good starting point.

> > It would also be great if we could apply the formatting in [67. Glossary](https://google.github.io/comprehensive-rust/glossary.html) — it's done via a stylesheet in the HTML, but that is of course lost in the PDF.
> 
> One option might be to format the glossary with [`<dl>`, `<dt>`, and `<dd>` HTML elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl), which Pandoc turns into something reasonable when converting from HTML:

Right, if we can include HTML, then things should be easier — however, I very much want to avoid HTML since it is hard to edit in the English version and it looks weird and complicated for the translators.

I think we can ignore the glossary formatting for now.

book.toml
```diff
@@ -29,6 +29,18 @@ verbose = false # Report timing information.
...
+
+[output.pandoc.profile.pdf]
+output = "comprehensive-rust.pdf"
+pdf-engine = "lualatex"
+
+[output.pandoc.profile.pdf.variables]
+geometry = ["margin=1.25in"]
+linkcolor = "blue"
+urlcolor = "red"
+
```
mgeisler, line 43: Just so we have it here: I would like to try using the Nota fonts by default since they seem to do what they say:

```toml
[output.pandoc.profile.pdf.variables]
mainfont = "Noto Serif"
sansfont = "Noto Sans"
monofont = "Noto Sans Mono"

CJKmainfont = "Noto Serif CJK SC"
CJKsansfont = "Noto Sans CJK SC"
CJKmonofont = "Noto Sans Mono CJK SC"
```

You would need to more logic in `publish.yml` to either set or clear the `CJK*` variables for non-CJK languages.

max-heller: Worked around the inability to set CJK* with env vars by passing Pandoc a language specific file with extra options (`pandoc/zh-CN.yaml`, etc.). [Rendered](https://max-heller.com/comprehensive-rust/zh-CN/comprehensive-rust.pdf)

mgeisler: Wow, very nicely done!

mgeisler (approved):
Thanks a lot @max-heller for the many revisions! I suggest we merge this now so we can see how it works day-to-day.

mgeisler:
Just a small note on this: I consider the PDF generation "best effort" in the sense that it's super cool to have, but not critical for the classroom teaching we do most with the course. I don't plan on working a lot on #1708 and #1709 myself, but perhaps someone from the community will find the time for this.

PR #1576, opened 2023-12-11
Comment PRs with updated schedule information

djmitche:
This adds a GH action to add a comment to every PR giving the updated course schedule with the PR merged.

To accomplish this, I broke `mdbook-course` into a library and two binaries, allowing the mdbook content to be loaded dynamically outside of an `mdbook build` invocation.

I think this is a net benefit, but possible improvements include:
 * diffing the "before" and "after" schedules and only making the comment when those are not the same (or replacing the comment with "no schedule changes")
 * including per-segment timing behind `<details>` (with a few minutes effort I couldn't get this to play nicely with the markdown lists)

### DIFF
--- a/.github/workflows/course-schedule.yml
+++ b/.github/workflows/course-schedule.yml
@@ -2,7 +2,7 @@ name: "Course Schedule Updates"
 on:
   pull_request:
     paths:
-    - 'src/**.md'
+      - "src/**.md"
 
 jobs:
   course-schedule:
--- a/.github/workflows/course-schedule.yml
+++ b/.github/workflows/course-schedule.yml
@@ -1,6 +1,8 @@
 name: "Course Schedule Updates"
 on:
-  - pull_request
+  pull_request:
+    paths:
+    - 'src/**.md'
 
 jobs:
   course-schedule:
--- a/.github/workflows/course-schedule.yml
+++ b/.github/workflows/course-schedule.yml
@@ -1,0 +1,25 @@
+name: "Course Schedule Updates"
+on:
+  - pull_request
+
+jobs:
+  course-schedule:
+    runs-on: ubuntu-latest
+    name: Make Course Schedule Comment
+    permissions:
+      pull-requests: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup Rust cache
+        uses: ./.github/workflows/setup-rust-cache
+
+      - name: Generate Schedule
+        run: cargo run -p mdbook-course --bin course-schedule > course-schedule.md
+
+      - name: Comment PR
+        uses: thollander/actions-comment-pull-request@v2
+        with:
+          filePath: course-schedule.md
+          comment_tag: course-schedule
--- a/mdbook-course/Cargo.toml
+++ b/mdbook-course/Cargo.toml
@@ -8,18 +8,6 @@ publish = false
 repository = "https://github.com/google/comprehensive-rust"
 description = "An mdbook preprocessor for comprehensive-rust."
 
-[[bin]]
-name = "course-schedule"
-path = "src/bin/course-schedule.rs"
-
-[[bin]]
-name = "mdbook-course"
-path = "src/bin/mdbook-course.rs"
-
-[lib]
-name = "mdbook_course"
-path = "src/lib.rs"
-
 [dependencies]
 anyhow = "1.0.79"
 clap = "4.4.12"
--- a/mdbook-course/Cargo.toml
+++ b/mdbook-course/Cargo.toml
@@ -8,6 +8,18 @@ publish = false
 repository = "https://github.com/google/comprehensive-rust"
 description = "An mdbook preprocessor for comprehensive-rust."
 
+[[bin]]
+name = "course-schedule"
+path = "src/bin/course-schedule.rs"
+
+[[bin]]
+name = "mdbook-course"
+path = "src/bin/mdbook-course.rs"
+
+[lib]
+name = "mdbook_course"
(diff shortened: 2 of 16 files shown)


.github/workflows/course-schedule.yml
```diff
@@ -0,0 +1,25 @@
+name: "Course Schedule Updates"
+on:
+  - pull_request_target
```
djmitche, line 3: Note that the schedule in this PR won't update, since this directive indicates that the action should run based on configuration in the target branch -- and it hasn't been merged yet.

djmitche: We should probably think about the implications of this!

mgeisler: We use

```yaml
on:
  pull_request:
```

in the `build.yml`. The difference is that `pull_request` run the action in the context of the (future) merge commit, whereas `pull_request_target` runs at the [base of the branch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target).

Are you running it at the base to enable the read/write behavior of the `GITHUB_TOKEN`?

djmitche:
Hm, it appears that jobs from pull_request and pull_request_target cannot have inter-dependencies because they are in different workflows. So I can't upload an artifact in a `pull_request` workflow and download it in `pull_request_target`.

I think [reusing workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows) is more of a subroutine call, and can't cross security boundaries, which is why we need two workflows here.

Artifacts are handled as ZIP files (e.g., https://github.com/google/comprehensive-rust/actions/runs/7215251310) so we can't just link to an artifact and hope it springs into existence before someone clicks the link.

Are there any GitHub actions gurus who might be able to give me a hint here?

.github/workflows/course-schedule.yml
mgeisler: I don't think it's necessary to split it. Pull requests from new contributors have to be approved before they run — precisely to protect against injection of malicious build steps.

So it should be fine to add the necessary `permissions:` key to a regular build: we check that the job doesn't change in strange ways before we approve PR builds for new contributors. 

That's the theory, as far as I understand it :smile: The real problem is if this understanding is wrong and/or if it's not shared with others who have write access to the repository.

djmitche:
@mgeisler do could you have another look here?

mdbook-course/Cargo.toml
```diff
@@ -8,6 +8,18 @@ publish = false
...
+path = "src/bin/course-schedule.rs"
+
+[[bin]]
+name = "mdbook-course"
+path = "src/bin/mdbook-course.rs"
+
+[lib]
+name = "mdbook_course"
+path = "src/lib.rs"
+
```
mgeisler, line 22: This should not be necessary: what you write here is what Cargo deduces by default (as far as I know). I tested it locally and it seems to work fine without:

```suggestion
```

.github/workflows/course-schedule.yml
```diff
@@ -0,0 +1,25 @@
+name: "Course Schedule Updates"
+on:
+  - pull_request
+
+jobs:
+  course-schedule:
+    runs-on: ubuntu-latest
+    name: Make Course Schedule Comment
+    permissions:
+      pull-requests: write
```
mgeisler, line 10: From [Permissions for the GITHUB_TOKEN](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token), I think this will be ineffective for PRs that come from forks.

I guess we'll figure it out soon after merging it :smile:

mgeisler (approved):
Thanks, I'm looking forward to seeing this in action.

.github/workflows/course-schedule.yml
```diff
@@ -0,0 +1,25 @@
+name: "Course Schedule Updates"
+on:
+  - pull_request
```
mgeisler, line 3: Perhaps this will be nice to avoid triggering it on other files:

```suggestion
on:
  pull_request:
    paths:
    - 'src/**.md'
```

Down the road, I could also imagine changing the logic to not add a comment unless the time changed. So if someone fixes a typo, they probably don't need the timing information.

.github/workflows/course-schedule.yml
djmitche: Maybe I misunderstood earlier suggestions - does this not run when we click "Run Actions" for an untrusted fork?

djmitche:
Hm, do I need to hand-apply the diff from that "Test / format" failure? `dprint fmt` doesn't fix it.

.github/workflows/course-schedule.yml
mgeisler: It does run, but it fails because the token is downgraded. That's how I interpret the error I just saw in #1693:

image

mgeisler:
> Hm, do I need to hand-apply the diff from that "Test / format" failure? `dprint fmt` doesn't fix it.

It should, `dprint fmt` should update the files in your working copy.

.github/workflows/course-schedule.yml
mgeisler: This makes me think that we should remove this again, sorry!

mgeisler:
> > Hm, do I need to hand-apply the diff from that "Test / format" failure? `dprint fmt` doesn't fix it.
> 
> It should, `dprint fmt` should update the files in your working copy.

Will be fixed with #1694.

djmitche:
> Per discussion with @mgeisler I've updated this to just do the simple thing, and rely on reviewers verifying PRs and clicking the run-actions button to check for malicious behavior.

So it seems this was not a working solution :(

Issue #1565, opened 2023-12-07
Course feedback from students


marshallpierce:
Feedback from day 1: 

- [x] exercise 10.6 (elevator) wasn't very clear as to what the desired end state was -- not confusion over enum syntax per se, but more like what the success criteria was. 
- [x] The exercise requires `#[derive(Debug)`, so maybe we should provide an empty `Direction` with that attribute already there, and have them fill in that type and add more? _fixed in #1574_

marshallpierce:
# Day 2 and 3 feedback

## 12.3 Expression evaluation

- Uses Box before we introduce it, so `*` needed for `eval(*left)` is nonobvious
- The `match`es in `eval` is a bit awkward, and doesn't make use of the `let` control flow in the section. It uses destructuring, but not in a particularly enlightening way
- unused code warnings led to questions, so this sample should likely also have the allow(unused) snippet applied
- Discussion about error handling and how it's different from idiomatic error handling led to some comments that they'd rather just use real Result, even though it's not been covered yet

## 13.5 GUI

- More time spent on formatting text than on traits

## 19.8 

- `as_dependency` semantics weren't clear about whether it was for the package itself or one of its dependencies

## 20.1 Box

- `Cons` is confusing to the non-Lisp folks

## 20.3 Binary tree

- The BinaryTreeNode / BinaryTree terminology was confusing due to a node itself having left and right trees. Perhaps a type alias for the wrapped Option<...> type would address this.
- Same concerns about unused code warnings as above
- C++ folks found it odd that has() didn't take a reference

## 22.4 Health statistics

- Would have been helpful to have comments on struct fields
- u32 to i32 conversions are bad (note that the solution doesn't follow my suggestion to avoid `as`). Could use u16 to allow .into() conversion to i32, and still cover all conceivable blood pressures.

## 23.3 Lifetime annotations

- First sentence seems backwards: values/references must outlive the lifetimes that refer to them, not the other way around
- Sample not editable, probably accidentally
- Could use more clarity on why or when people would want to use lifetimes -- when is it worth going to the trouble of not Cloning, etc. 

## 23.6 Protobuf

- Relies on error handling, ?, etc, but nobody was put off by that, lending more weight to earlier comments to use real Result instead of similar, simpler types
- More about error handling that lifetimes or slices  -- one of the strongest consensus for any of the feedback here
- Some trouble figuring out how to get started on the problem
[...]

PR #1582, opened 2023-12-12
Update expression-evaluation exercise: more patterns, more enums

djmitche:
This modifies the exercise to use the standard `Result` type, based on feedback that students could understand it sufficiently at this point in the course.

Addresses #1565.

### DIFF
--- a/src/pattern-matching/exercise.rs
+++ b/src/pattern-matching/exercise.rs
@@ -18,15 +18,26 @@
 /// An operation to perform on two subexpressions.
 #[derive(Debug)]
 enum Operation {
+    /// Add two Int values, or concatenate two String values.
     Add,
+    /// Subtract two Int values.
     Sub,
-    Mul,
-    Div,
 }
 // ANCHOR_END: Operation
 
+// ANCHOR: Value
+/// A value, the result of evaluating expression.
+#[derive(Debug, PartialEq, Eq)]
+enum Value {
+    /// An integer value.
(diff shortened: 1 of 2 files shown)


src/pattern-matching/exercise.rs
```diff
@@ -18,15 +18,26 @@
...
+    /// An integer value.
+    Int(i64),
+    /// A string value.
+    String(String),
+}
+// ANCHOR_END: Value
+
 // ANCHOR: Expression
-/// An expression, in tree form.
+/// A numerical expression, in tree form.
```
marshallpierce, line 40: It's no longer just numerical

marshallpierce (approved):

hurryabit:
I think using `Result` from `std` is a good idea. I know you're trying to make the expression language more complex to have more pattern matching. Without that, the solution still would have two cases of pattern matching:

1. Figure out if an expression is a binary operation or just a number literal (or atom).
2. Figure out the operation used.

IMO, that's enough. Making it more complicated risks confusing the students without gaining any more insights.

mgeisler:
> IMO, that's enough. Making it more complicated risks confusing the students without gaining any more insights.

For context, Martin is currently teaching the 4-day class here in Zurich, so this seems like valuable feedback to me.

mgeisler:
> It also uses the standard `Result` type, based on feedback that students could understand it sufficiently at this point in the course.

Very nice!

PR #1585, opened 2023-12-13
Rename test module into tests in smart pointer exercise

hurryabit:
If I'm not mistaken, the naming convention for modules containing only tests is to call them `tests` (rather than `test`).


djmitche (approved):
https://doc.rust-lang.org/book/ch11-03-test-organization.html confirms your assertion.

PR #1587, opened 2023-12-13
Use cmp + match in binary tree example

hurryabit:
Currently, the implementation uses if-then-else chains and `<` and `>`.
This is not the most idiomatic Rust. Instead, we can use `cmp` and
`match` to make the code easier to read.


djmitche (approved):
Thanks!

PR #1589, opened 2023-12-13
Use a type alias in binary tree exercise

djmitche:
As suggested by @marshallpierce and @hurryabit

### DIFF
--- a/src/smart-pointers/exercise.rs
+++ b/src/smart-pointers/exercise.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 // ANCHOR: solution
-use std::cmp::Ordering::*;
+use std::cmp::Ordering;
 
 // ANCHOR: types
 /// A node in the binary tree.
@@ -25,7 +25,8 @@ struct Node<T: Ord + Copy> {
 }
 
 /// A possibly-empty subtree.
-type Subtree<T> = Option<Box<Node<T>>>;
+#[derive(Debug)]
+struct Subtree<T: Ord + Copy>(Option<Box<Node<T>>>);
 
 /// A container storing a set of values, using a binary tree.
 ///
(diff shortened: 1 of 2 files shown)


src/smart-pointers/exercise.rs
```diff
@@ -13,66 +13,81 @@
...
+}
 // ANCHOR_END: types
 
 impl<T: Ord + Copy> BinaryTree<T> {
     fn new() -> Self {
-        Self(None)
+        Self { root: None }
     }
 
     fn insert(&mut self, value: T) {
```
djmitche, line 44: For each of `insert`, `has`, and `len`, the logic here in `BinaryTree` is reproduced below in `Node`. This definitely feels like the new organization of the types means that the implementation always starts at the "wrong end".

hurryabit:
This feels way more complex to me than what we had before. I'm not sure, this is a good idea.

djmitche:
I've simplified this a little now, with three types:
 * BinaryTree -- public type with pass-through methods
 * Subtree -- type with meat of the method implementations
 * Node -- container

What made the previous implementation complex was that the methods really belong on the `Option<Box<..>>`, but it put the methods on the boxed type.  However, the public type needs to be the `Option<Box<..>>` type. So this implementation puts that one at the top level, at the cost of adding some pass-through methods.

I'm still not sure this is better, though -- what do you think?

marshallpierce:
There's a bit of machinery here to unpack for the reader, but despite that I think I still am in favor of this direction: it's _explicit_, the names are more representative of their roles, and it provides the opportunity to have different public API than what a subtree might expose, which I think is ultimately what you'd want in a "real" thing anyway. Not that that's a major motivator for these exercises, but it's not without value showing how that would work. IME students can figure out logic easily enough, but if the delegation proves to be confusing, we could just provide the BinaryTree impl.

hurryabit:
I like the new version a lot, particularly compared to the draft I criticised in my previous comment. I don't think the delegation is a problem. These functions all follow a similar pattern and are easy to parse.

djmitche:
Cool, let's keep it then!

djmitche:
Apologies, this still needs an "official" review before I can merge it..

marshallpierce (approved):

PR #1590, opened 2023-12-13
Fix typo in lifetime annotations section

hurryabit:
The code example was meant to be `editable` but was marked as
`eitable`.


djmitche (approved):
Thanks!

PR #1591, opened 2023-12-13
Remove unnecessary syntax in protobuf example

hurryabit:
`Ok(x?)` has the same outcome as `x` and save an unpacking/repacking cycle.

`x as usize` has no effect when `x` already has type `usize`.

In `x < 4usize` the `usize` is unnecessary when `x` already has type `usize`.


src/slices-and-lifetimes/exercise.rs
```diff
@@ -62,6 +62,10 @@ struct Field<'a> {
     value: FieldValue<'a>,
 }
 
+trait ProtoMessage<'a>: Default + 'a {
```
djmitche, line 65: This is a pretty interesting form of lifetime annotation -- I think this will generate nice discussion!

djmitche (commented):
This is later in the course, so I think bringing a couple topics together is OK. The updated exercise doesn't seem more difficult than previously, but it does seem like it will take more time.  We might address that by including the first two or three lines of `parse_field` in the given code. Maybe
```rust
fn parse_field(data: &[u8]) -> Result<(Field, &[u8]), Error> {
    let (tag, remainder) = parse_varint(data)?;
    let (field_num, wire_type) = unpack_tag(tag)?;
    let (fieldvalue, remainder) = match wire_type {
        _ => todo!()
    }
    todo!()
}
```
this would save students understanding the header format of a proto field.

Does doing so omit too much of the slices-and-lifetimes goodness of the exercise?

src/slices-and-lifetimes/exercise.rs
mgeisler: Yeah! Am I understanding it correctly that with

```rust
trait ProtoMessage<'_>: Default
```

you would end up with `'static` in the place of `'_`? I believe I've run into this situation before...

djmitche: That's my understanding. Without the lifetime annotation, you could have:
```rust
fn f<M: ProtoMessage>(m: M) {
  ALL_MESSAGES_EVER_SEEN.push(m)
}
```
and that has to be check-able without reference to the concrete type of M.

djmitche (approved):

hurryabit:
> This is later in the course, so I think bringing a couple topics together is OK. The updated exercise doesn't seem more difficult than previously, but it does seem like it will take more time. We might address that by including the first two or three lines of `parse_field` in the given code. Maybe
> 
> ```rust
> fn parse_field(data: &[u8]) -> Result<(Field, &[u8]), Error> {
>     let (tag, remainder) = parse_varint(data)?;
>     let (field_num, wire_type) = unpack_tag(tag)?;
>     let (fieldvalue, remainder) = match wire_type {
>         _ => todo!()
>     }
>     todo!()
> }
> ```
> 
> this would save students understanding the header format of a proto field.
> 
> Does doing so omit too much of the slices-and-lifetimes goodness of the exercise?

I think that's an excellent idea. I'll prepare a PR for this.

PR #1592, opened 2023-12-14
Remove unnecessary syntax in protobuf example

hurryabit:
`Ok(x?)` has the same outcome as `x` and save an unpacking/repacking cycle.

`x as usize` has no effect when `x` already has type `usize`.

In `x < 4usize` the `usize` is unnecessary when `x` already has type `usize`.


mgeisler (approved):
Thanks, this looks nice!

PR #1594, opened 2023-12-16
Simplify some code in the parsing exercise

hurryabit:
`let Some(x) = E else { return None; }` is equivalent to
`let Some(x) = E?`. That latter is shorter and more idiomatic, so let's
use that.

A pattern of the form `c @ P1 | c @ P1` has the disadvantage that the
name `c` is repeated. Let's replace it with `c @ (P1 | P2)`.

An `x: Option<Result<T, E>>` can be handled more succinctly with
`x.ok_or(...)??`.

### DIFF
--- a/src/error-handling/exercise.rs
+++ b/src/error-handling/exercise.rs
@@ -61,9 +61,7 @@ impl<'a> Iterator for Tokenizer<'a> {
     type Item = Result<Token, TokenizerError>;
 
     fn next(&mut self) -> Option<Result<Token, TokenizerError>> {
-        let Some(c) = self.0.next() else {
-            return None;
-        };
+        let c = self.0.next()?;
         match c {
             '0'..='9' => {
                 let mut num = String::from(c);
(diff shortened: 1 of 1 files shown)


mgeisler:
This is for [Exercise: Rewriting with Result](https://google.github.io/comprehensive-rust/error-handling/exercise.html), right? I believe the "goal" of this code is to be un-idiomatic a first so that people can rewrite it like you do.

We should probably add a comment in the code itself to warn others from trying to clean it up :smile:

hurryabit:
Do we want the code to be unidiomatic regarding things that are not related to `panic` vs `Result`? I find this a little problematic when it comes to the repeated variable in the patterns. It's hard for people to know that this can be done better.

Even if we want to give people an opportunity to write things in a better style, we should at least have the solution in such a better style, shouldn't we?

djmitche:
You're right -- these particular parts were just errors on my part, not intentional shortcomings to be corrected by students.

djmitche (approved):

mgeisler:
> Do we want the code to be unidiomatic regarding things that are not related to `panic` vs `Result`?

I see now! I had no understood this point, thanks for explaining!

PR #1599, opened 2023-12-19
fix: change "widley" to "widely"

gregatgoogle:


lb034582341 (approved):
Thanks!

djmitche (approved):

djmitche:
I'll update this to only change the original (in `src/`)

PR #1600, opened 2023-12-20
Expand "thiserror and anyhow" slide

### LINKED ISSUE #1597: Need explanation how `context` and `with_context` operate on non-anyhow types
Something unusual is happening in the example on the `thiserror` and `anyhow` page that needs explanation.
https://docs.rs/anyhow/latest/anyhow/trait.Context.html

```
    fs::File::open(path)
        .with_context(|| format!("Failed to open {path}"))?
```

`fs::File::open(path)` returns a value of type `core::result::Result<std::fs::File, std::io::error::Error>`. Then `.with_context` is called on that value and produces a value of type `core::result::Result<std::fs::File, anyhow::Error>`

`with_context` takes `self` as the first argument i.e. it should be the original result:
https://docs.rs/anyhow/latest/anyhow/trait.Context.html

But that result is not aware of `anyhow`! However, there is something interesting at the bottom of that page. It mentions "Implementations on Foreign Types" which appears to be what we have here. I've been using Rust for months, I've read several books about Rust, and yet it appears completely new to me.

Moreover, it appears for go against the orphan rule mentioned here: https://doc.rust-lang.org/book/ch10-02-traits.html

I even copied and modified the example to check that the return value plays no role:

```
use anyhow::Context;
use std::fs;

fn print_type_of<T>(tag: &str, _: &T) {
    println!("{tag}: {}", core::any::type_name::<T>());
}

fn main() {
    let result = fs::File::open("config.dat");
    print_type_of("result", &result);
    let result_ctx = result.context("Failed to open");
    print_type_of("result_ctx", &result_ctx);
}
```

It prints:

```
result: core::result::Result<std::fs::File, std::io::error::Error>
result_ctx: core::result::Result<std::fs::File, anyhow::Error>
```

`result_ctx` is not used except to print its type, so there should be no type inference.

I would really appreciate an additional new speaker note on that page mentioning "Implementations on Foreign Types".

proski:
Explain what traits `thiserror::Error` derives.

Explain how `.context()` and `.with_context()` operate on types that are not aware of `anyhow`.

Resolves #1597.

### DIFF
--- a/src/error-handling/thiserror-and-anyhow.md
+++ b/src/error-handling/thiserror-and-anyhow.md
@@ -41,13 +41,22 @@ fn main() {
 
 <details>
 
+## `thiserror`
+
 * The `Error` derive macro is provided by `thiserror`, and has lots of useful
-  attributes like `#[error]` to help define a useful error type.
-* `anyhow::Result<V>` is a type alias for `Result<V, anyhow::Error>`.
+  attributes to help define error types in a compact way.
+* The `std::error::Error` trait is derived automatically.
+* The message from `#[error]` is used to derive the `Display` trait.
+
+## `anyhow`
+
 * `anyhow::Error` is essentially a wrapper around `Box<dyn Error>`. As such it's again generally not
   a good choice for the public API of a library, but is widely used in applications.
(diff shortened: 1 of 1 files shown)


djmitche (approved):
Perfect, thanks!

PR #1601, opened 2023-12-20
Add missing mod declarations to modules solutions

hurryabit:
The solution for the modules exercise does not work since it is missing a a few module declarations. Here, we add them.


djmitche (approved):
Good catch - I should have seen that :)

PR #1602, opened 2023-12-20
Add directory tree to mod exercise solution

hurryabit:
The solution to the modules exercise comes with a bunch of files. I think it would  help the students to get a quick overview of the directory structure we've chosen first.


djmitche (approved):
That's great!

PR #1605, opened 2023-12-20
Make protobuf exercise slightly easier

hurryabit:
As @djmitche suggested in
https://github.com/google/comprehensive-rust/pull/1591#pullrequestreview-1782012715
it might be a little easier for the students if we give them a prefix of
the code required for `parse_field`. That's exactly what we do here.


djmitche (approved):
Looks good! The `WireType::Len` arm still requires thinking about `remainder` so should provide a bit of challenge.

PR #1606, opened 2023-12-20
Remove Copy bound from binary tree exercise

hurryabit:
There's absolutely no need for the element type of the binary tree to be `Copy`. Let's remove this bound hence.


djmitche (approved):
I had done this to avoid having to think about ownership, but your feedback suggests this actually confused students, so I agree that removing the bound is a good idea.

PR #1607, opened 2023-12-20
Let BinaryTreeL::has take a reference

hurryabit:
Currently, `BinaryTree::has` takes its argument by value. This is a pretty unrealistic API for a Rust library. Let's fix this and take the argument by reference instead.


djmitche (approved):
Looks good! The tests also provide a nice example of references to temporaries.

PR #1609, opened 2023-12-20
macros.md - fix typo

btoews:


mgeisler:
Thanks for the patch! I see it got fixed later in #1638.

PR #1611, opened 2023-12-21
Chromium: Add a section about the `chromium::import!` macro.

anforowicz:


anforowicz:
@adetaylor, can you PTAL?

/cc @danakj

anforowicz:
I guess that we may eventually need to also update the example solutions, but this is managed outside of this GitHub repo.

adetaylor (approved):

PR #1616, opened 2023-12-27
Include notes on when to use lifetime annotations

djmitche:
This addresses 

> Could use more clarity on why or when people would want to use lifetimes -- when is it worth going to the trouble of not Cloning, etc.

in #1565.


mgeisler (approved):
Thanks, I like this brief explanation.

PR #1635, opened 2024-01-03
fix mdbook redirection table

uta8a:
Link `overloading` in Speaker note in URL: https://google.github.io/comprehensive-rust/hello-world/hello-world.html is broken.

- broken: https://google.github.io/comprehensive-rust/hello-world/basic-syntax/functions-interlude.html (404)
- correct: https://google.github.io/comprehensive-rust/control-flow-basics/functions.html

The reason why it is broken is below.

Original markdown is here https://github.com/google/comprehensive-rust/blob/43474d27d1276e8a88307a1686ee3f36cd77cf7f/src/hello-world/hello-world.md?plain=1#L36-L37

```
[overloading](basic-syntax/functions-interlude.md)
```

Page path is `/hello-world/hello-world.html`, so overloading link become `/hello-world/basic-syntax/functions-interlude.html`, which is 404.

I feel it is better to use redirect to exact path, so I edit `book.toml` and make link redirect to `/basic-syntax/functions-interlude.html`, which re-redirect to path `/control-flow-basics/functions.html`

https://github.com/google/comprehensive-rust/blob/43474d27d1276e8a88307a1686ee3f36cd77cf7f/book.toml#L114


djmitche (changes requested):
Thank you!

I think we should try to avoid redirect chains, so if you could change `book.toml` so that it redirects to the final location, that would be great.

Also, please update `hello-world.md` to point directly to the target page. The redirect is great in case someone bookmarked the page, but let's keep the links in the book pointing to the "real" location.

uta8a:
Thank you for your review!

I agree that redirects should be avoided. I fixed the docs accordingly.

Also, if I modify `hello-world.md`'s link, isn't it necessary to update book.toml? If so, I will remove the redirect part in `book.toml` I added.

djmitche (approved):
Thank you! I agree that the `book.toml` change isn't critical here. We have historically added such redirects in case someone had an old URL bookmarked or used in a lnik from elsewhere -- both of which are unlikely here. However, the redirect doesn't hurt anything!

PR #1639, opened 2024-01-03
Use `expect()` for error handling in `task.md`

proski:
The use of conditionals for error handling makes the code look mode complex than is.

Read the actual port number from the listener instead of hardcoding port 6142 in two places. That also demonstrates how to deal with ephemeral ports.


src/async/tasks.md
```diff
@@ -14,34 +14,27 @@ use tokio::net::TcpListener;
 #[tokio::main]
 async fn main() -> io::Result<()> {
     let listener = TcpListener::bind("127.0.0.1:6142").await?;
```
djmitche, line 16: If we want to demonstrate ephemeral ports, this should be `0` with a comment indicating what that means.

```suggestion
    let listener = TcpListener::bind("127.0.0.1:0").await?;
```

djmitche (changes requested):
I like the `.expect` / `.unwrap` approach a lot better than `?` in this case, as `tokio::spawn` will discard the `Result` from the async block.  As you probably saw, the speaker notes suggest factoring this async block out into a function, at which time `?` becomes more practical, with the async block responsible for calling the function and checking the `Result`.

djmitche (approved):

djmitche:
Ugh, it's flagging me as not having signed the CLA, since that suggestion used my default email address. Computers are awesome.

djmitche:
(I just squashed the commits)

Issue #1463, opened 2023-11-10
Translation impact of v2 changes


mgeisler:
# Head of `main`

Looking only at the translations linked in our language picker, I see these numbers right now using `main`:

```
pt-BR before: 1597 translated messages, 196 fuzzy translations, 727 untranslated messages.
zh-CN before: 1634 translated messages, 193 fuzzy translations, 693 untranslated messages.
zh-TW before: 1505 translated messages, 137 fuzzy translations, 878 untranslated messages.
ko before: 1968 translated messages, 204 fuzzy translations, 348 untranslated messages.
es before: 2018 translated messages, 112 fuzzy translations, 390 untranslated messages.
```

This is after updating the translations with the newest strings from `main`:

```sh
MDBOOK_OUTPUT='{"xgettext": {"pot-file": "messages.pot"}}' mdbook build -d po
for lang in pt-BR zh-CN zh-TW ko es; do
    msgmerge --update po/$lang.po po/messages.pot
    echo -n "$lang before: "; msgfmt -o /dev/null --stat po/$lang.po
done
```

# Head of `cr2-organization`

Updating to 9dcb262 on `cr2-organization` and running the same commands will show the impact of the branch in terms of messages that have been touched. I get these numbers:

```
pt-BR after: 1086 translated messages, 351 fuzzy translations, 1148 untranslated messages.
zh-CN after: 1160 translated messages, 320 fuzzy translations, 1105 untranslated messages.
zh-TW after: 1019 translated messages, 281 fuzzy translations, 1285 untranslated messages.
ko after: 1453 translated messages, 377 fuzzy translations, 755 untranslated messages.
es after: 1530 translated messages, 274 fuzzy translations, 781 untranslated messages.
```

In table form:

| Language | Translated before | %   | Translated after | %   |
|----------|-------------------|-----|------------------|-----|
| pt-BR    | 1597              | 63% | 1086             | 42% |
| zh-CN    | 1634              | 65% | 1160             | 45% |
| zh-TW    | 1505              | 60% | 1019             | 39% |
| ko       | 1968              | 78% | 1453             | 56% |
| es       | 2018              | 80% | 1530             | 59% |

This suggests that the new material touch about 500 messages, or 1/3 of the content.

# Using back-dated sources

Now, the above shows the raw delta between `main` and `cr2-organization`. The actual impact on the translations will be different: the translations are published using back-dated sources, sources that match the time of the last `msgmerge`.

[...]

Issue #262, opened 2023-01-23
Korean translation


mgeisler:
As discussed in #110 and #130, we should incorporate the Korean translation from https://github.com/keispace/comprehensive-rust-kr.

mgeisler:
@keispace, would you be able to open a PR with the `.po` file as you have it right now?

I would like to start working on publishing the translation under https://google.github.io/comprehensive-rust/, and this way you and others can get experience with the translation infrastructure.

mgeisler:
Fixed via #290 and #300. Thanks a lot @keispace and @jiyongp!

mgeisler:
I just tried browsing through the translation — I cannot read it, but it looks awesome to me! I would be happy to start publishing this as soon as possible. We just need to add the language to `publish.yml`. I'll also add a language picker to the theme, but that's not strictly necessary to start publishing it.

jiyongp:
Thanks, @keispace again. As a follow-up, I've created #302 (draft) to further revise the translation.

mgeisler:
For reference, there is a Korean translation of the Rust Book: https://github.com/rinthel/rust-lang-book-ko.

mgeisler:
Let me reopen this until we start publishing it.

jiyongp:
Current status: day 1, day 2, day 3 morning, and day 4 were merged. day 3 afternoon under review (#422).

mgeisler:
Hi @jiyongp and @jooyunghan, I've asked our translation team to translate the content on bare metal. When done, I'll put up a PR and I hope I can get you two to skim it over. The translators are probably not Rust experts, so I'm sure there are some things to fix :slightly_smiling_face:

jiyongp:
Thanks. That will work. 

Besides, I started updating the Korean translation this week. The progress is (and will be) slow, but just wanted to let you know.

mgeisler:
I created #925 to track the refresh.

henrif75:
I'm reopening the issue until we publish the v.2.

PR #1641, opened 2024-01-04
Refreshes `ko.po` file (#262) to v.2 without adding translations.

henrif75:
Updates the `ko.po` with the v.2 English strings, without adding new translations. 
It will set the base to merge the professional translations later (#1463).

To avoid disruptions on the live KO version, I'll merge the v2 into a new branch, to merge to main when all is done.


henrif75:
Actually, the file is wrong, I'll re-submit when I fix it.

PR #1658, opened 2024-01-05
Update `mdbook` and `mdbook-i18n-helpers`

mgeisler:
This also ports over the `mdbook` theme changes between version 0.4.35 and 0.4.36.

### DIFF
--- a/.github/workflows/install-mdbook/action.yml
+++ b/.github/workflows/install-mdbook/action.yml
@@ -8,7 +8,7 @@ runs:
     # The --locked flag is important for reproducible builds. It also
     # avoids breakage due to skews between mdbook and mdbook-svgbob.
     - name: Install mdbook
-      run: cargo install mdbook --locked --version 0.4.35
+      run: cargo install mdbook --locked --version 0.4.36
       shell: bash
 
     - name: Install mdbook-svgbob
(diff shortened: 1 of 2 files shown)


.github/workflows/install-mdbook/action.yml
```diff
@@ -8,15 +8,15 @@ runs:
     # The --locked flag is important for reproducible builds. It also
     # avoids breakage due to skews between mdbook and mdbook-svgbob.
     - name: Install mdbook
-      run: cargo install mdbook --locked --version 0.4.35
+      run: cargo install mdbook --locked --version 0.4.36
```
qwandor, line 11: Is there a corresponding change needed somewhere else to use this version when building locally?

qwandor (approved):
Huh, surprising that a theme change is needed for a patch revision like this, that seems like it should be considered a breaking change.

.github/workflows/install-mdbook/action.yml
mgeisler: We tell people to install the latest version and that's normally good enough.

mgeisler:
Hey @qwandor, thanks for the review!

> Huh, surprising that a theme change is needed for a patch revision like this, that seems like it should be considered a breaking change.

The theme change is actually backwards compatible — we would just not get the benefits without the theme change. We're probably also a bit unusual in the way that we version one file out of the set of files that make up the theme.

PR #1663, opened 2024-01-07
Clarify that pattern matching consumes enum variant by default

### LINKED ISSUE #1491: Variant payload: "The pattern binds _references_ to the fields"
https://google.github.io/comprehensive-rust/enums/variant-payloads.html

> The values in the enum variants can only be accessed after being pattern matched. The pattern binds references to the fields in the “match arm” after the =>. 

That sounds wrong to me. The variables used in `match` capture the enum contents. References can be requested using the `ref` keyword, but it's not used in the example.

Here's a slightly modified example (added `Box` and another `match`) that won't compile precisely because the binding is a move, not a borrow by default.

```
enum WebEvent {
    PageLoad,                 // Variant without payload
    KeyPress(Box<char>),      // Tuple struct variant
    Click { x: i64, y: i64 }, // Full struct variant
}

#[rustfmt::skip]
fn inspect(event: WebEvent) {
    match event {
        WebEvent::PageLoad       => println!("page loaded"),
        WebEvent::KeyPress(c)    => println!("pressed '{c}'"),
        WebEvent::Click { x, y } => println!("clicked at x={x}, y={y}"),
    }
    match event {
        WebEvent::PageLoad       => println!("page loaded"),
        WebEvent::KeyPress(c)    => println!("pressed '{c}'"),
        WebEvent::Click { x, y } => println!("clicked at x={x}, y={y}"),
    }
}

fn main() {
    let load = WebEvent::PageLoad;
    let press = WebEvent::KeyPress(Box::new('x'));
    let click = WebEvent::Click { x: 20, y: 80 };

    inspect(load);
    inspect(press);
    inspect(click);
}
```

proski:
Clarify that pattern matching consumes enum variant by default

Describe two ways to borrow the variant without consuming it.

Fixes #1491.

### DIFF
--- a/src/pattern-matching/destructuring.md
+++ b/src/pattern-matching/destructuring.md
@@ -64,9 +64,15 @@ Key points:
   errors when running the code. Point out the places where your code is now
   inexhaustive and how the compiler tries to give you hints.
 - The values in the enum variants can only be accessed after being pattern
-  matched. The pattern binds references to the fields in the "match arm" after
-  the `=>`.
+  matched.
 - Demonstrate what happens when the search is inexhaustive. Note the advantage
   the Rust compiler provides by confirming when all cases are handled.
+- Save the result of `divide_in_two` in the `result` variable and `match` it in
+  a loop. That won't compile because `msg` is consumed when matched. To fix it,
+  match `&result` instead of `result`. That will make `msg` a reference so it
+  won't be consumed. This
+  ["match ergonomics"](https://rust-lang.github.io/rfcs/2005-match-ergonomics.html)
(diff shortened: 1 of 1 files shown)


proski:
I described the "match ergonomics" first, as it's less intrusive and more modern. I did mention `ref` as the compiler still recommends it. I tried hard to keep it short and simple, so I shortened the item where the references were (incorrectly or confusingly) mentioned.

Strictly speaking, this is not quite correct:

> The values in the enum variants can only be accessed after being pattern matched.

Match guards can access the values _when_ the pattern is being matched. But I didn't want to add more complexity. Besides, it's a separate issue.

mgeisler (approved):
Thanks a lot, this is great!

Issue #1516, opened 2023-11-29
"Trait Objects" require understanding of Box


djmitche:
The traits-and-methods segment is already pretty complex. Perhaps we should cover these in the smart-pointers segment.

PR #1682, opened 2024-01-10
Replace GUI exercise with Logger

### LINKED ISSUE #1617: GUI Exercise is more about text formatting than trait objects
This is from feedback in #1565. It is in the "Methods and Traits" segment, so it doesn't necessarily need to focus on trait objects. But, something nifty with traits would be good.

I've not come up with a great idea yet. The best I've got is some kind of signal-processing DAG, like old-school analog synthesizers. So you'd have a few types that take `dyn SignalSource` trait objects, and themselves implement `SignalSource`. But, signal processing is a little complex for an example!

I'm open to ideas here!

djmitche:
This should be a bit simpler, and notably
 * does not require trait objects, which per #1516 should be moved later in the course
 * does not require a lot of futzing with string formatting

But all that hard work developing the GUI exercise is not for naught: it remains in the "Modules" segment, where students will get a chance to read some Rust code and reorganize it a little bit.

Fixes #1617.

R=mgeisler as the original author of the GUI exercise.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1356,6 +1356,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "modules"
+version = "0.1.0"
+
+[[package]]
 name = "native-tls"
 version = "0.2.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ members = [
   "src/std-types",
   "src/std-traits",
   "src/iterators",
+  "src/modules",
   "src/testing",
   "src/memory-management",
   "src/smart-pointers",
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -78,7 +78,7 @@
   - [Traits](methods-and-traits/traits.md)
   - [Deriving](methods-and-traits/deriving.md)
   - [Trait Objects](methods-and-traits/trait-objects.md)
-  - [Exercise: GUI Library](methods-and-traits/exercise.md)
+  - [Exercise: Generic Logger](methods-and-traits/exercise.md)
     - [Solution](methods-and-traits/solution.md)
 - [Generics](generics.md)
   - [Generic Functions](generics/generic-functions.md)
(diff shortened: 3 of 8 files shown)


djmitche:
`dprint fmt` on my system doesn't make any changes here. Any idea why it's failing in the PR and not locally?

mgeisler:
Thanks, I like this a lot! It makes the generics part much more focused and it shows a useful pattern.



> `dprint fmt` on my system doesn't make any changes here. Any idea why it's failing in the PR and not locally?

Not sure! I just tried it here:

```
% gh pr checkout 1682
remote: Enumerating objects: 17, done.
remote: Counting objects: 100% (17/17), done.
remote: Compressing objects: 100% (7/7), done.
remote: Total 17 (delta 10), reused 16 (delta 10), pack-reused 0
Unpacking objects: 100% (17/17), 3.73 KiB | 954.00 KiB/s, done.
From github.com:google/comprehensive-rust
 * [new ref]         refs/pull/1682/head -> issue1617
Switched to branch 'issue1617'
% dprint fmt
Formatted 1 file.
% git diff                                                           (~/src/comprehensive-rust)
diff --git a/src/modules/exercise.rs b/src/modules/exercise.rs
index 87bf452..09fbb19 100644
--- a/src/modules/exercise.rs
+++ b/src/modules/exercise.rs
@@ -84,8 +84,8 @@ impl Widget for Window {
 
         let inner_width = self.inner_width();
 
-        // TODO: Change draw_into to return Result<(), std::fmt::Error>. Then use the ?-operator
-        // here instead of .unwrap().
+        // TODO: Change draw_into to return Result<(), std::fmt::Error>. Then use the
+        // ?-operator here instead of .unwrap().
         writeln!(buffer, "+-{:-<inner_width$}-+", "").unwrap();
         writeln!(buffer, "| {:^inner_width$} |", &self.title).unwrap();
         writeln!(buffer, "+={:=<inner_width$}=+", "").unwrap();
```

Ah, I know why!

Wrapping of comments is an unstable `rustfmt` feature... We therefore run `dprint fmt` after a `rustup default unstable` in CI. I also see no output if I switch back to stable temporarily!

Let me try to fix this!

mgeisler:
I found a way to run `rustfmt` from Rust nightly, assuming you have `rustup` on the system. Please see #1694.

mgeisler (approved):
Thanks a lot!

PR #1683, opened 2024-01-11
Use Mutex as an example of interior mutability

### LINKED ISSUE #1512: "Interior Mutability" should use Mutex, not Cell/RefCell
Cell and RefCell are pretty advanced for introducing interior mutability -- a mutex is much more familiar territory.

Mention Cell and RefCell in the speaker notes as alternatives to Mutex for more specific situations.

djmitche:
Mutex is probably more broadly understood by people coming from other languages.

Fixes #1512.

### DIFF
--- a/src/borrowing/interior-mutability.md
+++ b/src/borrowing/interior-mutability.md
@@ -2,73 +2,59 @@
 minutes: 10
 ---
 
-<!-- NOTES:
-Introduce the concept, with an example based on Mutex showing an `&self` method doing mutation; reference Cell/RefCell without detail.
--->
-
 # Interior Mutability
 
-Rust provides a few safe means of modifying a value given only a shared
-reference to that value. All of these replace compile-time checks with runtime
-checks.
+In some situations, exclusive (`mut`) access cannot be represented at compile
+time. For example, data might be accessed from multiple threads. The "interior
+mutability" pattern allows exclusive access behind a shared reference, using
+a runtime check to enforce the borrowing rules.
 
-## `Cell` and `RefCell`
+## `Mutex`
 
-[`Cell`](https://doc.rust-lang.org/std/cell/struct.Cell.html) and
-[`RefCell`](https://doc.rust-lang.org/std/cell/struct.RefCell.html) implement
-what Rust calls _interior mutability:_ mutation of values in an immutable
-context.
-
-`Cell` is typically used for simple types, as it requires copying or moving
-values. More complex interior types typically use `RefCell`, which tracks shared
-and exclusive references at runtime and panics if they are misused.
+One standard type that supports interior mutability is `Mutex` (known as a lock
+in some other languages). Many threads can have a shared reference to a
+`Mutex`, and the lock operation performs a _runtime_ check that no other thread
+holds the lock before granting access to the value inside.
 
 ```rust,editable
-use std::cell::RefCell;
-use std::rc::Rc;
+use std::sync::Mutex;
 
-#[derive(Debug, Default)]
-struct Node {
-    value: i64,
-    children: Vec<Rc<RefCell<Node>>>,
+fn handle_hit(hit_counter: &Mutex<u32>) {
+    let mut num_hits = hit_counter.lock().unwrap();
+    *num_hits += 1;
 }
 
-impl Node {
-    fn new(value: i64) -> Rc<RefCell<Node>> {
-        Rc::new(RefCell::new(Node { value, ..Node::default() }))
-    }
-
-    fn sum(&self) -> i64 {
-        self.value + self.children.iter().map(|c| c.borrow().sum()).sum::<i64>()
-    }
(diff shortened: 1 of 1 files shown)


mgeisler:
Like @hurryabit said in the issue, I also feel it's weird to use a `Mutex` here: other languages typically don't let the mutex wrap the data like you do in Rust. So being familiar with [`std::mutex`](https://en.cppreference.com/w/cpp/thread/mutex) or a similar type might actually confuse people.

The `mutex.lock().unwrap()` pattern is strange as well and it's certainly something people ask about in the Concurrency class.

Talking about `Cell` feel equally clear to me: it's a storage location which you can `get` and `set` on. It is a very clear concept and it avoids the baggage of thinking about threads.

src/borrowing/interior-mutability.md
```diff
@@ -2,73 +2,59 @@
...
 
-Rust provides a few safe means of modifying a value given only a shared
-reference to that value. All of these replace compile-time checks with runtime
-checks.
-
-## `Cell` and `RefCell`
+In some situations, exclusive (`mut`) access cannot be represented at compile
+time. For example, data might be accessed from multiple threads. The "interior
+mutability" pattern allows exclusive access behind a shared reference, using a
+runtime check to enforce the borrowing rules.
```
mgeisler, line 10: For something like `Cell`, this is not actually 100% true. I believe `Cell` has no runtime overhead: because it does not let you get a reference to the contained value, it doesn't have to do any checks when you `get` or `set` it. The restriction is that it's less useful for non-`Copy` types.

src/borrowing/interior-mutability.md
```diff
@@ -2,73 +2,59 @@
...
--->
-
 # Interior Mutability
 
-Rust provides a few safe means of modifying a value given only a shared
-reference to that value. All of these replace compile-time checks with runtime
-checks.
-
-## `Cell` and `RefCell`
+In some situations, exclusive (`mut`) access cannot be represented at compile
```
mgeisler, line 7: Another thought I had while reading this is that "shared reference" and "exclusive reference" don't express the read-only and mutable access well. Could we change the slide to use "read-only reference" and "mutable reference" here instead?

src/borrowing/interior-mutability.md
djmitche: This is why I didn't want to use Cell to introduce interior mutability..

src/borrowing/interior-mutability.md
djmitche: We use those terms for references earlier in the course (https://google.github.io/comprehensive-rust/references/shared.html) and I think they are the preferred term in the Rust community. I can replace the parenthesized `(mut)` with `(mutable)` if that helps.

djmitche:
I went back to using just `RefCell`, with the same example, but improved the text and notes a little bit. WDYT?

mgeisler (approved):
Thanks, I think should be good.

mgeisler:
> I went back to using just `RefCell`, with the same example, but improved the text and notes a little bit. WDYT?

I think it looks pretty good. I really like your concise part about "if there are no references, borrowing rules cannot be broken". I had no thought of it like this myself.

Have you considered focusing on `Cell` only? As a way to introduce the main concept of a type `T` where you can call `set` on a `&T` value? To me, that is the essence of interior mutability: the typesystem says you have an shared read-only reference, and yet you can modify something. The precise default of `Cell` vs `RefCell` is less important in my opinion.

src/borrowing/interior-mutability.md
mgeisler: I honestly don't know what the canonical term is... but I tried looking it up:

- The Rust Book talks about _immutable_ and _mutable references_ in [References and Borrowing](https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#references-and-borrowing).
- The Rust Reference mentions _shared borrow_ and _mutable borrow_ in [Borrow operators](https://doc.rust-lang.org/stable/reference/expressions/operator-expr.html#borrow-operators).

src/borrowing/interior-mutability.md
mgeisler: Right, it's quite subtle and somewhat besides the point! Perhaps you could just drop talking about runtime cost here?

djmitche:
I did consider using `Cell` instead, but I think it's such a "thin" layer that it actually obscures the meaning -- it would be easy to take away an understanding that "interior mutability is this special `Cell::set` that takes &self but still modifies itself".

I think RefCell hits the sweet spot of being complex enough to illustrate the point, without being too complex.

mgeisler:
> I did consider using `Cell` instead, but I think it's such a "thin" layer that it actually obscures the meaning -- it would be easy to take away an understanding that "interior mutability is this special `Cell::set` that takes &self but still modifies itself".
> 
> I think RefCell hits the sweet spot of being complex enough to illustrate the point, without being too complex.

Alright, I'm happy with that. Let's get it in! :smile:

PR #1684, opened 2024-01-11
pt-BR: cr2 Day 1 translations

henrif75:
*Stats:*
Before: 1247 translated messages, 402 fuzzy translations, 1811 untranslated messages.
After: 1664 translated messages, 260 fuzzy translations, 1536 untranslated messages.
#317 #1463

### DIFF
--- a/po/pt-BR.po
+++ b/po/pt-BR.po
@@ -41,7 +41,7 @@ msgstr "O Ecossistema do Rust"
 
 #: src/SUMMARY.md:11
 msgid "Code Samples"
-msgstr "Amostras de Código"
+msgstr "Exemplos de Código"
 
 #: src/SUMMARY.md:12
 msgid "Running Cargo Locally"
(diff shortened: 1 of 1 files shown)


po/pt-BR.po
```diff
@@ -4748,30 +4830,28 @@ msgstr ""
...
 "A type alias creates a name for another type. The two types can be used "
 "interchangeably."
 msgstr ""
+"Um _alias_ de tipo cria um nome para outro tipo. Os dois tipos podem ser "
+"usados ​​de forma intercambiável."
 
 #: src/user-defined-types/aliases.md:13
 msgid "// Aliases are more useful with long, complex types:\n"
-msgstr ""
+msgstr "// _Aliases_ são mais úteis com tipos longos e complexos:\n"
```
joaovicmendes, line 4850: Should we be translating comments in the code snippets?

joaovicmendes (approved):
LGTM!

po/pt-BR.po
henrif75: I think so, because that's one of the changes with the mdbook, instead of adding the whole code snippet to the .po file, it's adding only the comments for potential localization.

henrif75:
> LGTM!

Thanks!

po/pt-BR.po
mgeisler: Yeah, _potential_ is the keyword: the translation system supports translation of comments and string literals in the code blocks. You should decide if you want to do this or not and document it somewhere for the language.

PR #1685, opened 2024-01-11
Describe the `select!` arm behavior step-by-step

proski:
The original text describes several steps in one sentence. It's barely decipherable. Use three sentences instead of one.


djmitche (approved):
Much better!

Issue #1677, opened 2024-01-08
Ambiguous "pointers to data on the stack"


proski:
https://google.github.io/comprehensive-rust/async/pitfalls/pin.html

I realize it's a complex topic, but would it be possible to use more clear language for the text that is already there?

> When you await a future, all local variables (that would ordinarily be stored on a stack frame) are instead stored in the Future for the current async block.

Why is the second "Future" capitalized?

Is it correct that the first future that the second future are different?

> If your future has pointers to data on the stack, those pointers might get invalidated.

This can be interpreted in two ways:

* The future has pointers on the stack. Those pointers point to some data.
* The future has pointers. Those pointers point to data on the stack.

How can the pointers get invalidated? Does the execution continue with the stack at a different address? Does some other code have write access to the future as it's awaiting?

> This is unsafe.

Maybe it should be written as "This would be unsafe" as it not a real danger for safe Rust code.

> Therefore, you must guarantee that the addresses your future points to don’t change.

This probably rules out the stack at a a different address, but I'm guessing here.

Who is that mischievous monkey that tries to move objects around?

> That is why we need to “pin” futures.

Looking at the example, it's the future we call (`timeout_fut`) that needs to be pinned. Yet its address is in the `worker()` scope. So it's not the case of a future that has pointers and needs to be pinned. One future has pointers (our code) but we are pinning another future (`timeout_fut`).

> Using the same future repeatedly in a select! often leads to issues with pinned values.

That's vague and scary. I think it means that the future cannot be replaced with another one and needs to be reset in place if possible. Thankfully, `tokio::time::sleep` can be `reset`. That could be mentioned in the notes below.

djmitche:
Thanks! These are good comments! I've got a bunch of issues like this to work through so this will go on the list. If you'd like to make a PR to fix this, that would be ❤️

djmitche:
Solved in #1687.

Issue #1678, opened 2024-01-08
Add links to background information in "more to explore"


djmitche:
Especially in the later bits of the course (day 4 afternoon), the text just can't teach these complex topics thoroughly. Let's add some links to blog posts, documentation pages, and other approaches to the content that might give students a different perspective to help unlock that "ah-hah" moment.

As suggested by @Chiogros in #100.

PR #1687, opened 2024-01-12
Better explanation why futures need to be pinned

proski:
Attempt to address #1677.

Expert review is needed. The new text is my best guess based on the original text and other explanations I could find online.

A few things to note:

* I'm trying to distinguish the future we return and the future we await. My assumption is that the stack contents goes to the future the code returns, not the future the code is awaiting.
* Readers could be worried if they need to pin the code they write. I'm reassuring them that the borrow checks would normally catch bad references.
* I'm intentionally avoiding the words that something is unsafe (or would be unsafe). The async Rust is safe.
* I'm trying to be clear that `Pin` is a protective wrapper around a pointer, not a mechanism that changes the pointer or the pointed object.
* Likewise, I don't want to give an impression that an unpinned pointer to a future is inherently unsafe or invalid. It just cannot be used to poll the future.
* I dropped the vague mention of the "issues", as it probably refers to the issue with replacing a future (as opposed to resetting it in place). It's already mentioned in the notes further on this page. It affects pinning on stack only, `Box::pin()` can be replaced.

### DIFF
--- a/src/async/pitfalls/pin.md
+++ b/src/async/pitfalls/pin.md
@@ -1,13 +1,17 @@
 # `Pin`
 
-When you await a future, all local variables (that would ordinarily be stored on
-a stack frame) are instead stored in the Future for the current async block. If
-your future has pointers to data on the stack, those pointers might get
-invalidated. This is unsafe.
+Async blocks and functions return types implementing the `Future` trait. The
+type returned is the result of a compiler transformation which turns local
+variables into data stored inside the future.
 
-Therefore, you must guarantee that the addresses your future points to don't
-change. That is why we need to "pin" futures. Using the same future repeatedly
-in a `select!` often leads to issues with pinned values.
+Some of those variables can hold pointers to other local variables. Because of
+that, the future should never be moved to a different memory location, as it
+would invalidate those pointers.
+
+To prevent moving the future type in memory, it can only be polled through a
+pinned pointer. `Pin` is a wrapper around a reference that disallows all
+operations that would move the instance it points to into a different memory
+location.
 
 ```rust,editable,compile_fail
 use tokio::sync::{mpsc, oneshot};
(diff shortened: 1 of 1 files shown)


src/async/pitfalls/pin.md
```diff
@@ -1,13 +1,20 @@
...
-a stack frame) are instead stored in the Future for the current async block. If
-your future has pointers to data on the stack, those pointers might get
-invalidated. This is unsafe.
-
-Therefore, you must guarantee that the addresses your future points to don't
-change. That is why we need to "pin" futures. Using the same future repeatedly
-in a `select!` often leads to issues with pinned values.
+All async blocks and async functions return types implementing the `Future`
+trait. When the async code awaits another future, the `Future` trait holds the
+local variables that would normally be located on the stack.
```
mgeisler, line 5: I don't think this should be linked to await specifically. My understanding is that the async function body is _compiled_ to a struct which `impl Future`. What was local variables in the function body become struct fields.

So what you see as a local variable in the code is never actually a variable in the code after the compiler transformation. Thus, I would not say that this happens when you call `await` since it happened already in the immediate desugaring.

An example of this desugaring can be seen in [Async fn as a Future](https://tokio.rs/tokio/tutorial/async#async-fn-as-a-future). The `State0`, `State1`, and `Terminated` enum variants correspond to different parts of the now-gone async function body. The variants hold the local variables that are alive in the different parts of the body.

mgeisler, line 5: Perhaps this would be a more direct way of saying what I suggested above:

```suggestion
Async blocks and functions return types implementing the `Future` trait.
The type returned is the result of a compiler transformation which turns
local variables into data stored in an enum.
```

mgeisler:
Hi @proski, I really like the new language here, I think it's a nice improvement!

One thing to keep in mind is that the pages are meant to be _slides_, so they should be brief. Longer explanations are best put in the notes at the bottom of the page (so they can be collapsed during normal teaching). So I would suggest having one 3-line paragraph as the intro text and then expand on that in the notes.

proski:
I reduced the top text to 3 paragraphs and moved some noted to the end.

Thank you for the link to the Tokio documentation, it demystified so many things for me!

I'm not sure why the "Course Schedule Updates" check is failing. I'll keep retrying until the CI check is fixed.

mgeisler:
> I reduced the top text to 3 paragraphs and moved some noted to the end.

Cool, I think it looks very nice now! Mentioning "self-referential" is great since it can give people a jumping-off point for further investigation. As @djmitche mentioned in #1678, we cannot expect to teach complex topics like this in detail on slides, but we can at least give people the relevant words to search for later.

Could you include a link to https://doc.rust-lang.org/std/pin/ somewhere (if we don't have it)? That should help people who want more background information.

> Thank you for the link to the Tokio documentation, it demystified so many things for me!

Yeah, me too :smile: 

> I'm not sure why the "Course Schedule Updates" check is failing. I'll keep retrying until the CI check is fixed.

Sorry, it's broken on our end... I noticed this yesterday in #1576. We'll probably have to remove the check, but please just ignore it for now.

mgeisler (approved):
I like this a lot, @djmitche, do you also want to take a look?

mgeisler:
Thanks for this, let's merge this now and iterate more on it later (if people feel it's needed). @fw-immunant, you'll be teaching the Concurrency class soon, so you might have opnions on this too.

PR #1676, opened 2024-01-08
ja: refresh ja.po file

kantasv:
Hi, ja-translators, #652 ! I've refreshed `ja.po` file as of today, Jan 8th 2024.
I used the following procedures. Could you review the diff? Thanks 😊 

```
MDBOOK_OUTPUT='{"xgettext": {"pot-file": "messages.pot"}}' \
  mdbook build -d po
```

```
msgmerge --update po/ja.po po/messages.pot
```

PR #1636, opened 2024-01-03
ja: Translate Chapter 60 (Shared State)

kantasv:
Hi, ja translation folks (#652 ). Here's an MR for the chapter "Shared States." Could you review the translations? any feedback would be appreciated. Thank you! 

Note: There are several lines containing outdated `msgid`, but we can refresh the `ja.po` file in a separate MR. We can slightly update corresponding `msgstr` afterwards.

PR #1690, opened 2024-01-13
ja: Translate Chapter 60 (Shared State)

kantasv:
Hi, ja translation team (#652 ). Here's an MR for the chapter "Shared States." Could you review the translations? any feedback would be appreciated. Thank you 😄  

cc: @keiichiw , @chikoski , @HidenoriKobayashi , @ternbusty 
(Retrieved translaftion draft #1636 Chapter 60 draft, after recent `ja.po` file refresh #1676 )


po/ja.po
```diff
@@ -18290,25 +18290,35 @@ msgid ""
...
+"に2つの型により行われます："
 
 #: src/concurrency/shared_state.md:6
 msgid ""
 "[`Arc<T>`](https://doc.rust-lang.org/std/sync/struct.Arc.html), atomic "
 "reference counted `T`: handles sharing between threads and takes care to "
 "deallocate `T` when the last reference is dropped,"
 msgstr ""
+"[`Arc<T>`](https://doc.rust-lang.org/std/sync/struct.Arc.html), アトミックに"
+"参照カウントする型 （atomic reference counted）を `T` とするとき : スレッド間"
```
kantasv, line 18303: Translation confidence: Low, Re: "atomic reference counted"

po/ja.po
```diff
@@ -18407,46 +18440,59 @@ msgid ""
 "[`PoisonError`](https://doc.rust-lang.org/std/sync/struct.PoisonError.html). "
 "You can call `into_inner()` on the error to recover the data regardless."
 msgstr ""
+"スレッドにパニックを起こした`Mutex`がある場合、保護すべきデータが整合性の欠け"
+"た状態にある可能性を伝えるため、`Mutex`は「ポイゾンされた」（\"poisoned\"）状"
+"態になります。ポイゾンされたMutexに対して `lock()` をコールすると、"
```
kantasv, line 18445: Referred to JA Rustbook's translation, i.e., "poison" -> "ポイゾン" https://doc.rust-jp.rs/rust-nomicon-ja/poisoning.html

po/ja.po
```diff
@@ -18290,25 +18290,35 @@ msgid ""
 "Rust uses the type system to enforce synchronization of shared data. This is "
 "primarily done via two types:"
 msgstr ""
+"Rustは共有されたデータの同期を強制するために型システムを利用します。これは主"
```
HidenoriKobayashi, line 18293: I would say: Rustは共有データを確実に同期するために

po/ja.po
HidenoriKobayashi: Hmm, this is a hard one. This basically means that T has a reference count and its count is managed atomically, right?

I would probably just leave it un-translated up to ':".

po/ja.po
```diff
@@ -18290,25 +18290,35 @@ msgid ""
...
+"参照カウントする型 （atomic reference counted）を `T` とするとき : スレッド間"
+"の共有を扱い、最後の参照がドロップされたとき `T` をデアロケートすることを担当"
+"する、"
 
 #: src/concurrency/shared_state.md:8
 msgid ""
 "[`Mutex<T>`](https://doc.rust-lang.org/std/sync/struct.Mutex.html): ensures "
 "mutually exclusive access to the `T` value."
 msgstr ""
+"[`Mutex<T>`](https://doc.rust-lang.org/std/sync/struct.Mutex.html): `T`の値へ"
```
HidenoriKobayashi, line 18312: Maybe: `T`型の変数への相互排他アクセス ?

Well, value is not always a variable, but if you are to put it in a Mutex, then I would guess it's okay to say that is a variable to make it sound more natural in Japanese.

po/ja.po
```diff
@@ -18325,35 +18335,47 @@ msgid ""
...
 msgstr ""
+"`Arc` は\"Atomic Reference Counted\"の略で、アトミック操作を利用するという点"
+"で、`Rc`がスレッド安全になったバージョンのようなものです。"
 
 #: src/concurrency/shared_state/arc.md:31
 msgid ""
 "`Arc<T>` implements `Clone` whether or not `T` does. It implements `Send` "
 "and `Sync` if and only if `T` implements them both."
 msgstr ""
+"`Arc<T>` は `Clone` を実装します。このこは`T`が`Clone`を実装するしないに関係"
```
HidenoriKobayashi, line 18346: That's too pretty: このこ ;)

Probably something got deleted unintentionally?

po/ja.po
```diff
@@ -18325,35 +18335,47 @@ msgid ""
...
+"ありません。`T`が`Send`と`Sync`の両方を実装している場合で、かつその場合に限"
+"り、`Arc<T>` は両者を実装します。"
 
 #: src/concurrency/shared_state/arc.md:33
 msgid ""
 "`Arc::clone()` has the cost of atomic operations that get executed, but "
 "after that the use of the `T` is free."
 msgstr ""
+"`Arc::clone()`には実行されるアトミック操作のコストがあります。ただ、そのよう"
```
HidenoriKobayashi, line 18355: I would say: `Arc::clone()`にはアトミック操作のコストがかかります。ただ、その後は`T`の利用に関するコストはかかりません。

HidenoriKobayashi (commented):
Just some readability suggestions. You may want to double check with reviewers before actually making changes though.

po/ja.po
kantasv: Yes, that is my understanding. I agree, the reader only need to learn that Arc stands for atomic reference counted, rather than get confused with the translation?

FYI, the Rustbook JP translated as "原子的に参照する型." [状態共有並行性 - Rustbook](https://doc.rust-jp.rs/book-ja/ch16-03-shared-state.html#arct%E3%81%A7%E5%8E%9F%E5%AD%90%E7%9A%84%E3%81%AA%E5%8F%82%E7%85%A7%E3%82%AB%E3%82%A6%E3%83%B3%E3%83%88) but this sounds confusing. At least, "原子的"->"アトミック" is better in the programming context, IMO.

kantasv:
Hi, @HidenoriKobayashi thanks for your feedback. I applied changes :) 
If you're interested, could you be a codeowner of the `ja.po` file?

HidenoriKobayashi:
> Hi, @HidenoriKobayashi thanks for your feedback. I applied changes :) If you're interested, could you be a codeowner of the `ja.po` file?

Sure, if other folks are fine with that.

po/ja.po
```diff
@@ -18366,38 +18387,49 @@ msgid ""
...
 msgstr ""
+"Rustにおける`Mutex`とは、保護されるデータである、たった一つの要素から構成され"
+"たコレクションのようなものです。"
 
 #: src/concurrency/shared_state/mutex.md:33
 msgid ""
 "It is not possible to forget to acquire the mutex before accessing the "
 "protected data."
 msgstr ""
+"保護されたデータにアクセスする前に、相互排他を確保し忘れることはありません。"
```
keiichiw, line 18407:
```suggestion
"保護されたデータにアクセスする前に、ミューテックスを確保し忘れることはありません。"
```

It's common to say "ミューテックス" to refer the value of `Mutex<T>`. [Wikipedia](https://ja.wikipedia.org/wiki/%E3%83%9F%E3%83%A5%E3%83%BC%E3%83%86%E3%83%83%E3%82%AF%E3%82%B9)

po/ja.po
```diff
@@ -18407,46 +18439,59 @@ msgid ""
 "[`PoisonError`](https://doc.rust-lang.org/std/sync/struct.PoisonError.html). "
 "You can call `into_inner()` on the error to recover the data regardless."
 msgstr ""
+"スレッドにパニックを起こした`Mutex`がある場合、保護すべきデータが整合性の欠け"
```
keiichiw, line 18442:
```suggestion
"`Mutex`を保持したスレッドがパニックを起こした場合、保護すべきデータが整合性の欠け"
```

po/ja.po
```diff
@@ -18407,46 +18439,59 @@ msgid ""
...
 msgstr ""
+"`v: Arc<_>`は別のスレッドにムーブされる前に、`v2`としてクローンされる必要があ"
+"ります。`move` がラムダ式に追加されたことに注意してください。"
 
 #: src/concurrency/shared_state/example.md:57
 msgid ""
 "Blocks are introduced to narrow the scope of the `LockGuard` as much as "
 "possible."
 msgstr ""
+"ブロック文は`LockGuard`のスコープを可能な限り狭めるために導入されています。"
```
keiichiw, line 18494:
```suggestion
"ブロックは`LockGuard`のスコープを可能な限り狭めるために導入されています。"
```

In Rust, block is an expression (式), not a statement (文). So, we don't say ブロック文.
See https://doc.rust-lang.org/reference/expressions/block-expr.html.
Anyway, we can just say "ブロック" here.

keiichiw (changes requested):
Thanks for the PR!

keiichiw:
> > Hi, @HidenoriKobayashi thanks for your feedback. I applied changes :) If you're interested, could you be a codeowner of the `ja.po` file?
> 
> Sure, if other folks are fine with that.

I'm fine with adding Hidenori to the owner file. Feel free to create a PR

po/ja.po
keiichiw: In your original translation, it can be read that a Mutex can cause panick, but it doesn't make sense as Mutex is a value.

kantasv:
Thanks @keiichiw -san for concrete feedback and suggestions with references, which helps a lot! I have deliered changes. Could you reveiw it once again? Thank you!

keiichiw (approved):
LGTM! Thank you for the change

PR #1704, opened 2024-01-15
Move `pandoc/` folder out of the way

mgeisler:
The config fragments there are only used in the GitHub actions, so we can stash them way in the `.github/` folder.

@max-heller, I guess this should be fine with the new infrastructure?


max-heller (commented):
Looks good to me!

mgeisler:
> Looks good to me!

Thanks!

djmitche (approved):

.github/workflows/build.sh
```diff
@@ -31,11 +31,15 @@ else
...
+    if [ -f ".github/pandoc/$book_lang.yaml" ]; then
+        export MDBOOK_OUTPUT__PANDOC__PROFILE__PDF__DEFAULTS=".github/pandoc/$book_lang.yaml"
+    fi
 fi
 
 mdbook build -d "$dest_dir"
-[ -f "$dest_dir/pandoc/pdf/comprehensive-rust.pdf" ] && mv "$dest_dir/pandoc/pdf/comprehensive-rust.pdf" "$dest_dir/html/"
+if [ -f "$dest_dir/pandoc/pdf/comprehensive-rust.pdf" ]; then
+    mv "$dest_dir/pandoc/pdf/comprehensive-rust.pdf" "$dest_dir/html/"
+fi
```
mgeisler, line 42: @max-heller, I wanted to ask about this but forgot: when would the PDF not exist?

When `build.sh` is run in CI, the `mdbook-pandoc` plugin should always be installed and so the PDF should always exist, no?

If that's true, then I would like to remove the conditional.

max-heller: If this is only ever called from the CI job with `mdbook-pandoc` installed then yes, it should always exist (unless the build fails, which will abort the script before reaching this check)

mgeisler: Excellent, then I'll take it out for now: #1713. I have toyed with the idea of suggesting to people to use `build.sh` for local development, but for now it's only used in CI.

@sakex has been working on a new `mdbook` output format which would build all translations in one go: https://github.com/google/mdbook-i18n-helpers/pull/84. That would need to learn the logic of `build.sh` and then this could become the suggest way for people to build things locally.

PR #1710, opened 2024-01-16
split typos and formatting checks into separate workflows

djmitche:
Without this it's easy for failures of one of these checks to hide failures of the other.


.github/workflows/build.yml
```diff
@@ -34,6 +34,17 @@ jobs:
         with:
```
mgeisler, line 34: The call here should go away now.

mgeisler (approved):
Great idea, you should just remove the dupliate `typos` check first.

PR #1713, opened 2024-01-17
Always install and run `mdbook-pandoc`

mgeisler:
Before, we only installed and ran `mdbook-pandoc` when publishing the course — which means that a PR change could accidentally break the publish workflow. We now test that it works on every PR.

From a discussion in #1704.

### DIFF
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -37,9 +37,7 @@ else
 fi
 
 mdbook build -d "$dest_dir"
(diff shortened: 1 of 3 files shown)


mgeisler:
Ah, the build fails... turns out we don't install `mdbook-pandoc` for our regular build workflow! :smile: 

I guess we should install it so that we know that we don't accidentally break the publish step with one of our updates.

qwandor (approved):

PR #1723, opened 2024-01-18
Update to embedded-hal 1.0

qwandor:


randomPoison:
I just merged a couple of PRs that bump the crate version, but I'd like to merge your cleanup of `embedded-hal.md`. @qwandor can you rebase this on top of main? I'll approve and merge after that :)

qwandor:
We can't update to `embedded-hal` 1.0 until the `microbit` crate updates to support it. I've sent https://github.com/nrf-rs/microbit/pull/123 to update that, but that in turn depends on https://github.com/nrf-rs/nrf-hal/pull/431 to update `nrf52833-hal`.

randomPoison:
Ah, that's good to know! I merged #1720 and #1721 thinking that the fact that CI passed meant it didn't break anything, but I'll roll those back to avoid breaking the embedded class in the meantime!

qwandor:
I don't think those actually broke anything, because the examples weren't using anything from embedded-hal directly but rather depending on the re-export from the HAL crate.

mgeisler (approved):
Looks good to me, thanks! I really like how you're consistently linking to the crates you mention.

PR #1724, opened 2024-01-18
Clarify usage for thiserror-and-anyhow

nyurik:
IMO it is better to explain to users the common usage / applicability domain for anyhow and thiserror crates


djmitche (changes requested):
The textual change looks good. However, the rust formatting is in third-party code that should not be reformatted. Can you remove that diff from the PR?

nyurik:
@djmitche thx, fixed

djmitche (approved):
Formatting fixed, so that should auto-merge. Thanks!

PR #1746, opened 2024-01-24
es: refresh translation for January 2024

mgeisler:
This is a clean refresh, requires only a syntaxtical review.


deavid (approved):
Please let's think if there's a way to avoid such large diffs in the future. Maybe we should refresh more often? or is this the effect of the new v2 guide that we reworked everything, so it's a one-off?

If it's a one off, it's okay. If this is because we're not refreshing the translations often enough, maybe we should think about this.

mgeisler:
> Please let's think if there's a way to avoid such large diffs in the future. Maybe we should refresh more often? or is this the effect of the new v2 guide that we reworked everything, so it's a one-off?

Good point! The v2 rewrite is responsible for most, but not all, of it. More frequent rewrites will help spread it out.

I should have included it above, but I use the [GitHub CLI](https://cli.github.com/) to skim such PRs:

```
gh pr diff 1746 | bat -l patch
```

That actually works unlike the web UI :smile:

mgeisler:
Another source of the large diff is the changes to the line numbers:


```diff
@@ -12,7 +12,7 @@ msgstr ""
 "Plural-Forms: nplurals=1; plural=0;\n"
 "X-Generator: Poedit 3.4\n"
 
-#: src/SUMMARY.md:4 src/index.md:1
+#: src/SUMMARY.md:3 src/index.md:1
 msgid "Welcome to Comprehensive Rust 🦀"
 msgstr "Te damos la bienvenida a Comprehensive Rust 🦀"
 
@@ -48,988 +48,1211 @@ msgstr "Ejemplos de Código"
 msgid "Running Cargo Locally"
 msgstr "Ejecutar Cargo a nivel local"
 
-#: src/SUMMARY.md:15
+#: src/SUMMARY.md:16
 msgid "Day 1: Morning"
 msgstr "Día 1: mañana"
 
-#: src/SUMMARY.md:19 src/SUMMARY.md:79 src/SUMMARY.md:134 src/SUMMARY.md:192
-#: src/SUMMARY.md:218 src/SUMMARY.md:268
+#: src/SUMMARY.md:18 src/SUMMARY.md:44 src/SUMMARY.md:70 src/SUMMARY.md:93
+#: src/SUMMARY.md:119 src/SUMMARY.md:138 src/SUMMARY.md:158 src/SUMMARY.md:184
+#: src/SUMMARY.md:207 src/SUMMARY.md:244 src/SUMMARY.md:286 src/SUMMARY.md:337
 msgid "Welcome"
 msgstr "Te damos la bienvenida"
```

I have done some work towards minimizing this (https://github.com/google/mdbook-i18n-helpers/pull/145), but we need to update our documentation to make people use this and we should also fix https://github.com/google/mdbook-i18n-helpers/issues/154.

deavid:
I looked at the diff on console, but still, it is very hard to prove/disprove that the PR does what it says. It is way too long, and things have moved.

Glad to see that you're working on making this better and this is mostly due to the v2 rewrite.

PR #1751, opened 2024-01-24
fa: refresh translation for January 2024

mgeisler:
This is a clean refresh, requires only a syntaxtical review.

```
gh pr diff 1751 | bat -l patch
```


po/fa.po
DannyRavi, line 1: I think it is better to use bidirectional text instead of HTML tags, which is available in utf-8.

https://en.wikipedia.org/wiki/Bidirectional_text

DannyRavi, line 1: Everything looks good, I think it is better to  use bidirectional text instead of HTML tags for RTL languages like Farsi. for more information :
[Bidirectional_text](https://en.wikipedia.org/wiki/Bidirectional_text)

DannyRavi (approved):

po/fa.po
DannyRavi: Hi @mgeisler.
 A solution for text that contains a mixture of letters RTL and LTR is to use [RLM](https://en.m.wikipedia.org/wiki/Right-to-left_mark) and [LRM](https://en.m.wikipedia.org/wiki/Left-to-right_mark) and [bidirectional](https://en.m.wikipedia.org/wiki/Bidirectional_text) text. With the help of hidden characters

PR #1752, opened 2024-01-24
ja: refresh translation for January 2024

mgeisler:
This is a clean refresh, requires only a syntactical review.

```
gh pr diff 1752 | bat -l patch
```

### DIFF
--- a/po/ja.po
+++ b/po/ja.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: Comprehensive Rust 🦀\n"
-"POT-Creation-Date: 2024-01-08T14:58:00+09:00\n"
+"POT-Creation-Date: 2024-01-24T13:24:49+01:00\n"
 "PO-Revision-Date: 2023-06-06 13:18+0900\n"
 "Last-Translator: Kenta Aratani <kentaaratani@coinez.jp>\n"
 "Language-Team: Japanese <translation-team-ja@lists.sourceforge.net>\n"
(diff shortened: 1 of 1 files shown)


kantasv:
Hi, Martin thanks for refreshing the ja.po file :)

kantasv:
Hi, I have skimmed through the changes; I can see that line numbers are removed, probably due to the recent change in `mdbook-i18n-helpers`, where granularity is set to zero. If that is intended, I guess it's LGTM! 

https://github.com/google/mdbook-i18n-helpers/pull/145

FYI: It seems it-translation has line numbers in the updated po files. https://github.com/google/comprehensive-rust/pull/1738

mgeisler:
> If that is intended, I guess it's LGTM!

Yes, that is on purpose — but you can totally let me know if you don't like this! I'm going to propose that we generate the files like that going forward: #1753. Please let me know on that PR if this is a good/bad idea :smile: 



> FYI: It seems it-translation has line numbers in the updated po files. #1738

Yes, you're a spot on! I only thought about removing the line numbers _after_ making all the PRs... so I updated each of them, but the Italian one had already been merged. It'll be updated at some point when @detro or someone else regenerates the file using the updated instructions from #1753.

mgeisler:
> Hi, Martin thanks for refreshing the ja.po file :)

No problem :smile: A big reason for doing this is to fix the broken links that exist between the published translations. Right now,

- https://google.github.io/comprehensive-rust/references/shared.html

links to

- https://google.github.io/comprehensive-rust/ko/references/shared.html
- https://google.github.io/comprehensive-rust/es/references/shared.html
- and several other languages in the :globe_with_meridians: drop-down menu

However, the Korean link is a 404 since that translation hasn't been updated yet. @henrif75 has been busy getting the Korean translation updated via our Localization team, so this will be fixed soon. Other translations are being fixed by these PRs.

Btw, if you think it would be helpful, then we could send the Japanese translation to the Localization team as well.

kantasv (approved):

kantasv:
Thanks, I approved it. Btw, @mgeisler , about the localization team, yes it would be very helpful if possible :)

henrif75:
Hi @kantasv I'll send the remaining content to the localization team.
When it's back I'll send the PRs for the ja.po CODEOWNERS for review.

PR #1756, opened 2024-01-24
Move Trait Objects to the Smart Pointers segment

### LINKED ISSUE #1516: "Trait Objects" require understanding of Box

djmitche:
Fixes #1516.


djmitche:
Verified via #1757 that trait objects are not used before the Smart Pointers slide.

mgeisler (approved):
Thanks for updating the redirects!

PR #1757, opened 2024-01-24
Add a 'course-content' binary

djmitche:
This allows dumping the course content in order, which I've needed several times to verify whether a concept is used before it is covered -- for example in #1516 I want to make sure we don't use trait objects before the Smart Pointers section.


mgeisler (approved):
Looks useful!

PR #1772, opened 2024-01-30
Actually protect against deadlock in Dining Philosophers async exercise

proski:
Swapping forks for one philosopher feels like cheating. It's like one of the philosophers is facing away from the table. Or perhaps it's the only right-handed philosopher at the table.

More importantly, there is no effective mechanism to prevent deadlocks. Add that mechanism, it's useful for learning Rust.

The new code demonstrates let-else, drop and returning values from a loop. `std::mem::swap` remains in the thread version of the Dining Philosophers exercise for now.

This also fixes compilation. `left_fork` and `right_fork` had to be `mut` in `main()` for the workaround to compile.

### DIFF
--- a/src/exercises/concurrency/dining-philosophers-async.rs
+++ b/src/exercises/concurrency/dining-philosophers-async.rs
@@ -41,13 +41,28 @@ impl Philosopher {
 
     // ANCHOR: Philosopher-eat
     async fn eat(&self) {
-        // Pick up forks...
+        // Keep trying until we have both forks
         // ANCHOR_END: Philosopher-eat
-        let _first_lock = self.left_fork.lock().await;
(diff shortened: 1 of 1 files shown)


djmitche (approved):
I would contest the notion that there's no way to remove deadlock -- enforcing a strict order on locks avoids deadlock, and that's precisely what the existing solution accomplished. The new solution is arguably worse, as it does not guarantee progress -- if the timing worked out _just wrong_ two threads could continue to loop, acquiring and releasing conflicting utensils and sleeping forever.

All of which probably makes for a nice topic of discussion while reviewing the exercise solution. I do like the use of `loop` with an expression value, and `let else`!

PR #1773, opened 2024-01-31
Added an Alternative way for printing struct and declaring a string.

mani-chand:
As a learner I want to learn every way of doing a work. 
Here,
Knowing to_string method is good for learner to use at some situation. 
Knowing print complete struct object can be useful while debug.


djmitche (changes requested):
Thank you!

In general, we try to introduce the string-formatting gradually in the context of other content. The debug format (`:?`) first appears in the [`Strings` slide](https://google.github.io/comprehensive-rust/types-and-values/strings.html). Instead of including that as a comment here, please add a sentence to the speaker notes on the `Strings` slide mentioning it.

The `to_string` method doesn't appear often in the course. It might make most sense to include it in the [`String` slide](https://google.github.io/comprehensive-rust/std-types/string.html) , again by adding a sentence to the speaker notes covering both `String::from` and `ToString::to_string`.

mani-chand:
Hello djmitche , Can you please check #1775 pull request.

mani-chand:
> Thank you!
> 
> In general, we try to introduce the string-formatting gradually in the context of other content. The debug format (`:?`) first appears in the [`Strings` slide](https://google.github.io/comprehensive-rust/types-and-values/strings.html). Instead of including that as a comment here, please add a sentence to the speaker notes on the `Strings` slide mentioning it.
> 
> The `to_string` method doesn't appear often in the course. It might make most sense to include it in the [`String` slide](https://google.github.io/comprehensive-rust/std-types/string.html) , again by adding a sentence to the speaker notes covering both `String::from` and `ToString::to_string`.

Hello djmitche , Did changes according to your request.

src/user-defined-types/named-structs.md
```diff
@@ -14,10 +14,13 @@ struct Person {
...
 fn describe(person: &Person) {
     println!("{} is {} years old", person.name, person.age);
+        // Or 
+    println!("{:?}",person) // This way you can print the complete struct object.
+     
 }
 
 fn main() {
-    let mut peter = Person { name: String::from("Peter"), age: 27 };
+    let mut peter = Person { name: String::from("Peter"), age: 27 }; // This can also be done by -> let mut peter = Person { name:"Peter".to_string(),age:27 }
```
djmitche: Please also remove these changes, as they are replaced with the suggested changes.

src/types-and-values/strings.md
```diff
@@ -23,34 +23,33 @@ fn main() {
     //println!("{:?}", &sentence[12..13]);
 }
 ```
-
```
djmitche, line 26: Please restore these deleted newlines.

src/types-and-values/strings.md
```diff
@@ -23,34 +23,33 @@ fn main() {
...
 - The `&` in `&str` indicates that this is a reference. We will cover references
   later, so for now just think of `&str` as a unit meaning "a read-only string".
-
+  ```rust,editable
+  let greeting = "Greetings" // -> This type of string declaration is by default comes under &str type.
+  let mut sentence:String = greeting; // Output : Error because `greeting` is &str type. 
+  // You can convert `&str` to `String` by `to_string()` function.
+  // i.e 
+  let mut sentence = greeting.to_string()
+  ```
```
djmitche, line 40: I think this is too early in the course to talk about `to_string` -- it would be better handled later, in the `String` slide in the standard-library types segment.

src/types-and-values/strings.md
```diff
@@ -23,34 +23,33 @@ fn main() {
...
-
   <!-- mdbook-xgettext: skip -->
   ```rust,editable
   fn main() {
       println!(r#"<a href="link.html">link</a>"#);
       println!("<a href=\"link.html\">link</a>");
   }
   ```
-
+- ``` println!("{:?}",person)``` -> This way you can print the complete struct object,array,vector or any type of large data.Here ```person``` can be array,struct or vector
```
djmitche, line 54: The last line of the example in the slide already contains an example:
```
    println!("{:?}", &sentence[0..5]);
```
so another one is not necessary here -- and anyway, `person` isn't defined so the example you have added is pretty confusing. I think this speaker note can just say that `{:?}` produces a debug-formatted result, with more detail than the `{}` form a few lines earlier. The key is to have instructors bring this to students' attention.

djmitche (changes requested):

mani-chand:
Hello, are you still adding more modules about rust like axum tutorial something like that. I would like to add the basics of axum(Rust backend framework) to this comprehension.

src/types-and-values/strings.md
mani-chand: Hello should I add this in String slide in standard-library types segment.

src/types-and-values/strings.md
mani-chand: I think this should also be added in standard-library types segment with printing array , struct , vector as an example.

djmitche: This isn't related to the String type -- it's about string formatting -- so I don't think it belongs in the standard-library types section.

djmitche:
We've added some specific courses after the fundamentals course, but they are on larger themes: Android, bare-metal, and concurrency.  I don't think a course on a specific framework would fit into that model very well -- in fact, with a few minor exceptions we have tried to avoid recommending specific crates.

mani-chand:
Hello , Do I have to change any thing. Let me know.

djmitche:
Please see my earlier comments. Each of these points should have 1 sentence (not a code sample) in the speaker notes, without unnecessary whitespace changes.

mani-chand:
Hello , Done changes without extra spaces.

src/std-types/string.md
```diff
@@ -54,5 +54,8 @@ all `str` methods on a `String`.
     out-of-bounds.
   - To a substring by using `s3[0..4]`, where that slice is on character
     boundaries or not.
+- let s4 = "Greetings"; // -> This type of string declaration is by default comes under &str type.
+- let mut s5:String = s4; // Output : Error because `greeting` is &str type.  
+- let mut s5:String = s4.to_string(); // This way you can convert `&str` to `String` by `to_string()` function.
```
djmitche, line 59: These aren't sentences :)

```suggestion
  - Many types can be converted to a string with the [`to_string`](https://doc.rust-lang.org/std/string/trait.ToString.html#tymethod.to_string) method. This trait is automatically implemented for all types that implement `Display`, so anything that can be formatted can also be converted to a string.
```

src/std-types/vec.md
```diff
@@ -50,6 +50,9 @@ methods on a `Vec`.
...
   remove the last element.
 - Slices are covered on day 3. For now, students only need to know that a value
-  of type `Vec` gives access to all of the documented slice methods, too.
+  of type `Vec` gives access to all of the documented slice methods, too. 
+- let mut v4 = Vec::new();    
+- v4.push(User{username:"Ron".to_string(),email:"ron@mail.com".to_string(),password:"******".to_string()}); // We can add as many - users we want in this vector.
+- println!("{:?}",v4)
 
-</details>
+</details>
```
djmitche, line 58: `to_string` isn't related to `Vec`, which is the topic of this slide. Please remove this diff.

djmitche (commented):
I've added a suggested change for `src/std-types/string.md`.

In earlier comments, I suggested adding information about the `{:?}` formatting directive to `src/types-and-values/strings.md`. That seems to have disappeared -- please add that back in?

Thank you for your hard work on this! I know multiple rounds of review can be frustrating -- it happens for my PRs, too! A good technique for avoiding this is to make a list of the changes a review has requested, and check those off one-by-one as you accomplish them. Or, if a suggestion doesn't make sense or you don't agree with it, ask a question about it or explain why you don't agree.

src/std-types/vec.md
mani-chand: I wanted to demonstrate usage of vector of structs in that view I used to_string method not to explain it.

mani-chand:
> I've added a suggested change for `src/std-types/string.md`.
> 
> In earlier comments, I suggested adding information about the `{:?}` formatting directive to `src/types-and-values/strings.md`. That seems to have disappeared -- please add that back in?
> 
> Thank you for your hard work on this! I know multiple rounds of review can be frustrating -- it happens for my PRs, too! A good technique for avoiding this is to make a list of the changes a review has requested, and check those off one-by-one as you accomplish them. Or, if a suggestion doesn't make sense or you don't agree with it, ask a question about it or explain why you don't agree.

Thanks for that, I know it's same hard for you foo.Did changes accordingly.

mani-chand:
Hello did according to the changes.

mani-chand:
Hello,If you me want to add usage of struct in vectors at (https://github.com/google/comprehensive-rust/blob/main/src/std-types/vec.md)[std-types vec.md].As a complete program.

mani-chand:
Hello , Done with all formatting. I Think everything looks goods. Should I need to do some thing.

mani-chand:
Hello, can you please confirm the changes.

mani-chand:
> We try to keep the code examples in slides very minimal, and typically without explanations in comments. I think both of the changes you've suggested should be included only in the speaker notes, so that instructors can mention them while teaching the slide.
> 
> Please try to eliminate other changes, including whitespace changes or changes to slide content.

Hello, Can you please review the request.

djmitche (changes requested):
Thanks for the revisions!

In the typical open-source contribution process, a contribution should be focused on one change. The reviewer makes comments on the proposed changes, and the author then either makes the suggested changes or offers a counterargument. The reviewer usually has a broader view; in this case, I've taught the class several times so I am familiar with the content and its pacing, and with how students engage with the material. 

If new ideas come up, those should be introduced in a different PR, so that the PR can continue to focus on the initial topic and eventually get merged or closed.  All of this applies whether the PR is for code in a programming language or plain text.

In this case, we began by talking about string formatting and `to_string`, and are now talking about `to_string` and vectors of structs. The latter should be proposed in a new PR (and, fair warning, that will need a good pedagogical justification for its inclusion, and especially for inclusion in a longer code sample, rather than mention in the speaker notes).

The `to_string` changes look good, but the conversation about string formatting is unresolved: it no longer appears in the "Files changed" view of this PR, but I didn't see any comments from you with a reason to remove it.

So, changes requested:
 * Remove the changes to `vec.md`, and if desired make a new PR containing those changes and a reason to include them in the course.
 * Update the speaker notes in `src/types-and-values/strings.md` to cover `{:?}`, or offer a reason for dropping it from this PR.

mani-chand:
Is formatting and changes are good.

djmitche (approved):
Looks good! I ran `dprint fmt` to fix the formatting, and will merge once that's complete.

Thanks for your work on this!

mani-chand:
Thanks for your support.

PR #1774, opened 2024-01-31
pt-BR: add myself to CODEOWNERS

jcvicelli:


mgeisler:
Hi @jcvicelli, thanks so much for wanting to help out! I would encourage you to say hello on #317 so you can connect with the other translators.

There are a lot of PRs to review and there will be lots of small and large typos to correct, so we can use all the help we get :smile:

mgeisler (approved):

PR #1775, opened 2024-01-31
Added an Alternative way for printing struct and declaring a string

mani-chand:
### This pull request is #1773 pull request update
As a learner I want to learn every way of doing a work.
Here,

- Knowing to_string method is good for learner to use at some situation.
- Knowing print complete struct object can be useful while debug.

### DIFF
--- a/src/std-types/string.md
+++ b/src/std-types/string.md
@@ -2,57 +2,63 @@
 minutes: 10
 ---
 
-# String
+# Strings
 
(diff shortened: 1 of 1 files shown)


djmitche (commented):
I'm not sure what's happened here -- it looks like you've copied `strings.md` (early introduction to strings) to `string.md` (about the `String` type).

Also, please update the existing PR, rather than creating a new one, so we can better track the conversation.

Issue #1464, opened 2023-11-10
Size and style of slides in v2


mgeisler:
# Large Slides

Many slides are now bigger because they include more details.

| Before | After |
| ------ | ----- |
| ![image](https://github.com/google/comprehensive-rust/assets/89623/de0c27c3-36e0-482c-a730-0150ee3c4b1c) | ![image](https://github.com/google/comprehensive-rust/assets/89623/6169c3cf-78e9-4fe1-9642-e01918c70de5) |

What is the strange red rectangle? It's the "aspect ratio helper" which you can enable in `book.toml`. From a bit of experimentation in a meeting room, this rectangle shows roughly how much content we can fit on a screen.

The existing slides don't always fit! They have gained weight over time, as more and more facts and details have been added. That's very relatable and a comes from a good place: people want to cram in as much great information as possible :slightly_smiling_face: 

An example from the new course is the pages on structs. It used to have 21 lines of code and it now has 30 lines. It didn't fit super well before, and now it's really too large:

| Before | After |
| ------ | ----- |
| ![image](https://github.com/google/comprehensive-rust/assets/89623/2decc374-edf2-4137-adbf-960b932f330e) | ![image](https://github.com/google/comprehensive-rust/assets/89623/30f8cb56-ec68-4d16-83d5-9950f7b82608) |

My impression is that less than 10 lines of code is best. 15 lines can work, but then you only have room for 1-2 lines of output!

# Combined Slides

Some slides in the new course combine content what was shown on separate slides before. Compare

| Before | After |
| ------ | ----- |
| ![image](https://github.com/google/comprehensive-rust/assets/89623/d4583d0b-337a-4d72-a90d-1d30181de07d) | ![image](https://github.com/google/comprehensive-rust/assets/89623/4dd8aa90-a86b-4e6e-b5de-cf6529f08287) |
| ![image](https://github.com/google/comprehensive-rust/assets/89623/b9e9f302-5b90-4633-b18e-55f3eb75e965) | |

Funny enough, this style was actually how I wrote the very first version of the course :smile: I thought I could have ~10 long-form pages, each covering a bigger theme. I was then going to scroll down on the page during class.

However, the feedback from the initial review was that this would be hard to follow in class. Later, I've seen people struggle to follow along on the current slides that are too big. Scrolling up and down on a large screen in projected in front of 20 people is bad experience.

[...]

PR #1787, opened 2024-02-06
Added a different way of vector intialization.

mani-chand:
I am taking example datatype as String you can intialize with any datatype that is supported by rust.
Specifying ```Vec::<String>::new()``` explicitly indicates that variable is intended to be a vector containing String elements. This can be helpful for documenting the code and ensuring that variable is used correctly throughout the codebase.

### DIFF
--- a/src/std-types/vec.md
+++ b/src/std-types/vec.md
@@ -20,6 +20,9 @@ fn main() {
     // Canonical macro to initialize a vector with elements.
     let mut v3 = vec![0, 0, 1, 2, 3, 4];
 
+    // specifying the type when initializing a vector.
+    let mut v4 = Vec::<String>::new();
+
     // Retain only the even elements.
     v3.retain(|x| x % 2 == 0);
     println!("{v3:?}");


djmitche (changes requested):
I'm not sure this is adding anything to the course. In general, changes to the text should come with some rationale: either missed content, out-of-order content, or observed student confusion.

src/std-types/vec.md
```diff
@@ -20,6 +20,9 @@ fn main() {
     // Canonical macro to initialize a vector with elements.
     let mut v3 = vec![0, 0, 1, 2, 3, 4];
 
+    // specifying the type when initializing a vector.
+    let mut v4 = Vec::<String>::new();
```
djmitche, line 24: A more idiomatic way to do this would be
```rust
    let mut v4: Vec<String> = Vec::new();
```
which was covered in the type-inference slide on day 1.

The turbofish operator (`::<String>`) is introduced in the speaker notes in the impl trait slide in the "Generics" segment and further refined in the sections on FromIterator. I think that's enough coverage for the moment.

djmitche (changes requested):
Oops, forgot to click "Save" on the comment :)

djmitche:
@mani-chand what do you think about doing some work on #1464? A PR fixing the sizes of the slides in a single segment would be a good place to start. This is something really valuable to students, as they can see the course content on the screen without scrolling.

I noticed yesterday while making some other changes that the "Control Flow Basics" segment has some very long slides.

If you're interested, make a comment in that PR and I'll assign it.

mani-chand:
> @mani-chand what do you think about doing some work on #1464? A PR fixing the sizes of the slides in a single segment would be a good place to start. This is something really valuable to students, as they can see the course content on the screen without scrolling.
> 
> I noticed yesterday while making some other changes that the "Control Flow Basics" segment has some very long slides.
> 
> If you're interested, make a comment in that issue and I'll assign it.

Sure I am interested in working on it.

src/std-types/vec.md
mani-chand: > A more idiomatic way to do this would be
> 
> ```rust
>     let mut v4: Vec<String> = Vec::new();
> ```
> 
> which was covered in the type-inference slide on day 1.
> 
> The turbofish operator (`::<String>`) is introduced in the speaker notes in the impl trait slide in the "Generics" segment and further refined in the sections on FromIterator. I think that's enough coverage for the moment.

Oh! That's my bad i haven't  noticed it. Then lets just close the PR.

mani-chand: What do you say.

djmitche: Sounds good!

PR #1793, opened 2024-02-08
Added break-continue section.

mani-chand:
This is a contribution of a break-continue section for Comprehensive Rust.

image
image


mani-chand:
Hello @djmitche , Can you please look into this. What do you say ? is it good.

djmitche (commented):
It looks like you've broken this up into `break` and `continue`. I wonder if a different way to break it up would be a simple example that uses both keywords, and a more complex example on the second sub-slide illustrating that `loop` can have a label and a value?

At any rate, at least one of the examples should use `continue` -- can you think of a good example?

djmitche:
(To clarify, "I wonder if.." means it's OK if you don't think that's a good idea -- just explain why)

src/control-flow-basics/break-continue/continue.md
```diff
@@ -0,0 +1,29 @@
...
+        }
+        if a < b {
+            b -= a;
+        } else {
+            a -= b;
+        }
+    };
+    println!("{result}");
+}
+```
```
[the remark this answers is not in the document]
mani-chand: I thought the same but went on dividing the code. If you allow me I write an example for continue on continue slide.

mani-chand:
> (To clarify, "I wonder if.." means it's OK if you don't think that's a good idea -- just explain why)

I will think of some example for continue and commit it.

mani-chand:
> It looks like you've broken this up into `break` and `continue`. I wonder if a different way to break it up would be a simple example that uses both keywords, and a more complex example on the second sub-slide illustrating that `loop` can have a label and a value?
> 
> At any rate, at least one of the examples should use `continue` -- can you think of a good example?

I will add my own example on continue slide.

mani-chand:
Any change needed ? let me know.

djmitche (commented):
This is looking good! However, I think the first sub-slide is actually more complex than the second. Let's switch the order, with `continue` first and `break` (along with labels and `loop` taking a value) second.

src/control-flow-basics/break-continue.md
```diff
@@ -4,52 +4,28 @@ minutes: 4
...
-        } else {
-            a -= b;
-        }
-    };
-    println!("{result}");
-}
-```
-
 Both `continue` and `break` can optionally take a label argument which is used
 to break out of nested loops:
```
djmitche, line 11: This sentence doesn't belong here anymore, as this example doesn't have labels.

djmitche (commented):
As the PR stands now, both sub-slides use both `break` and `continue`. I think that's actually fine, though - the first slide can be the basics, and the second sub-slide the more advanced stuff.

* `src/control-flow-basics/break-continue.md` can introduce both keywords with the simple example. So the description of `break` should move to that slide too ("If you want to exit..")
* The sub-slide, maybe renamed to `src/control-flow-basics/break-continue/labels.md`, can introduce labels and loops with a value. I think it's fine to do both of these things in one example, but the existing example doesn't actually demonstrate loops with a value. So, add a value to the `break` statement in that example.

mani-chand:
Any change needed ? let me know.

src/SUMMARY.md
```diff
@@ -35,6 +35,7 @@
     - [`for`](control-flow-basics/loops/for.md)
     - [`loop`](control-flow-basics/loops/loop.md)
   - [`break` and `continue`](control-flow-basics/break-continue.md)
+    - [`labels`](control-flow-basics/break-continue/labels.md)
```
djmitche, line 38: The backticks are just for Rust keywords, which "label" is not, so:
```suggestion
    - [Labels](control-flow-basics/break-continue/labels.md)
```

src/control-flow-basics/break-continue.md
```diff
@@ -4,52 +4,30 @@ minutes: 4
...
+            continue;
         }
+        println!("{}", i);
     }
 }
 ```
 
-In this case we break the outer loop after 3 iterations of the inner loop.
-
 <details>
```
djmitche, line 31: This speaker note now corresponds to the second sub-slide ("Labels").

djmitche (commented):
Almost there :)

src/SUMMARY.md
djmitche: Oh, and plz change this on the slide itself, too.

mani-chand:
sorry, my bad I did it again.

src/control-flow-basics/break-continue/labels.md
```diff
@@ -0,0 +1,35 @@
...
+fn main() {
+    let (mut a, b) = (52, 100);
+    let result = loop {
+        if a > b {
+            break a;
+        } else {
+            a += 20;
+        }
+        let mut i = 0;
+        'outer: for _x in 1..5 {
```
djmitche, line 16: This actually had me scratching my head for a few minutes -- the label says `'outer` but it's the _inner_ loop! Also, without a label `break` will break the innermost loop, so this doesn't actually show the power of labels. Can you update the example so that the outer loop is the one labeled? I think that will require breaking with a value in that case, too.

src/control-flow-basics/break-continue/labels.md
```diff
@@ -0,0 +1,35 @@
+# labels
```
djmitche, line 1:
```suggestion
# Labels
```

djmitche (commented):
Looking good, just need to tighten up the example a little bit..

mani-chand:
Hello @djmitche , It looks good, I think we can go with this. What do you say.

djmitche (commented):
This does run, but has some warnings, and in particular `result` is never used.

```
warning: unused variable: `j`
 --> src/main.rs:4:22
  |
4 |         '_inner: for j in 1..=200 {
  |                      ^ help: if this is intentional, prefix it with an underscore: `_j`
  |
  = note: `#[warn(unused_variables)]` on by default

warning: unused variable: `result`
 --> src/main.rs:8:17
  |
8 |             let result = loop {
  |                 ^^^^^^ help: if this is intentional, prefix it with an underscore: `_result`

warning: variable does not need to be mutable
 --> src/main.rs:2:10
  |
2 |     let (mut a, mut b) = (1, 1);
  |          ----^
  |          |
  |          help: remove this `mut`
  |
  = note: `#[warn(unused_mut)]` on by default

warning: `playground` (bin "playground") generated 3 warnings (run `cargo fix --bin "playground"` to apply 3 suggestions)
    Finished dev [unoptimized + debuginfo] target(s) in 0.63s
     Running `target/debug/playground`
```

I think there's a chance to make a more compelling example here, that actually sort-of calculates something. That might take some creativity!

mani-chand:
Hello @djmitche , Changed code , Can you please review the PR and suggest changes.

mani-chand:
hello @djmitche , I have added a code with some calculation . Its finds if a n*n matrix 1st row as the target value or not.

mani-chand:
hello @djmitche , I have added a code with some calculation . Its finds if a n*n matrix 1st row as the target value or not.

src/control-flow-basics/break-continue/labels.md
```diff
@@ -0,0 +1,34 @@
...
+            }
+            if s[i][j] == target_value_in_1st_row {
+                result = true;
+                break result;
+            }
+            j += 1;
+        };
+    }
+    print!("{}",result)
+}
```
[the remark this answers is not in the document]
mani-chand: > Hm, this is a good illustration of use of labels, but not of breaking with a value -- `break result` just ends up assigning `result` to `result`. I like the two-dimensional array though! We haven't covered arrays yet, but students should be familiar enough to understand this example. What if we simplified it a bit to just search for a value in that array? I think this is a pretty common use of break with labels.
> 
> ```
> fn main() {
>     let s = [[5, 6, 7], [8, 9, 10], [21, 15, 32]];
>     let mut found = false;
>     let target_value = 10;
>     'outer: for i in 0..=2 {
>         for j in 0..=2 {
>             if s[i][j] == target_value {
>                 found = true;
>                 break 'outer;
>             }
>         }
>     }
>     print!("{}", found)
> }
> ```
> 
> We can leave the bit about loop having a value just mentioned in the speaker notes. It's unusual to see it used in Rust code, so maybe students don't need to see it on day 1!

Hey,I have not seen your comment and requested for review.

mani-chand:
Hello @djmitche , I have updated the code .Sorry for requesting for review Before seeing your comment .

mani-chand:
I haven't got you properly but i think this what you wanted.

mani-chand:
If it is okay , Can you give a quick feedback or review like rate me on a scale of 10 kind of thing . Because I want to know like how mush good i am at this.

I felt this PR took bit a long.I don't it was my mistake.

djmitche (approved):
Looks good! Thanks again for sticking with this!

PR #1795, opened 2024-02-08
Update lifetime-elision.md

IP1llar:
In the details section, it is suggested to adjust the signature of the nearest function to "lie" about the lifetimes returned to:

fn nearest<'a, 'q'>(points: &'a [Point], query: &'q Point) -> Option<&'q Point> {

to demonstrate the compiler checks lifetimes for validity. However, the syntax for 'q is incorrect and it should instead be

fn nearest<'a, 'q>(points: &'a [Point], query: &'q Point) -> Option<&'q Point> {

### DIFF
--- a/src/slices-and-lifetimes/lifetime-elision.md
+++ b/src/slices-and-lifetimes/lifetime-elision.md
@@ -61,7 +61,7 @@ references in its arguments that requires explicit annotation.
 Try adjusting the signature to "lie" about the lifetimes returned:
 
 ```rust,ignore
-fn nearest<'a, 'q'>(points: &'a [Point], query: &'q Point) -> Option<&'q Point> {
+fn nearest<'a, 'q>(points: &'a [Point], query: &'q Point) -> Option<&'q Point> {
 ```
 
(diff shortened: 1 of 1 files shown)


djmitche (approved):
Thanks! I blame someone's "helpful" editor :)

src/slices-and-lifetimes/lifetime-elision.md
```diff
@@ -61,7 +61,7 @@ references in its arguments that requires explicit annotation.
 Try adjusting the signature to "lie" about the lifetimes returned:
 
 ```rust,ignore
```
mgeisler, line 63: We could perhaps take out the `ignore` part here — and fill the body with a `todo!()`?

That should have prevented this kind of tiny typo from occurring. Alternatively, we could put the Rust code in a separate file and include just the line needed to demonstrate the point.

Just some food for thought in case @IP1llar has some free time and want to send us more PRs :smile:

mgeisler:
Thanks a lot @IP1llar for fixing this! We have so many pages now that we're super dependent on folks like you to help us fix typos like this.

PR #1797, opened 2024-02-08
Improved support for commenting when course timings change.

djmitche:
This
 * Works with the GitHub permissions model (at least in local testing on another temporary repo)
 * Only comments when the timings actually change

### DIFF
--- a/.github/workflows/course-schedule-comment.yml
+++ b/.github/workflows/course-schedule-comment.yml
@@ -23,10 +23,11 @@ jobs:
 
       - name: "Generate Schedule on upstream branch"
         run: |
-          mkdir -p ./course-schedule
           cargo run -p mdbook-course --bin course-schedule > upstream-schedule
 
       - name: "Download artifact from PR workflow"
+        # actions/download-artifact@v4 cannot do this without being given a PAT, although that
+        # is not required for public forked repositories.
         uses: actions/github-script@v7.0.1
         with:
           script: |
@@ -56,13 +57,14 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
             var fs = require('fs');
+            var pr_number = Number(fs.readFileSync('pr-number'));
             var upstream = fs.readFileSync('upstream-schedule').toString();
             var schedule = fs.readFileSync('schedule').toString();
             if (upstream != schedule) {
               await github.rest.issues.createComment({
                 owner: context.repo.owner,
                 repo: context.repo.repo,
-                issue_number: context.payload.workflow_run.pull_requests[0].number,
+                issue_number: pr_number,
                 body: schedule,
               });
             }
--- a/.github/workflows/course-schedule-comment.yml
+++ b/.github/workflows/course-schedule-comment.yml
@@ -1,0 +1,68 @@
+# Based on https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
+#
+# This runs with the full repo permissions, but checks out the upstream branch, not the PR.
+# Do not run arbitrary code from the PR here!
+name: Comment PR with Course Schedule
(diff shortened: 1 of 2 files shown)


.github/workflows/course-schedule-comment.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      - name: "Setup Rust cache"
+        uses: ./.github/workflows/setup-rust-cache
+
+      - name: "Generate Schedule on upstream branch"
+        run: |
+          mkdir -p ./course-schedule
+          cargo run -p mdbook-course --bin course-schedule > upstream-schedule
+
+      - name: "Download artifact from PR workflow"
```
kdarkhan, line 29: Nit: I think this can be simplified by using [download-artifact](https://github.com/actions/download-artifact) action. When I was implementing my workflow, I wasn't aware of this.

.github/workflows/course-schedule-comment.yml
```diff
@@ -0,0 +1,68 @@
...
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            var fs = require('fs');
+            var upstream = fs.readFileSync('upstream-schedule').toString();
+            var schedule = fs.readFileSync('schedule').toString();
+            if (upstream != schedule) {
+              await github.rest.issues.createComment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: context.payload.workflow_run.pull_requests[0].number,
```
kdarkhan, line 65: This might be flaky as I discovered. In `mdbook-i18n-helpers` I fixed it by https://github.com/google/mdbook-i18n-helpers/pull/168

According to my testing, the root cause of the issue is that `context.payload.workflow_run.pull_requests` array is not always populated. Some discussion on this issue https://github.com/orgs/community/discussions/25220

.github/workflows/course-schedule-comment.yml
djmitche: I'll give it a try! The linked article seemed to say that wouldn't work, but on a closer reading maybe that's not true. Also, it's quite an old article so maybe it works now :)

kdarkhan: I haven't tried it, so it might indeed be not working :)

.github/workflows/course-schedule-comment.yml
kdarkhan: Yes, debugging them is very cumbersome. It is possible to create a PR against your own fork though. That's what I did in my fork https://github.com/kdarkhan/mdbook-i18n-helpers/pulls?q=is%3Apr+is%3Aclosed

.github/workflows/course-schedule-comment.yml
djmitche: Ah, it requires a PAT with read access to the source repository, and won't accept the `repository` input parameter without one -- even if that repository might be public.

kdarkhan: I am ok with not implementing it as well if it is not easy.

djmitche: (It's possible that the script I've used _also_ won't work with a different, even public, repository -- we'll see!)

.github/workflows/course-schedule-comment.yml
djmitche: Yeah, that's how I was testing -- but I think it doesn't accurately simulate the permissions of a cross-owner PR.

kdarkhan:
I don't have approval permissions for this repo but LGTM.

qwandor (approved):

mgeisler:
> I don't have approval permissions for this repo but LGTM.

Let's fix this, I've invited you now!

PR #1813, opened 2024-02-12
Testing timing comments

djmitche:

PR #1814, opened 2024-02-12
Put a header on schedule comments

djmitche:
This will add a little more context. See the current result in https://github.com/google/comprehensive-rust/pull/1813.


mgeisler (approved):
Looks good to me!

Issue #1511, opened 2023-11-29
"Traits" should include more info


djmitche:
* Associated Types
* Trait Parameters
* Trait "inheritance"

mgeisler:
> * Associated Types

Yes! This always comes up because `Iterator` has one.

> * Trait Parameters
> * Trait "inheritance"

Less so, but I would be happy to cover it in the notes.

PR #1815, opened 2024-02-13
Expand Traits, add associated types

djmitche:
This breaks the "Traits" slide into three smaller sub-slides. It also addresses part of #1511 by explicitly addressing associated types.

### DIFF
--- a/src/methods-and-traits/traits/implementing.md
+++ b/src/methods-and-traits/traits/implementing.md
@@ -28,7 +28,11 @@ fn main() {
 
 <details>
 
-- Traits are implemented in an `impl <trait> for <type> { .. }` block.
+- Types must implement the trait with an `impl <trait> for <type> { .. }` block.
+
+- Unlike Go interfaces, just having matching methods is not enough: a `Cat` type
+  with a `talk()` method would not automatically satisfy `Pet` unless it is in
+  an `impl Pet` block.
 
 - Traits may specify pre-implemented (provided) methods and methods that users
   are required to implement themselves. Provided methods can rely on required
--- a/src/methods-and-traits/traits/implementing.md
+++ b/src/methods-and-traits/traits/implementing.md
@@ -28,14 +28,15 @@ fn main() {
 
 <details>
 
-- To implement `Trait` for `Type`, you use an `impl Trait for Type { .. }` block.
+- To implement `Trait` for `Type`, you use an `impl Trait for Type { .. }`
+  block.
 
 - Unlike Go interfaces, just having matching methods is not enough: a `Cat` type
   with a `talk()` method would not automatically satisfy `Pet` unless it is in
   an `impl Pet` block.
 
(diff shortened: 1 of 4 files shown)


djmitche:
@mani-chand, please review, as this is similar to work you've been doing?

src/methods-and-traits/traits/implementing.md
mani-chand, line 1: image

Looks good . I think we can omit this point because, I think the example above it self is explains the syntax. As, everyone learns rust after some programming language.

You can add point like from above only `Dog` struct can use this `Pet trait` no other struct can use this something like this. What do you say .

mani-chand:
Hey @djmitche , You can merge with the change, I suggested. What do you think.

src/methods-and-traits/traits/implementing.md
djmitche: That's a good point - this is different from Go "interfaces" where just implementing the right methods is enough.

mani-chand: Sounds good. okay then, You can merge the  PR. That all, I had.

mani-chand: Ohh, appologies for my previous deleted comment, I misread it.

djmitche:
Thanks for the review! Looks like I still need someone on the official reviewers list, too :)

src/methods-and-traits/traits/implementing.md
```diff
@@ -0,0 +1,41 @@
...
+
+fn main() {
+    let fido = Dog { name: String::from("Fido"), age: 5 };
+    fido.greet();
+}
+```
+
+<details>
+
+- Types must implement the trait with an `impl <trait> for <type> { .. }` block.
```
mgeisler, line 31: I'm not a fan of the mix between real syntax and pseudo-code, especially when Rust uses `<...>` for real things :smile: 

What about:

```suggestion
- To implement `Trait` for `Type`, you use an `impl Trait for Type { .. }` block.
```

mgeisler (approved):

src/methods-and-traits/traits/implementing.md
```diff
@@ -0,0 +1,41 @@
...
+
+- Unlike Go interfaces, just having matching methods is not enough: a `Cat` type
+  with a `talk()` method would not automatically satisfy `Pet` unless it is in
+  an `impl Pet` block.
+
+- Traits may specify pre-implemented (provided) methods and methods that users
+  are required to implement themselves. Provided methods can rely on required
+  methods. In this case, `greet` is provided, and relies on `talk`.
```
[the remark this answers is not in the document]
djmitche: TIL! I wonder where I picked up the term "provided"..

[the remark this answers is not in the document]
mgeisler: Sorry that the wrapping is off (most likely).

PR #1816, opened 2024-02-14
Make content of variable "name" an actual name

errge:
In the example, somebody who is trying to understand this code has to follow a lot of moving pieces: prefix, name, make_greeter, and then also learn the new move concept at the same time.

It'd be better if the content were to actually match the name of the variables.  "Hi" is a good prefix, but "there" is not a name, so let's go with an actual name.


djmitche (approved):
Good fix, thanks!

PR #1817, opened 2024-02-14
Change String representation to actual

errge:
Tested with rust 2015, 2018, 2021 and 2024, on amd64 musl, amd64 glibc and aarch64 musl, all of them represent Strings with (ptr, capacity, len).

This is an internal implementation detail, that shouldn't be exposed anyway, so it's no big deal, but in the speaker notes, we provide a debugging tool for demonstration purposes, so at least that should have a correct output.

### DIFF
--- a/src/memory-management/review.md
+++ b/src/memory-management/review.md
@@ -29,17 +29,17 @@ fn main() {
 ```
 
 ```bob
- Stack                             Heap
-.- - - - - - - - - - - - - -.     .- - - - - - - - - - - - - - - -.
-:                           :     :                               :
+ Stack
+.- - - - - - - - - - - - - -.      Heap
+:                           :     .- - - - - - - - - - - - - - - -.
 :    s1                     :     :                               :
-:   +-----------+-------+   :     :   +----+----+----+----+----+  :
-:   | ptr       |   o---+---+-----+-->| H  | e  | l  | l  | o  |  :
-:   | len       |     5 |   :     :   +----+----+----+----+----+  :
-:   | capacity  |     5 |   :     :                               :
 :   +-----------+-------+   :     :                               :
-:                           :     `- - - - - - - - - - - - - - - -'
-`- - - - - - - - - - - - - -'
+:   | capacity  |     5 |   :     :   +----+----+----+----+----+  :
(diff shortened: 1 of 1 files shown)


errge:
Tested resulting website with mdbook serve, modified diagram renders correctly.

errge:
Note: `src/memory-management/move.md` and `src/smart-pointers/trait-objects.md` also has this order in diagrams, I plan to fix them, when I fully reviewed those chapters of the slides/course.

I prefer to put those in separate pull requests, as here, the important fix is the debugging Rust code, because that was just incorrect.

src/memory-management/review.md
```diff
@@ -65,8 +65,8 @@ fn main() {
     // String provides no guarantees about its layout, so this could lead to
     // undefined behavior.
     unsafe {
-        let (ptr, capacity, len): (usize, usize, usize) = std::mem::transmute(s1);
-        println!("ptr = {ptr:#x}, len = {len}, capacity = {capacity}");
+        let (capacity, ptr, len): (usize, usize, usize) = std::mem::transmute(s1);
+        println!("capacity = {capacity}, ptr = {ptr:#x}, len = {len}");
```
djmitche, line 69: I wonder if it would be better to use [`String::into_raw_parts](https://doc.rust-lang.org/std/string/struct.String.html#method.into_raw_parts)` here? What do you think?

djmitche (approved):
Nice fix, thanks! I agree this is a detail, but still best to get it right.

src/memory-management/review.md
errge: Hmm, thanks for making me aware of this, I'm just learning rust, so I didn't know about it.

It's cool, that the string library has a feature for this, and it's a cool alternative.

I wouldn't go for it for two reasons:

  - weak reason: it's still unstable feature, so you have to enable it, otherwise it's a compile error
  - stronger reason: the currently demonstrated method works with any other type that one knows the representation for, so it has a bit more of an educational value, and there is ample warning there to say that this is for debugging, not for production

djmitche: Both good points!

PR #1819, opened 2024-02-15
Reformulate speaker notes regarding Box

errge:
The first change is to reformulate the English in a way, that emphasizes, that this is not a decision of the compiler, but the impossibility of computing an infinite value (e.g. changed the language from "not compute" to "would not be able to compute").

The second change is to fix the error message, of course the error message from the compiler is "recursive withOUT indirection", as "recursive with indirection" is actually what we want.

### DIFF
--- a/src/smart-pointers/box.md
+++ b/src/smart-pointers/box.md
@@ -74,15 +74,16 @@ fn main() {
     so only the pointer is moved.
 
 - If `Box` was not used and we attempted to embed a `List` directly into the
-  `List`, the compiler would not compute a fixed size of the struct in memory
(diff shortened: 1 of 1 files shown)


errge:
Happy to discuss these changes, but tried to put the new formulations in a way, in which I would have understood them faster when I first started to think what this section is trying to say.

djmitche (approved):

djmitche:
Can you run `dprint fmt` on this?

errge:
Thanks for the pointer @djmitche , stuff is passing tests now.

PR #1825, opened 2024-02-17
Fix style in glossary.md

mobyw:
`id` of `h1` title in the translated version will be changed to the translated text. 
Fixed `id` configured here will not take effect in the translated version.

Correct style: <https://google.github.io/comprehensive-rust/glossary.html>
Wrong style (zh-CN): <https://google.github.io/comprehensive-rust/zh-CN/glossary.html>

### DIFF
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -5,19 +5,19 @@ terms. For translations, this also serves to connect the term back to the
 English original.
 
 <style>
-h1#glossary ~ ul {
+h1 ~ ul {
     list-style: none;
     padding-inline-start: 0;
 }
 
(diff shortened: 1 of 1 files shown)


mgeisler (approved):
Hi @mobyw, wow, thanks for finding this!

I think this should work fine — though I was hoping to use a more specific match than the `h1 ~ ul`. Could you check if we can use [Heading attributes](https://rust-lang.github.io/mdBook/format/markdown.html#heading-attributes) to hard-code the heading for the translations?

Let's merge this now and then we can follow up with another PR later, if possible.

Issue #1476, opened 2023-11-13
Keep state of the Rust Playgrounds when navigating between slides


mgeisler:
Today, the embedded Playgrounds are reset when you navigate between slides. This has caused problems: if people navigate away from a slide to look something up, they've suddenly lost their work.

This is actually the reason why we don't make the exercises editable directly in the slides: too many people have lost work because of this.

We should be able to store the current state in the browser's local storage.

jnovikov:
I would like to help with it.

mgeisler:
> I would like to help with it.

Hi @jnovikov, that would be awesome! I hope the idea is somewhat clear, otherwise please let me know :smile: The problem right now is that using Arrow-Left or Arrow-Right (or clicking links) navigates away from the current slide: when you navigate back, the playground is back at its initial state. The idea is to prevent this by storing the state in [local storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage).

I don't know much about the editor here, except that `mdbook` uses this one: https://ace.c9.io/.

mani-chand:
Hey @mgeisler , I would like to give a try .If it Is still open.

djmitche:
It's all yours! I'm not sure anyone has in mind a good way to do this, so the first step is coming up with an idea..

mani-chand:
> It's all yours! I'm not sure anyone has in mind a good way to do this, so the first step is coming up with an idea..

I looked at it there is a `div` which as has some child html elements in them all our code lies. 

I will store all children's of that `parent div`
in an array using `Arraysfrom(document.getElementsByClassName('{that parent div class name}')[0].children)`.

Then I will make a object with the URL as key and array of html elements as value. 

I will store that array object (Assuming user make changes in all playgrounds) in browser local storage. 

Every time loads the page, I will check the object is that url present in that key. 

If present then I will display object value code. 
 
else I will start assigning that code to new key in the object. 


Process of displaying code : 
 Delete all children of parent div 
 add all childrens in object one by one using loop.

djmitche:
That sounds like it will work -- try it out!

mani-chand:
> > It's all yours! I'm not sure anyone has in mind a good way to do this, so the first step is coming up with an idea..
> 
> I looked at it there is a `div` which as has some child html elements in them  all our code lies as innerHTML/innerText.
> 
> I will store all children's of that `parent div`
[...]

Issue #1289, opened 2023-10-02
Update exercise boiler plate text to suggest working on the code locally


mgeisler:
As pointed out in https://github.com/google/comprehensive-rust/pull/1288#issuecomment-1743131185, we're giving mixed signals about how the exercises should be solved: in the Rust Playground or locally with Cargo?

This is mostly a remnant of how the course has evolved: it started out being very focused on the Playground, but later we started recommending a more regular Rust setup.

We should update the boiler plate text to be more consistent and to encourage people to use a local installation.

To streamline things, we could include the same text (using the mdbook `{{#include }}` syntax) from all the exercises and thus ensure everything is consistent.

professorabhay:
Hey, @mgeisler! Does that still need to be resolved? 
I would like to hands on this. 
Kindly provide some other details about that.

mgeisler:
Hi @professorabhay! Sorry for the slow reply... yes, I still think this is relevant.

Compare [Exercise: Health Statistics](https://google.github.io/comprehensive-rust/borrowing/exercise.html) and [Exercise: Builder Type](https://google.github.io/comprehensive-rust/memory-management/exercise.html).

The first exercise says

> Copy the code below to https://play.rust-lang.org/ and fill in the missing method:

but the second exercise says

> Fill in the missing pieces.

It would be nice to standardize this. It would also be good to more clearly guide the course participants towards using a local Rust installation. That should probably be a VS Code + Rustanalyzer setup.

I've been punting on this since I use Emacs myself and since I didn't want to tell people what to use :smile: But it would be good for people to use a local installation since they get a more full development environment.

Updating the course to use consistent text would be a good first step, so I suggest starting with that.

professorabhay:
No problem @mgeisler 😃
Btw, I get your point and will propose changes in the PR asap.

professorabhay:
![image](https://github.com/google/comprehensive-rust/assets/76243309/c2fb9b10-2cce-4313-962e-a16d19b8c7c4)

What do you think about that? @mgeisler

mgeisler:
Hi @professorabhay! Thanks for pinging me, I've been overwhelmed with PRs and comments on them here... I guess that's a good problem to have :smile: 

I like your text about VS Code and rust-nalyzer a lot, well done!

I would suggest adding it just once. You can add it to [Using Cargo](https://google.github.io/comprehensive-rust/cargo.html) under the Installation section.

[...]

PR #1827, opened 2024-02-18
Add missing editable attribute to exercise

titong0:
I was reading the docs and I stumbled upon this.
I'm unsure about when exercises should be solved directly on the site and when they should be copied and pasted into a playground. But since the previous chapter's exercise is solvable in the site [(fibonacci sequence)](https://github.com/google/comprehensive-rust/blob/0cb7f496b542d76468fa950718f57f4037abc8fa/src/types-and-values/exercise.md?plain=1#L15), I figured this one might be missing the attribute. Anyway amazing work on the docs!

### DIFF
--- a/src/control-flow-basics/exercise.md
+++ b/src/control-flow-basics/exercise.md
@@ -25,7 +25,7 @@ For example, beginning with _n<sub>1</sub>_ = 3:
 Write a function to calculate the length of the collatz sequence for a given
 initial `n`.
 
-```rust,should_panic
+```rust,editable,should_panic
 {{#include exercise.rs:collatz_length}}
   todo!("Implement this")
 }


mgeisler (approved):
Thanks @titong0!

We normally encourage people to do the exercises in the Rust Playground (since it will keep state), but this one seems so short that it can be done directly on the page.

mgeisler:
> I'm unsure about when exercises should be solved directly on the site and when they should be copied and pasted into a playground. But since the previous chapter's exercise is solvable in the site [(fibonacci sequence)](https://github.com/google/comprehensive-rust/blob/0cb7f496b542d76468fa950718f57f4037abc8fa/src/types-and-values/exercise.md?plain=1#L15), I figured this one might be missing the attribute.

Good question... I'm happy to put it in here as well. #1476 is about why we don't have the `editable` attribute for most exercises and #1289 is about making the text before each exercise more uniform and standardized. I think we could use some help with both, in case you are interested :smile: 

> Anyway amazing work on the docs!

Thank you so much, I'll forward this to the rest of the team!

mgeisler:
@titong0, if you can sign the CLA, then we can merge this!

titong0:
I'm sorry, I'm having trouble with this CLA thing. I have signed the agreement with the same email I am registered with on github. However, since I have email privacy enabled, my commits are associated with some sort of github-only "fake" email with the address `68573622+titong0@users.noreply.github.com.`. Is there any way I could tell google that it should check for *that* email instead of my actual, personal email?

mgeisler:
> I'm sorry, I'm having trouble with this CLA thing. I have signed the agreement with the same email I am registered with on github. However, since I have email privacy enabled, my commits are associated with some sort of github-only "fake" email with the address `68573622+titong0@users.noreply.github.com.`. Is there any way I could tell google that it should check for _that_ email instead of my actual, personal email?

Hey @titong0, it works now! It turns out that we just needed to tell the bot to rescan the PR! It's obscure, but there is actually a small link for both you and me on to do this on [the failed run](https://github.com/google/comprehensive-rust/pull/1827/checks?check_run_id=21706540326).

So all good now, thank you for the contribution!

mgeisler:
Thanks a lot!

PR #1830, opened 2024-02-20
Make the "break with label" example more illustrative

marjakh:
In the old version, using "break 'outer;" and using "break;" (without the label) produce the same output.

This version fixes that to make the example more illustrative.

### DIFF
--- a/src/control-flow-basics/break-continue/labels.md
+++ b/src/control-flow-basics/break-continue/labels.md
@@ -7,16 +7,18 @@ to break out of nested loops:
 fn main() {
     let s = [[5, 6, 7], [8, 9, 10], [21, 15, 32]];
     let mut found = false;
+    let mut elements_searched = 0;
     let target_value = 10;
     'outer: for i in 0..=2 {
         for j in 0..=2 {
+            elements_searched += 1;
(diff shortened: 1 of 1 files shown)


mgeisler:
> In the old version, using "break 'outer;" and using "break;" (without the label) produce the same output.
> 
> This version fixes that to make the example more illustrative.

Good point! Could you remove the `found` variable now since it's probably not needed? I'm always trying hard to minimize the number of lines used by the code samples (to avoid scrolling up and down during a class).

src/control-flow-basics/break-continue/labels.md
```diff
@@ -6,17 +6,17 @@ to break out of nested loops:
...
         for j in 0..=2 {
+            elements_searched += 1;
             if s[i][j] == target_value {
-                found = true;
                 break 'outer;
             }
         }
     }
-    print!("{}", found)
+    print!("elements searched: {}", elements_searched);
```
mgeisler, line 19: Let me inline this like we do in most other examples of the course:

```suggestion
    print!("elements searched: {elements_searched}");
```

mgeisler (approved):
Awesome, thanks a lot for the quick fix!

PR #1832, opened 2024-02-20
Update exercise to add also the tests, in a TDD style

jcvicelli:
Having the tests only in the result brings less value, so the participants can do it following TDD practices


randomPoison (approved):
This seems like a good change to me. I'm not sure why we weren't including the test in the exercise materials, that's generally how the exercises are setup.

djmitche:
Probably just a mistake -- thanks!

PR #1833, opened 2024-02-20
Change hardcoded vector update to a for loop

jcvicelli:
iterating over the vector instead of hardcoding each item

### DIFF
--- a/src/references/exercise.rs
+++ b/src/references/exercise.rs
@@ -25,7 +25,6 @@ fn magnitude(vector: &[f64; 3]) -> f64 {
 /// Change the magnitude of the vector to 1.0 without changing its direction.
 fn normalize(vector: &mut [f64; 3]) {
     let mag = magnitude(vector);
-
     for item in vector {
         *item /= mag;
     }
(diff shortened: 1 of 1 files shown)


src/references/exercise.rs
```diff
@@ -25,9 +25,10 @@ fn magnitude(vector: &[f64; 3]) -> f64 {
 /// Change the magnitude of the vector to 1.0 without changing its direction.
 fn normalize(vector: &mut [f64; 3]) {
     let mag = magnitude(vector);
-    vector[0] /= mag;
-    vector[1] /= mag;
-    vector[2] /= mag;
+
```
mgeisler, line 28: Nice! Let me remove the blank line to space on vertical space:
```suggestion
```

mgeisler (approved):
Thanks! I hope we've covered `for` loops at this point?

djmitche:
The `course-content` binary is a good way to check that.

In this case, yes, `for` has been covered -- but only over ranges. Maybe we could compromise between idiomaticity and course flow with something like

```rust
for i in 0..3 {
  vector[i] /= mag;
}
```

On the other hand, this is in the solution to an exercise. If a student notices the for loop and asks, then the instructor can briefly explain it and circle back during the Iterators segment. And if nobody notices -- the syntax reads very naturally, so maybe students aren't surprised -- no problem!

mgeisler:
> The `course-content` binary is a good way to check that.

Ah yes, I remember you created this a while ago :smile: As you can see, I haven't tried using it yet.

PR #1835, opened 2024-02-20
Update user-defined-types/exercise.md

MaerF0x0:
Allow us to edit so we can complete the exercise :)


randomPoison (approved):
Ah, good change! I think in general we should make all code blocks editable, I often want to edit code blocks for demonstration purposes.

PR #1805, opened 2024-02-10
Rendered PDF improvements

max-heller:
Updates `mdbook-pandoc` to [0.5.0](https://github.com/max-heller/mdbook-pandoc/releases/tag/v0.5.0) to use some new features:
- Table wrapping (closes #1709)
  
  The wrapping logic mirror's [Pandoc's](https://pandoc.org/MANUAL.html#extension-pipe_tables) (i.e., compare the relative widths of the `---` columns to determine column widths) so tables in the source may need tweaking to get the proportions right
- Replace broken links in PDF with corresponding links to https://google.github.io/comprehensive-rust/

PR #1836, opened 2024-02-21
Link to PDF version of course from first page

max-heller:
Re https://github.com/google/comprehensive-rust/pull/1805#pullrequestreview-1888804528

### DIFF
--- a/src/index.md
+++ b/src/index.md
@@ -12,7 +12,8 @@ generics and error handling.
 > <https://google.github.io/comprehensive-rust/>. If you are reading somewhere
 > else, please check there for updates.
 >
-> The course is also available [as a PDF](https://google.github.io/comprehensive-rust/comprehensive-rust.pdf).
+> The course is also available
+> [as a PDF](https://google.github.io/comprehensive-rust/comprehensive-rust.pdf).
 
 The goal of the course is to teach you Rust. We assume you don't know anything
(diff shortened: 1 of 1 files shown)


src/index.md
```diff
@@ -11,6 +11,9 @@ generics and error handling.
 > The latest version of the course can be found at
 > <https://google.github.io/comprehensive-rust/>. If you are reading somewhere
 > else, please check there for updates.
+>
+> The course is also available
+> [as a PDF](https://google.github.io/comprehensive-rust/comprehensive-rust.pdf).
```
mgeisler, line 16: I think a relative link will work better: it should be correct for all languages out of the box:

```suggestion
> [as a PDF](comprehensive-rust.pdf).
```

max-heller: I wasn't sure if that'd be an issue when the PDF isn't available (e.g., building the book locally with `mdbook build`). Is it fine if the link is broken in such cases?

mgeisler: Good question... I would say yes, building things locally is something people can (and should) do while developing the material, but the publishing workflow is the one that actually counts.

The `build.sh` script is a little step towards formalizing this. @sakex has been working on #1236, which will let `mdbook` do all the work for all the translations.

mgeisler (approved):
Thanks!

PR #1838, opened 2024-02-21
Move hello world example to second section

randomPoison:
The current location for the hello world example is awkward since the two surrounding pages are both talking about Rust from a high level without talking about syntax at all. I think starting the second section (types and values) with the hello world example would help the first section flow more smoothly.


djmitche (approved):
Makes sense to me!

Issue #1834, opened 2024-02-20
Update the same comment with schedule changes


djmitche:
We should only add a new schedule comment if one does not already exist.

PR #1839, opened 2024-02-21
Tweak timings of exercises to better reflect teaching times

randomPoison:
A few of the early exercises had much larger estimates than were actually necessary for the super simple early exercises. I've gone through and reviewed the time estimates for exercises and tweaked the estimates based on how much time students have actually needed in my classes so far.

### DIFF
--- a/src/borrowing/exercise.md
+++ b/src/borrowing/exercise.md
@@ -1,5 +1,5 @@
 ---
-minutes: 30
+minutes: 20
 ---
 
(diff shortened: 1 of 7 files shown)


mgeisler (approved):
Thanks for this, it's super useful to get some real-life data on this!

djmitche:
Yes, that's fantastic! I'm sorry the schedule comments are so verbose.. I'll work on that! (#1834)

PR #1841, opened 2024-02-21
Flesh out `for` loop example

randomPoison:
* Add example of iterating over a collection.
* Update speaker notes to call out the use of iterators.

I think it's useful to call out that `for` loops primarily are used to iterate over a collection of objects, even though we haven't yet talked about any concrete collection types at this point. I think using the array literal syntax is simple enough to understand that it should be quick to explain when we get to this slide.

### DIFF
--- a/src/control-flow-basics/loops/for.md
+++ b/src/control-flow-basics/loops/for.md
@@ -1,19 +1,25 @@
 # `for`
 
 The [`for` loop](https://doc.rust-lang.org/std/keyword.for.html) iterates over
-ranges of values:
+ranges of values or the items in a collection:
 
 ```rust,editable
 fn main() {
     for x in 1..5 {
         println!("x: {x}");
     }
+
+    for elem in [1, 2, 3, 4, 5] {
+        println!("elem: {elem}");
+    }
 }
 ```
 
 <details>
 
-- We will discuss iteration later; for now, just stick to range expressions.
+- Under the hood `for` loops use a concept called "iterators" to handle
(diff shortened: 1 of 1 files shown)


randomPoison:
Oh, I forgot that we demonstrate using a `for` loop on an array on the slide after we introduce the syntax for arrays. I kind of think it's still worth it to add the array example to the `for` slide to better emphasize the nature of for loops, but let me know if this seems too redundant.

src/control-flow-basics/loops/for.md
```diff
@@ -1,19 +1,25 @@
...
 
 ```rust,editable
 fn main() {
     for x in 1..5 {
         println!("x: {x}");
     }
+
+    for elem in [1, 2, 3, 4, 5] {
+        println!("elem: {elem}");
+    }
```
mgeisler, line 14: I think this was part of the strategy of only using value types at first.

However, I can see we break this again in [6.3.1. Labels](https://google.github.io/comprehensive-rust/control-flow-basics/break-continue/labels.html) where we use more arrays.

I agree that the syntax is simple and understandable. I would be inclined to remove the `1..5` range syntax since I find it more complex: as a new person, I would wonder what this becomes and if the `..` part is part of `for`-loop syntax (it's not).

djmitche (approved):

src/control-flow-basics/loops/for.md
mgeisler: Thanks, I asked Dmytro as well and he also likes it!

PR #1842, opened 2024-02-22
Add a dev theme to help with slide aspect ratio

### LINKED ISSUE #1796: Show/hide the sizing box without re-running `mdbook`
#1464 describes a way to enable a red box in development mode, to determine whether a slide will fit onscreen without scrolling.

Let's define a keystroke or hidden button that will toggle this box, without any special configuration in `book.toml`.  This will make it easier to use this tool while developing or reviewing a PR, without accidentally committing changes to `book.toml`.

CC @mani-chand who is working on fixing slides for #1464.

mani-chand:
Added a dev theme to help with slide aspect ratio while updating content or for reviewing PRs.

Fixes #1796.


djmitche:
Can you update the PR description to summarize what this does and why? PRs are often used as reference later, to answer questions like "why was this done this way?" or "what does this code do?". "Check on your local and merge it" doesn't really answer that question.  Please also update the title of the PR to be a short summary of the description. Maybe "Add a dev theme to help with slide aspect ratio" (note the present tense "Add" instead of past-tense "Added").

mani-chand:
> Can you update the PR description to summarize what this does and why? PRs are often used as reference later, to answer questions like "why was this done this way?" or "what does this code do?". "Check on your local and merge it" doesn't really answer that question.  Please also update the title of the PR to be a short summary of the description. Maybe "Add a dev theme to help with slide aspect ratio" (note the present tense "Add" instead of past-tense "Added").

Done 👍 .

djmitche (commented):
Nice work -- the result looks good!

I thought you were using a custom theme to avoid the need for a hide/show button, but I see that this implements both: even in the dev theme, the user can show/hide the red box. We should probably pick one way to do this: either ask users to switch to the dev theme to see the red box, or allow hiding/showing the box in any theme. Which do you think is a better choice?

The second thing we'll need to worry about is that `theme/book.js` gets replaced every time we update the version of mdBook we use. If we need to re-apply your changes every time, that will get to be pretty difficult. Could you find a way to minimize the changes required to `theme/book.js`? In particular, you can add a new stylesheet in `theme/css` and add it to the `additional-css` list in `book.toml`. Similarly, you can add a new JS file in `theme` and add it to the `additional-js` list in `book.toml`. The speaker-notes files are a good example to follow. In fact, there are no changes from the default `book.js` for the speaker notes support!

mani-chand:
> Nice work -- the result looks good!
> 
> I thought you were using a custom theme to avoid the need for a hide/show button, but I see that this implements both: even in the dev theme, the user can show/hide the red box. We should probably pick one way to do this: either ask users to switch to the dev theme to see the red box, or allow hiding/showing the box in any theme. Which do you think is a better choice?
> 
> The second thing we'll need to worry about is that `theme/book.js` gets replaced every time we update the version of mdBook we use. If we need to re-apply your changes every time, that will get to be pretty difficult. Could you find a way to minimize the changes required to `theme/book.js`? In particular, you can add a new stylesheet in `theme/css` and add it to the `additional-css` list in `book.toml`. Similarly, you can add a new JS file in `theme` and add it to the `additional-js` list in `book.toml`. The speaker-notes files are a good example to follow. In fact, there are no changes from the default `book.js` for the speaker notes support!

I will look in to the changes and adjust them as you suggested by tommorow EOD.

mani-chand:
> > Nice work -- the result looks good!
> > I thought you were using a custom theme to avoid the need for a hide/show button, but I see that this implements both: even in the dev theme, the user can show/hide the red box. We should probably pick one way to do this: either ask users to switch to the dev theme to see the red box, or allow hiding/showing the box in any theme. Which do you think is a better choice?
> > The second thing we'll need to worry about is that `theme/book.js` gets replaced every time we update the version of mdBook we use. If we need to re-apply your changes every time, that will get to be pretty difficult. Could you find a way to minimize the changes required to `theme/book.js`? In particular, you can add a new stylesheet in `theme/css` and add it to the `additional-css` list in `book.toml`. Similarly, you can add a new JS file in `theme` and add it to the `additional-js` list in `book.toml`. The speaker-notes files are a good example to follow. In fact, there are no changes from the default `book.js` for the speaker notes support!
> 
> I will look in to the changes and adjust them as you suggested by tommorow EOD.

So, I will make a button in left part of navbar and allow hiding/showing the box in any theme and i will remove dev theme because i am geting problem because majority of theme code work only in book theme.

mani-chand:
https://github.com/google/comprehensive-rust/assets/65899331/8964ffe6-6e55-49a6-88a7-cb6c552b3d5f


Did changes , let me know if it is good or any changes required.

mani-chand:
> Static.and.Const.-.Comprehensive.Rust.-.Google.Chrome.2024-02-26.22-05-00.mp4 
> Did changes , let me know if it is good or any changes required.

Do check at your local.

theme/redbox.js
```diff
@@ -0,0 +1,31 @@
+// Create a new div element
```
djmitche, line 1: It's typical to include code like this in an [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) so that all of these variables don't end up in the global JS namespace. There are examples of this in `theme/book.js`.

theme/redbox.js
```diff
@@ -0,0 +1,31 @@
...
+hideShowButton.type = "button";
+hideShowButton.title = "showing/hiding red box";
+hideShowButton.id = "Dev";
+var navbarButtons = document.getElementsByClassName("left-buttons");
+navbarButtons[0].insertBefore(hideShowButton, navbarButtons.firstChild);
+document.getElementById("aspect-ratio-helper").style.display = "none";
+hideShowButton.addEventListener("click", function () {
+  if (document.getElementById("aspect-ratio-helper").style.display === "none") {
+    document.getElementById("aspect-ratio-helper").style.display = "block";
+    hideShowButton.innerHTML = "User";
```
djmitche, line 26: Rather than using text as the `innerHTML` here, let's use an icon. That can either be with a font-awesome element (`<i class="fa fa-something">`) or just a unicode character. I kind of like ⬚!

theme/redbox.js
```diff
@@ -0,0 +1,31 @@
...
+// Get the parent element where you want to append the new div
+var parentElement = document.body; // Change this to your desired parent element
+// Append the new div to the parent element
+parentElement.insertBefore(newDiv, parentElement.firstChild);
+// Create the button element
+var hideShowButton = document.createElement("button");
+hideShowButton.innerHTML = "redbox";
+hideShowButton.className = "icon-button";
+hideShowButton.type = "button";
+hideShowButton.title = "showing/hiding red box";
```
djmitche, line 18: This is shown when hovering over the button, so let's explain a bit about what the red box does:
```suggestion
hideShowButton.title = "Outline the area that fits on one screen while teaching the course.";
```

book.toml
```diff
@@ -31,7 +31,6 @@ verbose = false # Report timing information.
 
 [output.pandoc]
 optional = true
-hosted-html = "https://google.github.io/comprehensive-rust/"
```
djmitche, line 34: Was this removed on purpose? If not, please restore :)

djmitche (changes requested):
Looking good! Can you also edit `STYLE.md` to suggest using this new tool, and remove the bit about uncommenting things in `book.toml`? Please also remove those comments in `book.toml` and the aspect-ratio Python script.

mani-chand:
I haven't checked it by running in my local.Ignore it you find something wrong.

theme/redbox.js
mani-chand: Done

theme/redbox.js
mani-chand: Not sure how the icon is reflecting on web page.

theme/redbox.js
mani-chand: Done.

book.toml
mani-chand: Done. I don't know how it got removed.

STYLE.md
```diff
@@ -15,14 +15,9 @@ important to keep this in mind when adding content: we only have limited
...
-lines in the `book.toml` file to have a red rectangle rendered on top of all
-pages:
-
-```toml
-[preprocessor.aspect-ratio-helper]
-command = "./aspect-ratio-helper.py"
-```
+You can test the amount of space available using a simple tool.This tool can be
+used by clicking a toggle button right to the search button on left side of
+navbar.
```
djmitche, line 20:
```suggestion
You can test the amount of space available using a simple tool. This tool can be
used by clicking a toggle button next to the search button on left side of the
navbar.
```

theme/redbox.js
```diff
@@ -0,0 +1,38 @@
...
+  var nestedDiv = document.createElement("div");
+  // Append the nested div to the new div
+  newDiv.appendChild(nestedDiv, newDiv.firstChild);
+  // Get the parent element where you want to append the new div
+  var parentElement = document.body; // Change this to your desired parent element
+  // Append the new div to the parent element
+  parentElement.insertBefore(newDiv, parentElement.firstChild);
+  // Create the button element
+  var hideShowButton = document.createElement("button");
+  hideShowButton.innerHTML = '<i class="fa-solid fa-square fa-lg"></i>';
```
djmitche, line 16:
```suggestion
  hideShowButton.innerHTML = '<i class="fa fa-square-o"></i>';
```
With `fa-lg`, this icon doesn't sit at the same level as the others:

image

I think the `fa` is required -- it's present on the other `i` elements, anyway.  And, the `-o` at the end makes it hollow. These last two points may be different for more recent versions of Font awesome, but it appears mdbook uses [v4](https://fontawesome.com/v4/icons/).

djmitche (commented):
This is looking great. If you want to add these last two changes (and formatting) we can get it merged. Nice work on this!!

mani-chand:
image
image

Used two different icons for toggle button one for hiding and another showing redbox. I feel it looks better.

mani-chand:
Hello @djmitche , I think you can merge now.

mani-chand:
#1859 Can you look at this too.

mani-chand:
Hello @djmitche , Just let me know , If there are any changes needed.

.github/workflows/build.sh
```diff
@@ -37,6 +37,10 @@ else
 fi
 
 mdbook build -d "$dest_dir"
+
+# Disable the redbox button in built versions of the course
+echo '// disabled' > "${dest_dir}/html/theme/redbox.js"
```
djmitche, line 42: I think it will be easiest to just erase this JS file in the built version. If this looks good to you and @mgeisler then I think we can merge!

djmitche (approved):

.github/workflows/build.sh
mani-chand: Sorry, I don't know,I am not sure about it.Lets wait for the review of @mgeisler .

mgeisler, line 42: Thanks, this is a great idea! I would refer back to the script like this:

```suggestion
echo '// Disabled in published builds, see build.sh' > "${dest_dir}/html/theme/redbox.js"
```

mgeisler: I think this is a nice approach, thanks @djmitche.

Since the feature is entirely triggered by the configuration in `book.toml`, a different solution would be to manipulate this file. This can be done with environment variables which override the `book.toml` file.

Unfortunately, it's not super easy to remove the `redbox.js` and `redbox.css` entries from the two lists... I was playing with a solution that uses [`toml`](https://github.com/gnprice/toml-cli) and [`jq`](https://jqlang.github.io/jq/) — here is how to get the other JS files:

```
toml get book.toml 'output.html.additional-js' | jq 'map(select(test("redbox") | not))'
```

that value could be put back into a `MDBOOK_OUTPUT__HTML__ADDITIONAL_JS` environment variable in `build.sh` and this will also disable the button (the same should then be done for the `additional-css` key).

This is probably overkill and I like your simple solution very much.

mgeisler (approved):

theme/book.js
```diff
@@ -721,4 +721,4 @@ function playground_text(playground, hidden = true) {
         updateBorder();
         document.addEventListener('scroll', updateBorder, { passive: true });
     })();
-})();
+})();
```
mgeisler, line 724:
```suggestion
})();
```

book.toml
```diff
@@ -26,8 +26,6 @@ verbose = false # Report timing information.
 # pages. This will show you an estimate of what the course
 # participants can see during the presentation.
 #
```
mgeisler, line 28: This should go away too:

```suggestion
```

book.toml
```diff
@@ -23,11 +23,6 @@ class = "bob"
 verbose = false # Report timing information.
 
 # Enable this preprocessor to overlay a large red rectangle on the
```
mgeisler, line 25: More cleanups in the old `book.toml` file:

```suggestion
```

PR #1855, opened 2024-02-28
Make diagram consistent with code

nyurik:


randomPoison (approved):
Thanks for making this more consistent!

PR #1856, opened 2024-02-28
Update course-schedule PR comments in-place

### LINKED ISSUE #1834: Update the same comment with schedule changes

djmitche:
Instead of always adding a new comment, just update the existing comment if one exists.

This fixes #1834.


djmitche:
A course schedule!

djmitche:
I tested this on the PR and the JS ran and updated the comment. We can't know for sure that it works with all of the GitHub permissions stuff until merging, but I'm pretty confident.

mgeisler (approved):
Looks nice, thanks!

PR #1824, opened 2024-02-16
Added Array and tuples

mani-chand:
#1464 issue PR

PR #1859, opened 2024-02-29
Divided Array and tuples

mani-chand:
#1824 PR Changes


djmitche (changes requested):
This looks great! I think it's a good idea to make this two top-level slides, rather than a sub-slide.

The only thing I'd request to change is the order of the two slides. Tuples come first right now, but the speaker notes start with "Like arrays .." I think it would be easier to understand if the order were swapped, with arrays coming first. I don't think any other changes are required to the content, other than swapping the order.

mani-chand:
Hello @djmitche , I think now you merge the PR.

djmitche (approved):
Great -- thank you again!

mani-chand:
> Great -- thank you again!

No need to thank me . I just love doing technical things . I would like you to thank for giving me ideas, helping ad suggestions.

PR #1882, opened 2024-03-06
Do not show trait vtables as located in the heap in memory diagram

fw-immunant:
This diagram is misleading and I often explain that the character data of string literals resides in the executable's static data, with vtables working the same.


djmitche (approved):
Good fix, thanks!

mgeisler:
Yeah, thanks for fixing this!

PR #1883, opened 2024-03-06
concurrency: Update async trait slide

fw-immunant:
Clarify that the basics are stable, but dyn support is still missing. This slide was outdated and didn't explain the current state of stable correctly.


djmitche (approved):
Nice!

PR #1884, opened 2024-03-06
Tweak solution to iter exercise

fw-immunant:
Because `.zip()` is limited to the shorter length, the `.take()` call here is unnecessary. When explaining this solution I don't want to have to explain a call to a method that, used as it is, does nothing.


djmitche (approved):
Thanks!

PR #1885, opened 2024-03-06
concurrency: Add detailed teaching notes for welcome and threads slides

fw-immunant:
These follow the flow of what I actually teach, which spends a significant amount of time on the latter slide. I think it's worthwhile to have a real flow documented in the teaching notes, both to make sure nothing gets forgotten and to structure the experience of teaching.


djmitche (approved):
Nice, thank you!

PR #1887, opened 2024-03-07
Fix typo in associated-types.md

nullism:
A minor nitpick, but as someone new to the language I did spend a bit more time than I'd like to admit trying to understand the meaning of `allow` before realizing this is likely a typo. Maybe I still don't understand, and in that case I'd appreciate a correction!


mgeisler (approved):
Thanks for fixing that! These typo fixes are invaluable for us since it's hard to catch your own mistakes when you read the text :smile:

PR #1898, opened 2024-03-08
Move slices and strings to references section

randomPoison:
This PR moves the slides for slices and strings into the day 1 section on references. This seems like the more natural place to introduce slices since slices are a type of reference. It then also made sense to me to follow that with the introduction of `&str` and `String`, since students now have the context to understand what a "string slice" is. I also removed the strings slide from the types and values section since it didn't make sense to cover the same topic twice in the same day. I tested this new organization in my class on Wednesday and it didn't cause day 1 to take too long.


djmitche (approved):
Nice -- I think this makes a lot more sense!

PR #1899, opened 2024-03-08
Remove generics from logger exercise

randomPoison:
The logger exercise comes before the section on generics, and the purpose of the exercise is for students to get practice writing a trait implementation, so using generics in the solution is a source of confusion for students. I've removed the generic and made `VerbosityFilter` directly hold a `StderrLogger`.


fw-immunant (approved):
My experiences with the material ordering and student responses here are similar. Thanks for improving this--looks good to me.

PR #1900, opened 2024-03-08
Simplify generic min exercise

randomPoison:
I was a bit dissatisfied with the test code for the generic min exercise. We were supposed to be testing that the student wrote the generic function correctly, but we only test their `min` function with one type. I rewrote the exercise to test against multiple types, which required that we use the regular `Ord` trait rather than a custom one. I tend to prefer using the real items from `std` when we can because it's a good way to get students familiar with the standard library. I also removed the custom `Citation` type since it wasn't really important to the exercise.

### DIFF
--- a/src/generics/exercise.md
+++ b/src/generics/exercise.md
@@ -5,12 +5,21 @@ minutes: 10
 # Exercise: Generic `min`
 
 In this short exercise, you will implement a generic `min` function that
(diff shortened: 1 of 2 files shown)


djmitche (approved):
Looks good! I was perhaps too conservative in not wanting to use the std type until it was introduced.

PR #1917, opened 2024-03-11
Saving playground code state

mani-chand:
The embedded Playground's code reset when you navigate between slides. This has caused problems: if people navigate away from a slide to look something up, they've suddenly lost their work.


djmitche:
from @fw-immunant in https://github.com/google/comprehensive-rust/issues/1476#issuecomment-1989150255
> It seems like the text editor widget (ace_text) should have some API for getting/setting text contents, rather than digging around in its internal markup.
> 
> From the [docs](https://ace.c9.io/#nav=howto), it looks like these suffice: `editors[0].getValue();`/`editors[0].setValue("fn main(){\n\n}", -1);`.
> 
> While teaching I rely on being able to reset the code samples to their original contents (both while discussing a slide and before teaching the course a second time), so I'd like to make sure there's still a straightforward way to do so before this lands.

theme/redbox.js
```diff
@@ -36,3 +36,69 @@
     }
   });
 })();
+
+function setCodeToPlayground() {
+  const code = JSON.parse(localStorage.getItem(window.location.href));
+  console.log(code);
```
fw-immunant, line 42: Remove debugging printing.

theme/redbox.js
```diff
@@ -36,3 +36,69 @@
...
+        //span.classList = spanChild[j].classes
+        for (let cls in spanChild[j].classes) {
+          span.classList.add(spanChild[j].classes[cls]);
+        }
+        span.innerText = spanChild[j].text;
+        div.insertBefore(span, div.lastChild);
+      }
+      playground.insertBefore(div, playground.lastChild);
+    }
+  }
```
fw-immunant, line 71: Use the API of the editor component rather than poking into its portion of the DOM:
```js
editors[0].setValue("fn main(){\n\n}", -1);
```
or similar.

theme/redbox.js
```diff
@@ -36,3 +36,69 @@
...
+  localStorage.removeItem(window.location.href);
+}
+
+window.onunload = setTimeout(setCodeToPlayground, 5000);
```
fw-immunant, line 75: I believe this is a type error. Isn't `window.onunload` supposed to be a callable? `setTimeout` returns an integer timeout id.

What is this intended to do?

theme/redbox.js
```diff
@@ -36,3 +36,69 @@
...
+        classes: playground[i].children[j].classList,
+        styles: playground[i].children[j].style,
+      };
+      line.push(codeList);
+    }
+    code.push([parentCodeList, line]);
+  }
+  console.log(code);
+  //localStorage.removeItem(window.location.href)
+  localStorage.setItem(window.location.href, JSON.stringify(code));
```
fw-immunant, line 102: Again, remove debugging prints and use the public API of the editor rather than accessing the details of its implementation in the DOM.

I don't think I see a reason to store a serialized JSON object in localStorage if the object is just the code--it should be a single string if I understand correctly what information is being persisted.

fw-immunant (changes requested):
See individual comments.

I think for usability reasons this feature also requires that there's a way to reset the code on an individual page to its original state, the way reloading the page did previously. In addition, after teaching the full course, instructors need a way to "erase the blackboards" and reset all code samples to their pristine state.

mani-chand:
I will work on it.

mani-chand:
Hello @djmitche  , Its working fine. Just check it once and let me know the changes.

mani-chand:
> Hello @djmitche  , Its working fine. Just check it once and let me know the changes.

Should I rename the file as `additional. js`.

djmitche:
> Should I rename the file as `addtional.js`. Because `redbox.js' doesn't sound perfect.

This sounds like a good idea.

I've got a lot of my "regular" job to do this week, before the next version of Chromium is released, so I'll be a little slow to review here. If @fw-immunant has time to take a look, that'd be great -- otherwise I'll get to it by-and-by. Sorry for the delay!

mani-chand:
> > Should I rename the file as `addtional.js`. Because `redbox.js' doesn't sound perfect.
> 
> This sounds like a good idea.
> 
> I've got a lot of my "regular" job to do this week, before the next version of Chromium is released, so I'll be a little slow to review here. If @fw-immunant has time to take a look, that'd be great -- otherwise I'll get to it by-and-by. Sorry for the delay!

Its ok with delay. Take your time and review the PR.

mani-chand:
> > Should I rename the file as `addtional.js`. Because `redbox.js' doesn't sound perfect.
> 
> This sounds like a good idea.
> 
> I've got a lot of my "regular" job to do this week, before the next version of Chromium is released, so I'll be a little slow to review here. If @fw-immunant has time to take a look, that'd be great -- otherwise I'll get to it by-and-by. Sorry for the delay!

All the best for next version of Chromium .

book.toml
```diff
@@ -46,7 +46,7 @@ urlcolor = "red"
 
 [output.html]
 curly-quotes = true
-additional-js = ["theme/speaker-notes.js", "theme/redbox.js"]
+additional-js = ["theme/speaker-notes.js", "theme/additional.js"]
```
djmitche, line 49: Ah, I thought you meant adding a _new_ JS file for this functionality, rather than moving the existing red-box functionality too. Maybe keep `theme/redbox.js` and move this new functionality into `theme/save-playgrounds.js`?

theme/additional.js
```diff
@@ -36,3 +36,33 @@
...
+    pre_block
+  ) {
+    let code_block = pre_block.querySelector("code");
+    let editor = window.ace.edit(code_block);
+    let code = editor.getValue();
+    codes.push(code);
+  });
+  localStorage.setItem(window.location.href, JSON.stringify(codes));
+}
+setInterval(getCodeFromPlayground, 900);
```
djmitche, line 68: This seems like it will wake up the browser frequently even if nothing has changed. I wonder if we could either do this on some event indicating the browser is navigating away (maybe https://developer.mozilla.org/en-US/docs/Web/API/Window/pagehide_event?) or on a timer after the user has modified something in the editor.

theme/additional.js
```diff
@@ -36,3 +36,33 @@
...
+    ) {
+      let code_block = pre_block.querySelector("code");
+      let editor = window.ace.edit(code_block);
+      editor.setValue(codes[i]);
+      editor.clearSelection();
+      i += 1;
+    });
+  }
+}
+setCodeToPlayground();
```
djmitche, line 55: minor point: this is a bit hard to spot here. Please move the whole thing to an IIFE, like the red-box functionality, and put the immediately-evaluated statements at the bottom:
```js
(function savePlaygrounds() {
  function setCodeToPlayground() ..
  function getCodeFromPlaground() ..

  setCodeToPlayground();
  setInterval(getCodeFromPlayground, 900);
})()
```

djmitche (commented):
Looking good! I made a few suggestions, but the part I don't know is how best to either detect the changes or respond to navigation. I know that `onunload` is deprecated, and `onbeforeunload` is for displaying dialogs and not for saving data.

mani-chand:
Hello @djmitche , I have did changes according to the changes required . Let me know any extra changes required. `pagehide` event is working fine once test it.

djmitche (commented):
Great work!! This works just fine for me, and I can see in the debugger that it's setting localStorage reliably, but only when necessary.

As @fw-immunant mentioned, though, we need a way to reset all playgrounds to their default state before this lands -- otherwise instructors will be stuck with modified content from the last time they taught the course.

Maybe we could convert the aspect-ratio icon into an "instructor menu" and put the aspect-ratio item and the reset-playgrounds item in that menu. What do you think?

djmitche:
I think it will be easiest to do it here, since this PR can't be merged until that's ready.

mani-chand:
Ok. I will do it in this PR.

mani-chand:
Hello @djmitche ,  It looks good once check and just let me know any changes required.

mani-chand:
> Hello @djmitche ,  It looks good once check and just let me know any changes required.

It will set all playground except the using playground (page you are in)

fw-immunant:
What calls the `handleInstructor` function here?

mani-chand:
> What calls the `handleInstructor` function here?

reference : 

https://developer.mozilla.org/en-US/docs/Glossary/IIFE

```
(function handleInstructor() {

})() 
```

mani-chand:
> ~What calls the `handleInstructor` function here?~ Never mind, it's an IIFE.

Let me know, If you had any changes.

theme/redbox.js
```diff
@@ -1,38 +1,101 @@
...
+    redBoxItem.insertBefore(redBoxButton, redBoxItem.lastChild);
+    playgroundStateItem.insertBefore(
+      playgroundStateButton,
+      playgroundStateItem.lastChild
+    );
+
+    instructorMenu.innerHTML =
+      '<i class="fa fa-ellipsis-v" aria-hidden="true"></i>';
+    redBoxButton.innerHTML = "redbox";
+    playgroundStateButton.innerHTML = "reset code";
```
djmitche, line 27: Let's be more specific about what is reset, and also include a title:
```suggestion
    redBoxButton.innerHTML = "aspect-ratio box";
    redBoxButton.title = "Outline the area that fits on one screen while teaching the course.";
    playgroundStateButton.innerHTML = "reset all playgrounds";
    playgroundStateButton.title = "Reset code in all playgrounds to its original value.";
```

theme/redbox.js
```diff
@@ -1,38 +1,101 @@
...
+
+    instructorMenuList.role = "menu";
+    redBoxItem.role = "none";
+    playgroundStateItem.role = "none";
+    redBoxButton.role = "menuitem";
+    playgroundStateButton.role = "menuitem";
+
+    redBoxButton.id = "redbox";
+
+    instructorMenuList.style.marginLeft = "55px";
```
djmitche, line 43: Minor detail: I think this would be easier to read if each item was grouped together, and all its properties set in the same place.

theme/redbox.js
```diff
@@ -1,38 +1,101 @@
...
-      document.getElementById("aspect-ratio-helper").style.display === "none"
-    ) {
-      document.getElementById("aspect-ratio-helper").style.display = "block";
-      hideShowButton.innerHTML = '<i class="fa fa-square"></i>';
-    } else {
-      document.getElementById("aspect-ratio-helper").style.display = "none";
-      hideShowButton.innerHTML = '<i class="fa fa-square-o"></i>';
-    }
-  });
+(function handleInstructor(params) {
```
djmitche, line 1: I don't think this function needs `params`!

It probably makes sense to move this to _yet another_ `.js` file, maybe `instructor-menu`?

For JS symbols that need to be reachable from other `.js` files, I think attaching them to `window` will work, even if it's a bit of a hack:
```js
(function savePlaygrounds() {
  ...
  window.resetAllPlaygrounds = function() { .. }
})()
```
```js
(function instructorMenu() {
    playgroundStateButton.addEventListener("click", () => {
      ...
      window.resetAllPlaygrounds();
    });
})()
```

theme/redbox.js
```diff
@@ -1,38 +1,101 @@
...
+      playgroundStateItem,
+      instructorMenuList.lastChild
+    );
+    redBoxItem.insertBefore(redBoxButton, redBoxItem.lastChild);
+    playgroundStateItem.insertBefore(
+      playgroundStateButton,
+      playgroundStateItem.lastChild
+    );
+
+    instructorMenu.innerHTML =
```
djmitche, line 24:
```suggestion
    instructorMenu.title = "Utilities for course instructors";
    instructorMenu.innerHTML =
```

theme/save-playgrounds.js
```diff
@@ -0,0 +1,33 @@
+(function savePlaygrounds() {
+  function setCodeToPlayground() {
+    var codes = JSON.parse(
+      localStorage.getItem(`${window.location.href}₹code`)
```
djmitche, line 4: clever!

djmitche (changes requested):
I'm not sure how to solve it, but I noticed that the theme menu (the paintbrush icon) disappears if you click outside the menu, while the instructor menu does not. It'd be great to fix that if it's an easy fix.

mani-chand:
Hello @djmitche  , Ignore the formatting and check it . Let me know any changes needed.

theme/save-playgrounds.js
mani-chand: 😎

.github/workflows/build.sh
```diff
@@ -40,6 +40,7 @@ mdbook build -d "$dest_dir"
 
 # Disable the redbox button in built versions of the course
 echo '// Disabled in published builds, see build.sh' > "${dest_dir}/html/theme/redbox.js"
+echo '// Disabled in published builds, see build.sh' > "${dest_dir}/html/theme/instructor-menu.js"
```
djmitche, line 43: Oh! I think we don't want this anymore, as we will need the "reset playgrounds" option in the published build. @mgeisler now that we have all of this in a sub-menu, what do you think?

theme/redbox.js
```diff
@@ -1,4 +1,4 @@
-(function redBoxButton() {
+function redBoxButton() {
```
djmitche, line 1: I think this should still be an IIFE, but at the end should have something like
```js
  ...
  window.redboxButtonClicked = redboxButtonClicked;
})();
```

Similarly, `save-playgrounds.js` would have
```js
  ...
  window.resetPlaygroundsClicked = resetPlaygroundsClicked;
})();
```

Then `instructor-menu.js` can have
```js
  redboxButton.addEventListener("click", e => window.redboxButtonClicked(e));
  playgroundStateButton.addEventListener("click", e => window.resetPlaygroundsClicked(e));
```

with all of the implementation of what to do when those buttons are clicked appearing in the appropriate file.

djmitche (changes requested):
This works great! My remaining comments are more about maintainability, and in particular separation of concerns. That is, the `instructor-menu.js` file should just set up the instructor menu and call global functions when the menu items are clicked. Everything specific to how those menu items are implemented should be in their own file. For example, the `₹code` trick should only appear in `save-playgrounds.js`.

mani-chand:
Hello @djmitche , check and Let me know any changes needed.

djmitche (approved):
I think this looks good! This was a lot of work and a lot of rounds of review, but I think the result is going to make teaching the course quite a bit better.

I'd like to get some feedback from @mgeisler regarding keeping the menu in builds before merging it.

A few minor points that could be addressed in a later PR if you'd prefer:

 * There's no feedback on clicking the "reset all playgrounds" to indicate that all playgrounds were reset. In fact, if one appears on the current slide, its contents aren't reset. I know `alert()` is lame, but maybe this is a good use-case for it?
 * It feels weird that the menu stays visible after clicking one of the items. Maybe hide it in the "click" handlers?

mani-chand:
> I think this looks good! This was a lot of work and a lot of rounds of review, but I think the result is going to make teaching the course quite a bit better.
> 
> I'd like to get some feedback from @mgeisler regarding keeping the menu in builds before merging it.
> 
> A few minor points that could be addressed in a later PR if you'd prefer:
> 
>  * There's no feedback on clicking the "reset all playgrounds" to indicate that all playgrounds were reset. In fact, if one appears on the current slide, its contents aren't reset. I know `alert()` is lame, but maybe this is a good use-case for it?
>  * It feels weird that the menu stays visible after clicking one of the items. Maybe hide it in the "click" handlers?

I can do 2nd point. If you say it should hide on clicking one option it is possible. I will do  it and push it.

mani-chand:
> I think this looks good! This was a lot of work and a lot of rounds of review, but I think the result is going to make teaching the course quite a bit better.
> 
> I'd like to get some feedback from @mgeisler regarding keeping the menu in builds before merging it.
> 
> A few minor points that could be addressed in a later PR if you'd prefer:
> 
>  * There's no feedback on clicking the "reset all playgrounds" to indicate that all playgrounds were reset. In fact, if one appears on the current slide, its contents aren't reset. I know `alert()` is lame, but maybe this is a good use-case for it?
>  * It feels weird that the menu stays visible after clicking one of the items. Maybe hide it in the "click" handlers?

`confirm` works fine I think.

mani-chand:
> > I think this looks good! This was a lot of work and a lot of rounds of review, but I think the result is going to make teaching the course quite a bit better.
> > 
> > I'd like to get some feedback from @mgeisler regarding keeping the menu in builds before merging it.
> > 
> > A few minor points that could be addressed in a later PR if you'd prefer:
> > 
> >  * There's no feedback on clicking the "reset all playgrounds" to indicate that all playgrounds were reset. In fact, if one appears on the current slide, its contents aren't reset. I know `alert()` is lame, but maybe this is a good use-case for it?
> >  * It feels weird that the menu stays visible after clicking one of the items. Maybe hide it in the "click" handlers?
> 
> for 1st point `confirm` works fine I think.

mani-chand:
> I think this looks good! This was a lot of work and a lot of rounds of review, but I think the result is going to make teaching the course quite a bit better.
> 
> I'd like to get some feedback from @mgeisler regarding keeping the menu in builds before merging it.
> 
> A few minor points that could be addressed in a later PR if you'd prefer:
> 
> * There's no feedback on clicking the "reset all playgrounds" to indicate that all playgrounds were reset. In fact, if one appears on the current slide, its contents aren't reset. I know `alert()` is lame, but maybe this is a good use-case for it?
> * It feels weird that the menu stays visible after clicking one of the items. Maybe hide it in the "click" handlers?

@djmitche , Once can you test it again.
 It feels weird that the menu stays visible after clicking one of the items. Maybe hide it in the "click" handlers? - Completed.

djmitche:
Nice! That might also address the feedback issue. I'm happy to merge this as-is, if others are OK.

theme/save-playgrounds.js
```diff
@@ -0,0 +1,46 @@
...
+    var codes = [];
+    Array.from(document.querySelectorAll(".playground")).forEach(function (
+      pre_block
+    ) {
+      let code_block = pre_block.querySelector("code");
+      let editor = window.ace.edit(code_block);
+      let code = editor.getValue();
+      codes.push(code);
+    });
+    localStorage.setItem(`${window.location.href}₹code`, JSON.stringify(codes));
```
mgeisler, line 29: Has it already been discussed what the `₹` character is doing here? It looks a bit strange to me :smile: 

Have you considered using just the [`pathname`](https://developer.mozilla.org/en-US/docs/Web/API/Location/pathname) instead of the full URL? Since local storage is different for different hosts, I don't think we need the full URL here.

mgeisler: Oh, if the `₹` is necessary somehow, then I would suggest creating a new function to construct the key. You can then document the use there — and also call it from both places that need this key.

theme/instructor-menu.js
```diff
@@ -0,0 +1,71 @@
+(function handleInstructor() {
+  function handleInstructorMenu() {
```
mgeisler, line 2: Did you discuss simply adding this to the theme instead? Since this will be a permanent thing on all pages, it should go into [index.hbs](https://github.com/google/comprehensive-rust/blob/main/theme/index.hbs) or [head.hbs](https://github.com/google/comprehensive-rust/blob/main/theme/head.hbs).

That would simplify the code a lot since you can just write out the HTML directly.

mgeisler: Why was this not done for the red box, you might ask? Because that was not supposed to be rendered for people looking at the course in the published version.

theme/save-playgrounds.js
mani-chand: > Oh, if the `₹` is necessary somehow, then I would suggest creating a new function to construct the key. You can then document the use there — and also call it from both places that need this key.

Sorry, I didn't understand creating of "another function for key creating".

PR #1918, opened 2024-03-12
ko: Merge CR v.2 Korean translation back into main

henrif75:
This PR merges the Korean (ko) v.2 translation back to main.
This requires just a syntactical review, as the partial PRs into this branch were reviewed already.

**The Chromium translation is still pending but I realized it's better to get this out and start benefiting users in the other modules.**

gh pr diff 1918 | bat -l patch

#1463 #284


jiyongp (approved):
Yeah, I agree with you on that landing this is more important than fixing the CR part.

PR #1924, opened 2024-03-16
Removed destructuring about struct

mani-chand:
#1464 issue . Let's limit this section to arrays and tuples. Destructuring in structs explained in [Day-2  Morning]((https://google.github.io/comprehensive-rust/pattern-matching/destructuring.html)).


djmitche (approved):
Nice catch, thanks! I see that destructuring chapter in day 2 also refers back to this one ("Like tuples, ..")

PR #1927, opened 2024-03-20
A simple fix in the day

alisoufali:


djmitche (commented):
Thanks! Unfortunately, we do require a CLA even for a one-character change like this. But that will cover further contributions you might make!

alisoufali:
> Thanks! Unfortunately, we do require a CLA even for a one-character change like this. But that will cover further contributions you might make!

Hi I just did handled the CLA. Do I need to make another commit to trigger the checks

djmitche (approved):
Nope, all set - thanks!

PR #1930, opened 2024-03-22
Move mpsc::Sender to the list of Send + Sync types

rbehjati:
As of 1.72.0, `mpsc::Sender` [is `Sync`](https://doc.rust-lang.org/std/sync/mpsc/struct.Sender.html#impl-Sync-for-Sender%3CT%3E).


djmitche (approved):
Nice! I don't want to lean too heavily into naming versions, but mentioning it here-and-there is a nice way of indicating that things in Rust do get better over time.

PR #1931, opened 2024-03-22
Update labeler.yml with Vietnamese

henrif75:


qwandor (approved):

henrif75:
That was fast @qwandor, thanks!

PR #1934, opened 2024-03-24
Fix misspelling of implementing

osmanbowser:
Should fix the page title as well


mgeisler (approved):
Thanks for catching this!

mgeisler:
> Should fix the page title as well

This fixes the outline in the left side — luckily the page title is already spelled correctly.

osmanbowser:
> > Should fix the page title as well
> 
> This fixes the outline in the left side — luckily the page title is already spelled correctly.

I actually noticed the typo in the site title in my Chrome tab. 

Then looking through the markdown file I found the typo that I assume is somehow connected to the site title.

It's an interesting setup that's for sure. Thanks for your help Martin!

PR #1935, opened 2024-03-24
Temporarily disable saving playground state

mgeisler:
When browing around on https://google.github.io/comprehensive-rust/, I sooner or later end up in a state where the local storage has `[]` stored for a page with one or more playgrounds. The effect of this is that the code is removed from the page!

I am not sure why this happens, but I’m afraid the code here needs more testing. I’m teaching a class Monday morning, so I’ll disable the code from #1917 for now.

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -48,9 +48,6 @@ urlcolor = "red"
 curly-quotes = true
 additional-js = [
   "theme/speaker-notes.js",
-  "theme/redbox.js",
-  "theme/save-playgrounds.js",
-  "theme/instructor-menu.js",
 ]
 additional-css = [
   "theme/css/svgbob.css",


mgeisler:
Cc @djmitche, @fw-immunant, and @mani-chand. I merged this now so I can ensure the course works for the class I'm teaching in ~12 hours.

I only saw this after browsing around on the site for 30 minutes or so. I don't really know why it would happen, but I guess the event that saves the playgrounds is unreliable. I would suggest looking into saving the state on _changes_ to the playgrounds instead of trying to detect navigating away from the pages.

I've seen problems with detecting the `pagehide` event in the past: the event is not [reliably fired](https://developer.mozilla.org/en-US/docs/Web/API/Window/pagehide_event#usage_notes) because it has been misused in the past.

I ran into another problem: clearing saved playground data doesn't fully work. Navigating away from the current page will save the playgrounds on the page again.

PR #1941, opened 2024-03-25
Improve readability of Multi-threaded Link Checker.

raffomartini:


djmitche (approved):
Thanks!

djmitche:
I see that `dprint` suggested adding `\` to each line, but I'm not sure why it did that. The result is that each sentence is on its own line in the rendered markdown, which seems unnecessary. I'll run `dprint fmt` locally and push the result.

mgeisler:
> I see that `dprint` suggested adding `\` to each line, but I'm not sure why it did that. The result is that each sentence is on its own line in the rendered markdown, which seems unnecessary. I'll run `dprint fmt` locally and push the result.

Strange indeed!

The reason is that ending a line with a double-space (as done in https://github.com/google/comprehensive-rust/pull/1941/commits/cae87dbebe0de26c03e330cbad2da9eeec94bb75) is a way to force a linebreak in Markdown. Ending the line with a backslash is the same and `dprint` normalizes the Markdown to the latter.

I agree that we shouldn't force line breaks here, so thank you for fixing this!

PR #1942, opened 2024-03-25
Saving the playground state in local storage

mani-chand:
#1476 issue. Updated the function call(`getCodeFromPlayground`) in `save-playground.js` file from `pagehide` event to change event in ace editor.


mani-chand:
Hello @djmitche @mgeisler, Look at the changes made.

mani-chand:
> Hello @djmitche, @mgeisler, Look at the changes made.

djmitche (commented):
This makes sense to me, but I'd like @mgeisler to take a look and try it out before we merge again.

theme/save-playgrounds.js
```diff
@@ -23,13 +23,18 @@
     ) {
       let code_block = pre_block.querySelector("code");
       let editor = window.ace.edit(code_block);
-      let code = editor.getValue();
-      codes.push(code);
+      editor.session.on("change", function () {
+        let code = editor.getValue();
+        codes.push(code);
```
mgeisler, line 28: Hi @mani-chand, thanks for looking more at this!

Will this not push the text into `codes` on every change to any editor on the page? That sounds wrong to me: I would expect the `change` event to update a single item in the `codes` list.

More importantly, how have you tested this? It was a little hard to trigger the problem with the old code, so it's important that we find a way to improve the testing of any new code.

That will probably be something simple like "I ran `mdbook serve --open`, and browsed back and forth 10 times between two pages". If that "procedure" can reproduce the old error, then we can use it to test the new code.

mani-chand: Hello @mgeisler , I think, I did the changes you wanted. Could you verify them and suggest any changes if needed.

About testing, In old code , we used  ` javascript event listener` to save the code , now, we are using `ace` editor's `change` event this time it won't  be a problem.

I did normally checked in my local , It worked good for me . Just check at your local and let me know.

mgeisler: Hi @mani-chand, thanks for the update, I'll take a look in a minute.

An important first question is: are you able to reliably reproduce the error _before_ this fix?

mani-chand: No, But i think the error came because of closing window/tab directly changing the code.

mani-chand: > Hi @mani-chand, thanks for the update, I'll take a look in a minute.
> 
> An important first question is: are you able to reliably reproduce the error _before_ this fix?
> 
> 

But if you want me to do, I will try and reproduce the error.

mani-chand: As per https://github.com/google/comprehensive-rust/pull/1935#issuecomment-2016969529

Hey @mgeisler , I have used the old code (#1917) and tried to reproduce the error(Not saving the code on closing the page/tab/window).But I am not able to replicate it.

As per old code : 
> When you reset the playground it will clear all including the current page(But it didn't reset the current page playground code.) but when you leave the current page `pagehide` event triggers and saves the state even there is no change in the code.

mgeisler: > But I am not able to replicate it.

Okay, that is interesting :slightly_smiling_face: 

The problem I saw with the old code was that the pages would become empty after a while. I noticed this while browsing the rendered book — the next day some of the students told me that they had also noticed the missing code blocks.

If we cannot reproduce the old error, then I don't feel confident in a PR that tries to fix it. The problem is that we won't know if the PR is really fixing the error if we don't know how to reproduce the error.

It would be helpful if you can figure out how to reliably reproduce the error and then test the same procedure on the new code.

mani-chand: Ok, I got it.

mani-chand: > > But I am not able to replicate it.
> 
> Okay, that is interesting :slightly_smiling_face: 
> 
> The problem I saw with the old code was that the pages would become empty after a while. I noticed this while browsing the rendered book — the next day some of the students told me that they had also noticed the missing code blocks.
> 
> If we cannot reproduce the old error, then I don't feel confident in a PR that tries to fix it. The problem is that we won't know if the PR is really fixing the error if we don't know how to reproduce the error.
> 
> It would be helpful if you can figure out how to reliably reproduce the error and then test the same procedure on the new code.
> 

Hello , It may take time to complete this because of my final exams.

mani-chand: Hello @mgeisler , Found the bug 🎉 . All non-editable exercise playground codes went missing/cleared . It is may be because of `ACE API ` i am trying to replace the playground code with the same code in non-editable playground it get cleared. (because of `pagehide` event excercise code will save in localstorage and when we go back to the exercise again `savePlayground` function will try to replace the same code but because of non-editable playground coide will get cleared).

Can you please verify they and confirm me.

mani-chand: > Hello @mgeisler , Found the bug 🎉 . All non-editable exercise playground codes went missing/cleared . It is may be because of `ACE API ` i am trying to replace the playground code with the same code in non-editable playground it get cleared. (because of `pagehide` event excercise code will save in localstorage and when we visit the exercise again `savePlayground` function will try to replace the same code but because of non-editable playground coide will get cleared).
> 
> Can you please verify that and confirm me.

To reproduce the issue :
  > Go to a non - editable playground and refresh the page 
or
 >  Go to a non - editable playground and  go to next page and visit again.

mani-chand:
Hello @mgeisler, Can you give me a update on it.

mgeisler:
> Hello @mgeisler, Can you give me a update on it.

Hi @mani-chand, I haven't forgotten about this — I spent some time looking at it today and I think you're right that there is a problem with the pages that have non-editable playgrounds. However, I don't understand what the problem is... I don't see any exceptions in the JavaScript console, but I do see the code block disappear.

This makes me nervous about changing this code more. I will look more next week!

mani-chand:
> > Hello @mgeisler, Can you give me a update on it.
> 
> However, I don't understand what the problem is... I don't see any exceptions in the JavaScript console, but I do see the code block disappear.
> 

Hello @mgeisler , its because of pagehide event excercise code will save in localstorage and when we visit the exercise(non editable playground) again `savePlayground` function will try to replace the same code but because of non-editable playground code will get cleared.

mgeisler (approved):
Thanks for sticking with it! I tested it locally and it works really well :tada:  Let us get it merged and out into the hands of people so we can get it tested widely.

PR #1944, opened 2024-03-26
Update move.md

cdman:
Fix the String length (and capacity) to make it less confusing.


fw-immunant (approved):

djmitche (approved):
Well, that sure was confusing!

PR #1946, opened 2024-03-28
Fix example of niche optimization in Smart Pointers / Box<T>

DmitryMilk:
Example with linked list replaced by example with Option


djmitche (approved):
Nice!

PR #1961, opened 2024-04-07
Run `mdbook test` on translated code

mgeisler:
I made a small mistake during the refactor in #1492: I removed the `MDBOOK_BOOK__LANGUAGE` environment variable, which means that we ended up testing the original Rust code snippets again and again.


po/ko.po
```diff
@@ -3275,7 +3275,7 @@ msgstr "`&str` - 읽기 전용 문자열입니다. 문자열 리터럴은 이 
 
 #: src/types-and-values/strings.md
 msgid "\"Greetings\""
-msgstr "\"인사말\""
+msgstr "\"  인사말\""
```
jiyongp, line 3278:
```suggestion
msgstr "\"인사말\""
```

deavid (commented):
LGTM for changes in "es.po" file.

po/ko.po
mgeisler: Thanks @jiyongp for the careful review!

This change is unfortunately on purpose... it's essentially a bug in the [version of the page](https://google.github.io/comprehensive-rust/ko/types-and-values/strings.html). The problem is that we demonstrate slicing, so the example is

```rust
    let greeting: &str = "인사말";
    let planet: &str = "🪐";
    let mut sentence = String::new();
    sentence.push_str(greeting);
    sentence.push_str(", ");
    sentence.push_str(planet);
    println!("{:?}", &sentence[0..5]);
```

The `0..5` part is outside the control of you as a translator, but it ends up slicing into a character of `"인사말"`. Adding the spaces "fixes" this.

The [current page](https://google.github.io/comprehensive-rust/references/strings.html) uses different strings and different indices, but it will likely end up with the same problem.

A solution would be to use `let s3: &str = &s2[s2.len() - s1.len()..];` in the slicing example. That way the byte lenghts will update correctly with the translations.

mgeisler: I created #1963 to fix this in the English version.

jiyongp: Ack. Thanks for the detailed explanation. The rest of ko.po LGTM.

mgeisler: Thanks, that's good to hear!

mgeisler:
I've enabled auto-merge here, so if someone approves the PR, then we can get it merged and fix the broken code snippets.

jiyongp (approved):

PR #1962, opened 2024-04-07
Only save CI cache on `main` branch

mgeisler:
When a PR restores the cache, it will find the most recent cache from the `main` branch. This cache should be mostly up to date.

We are constantly at the 10 GB limit for our CI caches, so removing the extra caches from the many PR branches should ensure that we always have space for the `main` branch caches.


mgeisler:
Seems to be working, here is the Bengali translation which won't save the cache:
image

Compared with the English translation, which does save the cache:
image

Also, the tests that don't set `matrix.language` save the cache because of the check for `''`:
image

qwandor (approved):

PR #1964, opened 2024-04-07
Use larger GitHub runners for build actions

mgeisler:
We apparently have access to larger runners on the GitHub Enterprise plan.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ env:
 
 jobs:
   format:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -30,7 +30,7 @@ jobs:
         uses: dprint/check@v2.2
 
   typos:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


mgeisler:
The [standard GitHub-hosted runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) have 4 cores for Linux.

Using the 64-core runner removes ~1 minute for the per-language builds. Compare

- [Before, #1963](https://github.com/google/comprehensive-rust/actions/runs/8587866099/job/23532030623?pr=1963)
- [After, #1964](https://github.com/google/comprehensive-rust/actions/runs/8587927053/job/23532170916?pr=1964)

So the time per language shrinks from about 4.5 minutes to 3.5 minutes. This is not super impressive and I guess this is mainly because we have lots of serial work in our jobs: building the and testing the translation takes the same amount of time in both cases. It's basically only the `cargo install` steps that improve because of much better parallelism.

I'm now testing the publish workflow, but I expect similar results since this also mostly consist of single-threaded steps.

mgeisler:
The two publish jobs:

- [Before](https://github.com/google/comprehensive-rust/actions/runs/8587315790/job/23530810073): 17m 16s in total, 23s used for uploading the page.
- [After, this PR](https://github.com/google/comprehensive-rust/actions/runs/8588011201/job/23532347598): 17m 6s in total, without uploading the page.

I think this means that it's not worth using the larger runners for us.

mgeisler:
cc @qwandor, @djmitche, @randomPoison in case you had been wanting to experiment this this.

PR #1970, opened 2024-04-08
fr: removing semi-deleted former translations.

AdrienBaudemont:
fr: removing semi-deleted former translations.


mgeisler (approved):
Merci!

PR #1971, opened 2024-04-09
tr: day 1 morning translation

Enes1313:
Translated day-1-morning
Part of #500

### DIFF
--- a/po/tr.po
+++ b/po/tr.po
@@ -3342,7 +3342,7 @@ msgstr "\"büyükçe\""
 
 #: src/control-flow-basics/if.md
 msgid "\"huge\""
-msgstr "\"büyük\""
+msgstr "\"aşırı büyük\""
 
 #: src/control-flow-basics/if.md
 msgid ""
--- a/po/tr.po
+++ b/po/tr.po
@@ -1,8 +1,8 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: Comprehensive Rust 🦀\n"
-"POT-Creation-Date: 2024-03-03T14:32:49+03:00\n"
-"PO-Revision-Date: 2024-03-03 16:36+0300\n"
+"POT-Creation-Date: 2024-04-07T05:11:51+03:00\n"
+"PO-Revision-Date: 2024-04-09 12:31+0300\n"
 "Last-Translator: akerem@protonmail.com\n"
 "Language-Team: Turkish <gnome-turk@gnome.org>\n"
 "Language: tr\n"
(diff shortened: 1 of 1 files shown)


mgeisler:
Thanks @Enes1313! Let me try and find someone who can help review this!

Enes1313:
If you don't mind, I told a few friends to "come and review". I don't know when they will arrive :) I think you will find someone until one of them comes :)

po/tr.po
```diff
@@ -3220,193 +3228,235 @@ msgid ""
...
 
 #: src/control-flow-basics/if.md
 msgid "\"biggish\""
-msgstr ""
+msgstr "\"büyükçe\""
 
 #: src/control-flow-basics/if.md
 msgid "\"huge\""
-msgstr ""
+msgstr "\"büyük\""
```
alerque, line 3345: This probably shouldn't be the same as the translation for 'large'. Consider 'kocaman'.

po/tr.po
```diff
@@ -3882,18 +3955,18 @@ msgid ""
 "covered later."
 msgstr ""
```
alerque, line 3956: A number of strings are not translated here. I don't know if 100% coverage is required or what, but the PR says "fully" and there are a good chunk of empty strings near here.

alerque (commented):
I did not review this thoroughly, just skimmed through it and did some smoke checks. Most of what I landed on seemed pretty good, but there are quite a few empty strings.

po/tr.po
Enes1313: Thank you. Yes, it shouldn't be the same as "large" but I couldn't find a nice Turkish word instead. The word "kocaman" doesn't sound good. Maybe I could choose words like "dev" or "devasa". I need to consult a few people about this. Let me make a change based on the answers I get. I will make changes as soon as possible according to the answers I receive.

po/tr.po
alerque: Okay got it. I wasn't following where all the pieces were coming from or what the expectations on completeness were, just spot checking.

Kolay gelsin.

Enes1313: Teşekkür ederim :)

mgeisler: Empty strings are okay in a PO file — that just means that the entry isn't translated yet. The English text will be used in the interim.

Enes1313:
@alerque can you review again, please

alerque:
From a quick once over it looks good.

By the way in the future it would be a *LOT* easier to review PRs if you did multiple commits and first added a commit that updated the existing PO file with the current message ids and updated source, then after that have a commit that only adds/changes the actual translations. This PR is pretty hard to weed through what is changes that came in based on PO source updates and what was actually done as translation work.

Enes1313:
Oh, I have understood. Thank you.

alerque:
I'm not a maintainer and the review approval mechanism is disabled on this project for non-maintainers otherwise I would use it.

mgeisler (approved):
Thanks @Enes1313 for the work here and thanks @alerque for reviewing it!

alerque:
Sure.

mgeisler:
> Sure.

Thanks! You've now been invited and should be able to approve future PRs! You're welcome to add yourself to the `.github/CODEOWNERS` file so that you're automatically assigned to the Turkish PRs.

PR #1973, opened 2024-04-11
ru: translate days 1-2 section titles

1dimir:
* Add translations for the new extracted messages related to Day 1and Day 2 chapters;
* Fix numbering in translation of Day 4 titles;
* Use PO file translator comments for notes and reasons why one or another choice was made, mention alternatives;
* Do not translate type and trait names;
* Rename: Поток управления -> Управляющие конструкции;
* Rename: Перечисления и сопоставление с образцом. -> Деструктуризация;
* Rename: Ссылки -> Заимствование;
* Rename: Висячие ссылки -> Изменяемые ссылки;
* Rename: Вторая половина дня -> После обеда;
* Rename: Поток управления -> `Let` в потоке управления (src: Let Control Flow);
* Rename: Функции -> Обобщённые функции (src: Generic Functions);
* Rename: Обобщения (generics). -> Обобщённые типажи (src: Generic Traits);
* Cannot translate `Slices: &[T]` - possible bug

#326


1dimir:
I'm looking into failed `dprint` check.

1dimir:
> I'm looking into failed `dprint` check.

Resolved

po/ru.po
```diff
@@ -56,6 +56,8 @@ msgstr "День 1: Утро"
 msgid "Welcome"
 msgstr "Добро пожаловать"
 
+# There should be two separate keys (or additional context): Hello Rust and Hello, World.
```
mgeisler, line 59: Thanks to @dyoo, it is actually possible to embed translation comments in the Markdown files. See https://github.com/google/mdbook-i18n-helpers/pull/177 where he adds the documentation to our tooling.

I don't think we support setting the [translation context](https://www.gnu.org/software/gettext/manual/html_node/Contexts.html) yet. @dyoo, do you know?

@1dimir, if you have suggestions for changes to the English text (which is far from perfect!), then feel free to submit PRs.

dyoo: Yeah, we haven't added that, though I suspect it can follow a similar pattern to how we're doing comments.

So probably a sketch of supporting this would be:

* Define another comment directive to let authors specify the ctxt, similar to what we're doing [here](https://github.com/google/mdbook-i18n-helpers/blob/main/i18n-helpers/src/directives.rs#L17-L35)

* Collect and thread that state when we're [gathering groups](https://github.com/google/mdbook-i18n-helpers/blob/main/i18n-helpers/src/lib.rs#L154-L171) for translation.

* Finally, pass the `ctxt` [here](https://github.com/google/mdbook-i18n-helpers/blob/main/i18n-helpers/src/xgettext.rs#L43) and call the builder method for [setting the `ctxt`](
https://docs.rs/polib/latest/polib/message/struct.MessageBuilder.html#method.with_msgctxt).

So I think this would be fairly straightforward extension.

mgeisler: Thanks @dyoo, I took the above and used it to create https://github.com/google/mdbook-i18n-helpers/issues/184.

po/ru.po
```diff
@@ -122,51 +133,53 @@ msgstr "Выражение `if`"
 
 #: src/SUMMARY.md src/control-flow-basics/loops.md
 msgid "Loops"
-msgstr ""
+msgstr "Циклы"
 
 #: src/SUMMARY.md src/control-flow-basics/loops/for.md
 msgid "`for`"
-msgstr ""
+msgstr "`for`"
```
Makarevich, line 140: @mgeisler Does this actually need to be "translated"? The `msgid` and `msgstr` are exactly the same.

mgeisler: Making them the same is the correct way to say `for` should be remain unchanged as `for`.

You're correct that leaving the entry empty will have the same effect — but the entry will then show up as untranslated in various statistics. So it's better to do what @1dimir did here and translate it to the same thing.

dyeroshenko (approved):
NIT: For future PRs, can we try to have 1-2 commits per PR? It's just more convenient to review smaller changes. 

Other than that, LGTM and huge thank you!

PR #1976, opened 2024-04-11
Move generics to day 2 afternoon

randomPoison:
I'm consistently finding that pattern matching and methods/traits section fills up the morning, and I don't get to generics until the afternoon. This just moves the generics section to the afternoon to be a bit more accurate. This makes the estimated times a bit wonky, but I think the std types and traits sections are a bit over-estimated currently, and I find I have enough time in the afternoon to cover those and generics. At some point I'd like to go through and make those sections a bit more structured and make the time estimates more accurate.


mgeisler (approved):
Sounds good to me! Thanks for keeping the timing information in mind — it would be very useful to update it based on a recording at some point.

PR #1977, opened 2024-04-11
Simplify if-let example

randomPoison:
I find that `if let` makes the most sense to use when you don't have an `else` case, otherwise it's generally clearer to express the same thing with a `match`. This changes the `if let` example to be (arguably) a bit more idiomatic and less verbose.


djmitche (approved):
I suppose students will not assume that `else` isn't supported here, so this is fine!

PR #1980, opened 2024-04-11
Remove redundant trait bounds from counter exercise

randomPoison:
The trait bounds aren't needed on the struct definition, only the impl block. I think it'd be useful to show the difference here in order to show students how trait bounds for collection types are usually on the impl blocks rather than the type itself.


djmitche (approved):
Would it make sense to call this out in the speaker notes?

PR #1981, opened 2024-04-12
Add safety comments and use consistent format for existing ones.

qwandor:
We should have safety comments on all `unsafe` blocks, to set a good example.


djmitche (approved):
Good idea!

PR #1986, opened 2024-04-14
Add missing import for try-conversions.md example

ivanbuper:
Uncommenting the "write" line causes the example to fail compilation because the compiler can't find `fs`.


djmitche (commented):
Thanks! That will generate a warning regarding the unused import, though. Maybe it's better to use the full path in the out-commented code?

ivanbuper:
@djmitche Done, PTAL!

djmitche (approved):
Thanks!

PR #1995, opened 2024-04-15
Exercise has a single function to implement

### LINKED ISSUE #1993: Inconsistent text in CR2
"Hard-code both functions to operate on 3 × 3 matrices." in **tuples-and-arrays/exercise.md** text should change. In the relevant file before CR2, the implementation of two functions was requested.

djmitche:
Fixes #1993.


mgeisler (approved):
Thanks! The exercise used to ask you to implement a pretty print function too, but that's now gone.

Issue #1305, opened 2023-10-05
Evaluate translation platforms


mgeisler:
We have so far let translators come up with their own workflows. This has worked remarkably well, despite having to juggle huge PO files with around 20k lines.

There are alternatives to editing the PO files locally: there exists several different online translation platforms which offers a better interface plus automation. A quick search finds sites such as

- https://crowdin.com/
- https://transifex.com/
- https://lokalise.com/
- https://poeditor.com/

There are many others, so this is just a selection.

# Features

- Support for PO files:
  - It is important that they should not mangle the POT-Creation-Date header we rely on (#1243).
  - An easy way to upload new PO files, either automatically or manually.
- Free for open source projects.
- Ability to contribute changes back to this repository via a pull request.
  - I don't think we are allowed to give direct write-access to the platforms, so a PR will be ideal.

I think all of the ones listed above support these features.

I would like someone to evaluate the different platforms and some up with a suggestion for how we can use one. The goal is to make life easier for our translators.

mgeisler:
@henrif75, we talked about this at some point in the past, but I figured we could try and crowd-source this.

henrif75:
I put together a [simple spreadsheet](https://docs.google.com/spreadsheets/d/1uetSQ2zE-6YyKtqvZvE5gbS4qw4VrakRBMpNJr5Op-0/edit?usp=sharing) to collect the information.

mgeisler:
Hi @henrif75, I talked with @guardsmanpanda today about this topic: Bjørn has been looking into this as well and I hope we can move forward with trying out one of the platforms.

@deavid, feel free to chime in if you have input as well :smile:

GuardsmanPanda:
@deavid Later (this week I hope) I'll post some thoughts and a direction for further investigation / testing, but I am leaning towards suggesting that all translation work happens in the tool, it is much more convenient when sentences can be updated / approved individually.

But, there is still some thinking and tooling work to be done here, for example, right now the translation keys are just the English sentence, so even minor wording or grammar changes would be a 'new' translation key (save for some fancy matching magic), is that a problem we need to fix?

I am imagining the flow as follows.
1) Change to the course.
2) Push new phrases to the translation tool.
3) Update translations in the translation tool.
4) CI process rebuilds the course with the latest translation, grabbed directly from the tool API.

Not sure if step 2 happens auto magically or manually with human oversight, I lean towards the latter since I doubt we will make many course updates that break translations.

[...]

PR #1999, opened 2024-04-17
Add @alerque as a reviewer for the Turkish translation

alerque:
See invitation here: https://github.com/google/comprehensive-rust/pull/1971#issuecomment-2061199545

Also note that @enes1313 is marked as a code owner, but they are *not* a collaborator on the repository so the code owner features are non-functional and this is considered an error by the GH parser. Obviously there are two possible resolutions to that...


mgeisler (approved):
Welcome onboard!

mgeisler:
> Also note that @Enes1313 is marked as a code owner, but they are _not_ a collaborator on the repository so the code owner features are non-functional and this is considered an error by the GH parser. Obviously there are two possible resolutions to that...

Thanks for pointing that out! I've fixed it by inviting @Enes1313 to the repository. I hope to move to a translation platform in the future (#1305), but right now we're just letting people maintain the PO files here.

PR #2000, opened 2024-04-17
Add examples of the borrow checker catching errors

randomPoison:
I think it'd be helpful to actually demonstrate to students how the "sharing XOR mutability" rule actually prevents errors in practice, since right now we explain the rule but don't give much context as to why the rule is important.

### DIFF
--- a/src/borrowing/examples.md
+++ b/src/borrowing/examples.md
@@ -1,0 +1,37 @@
+---
+minutes: 3
+---
+
+# Borrow Errors
+
+As a concrete example of how these borrowing rules prevent memory errors,
+consider the case of modifying a collection while there are references to its
+elements:
+
+```rust,editable,compile_fail
+fn main() {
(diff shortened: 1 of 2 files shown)


mgeisler (approved):
Excellent example! I printed the `vec.capacity()` before and after and it changes from 5 to 10. So the example is extra good because the backing array is really deallocated by the push.

src/borrowing/examples.md
```diff
@@ -0,0 +1,33 @@
+# Borrow Errors
```
djmitche, line 1: This will need timing information, similar to other slides. Maybe 3 minutes?

djmitche (changes requested):
Agreed with @mgeisler regarding the content.

src/borrowing/examples.md
randomPoison: Added!

djmitche (approved):

PR #2001, opened 2024-04-17
Simplify the example code for `Clone`

randomPoison:
The previous example code for the `Clone` slide was a bit too complex in a way that obscured the fundamental point. I've replaced it with the `say_hello` example from the previous slide, but updated to demonstrate how cloning can address the borrow checker error. I also added a speaker note to mention that `Clone` performs a deep copy, which might be different from what students are used to if they come from a language like Python that does shallow copies by default.

### DIFF
--- a/src/memory-management/clone.md
+++ b/src/memory-management/clone.md
@@ -8,26 +8,29 @@ Sometimes you _want_ to make a copy of a value. The `Clone` trait accomplishes
 this.
 
 ```rust,editable
-#[derive(Default)]
-struct Backends {
-    hostnames: Vec<String>,
-    weights: Vec<f64>,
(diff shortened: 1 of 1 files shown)


mgeisler (approved):
Thanks for removing the complexity here!

I was considering adding an extra speaker note highlightin that `clone` is user-defined and thus you could also make a type do a shallow clone — such as `Rc` and `Arc`.

djmitche (approved):

randomPoison:
> I was considering adding an extra speaker note highlightin that `clone` is user-defined

Ah good idea, I added an additional speaker note to call that out.

djmitche (approved):

PR #2002, opened 2024-04-17
Slightly simplify binary tree exercise

randomPoison:
Give students a little more context for the binary tree exercise by giving them the wrapper methods on `BinaryTree` at the start and explicitly asking them to implement the methods on `Subtree`. I think this simplifies the exercise a bit and makes it a bit more focused for students.


djmitche (approved):
Nice fix, thanks!

PR #2005, opened 2024-04-18
Use tables to summarize course content

### LINKED ISSUE #1982: Timing information enters the translations
Hi @djmitche, I noticed in #1974 that the timing information enters the translations. This is in some sense working as intended, but I fear that it creates extra work for the translators.

On [Welcome to Day 1](https://google.github.io/comprehensive-rust/welcome-day-1.html#welcome-to-day-1), as an example, if we would format

- [Welcome](https://google.github.io/comprehensive-rust/welcome-day-1.html) (5 minutes)
- [Hello, World](https://google.github.io/comprehensive-rust/hello-world.html) (15 minutes)
- [Types and Values](https://google.github.io/comprehensive-rust/types-and-values.html) (40 minutes)
- [Control Flow Basics](https://google.github.io/comprehensive-rust/control-flow-basics.html) (40 minutes)

as a table

| Page | |
| --- | ---:|
| Welcome | 5 minutes |
| Hello, World | 15 minutes |
| Types and Values | 40 minutes |
| Control Flow Basics | 40 minutes |

then the page titles would be extracted on their own — and they would match the title we already extract from each page. So it would reuse the translation for the page, which is kind of neat :slightly_smiling_face: 

I don't have a good idea for how to include the page link like in the current formatting. We include the link in the translatable message, so this would defeat the attempt at unifying the message here with the page title.

djmitche:
This is more friendly to translation (as it can share the translation of the title).

This fixes #1982.

### DIFF
--- a/mdbook-course/src/course.rs
+++ b/mdbook-course/src/course.rs
@@ -255,7 +255,7 @@ impl Course {
                 duration(session.minutes())
             )
             .unwrap();
-            let mut segments = Table::new(["Segment".into(), "Time".into()]);
+            let mut segments = Table::new(["Segment".into(), "Duration".into()]);
             for segment in session {
                 // Skip short segments (welcomes, wrap-up, etc.)
                 if segment.minutes() == 0 {
@@ -308,7 +308,7 @@ impl Session {
     /// Generate a Markdown outline for this session, for placement at the given
     /// path.
     pub fn outline(&self) -> String {
-        let mut segments = Table::new(["Segment".into(), "Time".into()]);
+        let mut segments = Table::new(["Segment".into(), "Duration".into()]);
         for segment in self {
             // Skip short segments (welcomes, wrap-up, etc.)
             if segment.minutes() == 0 {
@@ -318,7 +318,7 @@ impl Session {
         }
         format!(
             "Including {BREAK_DURATION} minute breaks, this session should take about {}. It contains:\n\n{}",
-            duration(self.minutes()),segments)
+            duration(self.minutes()), segments)
     }
 
     /// Return the total duration of this session.
@@ -383,7 +383,7 @@ impl Segment {
     }
 
     pub fn outline(&self) -> String {
-        let mut slides = Table::new(["Slide".into(), "Time".into()]);
+        let mut slides = Table::new(["Slide".into(), "Duration".into()]);
         for slide in self {
             if slide.minutes() == 0 {
                 continue;
--- a/mdbook-course/src/course.rs
+++ b/mdbook-course/src/course.rs
@@ -36,10 +36,10 @@
 //! top-level item are treated as further slides in the same segment.
 
 use crate::frontmatter::{split_frontmatter, Frontmatter};
-use crate::markdown::{duration, relative_link};
+use crate::markdown::{duration, Table};
 use mdbook::book::{Book, BookItem, Chapter};
 use std::fmt::Write;
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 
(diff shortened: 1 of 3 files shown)


mdbook-course/src/course.rs
```diff
@@ -394,28 +382,19 @@ impl Segment {
         self.into_iter().map(|s| s.minutes()).sum()
     }
 
-    pub fn outline(&self, at_source_path: impl AsRef<Path>) -> String {
-        let mut outline = String::from("In this segment:\n");
+    pub fn outline(&self) -> String {
+        let mut slides = Table::new(["Slide".into(), "Time".into()]);
```
mgeisler, line 386: Would it be slightly more accurate to call this a "duration"?

```suggestion
        let mut slides = Table::new(["Slide".into(), "Duration".into()]);
```

mdbook-course/src/course.rs
```diff
@@ -313,24 +307,18 @@ impl Session {
...
-                duration(segment.minutes())
-            )
-            .unwrap();
+            segments.add_row([segment.name.clone(), duration(segment.minutes())]);
         }
-        writeln!(&mut outline,"\nIncluding {BREAK_DURATION} minute breaks, this session should take about {}", duration(self.minutes())).unwrap();
-        outline
+        format!(
+            "Including {BREAK_DURATION} minute breaks, this session should take about {}. It contains:\n\n{}",
+            duration(self.minutes()),segments)
```
mgeisler, line 321: I think your `rustfmt` has given up here :smile: 

```suggestion
            duration(self.minutes()), segments)
```

mdbook-course/src/course.rs
```diff
@@ -313,24 +307,18 @@ impl Session {
 
     /// Generate a Markdown outline for this session, for placement at the given
     /// path.
-    pub fn outline(&self, at_source_path: impl AsRef<Path>) -> String {
-        let mut outline = String::from("In this session:\n");
+    pub fn outline(&self) -> String {
+        let mut segments = Table::new(["Segment".into(), "Time".into()]);
```
mgeisler, line 311: Perhaps "duration" is better here?

```suggestion
        let mut segments = Table::new(["Segment".into(), "Duration".into()]);
```

mdbook-course/src/course.rs
```diff
@@ -245,32 +245,26 @@ impl Course {
...
         for session in self {
             writeln!(
                 &mut outline,
-                " * {} ({}, including breaks)",
+                " * {} ({}, including breaks)\n",
                 session.name,
                 duration(session.minutes())
             )
             .unwrap();
+            let mut segments = Table::new(["Segment".into(), "Time".into()]);
```
mgeisler, line 258:
```suggestion
            let mut segments = Table::new(["Segment".into(), "Duration".into()]);
```

mdbook-course/src/markdown.rs
```diff
@@ -57,6 +58,46 @@ pub fn duration(mut minutes: u64) -> String {
     }
 }
 
+/// Table implements Display to format a two-dimensional table as markdown,
+/// following https://github.github.com/gfm/#tables-extension-.
+pub struct Table<const N: usize> {
```
mgeisler, line 63: Nice! If you have time, it would be nice to have a small unit test below to show how the output looks like.

mgeisler (approved):
Thanks @djmitche!

mdbook-course/src/course.rs
djmitche: I think it doesn't do much in macro invocations? I was surprised by this too.

Issue #1536, opened 2023-11-30
Reorganize concurrency course


djmitche:
* Organize similar to Fundamentals: shorter segments each ending with an exercise
* Put the entire course under `src/concurrency` (it's currently under `src/{concurrency,async}`)
* Move exercises into the source directories as done with Fundamentals
* More, simpler exercises
* Try to factor out the "pitfalls" segment of the async session
* Cover sync primitives like Mutex (since they're used in dining philosophers)

fw-immunant:
I'm open to discussion, but I think the current format of the concurrency course works really well. What makes you feel that shorter segments with more exercises would be an improvement?

What other designs have you thought of for the "pitfalls" segment? One additional pitfall I usually discuss is lock poisoning, and how the stdlib embraced a pattern that adds a lot of noise to enable robustness while Tokio rejected the concern in favor of a simpler interface. That said, I usually discuss that topic with the solutions for the async philosophers exercise, where it's worthwhile to diff the solution against the threads-based implementation to see the reason for each difference.

We do presently have a slide about `Mutex<T>`.

djmitche:
> I'm open to discussion, but I think the current format of the concurrency course works really well. What makes you feel that shorter segments with more exercises would be an improvement?

Only that that was a major part of the v2 work on Fundamentals, and I think that's been well-received. It would still have "sync morning" and "async afternoon", and roughly the same number of segments, just with a smaller exercise after each segment. My sense is that these exercises are "too big" and take students wildly different amounts of time to solve, so optimizing for shorter, better-encapsulated exercises would help.

> What other designs have you thought of for the "pitfalls" segment? One additional pitfall I usually discuss is lock poisoning, and how the stdlib embraced a pattern that adds a lot of noise to enable robustness while Tokio rejected the concern in favor of a simpler interface. That said, I usually discuss that topic with the solutions for the async philosophers exercise, where it's worthwhile to diff the solution against the threads-based implementation to see the reason for each difference.

SUMMARY for my WIP is [here](https://github.com/djmitche/comprehensive-rust/blob/issue1536/src/SUMMARY.md). Basically it's spreading the "pitfalls" out across the places where they are relevant, instead of spending 45 minutes essentially talking exclusively about how awful async is.

Good point about mutex poisoning!

> We do presently have a slide about `Mutex<T>`.

---

[...]

PR #2007, opened 2024-04-19
Update Concurrency course witih times

djmitche:
As I mentioned in #1536:

 * Break into segments at approximately the places @fw-immunant put breaks
 * Move all of the files into `src/concurrency`
 * Add timings and segment/session metadata so course outlines appear

There's room for more work here, including some additional feedback from @fw-immunant after the session I observed, but let's do one step at a time :)


mdbook-course/src/replacements.rs
```diff
@@ -35,10 +35,10 @@ pub fn replace(
         return;
     };
     chapter.content = DIRECTIVE
-        .replace(&chapter.content, |captures: &regex::Captures| {
+        .replace_all(&chapter.content, |captures: &regex::Captures| {
             let directive_str = captures[1].trim();
             let directive: Vec<_> = directive_str.split_whitespace().collect();
-            match directive.as_slice() {
+            match dbg!(directive).as_slice() {
```
mgeisler, line 41: This is probably a forgotten `dbg!`?

src/SUMMARY.md
```diff
@@ -353,46 +353,50 @@
...
+- [Async Basics](concurrency/async.md)
+  - [`async`/`await`](concurrency/async/async-await.md)
+  - [Futures](concurrency/async/futures.md)
+  - [Runtimes](concurrency/async/runtimes.md)
+    - [Tokio](concurrency/async/runtimes/tokio.md)
+  - [Tasks](concurrency/async/tasks.md)
+- [Channels and Control Flow](concurrency/async/control-flow.md)
+  - [Async Channels](concurrency/control-flow/channels.md)
+  - [Join](concurrency/control-flow/join.md)
+  - [Select](concurrency/control-flow/select.md)
```
mgeisler, line 390: I somehow feel there should be an `async` in the URLs here to highlight that these slides are specific to that sub-topic. What do you think of calling this segment "Async Channels and Control Flow"? The folder would then be `concurrency/async-control-flow/`.

djmitche: That title is too long (won't fit in the sidebar), but I will change the folder name.

mgeisler: Thanks, that sounds good!

src/SUMMARY.md
```diff
@@ -353,46 +353,50 @@
...
+  - [Tasks](concurrency/async/tasks.md)
+- [Channels and Control Flow](concurrency/async-control-flow.md)
+  - [Async Channels](concurrency/async-control-flow/channels.md)
+  - [Join](concurrency/async-control-flow/join.md)
+  - [Select](concurrency/async-control-flow/select.md)
+- [Pitfalls](concurrency/pitfalls.md)
+  - [Blocking the Executor](concurrency/pitfalls/blocking-executor.md)
+  - [`Pin`](concurrency/pitfalls/pin.md)
+  - [Async Traits](concurrency/pitfalls/async-traits.md)
+  - [Cancellation](concurrency/pitfalls/cancellation.md)
```
mgeisler, line 395: Could we also add an `async-` here since the pitfalls are about async Rust? Not sure important, but it might be easier for people to find things when the URL says `async`.

mgeisler (approved):
Thanks so much for taking the time to clean this up and reshuffle it!

PR #2021, opened 2024-04-22
Updated struct and enums in pattern-matching.

mani-chand:
#1464 issue.


src/SUMMARY.md
```diff
@@ -76,7 +76,8 @@
 - [Welcome](welcome-day-2.md)
 - [Pattern Matching](pattern-matching.md)
   - [Matching Values](pattern-matching/match.md)
-  - [Destructuring](pattern-matching/destructuring.md)
+  - [Structs](pattern-matching/Structs.md)
+  - [Enums](pattern-matching/Enums.md)
```
mgeisler, line 80: Please use lower-case letters for the filenames:

```suggestion
  - [Structs](pattern-matching/structs.md)
  - [Enums](pattern-matching/enums.md)
```

src/pattern-matching/Enums.md
```diff
@@ -1,18 +1,6 @@
...
-# Destructuring
-
-Like tuples, structs and enums can also be destructured by matching:
-
-## Structs
-
-```rust,editable
-{{#include ../../third_party/rust-by-example/destructuring-structs.rs}}
-```
+# Enums
```
mgeisler, line 1: Is this missing the `minutes: NN` frontmatter the old file had?

src/pattern-matching/Structs.md
```diff
@@ -0,0 +1,21 @@
+---
+minutes: 8
```
mgeisler, line 2: The minutes here should be split between the two pages: perhaps 4 minutes for each?

mgeisler (changes requested):
Thanks a lot for the continued help here!

src/SUMMARY.md
mgeisler: Even better, please add "Destructuing" to the title here and below:

```suggestion
  - [Structs](pattern-matching/destructuring-structs.md)
  - [Enums](pattern-matching/destructuring-enums.md)
```

Please also make sure to add redirections in `book.toml` to preserve the existing URLs. This is important for search engine optimization (SEO) so people can continue to easily find the information.

src/pattern-matching/Structs.md
mani-chand: Ok.

mani-chand:
> Thanks a lot for the continued help here!

I just took leave in between for my exams. Got happy thinking I am gone😝.

src/SUMMARY.md
mani-chand: Need help with adding redirections in `book.toml` .

src/SUMMARY.md
```diff
@@ -89,10 +90,6 @@
   - [Deriving](methods-and-traits/deriving.md)
   - [Exercise: Generic Logger](methods-and-traits/exercise.md)
     - [Solution](methods-and-traits/solution.md)
-
-# Day 2: Afternoon
-
-- [Welcome](welcome-day-2-afternoon.md)
```
fw-immunant, line 95: Why is this being moved?

mani-chand: oops!, It's by mistake. I will revert it.

src/SUMMARY.md
```diff
@@ -150,7 +151,6 @@
 - [Borrowing](borrowing.md)
   - [Borrowing a Value](borrowing/shared.md)
   - [Borrow Checking](borrowing/borrowck.md)
-  - [Borrow Errors](borrowing/examples.md)
```
mgeisler, line 153: I think this is also lost by mistake?

book.toml
```diff
@@ -180,8 +180,8 @@ use-boolean-and = true
 "ownership/moves-function-calls.html" = "../memory-management/move.html"
 "ownership/shared-unique-borrows.html" = "../borrowing/shared.html"
 "pattern-matching/destructuring-arrays.html" = "../tuples-and-arrays/destructuring.html"
-"pattern-matching/destructuring-enums.html" = "../pattern-matching/destructuring.html"
-"pattern-matching/destructuring-structs.html" = "../pattern-matching/destructuring.html"
+"pattern-matching/destructuring-enum.html" = "../pattern-matching/destructuring-enums.html"
+"pattern-matching/destructuring-struct.html" = "../pattern-matching/destructuring-structs.html"
```
mgeisler, line 184: Please see the comment above the `redirect` table in the TOML file.

In this case you're renaming

- from `src/pattern-matching/destructuring.md`
- to `src/pattern-matching/destructuring-enums.md`

This means that the

- old URL is `pattern-matching/destructuring.html`
- and the new URL is `pattern-matching/destructuring-enums.html`

The `redirect` table is a bit weird, but when you have `somewhere/old.html = elsewhere/new.html`, you instruct `mdbook build` to create a small `somewhere/old.html` file with some JavaScript which will redirect the visitor to `elsewhere/new.html`.

Because the redirect happens when you visit `somewhere/old.html`, it is important to make it relative to `somewhere/old.html`. If you redirect to `elsewhere/new.html`, the final URL becomes `somewhere/elsewhere/new.html`.

In this case, you should redirect like this:

```suggestion
"pattern-matching/destructuring.html" = "destructuring-enums.html"
```

You should be able to double-check this locally by noticing that `mdbook build` outputs the redirection files. I think it might also work in `mdbook serve`?

src/SUMMARY.md
mani-chand: Yes.My system sucks some time alot sorry.

mani-chand:
I think this time it was perfect . Sorry for the trouble.

src/SUMMARY.md
```diff
@@ -76,7 +76,8 @@
 - [Welcome](welcome-day-2.md)
 - [Pattern Matching](pattern-matching.md)
   - [Matching Values](pattern-matching/match.md)
-  - [Destructuring](pattern-matching/destructuring.md)
+  - [Destructuring Structs](pattern-matching/destructuring-structs.md)
+  - [Destructuring Enums](pattern-matching/destructuring-enums.md)
```
djmitche, line 80: #1464 suggests making these sub-slides, but maybe the two slides makes more sense here?

djmitche (commented):
I'm fine with two slides or a slide and sub-slide, as long as the decision is intentional.

src/SUMMARY.md
mani-chand: From last 2 PRs, I have been making it into different slides. At start we used to make them as subslides ,But Not sure about the trasaction.

djmitche: Sounds good -- thanks for updating the timing!

djmitche (approved):

src/SUMMARY.md
mgeisler: Thanks!

mani-chand: Need a update on #1942 PR comment.

PR #2025, opened 2024-04-24
Introduce a `cargo clippy` run

mgeisler:
This might help catch stylistic problems in our Rust code.

Related to #2587.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,55 @@
+name: Lint
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  clippy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+
+      - name: Setup Rust cache
+        uses: ./.github/workflows/setup-rust-cache
+
+      - name: Clippy
+        run: cargo clippy -- -Dwarnings
+
+  format:
+    runs-on: ubuntu-latest
+    env:
+      NIGHTLY_VERSION: nightly-2025-09-01
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+
+      - name: Install formatting dependencies
+        run: |
+          sudo apt update
+          sudo apt install gettext yapf3
+
+      - name: Install pinned nightly for rustfmt
+        run: |
+          rustup toolchain install --profile minimal "$NIGHTLY_VERSION"
+          rustup component add rustfmt --toolchain "$NIGHTLY_VERSION"
+
+      - name: Check formatting
+        uses: dprint/check@v2.3
+
+  typos:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+
+      - name: Check for typos
+        uses: crate-ci/typos@v1.35.7
+        with:
+          config: ./.github/typos.toml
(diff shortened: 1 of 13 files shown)


antoniolinhart (approved):
Nice! This will help make style more consistent

third_party/rust-on-exercism/health-statistics.rs
```diff
@@ -37,12 +37,9 @@ impl User {
             patient_name: &self.name,
             visit_count: self.visit_count as u32,
             height_change: measurements.height - self.height,
-            blood_pressure_change: match self.last_blood_pressure {
-                Some(lbp) => {
-                    Some((bp.0 as i32 - lbp.0 as i32, bp.1 as i32 - lbp.1 as i32))
-                }
-                None => None,
-            },
```
djmitche, line 45: We should probably exclude third_party/ somehow..

djmitche (approved):
I was worried that this might "correct" some things that had been written as-is for didactic reasons, but it looks like those are handled well. It also seems that this isn't scanning code in `rust` blocks in the `.md` files, which probably helps.

third_party/rust-on-exercism/health-statistics.rs
mgeisler: Oh, good idea! Let me figure out how to do that before merging!

michael-kerscher: If the settings are done in the workspace [Cargo.toml](https://github.com/google/comprehensive-rust/blob/ab78a9e6b86b17f370185596bb3cbc63a156b7dd/Cargo.toml) and every Cargo.toml has a `workspace = true` in the `[lints]` section as described in https://doc.rust-lang.org/cargo/reference/workspaces.html#the-lints-table we could tell cargo clippy what to do in the third_party Cargo.toml.

This would enable a central configuration for clippy and gives the option to override specific settings for individual components

mgeisler:
Another thought, as discussed on #2410. We should try and enable Clippy as part of `dprint`. A short look in the README lead me to the [`clippy-driver`](https://github.com/rust-lang/rust-clippy?tab=readme-ov-file#using-clippy-driver), which accepts input on `stdin`.

third_party/rust-on-exercism/health-statistics.rs
mgeisler: > We should probably exclude third_party/ somehow..

Actually... now that I look at this again, I think we should also format/link/check this code. It is in `third_party/` to designate that it _originates_ from a third party project (with their own license terms).

However, we are allowed to modify this code — and we have done so in the past.

mgeisler: @djmitche, if you don't buy the reasoning here, then please let me know and we can reconsider.

> This would enable a central configuration for clippy and gives the option to override specific settings for individual components

Thanks @michael-kerscher, that is an interesting option that I hadn't considered. I would tend to prefer specific tweaks with in-line comments (to keep the tweak close to the code it touches), but you're right that a centralized setup can be nice if we end up with many tweeks.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,55 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+
+      - name: Setup Rust cache
+        uses: ./.github/workflows/setup-rust-cache
+
+      - name: Clippy
+        run: cargo clippy -- -Dwarnings
```
qwandor, line 23: You could consider using `actions-rs/clippy-check`, which will add clippy warnings as GitHub annotations.

djmitche: Good consideration, but note that it can only do so in PRs that aren't from a fork, which isn't how this repo works.

PR #2029, opened 2024-04-26
Add links to the Rust advanced testing workshop

LukeMathWalker:
Following up on https://github.com/google/googletest-rust/issues/376

### DIFF
--- a/src/other-resources.md
+++ b/src/other-resources.md
@@ -47,6 +47,9 @@ A small selection of other guides and tutorial for Rust:
   a series of small presentations covering both basic and advanced part of the
   Rust language. Other topics such as WebAssembly, and async/await are also
   covered.
+- [Testing in Rust: going beyond the basics](https://github.com/mainmatter/rust-advanced-testing-workshop):
+  a self-paced workshop that goes beyond Rust's built-in testing framework. It covers `googletest`, snapshot testing,
+  mocking as well as how to write your own custom test harness. 
 - [Beginner's Series to Rust](https://docs.microsoft.com/en-us/shows/beginners-series-to-rust/)
   and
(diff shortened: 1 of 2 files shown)


mgeisler (approved):
Thanks @LukeMathWalker! If you ever publish the material on a website, then it would be nice to update the links to point there.

Right now, I feel a bit lost when I see the GitHub repository because it wants me to explore the `exercises/` directory without any guidance. Or I can install some tools, but that's not the first thing I want to do when browing training material.

src/other-resources.md
```diff
@@ -47,6 +47,9 @@ A small selection of other guides and tutorial for Rust:
   a series of small presentations covering both basic and advanced part of the
   Rust language. Other topics such as WebAssembly, and async/await are also
   covered.
+- [Testing in Rust: going beyond the basics](https://github.com/mainmatter/rust-advanced-testing-workshop):
```
mgeisler, line 50: I think the name of the workshop is this:

```suggestion
- [Advanced testing for Rust applications](https://github.com/mainmatter/rust-advanced-testing-workshop):
```

LukeMathWalker:
> Thanks @LukeMathWalker! If you ever publish the material on a website, then it would be nice to update the links to point there.

We have that on our to-do list, but no definite date (yet!).
It is somewhat annoying to have to install tooling upfront, but the tooling is indeed what "sheperds" you through the exercises, so to say.

mgeisler:
> We have that on our to-do list, but no definite date (yet!). It is somewhat annoying to have to install tooling upfront, but the tooling is indeed what "sheperds" you through the exercises, so to say.

Yeah, I guess that makes sense! Thanks for adding the links, they should merge as soon as the tests are done.

LukeMathWalker:
> Looks like this needs a `dprint fmt` run -- once that's done, we can merge.

Done!

PR #2039, opened 2024-05-01
fix: rename scoped-threads.md in translations files

vohoanglong0107:
After merging #2007, the file `src/concurrency/scoped-threads.md` was renamed into `src/concurrency/threads/scoped.md`, but the changes were not reflected in all translations files.

There are others file affected by #2007. I will create a dedicated PR for each of these files to make the review process easier.

### DIFF
--- a/po/bn.po
+++ b/po/bn.po
@@ -1081,7 +1081,7 @@ msgstr "Concurrency: সকাল"
 msgid "Threads"
 msgstr "সম্পাদনার সুতো"
 
-#: src/SUMMARY.md src/concurrency/scoped-threads.md
+#: src/SUMMARY.md src/concurrency/threads/scoped.md
 msgid "Scoped Threads"
 msgstr "স্কোপড Threads"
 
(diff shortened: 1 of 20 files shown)


vohoanglong0107:
Sorry, I didn't mean to notify so many people 🙇‍♂️

mgeisler:
Since I have everyone's attention here :smile: The correct way to fix the comments in the `.po` files is to follow the steps under [Refreshing an Existing Translation](https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md#refreshing-an-existing-translation).

You should never have to modify anything but the `msgstr` fields by hand — the `msgid` fields and the comments are only modified by the translation tools.

mgeisler:
Oh, and I should have added this: @vohoanglong0107, thanks for helping out! It's super nice to have people like you join the project to clean up and improve the translations!

There are so many translations now that @henrif75 and I rely heavily on volunteers like you and 50+ other reviewers and translators to keep things working! Thanks everybody!

PR #2041, opened 2024-05-01
vi: fix renamed file scoped_threads.md

vohoanglong0107:
After merging https://github.com/google/comprehensive-rust/pull/2007, the file src/concurrency/scoped-threads.md was renamed into src/concurrency/threads/scoped.md, but the changes were not reflected in all translations files.

There are others file affected by https://github.com/google/comprehensive-rust/pull/2007. I will create a dedicated PR for each of these files to make the review process easier.

### DIFF
--- a/po/vi.po
+++ b/po/vi.po
@@ -1096,7 +1096,7 @@ msgstr ""
 msgid "Threads"
 msgstr ""
 
-#: src/SUMMARY.md src/concurrency/scoped-threads.md
+#: src/SUMMARY.md src/concurrency/threads/scoped.md
 msgid "Scoped Threads"
 msgstr ""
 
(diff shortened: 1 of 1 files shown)


mgeisler:
> the changes were not reflected in all translations files.

The filenames in the `.po` file comments are only used by translators to jump to the definition in the PO editor.

You should not update it by hand: you should only update the comments and the `msgid` fields using `msgmerge` as part of a translation file refresh. See [Refreshing an Existing Translation](https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md#refreshing-an-existing-translation) for how to do this.

mgeisler:
I will close this PR since there is a supported way of updating these comments.

mgeisler:
Cc @henrif75 in case you see other PRs like this. We might need to update our testing and documentation to make this clearer.

PR #2042, opened 2024-05-01
Update break-continue.md

prosper980:
The 'for' was confusing


djmitche:
just waiting on CLA

src/control-flow-basics/break-continue.md
```diff
@@ -9,7 +9,7 @@ If you want to immediately start the next iteration use
 
 If you want to exit any kind of loop early, use
 [`break`](https://doc.rust-lang.org/reference/expressions/loop-expr.html#break-expressions).
-For `loop`, this can take an optional expression that becomes the value of the
+Using the `loop`, this can take an optional expression that becomes the value of the
```
fw-immunant, line 12: "For" was confusing here, but it's not grammatically correct to use "the" to introduce what is effectively a proper noun here. We should probably say "With `loop`, ...".

PR #2043, opened 2024-05-01
es: Course length mistake

salvamunoz:
Edit course length from 3 days to 4 which is actually the total length


deavid:
The original english text does not mention the length. I would recommend to just rearrange the translation to match the same.

Also, you need to sign the CL as mentioned above.

mgeisler:
@salvamunoz, can you take out the mentioning of days like @deavid suggests? Then we can merge this!

po/es.po
```diff
@@ -1217,7 +1217,7 @@ msgid ""
 "course covers the full spectrum of Rust, from basic syntax to advanced "
 "topics like generics and error handling."
 msgstr ""
-"Este es un curso de Rust de tres días que ha desarrollado el equipo de "
+"Este es un curso de Rust de cuatro días que ha desarrollado el equipo de "
```
mgeisler, line 1220: @salvamunoz and @deavid, is this correct Spanish?

```suggestion
"Este es un curso de Rust que ha desarrollado el equipo de "
```

deavid:
Yes, that's correct Spanish

po/es.po
mgeisler: Pushed via command line commit.

mgeisler:
> Yes, that's correct Spanish

Thanks a lot! :tada:

mgeisler (approved):

PR #2044, opened 2024-05-01
Clarify `String` definition

Enes1313:
Changed string definitions in string.md and strings.md files according to discussion #2028

### DIFF
--- a/src/references/strings.md
+++ b/src/references/strings.md
@@ -11,7 +11,7 @@ Including `&str` as a way of representing a slice of valid utf-8
 We can now understand the two string types in Rust:
 
 - `&str` is a slice of UTF-8 encoded bytes, similar to `&[u8]`.
-- `String` has ownership over its heap-allocated buffer of UTF-8 encoded bytes.
+- `String` is an owned buffer of UTF-8 encoded bytes, similar to `Vec<T>`.
 
 <!-- Avoid using fixed integers when slicing since this breaks
 translations. Using the length of s1 and s2 is safe. -->
--- a/src/references/strings.md
+++ b/src/references/strings.md
@@ -11,7 +11,7 @@ Including `&str` as a way of representing a slice of valid utf-8
 We can now understand the two string types in Rust:
 
 - `&str` is a slice of UTF-8 encoded bytes, similar to `&[u8]`.
-- `String` is an owned, heap-allocated buffer of UTF-8 bytes.
+- `String` has ownership over its heap-allocated buffer of UTF-8 encoded bytes.
 
 <!-- Avoid using fixed integers when slicing since this breaks
 translations. Using the length of s1 and s2 is safe. -->
--- a/src/std-types/string.md
+++ b/src/std-types/string.md
@@ -4,7 +4,8 @@ minutes: 10
 
 # String
 
-[`String`][1] represents a widely used heap-allocated growable UTF-8 encoded string buffer:
+[`String`][1] represents a widely used heap-allocated growable UTF-8 encoded
+string buffer:
 
 ```rust,editable
 fn main() {
--- a/src/std-types/string.md
+++ b/src/std-types/string.md
@@ -4,7 +4,7 @@ minutes: 10
 
 # String
 
-[`String`][1] is the standard heap-allocated growable UTF-8 string buffer:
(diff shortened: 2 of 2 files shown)


src/std-types/string.md
```diff
@@ -4,7 +4,8 @@ minutes: 10
 
 # String
 
-[`String`][1] is the standard heap-allocated growable UTF-8 string buffer:
+[`String`][1] represents a widely used heap-allocated growable UTF-8 encoded
+string buffer:
```
fw-immunant, line 8: `String` represents a heap buffer, but the heap buffer itself is not widely used. I think the original "standard" really doesn't need to be said at all as this is a key part of the standard library that students don't need to be explicitly told is common--they'll have seen it many times.

fw-immunant (changes requested):
I hate to bikeshed something that's already been discussed but this prose is technically misleading. Remove the implication that the heap buffer itself is what is widely-used; in general I don't think we need to emphasize the commonness of `String` when describing what it is.

src/std-types/string.md
mgeisler: > I think the original "standard" really doesn't need to be said at all as this is a key part of the standard library that students don't need to be explicitly told is common--they'll have seen it many times.

That is a good point, thanks! I am very much in favor of simplifying the text and removing unnecessary words here.

I originally wrote "standard" to emphasize that `String` is the string type you want. Now that I think about it again, this is more something people would ask if they come from C++: in languages like Python, Java, Kotlin, ... there is much less discussion about which string type to use.

So I'm fine with dropping this point. How about this:

```suggestion
[`String`][1] is a heap-allocated growable UTF-8 encoded
string buffer:
```

src/references/strings.md
```diff
@@ -11,7 +11,7 @@ Including `&str` as a way of representing a slice of valid utf-8
 We can now understand the two string types in Rust:
 
 - `&str` is a slice of UTF-8 encoded bytes, similar to `&[u8]`.
-- `String` is an owned, heap-allocated buffer of UTF-8 bytes.
+- `String` has ownership over its heap-allocated buffer of UTF-8 encoded bytes.
```
mgeisler, line 14: Now that I see the two bullet points together, I feel it would be good for them to have a similar structure: "Foo _is_ X, similar to Y".

So what about this:

```suggestion
- `String` is an owned buffer of UTF-8 encoded bytes, similar to `Vec<T>`.
```

src/std-types/string.md
mgeisler: @Enes1313 and @djmitche, please let us know here and below what you think about the suggestions here.

Enes1313: Some of my thoughts:
"is a" : I think this is weird because in the sentence we call string as "string buffer". If we are going to use "is a", we should use "String is a ... string type" like in the Rust book.
"represent": I think this is okay.
"standard" - "widely used": It may be another sentence. Actually, I don't want to see :) I agree with @fw-immunant .

(My English is not very good)

src/references/strings.md
Enes1313: It is ok for me but I want to make a suggestion. Then the desired one can be selected.

```suggestion
- `String` is a type that has ownership over its heap-allocated buffer of UTF-8 encoded bytes.
```

mgeisler: Since we say "the two string types in Rust" just above, I think adding "is a type" here isn't necessary?

src/std-types/string.md
mgeisler: I'm using "is a" to describe what `String` consists of, like saying "a house is a place where you live and sleep".

We could simplify it more and say

```suggestion
[`String`][1] is a growable UTF-8 encoded string:
```

That avoids the `buffer` jargon and it removes the detail about where it is allocated.

Enes1313: This is ok for me! Thank you

src/references/strings.md
Enes1313: Yes you are right. I missed this.

src/std-types/string.md
djmitche: LGTM too

mgeisler (approved):

mgeisler:
@fw-immunant, can you merge this if you're happy with it?

fw-immunant (approved):

PR #2048, opened 2024-05-03
Add a schematic state machine implementing Future

djmitche:
@fw-immunant generated something like this on-the-fly in teaching the course, and I thought it was great. I think having a _schematic_ understanding of what's going on here helps students through some of the pitfalls. Particularly, it motivates `Pin`, which is where @fw-immunant did this derivation.

### DIFF
--- a/src/concurrency/async/futures.md
+++ b/src/concurrency/async/futures.md
@@ -36,9 +36,11 @@ pause until that Future is ready, and then evaluates to its output.
 - The `Future` and `Poll` types are implemented exactly as shown; click the
   links to show the implementations in the docs.
 
-- We will address `Pin` in the "Pitfalls" segment.
-
 - `Context` allows a Future to schedule itself to be polled again when an event
   such as a timeout occurs.
+
+- `Pin` ensures that the Future isn't moved in memory, so that pointers into
+  that future remain valid. This is required to allow references to remain valid
+  after an `.await`. We will address `Pin` in the "Pitfalls" segment.
 
 </details>
--- a/src/concurrency/async/futures.md
+++ b/src/concurrency/async/futures.md
@@ -36,14 +36,9 @@ pause until that Future is ready, and then evaluates to its output.
 - The `Future` and `Poll` types are implemented exactly as shown; click the
   links to show the implementations in the docs.
 
-- We will not get to `Pin` and `Context`, as we will focus on writing async
-  code, rather than building new async primitives. Briefly:
+- We will address `Pin` in the "Pitfalls" segment.
 
-  - `Context` allows a Future to schedule itself to be polled again when an
-    event occurs.
-
-  - `Pin` ensures that the Future isn't moved in memory, so that pointers into
-    that future remain valid. This is required to allow references to remain
-    valid after an `.await`.
+- `Context` allows a Future to schedule itself to be polled again when an event
+  such as a timeout occurs.
 
 </details>
--- a/src/concurrency/async/state-machine.md
+++ b/src/concurrency/async/state-machine.md
@@ -1,0 +1,154 @@
+---
+minutes: 10
+---
+
+# State Machine
+
+Rust transforms an async function or block to a hidden type that implements
+`Future`, using a state machine to track the function's progress. The details of
+this transform are complex, but it helps to have a schematic understanding of
+what is happening.
+
+```rust,editable
+use std::task::Poll;
+
+/// A simplified version of `std::future::Future`.
+trait Future {
+    type Output;
+    fn poll(&mut self) -> Poll<Self::Output>;
+}
+
+/*
+ * async fn roll_d10() -> u32 { 7 }
+ */
+
+/// Return a random number from 1 to 10.
+fn roll_d10() -> RollD10 {
+    RollD10
+}
+
+struct RollD10;
+
+impl Future for RollD10 {
+    type Output = u32;
+
+    fn poll(&mut self) -> Poll<Self::Output> {
+        Poll::Ready(7)
+    }
+}
+
+/*
+ * async fn two_d10(modifier: u32) -> u32 {
+ *     let first_roll = roll_d10().await;
+ *     let second_roll = roll_d10().await;
+ *     first_roll + second_roll + modifier
+ * }
+ */
+
+/// Sum two D10 rolls plus a modifier.
+fn two_d10(modifier: u32) -> TwoD10 {
+    TwoD10::Init { modifier }
+}
+
+enum TwoD10 {
+    // Function has not begun yet.
+    Init { modifier: u32 },
+    // Waitig for first `.await` to complete.
+    FirstRoll { modifier: u32, fut: RollD10 },
+    // Waitig for second `.await` to complete.
+    SecondRoll { modifier: u32, first_roll: u32, fut: RollD10 },
+}
+
+impl Future for TwoD10 {
+    type Output = u32;
+    fn poll(&mut self) -> Poll<Self::Output> {
+        loop {
+            match self {
+                TwoD10::Init { modifier } => {
+                    let fut = roll_d10();
+                    *self = TwoD10::FirstRoll { modifier: *modifier, fut };
+                }
+                TwoD10::FirstRoll { modifier, fut } => {
+                    if let Poll::Ready(first_roll) = fut.poll() {
+                        let fut = roll_d10();
+                        *self = TwoD10::SecondRoll {
+                            modifier: *modifier,
+                            first_roll,
+                            fut,
+                        };
+                    } else {
+                        return Poll::Pending;
+                    }
+                }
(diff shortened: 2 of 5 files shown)


src/concurrency/async/state-machine/recursion.md
```diff
@@ -0,0 +1,34 @@
+---
+minutes: 3
+---
+
+# Recursion
+
+An async function's future type _contains_ the futures for all functions it
+calls. This means a recursive async functions are not allowed.
```
fw-immunant, line 8: I would clarify this by invoking the analogy with recursive enums: they need an indirection to avoid being infinite-sized types. [Prior to Rust 1.77](https://blog.rust-lang.org/2024/03/21/Rust-1.77.0.html#support-for-recursion-in-async-fn), recursion in `async fn` was forbidden entirely and code had to forgo the `async fn` transform in favor of explicitly returning a `Box<impl Future<Output=...>>` type, but now only "bare" recursion (without an indirection) is forbidden.

src/concurrency/async/state-machine.md
```diff
@@ -0,0 +1,97 @@
...
+<details>
+
+While this code will run, it is simplified from what the real state machine
+would do. The important things to notice here are:
+
+- Calling an async function does nothing but construct a value, ready to start
+  on the first call to `poll`.
+- All local variables are stored in the function's future struct, including an
+  enum to identify where execution is currently suspended. The real generated
+  state machine would not initialize `i` to 0.
```
fw-immunant, line 87: I think it makes more sense for explanation to first trace the control flow so we can enumerate states and get in our heads something like "we'll have three possible execution states for this async fn, so it makes sense that its future type would be (morally) an enum with three variants". Then we can look at variable liveness at each of the awaits to determine the payload of each variant. This lets us get to "futures store live locals" without having to introduce the notion of liveness explicitly like a compilers course would--these are just the variables whose values we'll need to run the rest of the function.

I think it's better for an example of the state machine transform to use an async fn that doesn't have async in a loop, so that it's easy for readers to enumerate the entire set of states in their head.

So I might suggest an example more like this:
```rust
async fn send(s: String) -> usize {
    println!("{}", s);
    s.len()
}

async fn example(x: i32) -> usize {
    let double_x = x*2;
    let mut count = send(format!("x = {x}")).await;
    count += send(format!("double_x = {double_x}")).await;
    count
}
```

This gives us three states:
- an initial state holding the argument `x`
- a state holding `double_x` until we return from our first `await`
- a state holding `count` until we return from our second `await`

This is, I think, less confusing than a state we return to across iterations of our loop on `1..=count`.

fw-immunant (commented):
Some comments inline.

src/concurrency/async/state-machine.md
djmitche: Implicit in this suggestion is to use enum variants to represent the liveness of variables. That's great, but does require a bit more mucking about with pins than we want to present at this point. In fact, it uses `pin_project`, which we don't even talk about in the Pin section. Should I maybe revert this to a flat struct with a `fut` field, just to avoid this?

djmitche:
Reading through https://blog.yoshuawuyts.com/self-referential-types/ I see links to https://tmandry.gitlab.io/blog/posts/optimizing-await-1/ and https://tmandry.gitlab.io/blog/posts/optimizing-await-2/, all of which might help with this.

src/concurrency/async/state-machine.md
fw-immunant: I don't think it's so important to make this example have perfect fidelity or be compilable/runnable--pseudocode to capture the transformation would be fine, and would allow us to side-step `Pin` questions. The sub-future, if we really want to represent it, could just be a `std::future::Ready` or isomorphic. The most significant thing is not the state of the child future but the fact that at await points we capture live state.

djmitche:
https://natkr.com/2025-04-10-async-from-scratch-1/#fairdicepair might be helpful too

djmitche:
Sorry for the massive delay, @fw-immunant  -- PTAL?

fw-immunant:
Will check out tomorrow morning.

src/concurrency/async/futures.md
```diff
@@ -36,14 +36,9 @@ pause until that Future is ready, and then evaluates to its output.
...
-- We will not get to `Pin` and `Context`, as we will focus on writing async
-  code, rather than building new async primitives. Briefly:
+- We will address `Pin` in the "Pitfalls" segment.
 
-  - `Context` allows a Future to schedule itself to be polled again when an
-    event occurs.
-
-  - `Pin` ensures that the Future isn't moved in memory, so that pointers into
-    that future remain valid. This is required to allow references to remain
-    valid after an `.await`.
```
fw-immunant: I think we should keep the short description of `Pin` here while adding the mention that we cover it in more detail in the Pitfalls segment. We can't answer all questions here ("when will there be pointers into a future and why doesn't borrowck prevent such moves") but saying something concrete gives something to go on rather than leaving things fully opaque.

src/concurrency/async/state-machine.md
```diff
@@ -0,0 +1,154 @@
...
+
+fn main() {
+    let mut fut = two_d10(13);
+    loop {
+        if let Poll::Ready(result) = fut.poll() {
+            println!("result: {result}");
+            break;
+        }
+    }
+}
```
fw-immunant: I think it might be beneficial to use more pseudocode here and forfeit being able to run the example, as the behavior isn't surprising and I don't think tweaking the code is very elucidating here. We could remove the simplified Future definition and hand-wave the sub-future ("returns a random number when polled" is easy enough to take as a black box) and `main` isn't needed as this slide isn't about how to actually execute futures.

The first future with only one state doesn't showcase the state-machine transformation we care about here, and minimizing the amount of code involved helps clarify what's important to take away from this slide.

Then we would just have something like this:
```rust
use std::future::Future;
use std::task::{Poll, Context};
use std::pin::Pin;

/// Sum two D10 rolls plus a modifier.
async fn two_d10(modifier: u32) -> u32 {
    let first_roll = roll_d10().await;
    let second_roll = roll_d10().await;
    first_roll + second_roll + modifier
}

// desugars to (approximately):

/// Sum two D10 rolls plus a modifier.
fn two_d10(modifier: u32) -> TwoD10 {
    TwoD10::Init { modifier }
}

enum TwoD10 {
    // Function has not begun yet.
    Init { modifier: u32 },
    // Waitig for first `.await` to complete.
    FirstRoll { modifier: u32, fut: RollD10Future },
    // Waitig for second `.await` to complete.
    SecondRoll { modifier: u32, first_roll: u32, fut: RollD10Future },
}

impl Future for TwoD10 {
    type Output = u32;
    fn poll(mut self: Pin<&mut Self>, ctx: &mut Context) -> Poll<Self::Output> {
        loop {
            match *self {
                TwoD10::Init { modifier } => {
                    // Create future for first dice roll.
                    let fut = roll_d10();
                    *self = TwoD10::FirstRoll { modifier: modifier, fut };
                }
                TwoD10::FirstRoll { modifier, ref mut fut } => {
                    // Poll sub-future for first dice roll.
                    if let Poll::Ready(first_roll) = fut.poll(ctx) {
                    	// Create future for second roll.
                        let fut = RollD10Future::new();
                        *self = TwoD10::SecondRoll {
                            modifier: modifier,
                            first_roll,
                            fut,
                        };
                    } else {
                        return Poll::Pending;
                    }
                }
                TwoD10::SecondRoll { modifier, first_roll, ref mut fut } => {
                    // Poll sub-future for second dice roll.
                    if let Poll::Ready(second_roll) = fut.poll(ctx) {
                        return Poll::Ready(first_roll + second_roll + modifier);
                    } else {
                        return Poll::Pending;
                    }
                }
            }
        }
    }
}
```

djmitche:
Thanks -- updated!

fw-immunant (approved):

fw-immunant:
We have to mark the code block as no-run. Otherwise, LGTM.

djmitche:
Thanks for your help with this! With luck I'll find a way to teach the course again, but otherwise please do share any feedback on how this goes if/when you teach it.

fw-immunant:
> Thanks for your help with this! With luck I'll find a way to teach the course again, but otherwise please do share any feedback on how this goes if/when you teach it.

I taught it today, fresh off the presses! One thing that confused folks is the outer `loop` around the state machine transition, but we had a productive discussion there about how it either executes one or two times and how returning `Pending` more or fewer times can interact with the executor's scheduler. There are one or two typos I'll PR about. Otherwise I think it was a good addition in terms of providing a concrete demonstration of how things get compiled. One student asked if we can see this kind of output from `rustc`--but as far as I know the only way to do so involves dumping MIR, which is far from readable.

djmitche:
I wish there was such a thing, too, similar to `cargo expand`. Maybe if/when generators are stabilized that will be more practical? Anyway, I'm happy to LGTM a typo-fix PR :)

PR #2052, opened 2024-05-04
vi: Translate the std library String type

l0ngvh:
I couldn't find an effective way to translate `Trait` and `implement`, so both terms are going to be used as is in the translation.

### DIFF
--- a/po/vi.po
+++ b/po/vi.po
@@ -6177,8 +6177,10 @@ msgid ""
 "To a substring by using `s3[0..4]`, where that slice is on character "
 "boundaries or not."
 msgstr ""
-"Truy cập một chuỗi con bằng cách sử dụng `s3[0..4]`, với phạm vi của chuỗi "
-"con nằm trên ranh giới của ký tự và không nằm trên ranh giới của ký tự."
+"Truy cập một xâu con bằng cách sử dụng `s3[0..4]`, với phạm vi của xâu con "
+"con giao thoa với ranh giới của các ký tự cận biên, hoặc với các ký tự cận "
+"biên hoàn toàn nằm bên trong xâu con. Ví dụ, truy cập `s[5..7] với `s = "
+"\"Xin chào\"`."
 
 #: src/std-types/string.md
 msgid ""
--- a/po/vi.po
+++ b/po/vi.po
@@ -6047,32 +6047,35 @@ msgid ""
 "[`String`](https://doc.rust-lang.org/std/string/struct.String.html) is a "
 "growable UTF-8 encoded string:"
 msgstr ""
+"[`String`](https://doc.rust-lang.org/std/string/struct.String.html) là kiểu "
+"dữ liệu chuẩn sử dụng để lưu trữ chuỗi UTF-8 trên bộ nhớ heap, có thể được "
+"mở rộng:"
 
 #: src/std-types/string.md src/std-traits/read-and-write.md
 #: src/memory-management/review.md src/testing/unit-tests.md
 #: src/concurrency/threads/scoped.md
 msgid "\"Hello\""
-msgstr ""
+msgstr "\"Xin chào\""
 
 #: src/std-types/string.md
 msgid "\"s1: len = {}, capacity = {}\""
-msgstr ""
+msgstr "\"s1: độ dài = {}, kích thước = {}\""
 
 #: src/std-types/string.md
 msgid "'!'"
-msgstr ""
+msgstr "'!'"
 
 #: src/std-types/string.md
 msgid "\"s2: len = {}, capacity = {}\""
-msgstr ""
+msgstr "\"s2: độ dài = {}, kích thước = {}\""
 
 #: src/std-types/string.md
 msgid "\"🇨🇭\""
-msgstr ""
+msgstr "\"🇻🇳\""
 
 #: src/std-types/string.md
 msgid "\"s3: len = {}, number of chars = {}\""
-msgstr ""
+msgstr "\"s3: độ dài = {}, số ký tự = {}\""
 
 #: src/std-types/string.md
 msgid ""
@@ -6080,18 +6083,25 @@ msgid ""
 "string/struct.String.html#deref-methods-str), which means that you can call "
 "all `str` methods on a `String`."
 msgstr ""
+"`String` implement [`Deref<Target = str>`](https://doc.rust-lang.org/std/"
+"string/struct.String.html#deref-methods-str), có nghĩa tất cả các hàm của "
+"`str` đều có thể được gọi trên một biến kiểu `String`."
 
 #: src/std-types/string.md
 msgid ""
 "`String::new` returns a new empty string, use `String::with_capacity` when "
 "you know how much data you want to push to the string."
 msgstr ""
+"`String::new` trả về một chuỗi rỗng. Khi dữ liệu cần đẩy vào chuỗi đã được "
+"xác định trước, hãy sử dụng `String::with_capacity`."
 
 #: src/std-types/string.md
 msgid ""
 "`String::len` returns the size of the `String` in bytes (which can be "
 "different from its length in characters)."
 msgstr ""
+"`String::len` trả về kích thước của `String` theo byte (có thể khác với số "
+"ký tự trong chuỗi)."
 
 #: src/std-types/string.md
 msgid ""
@@ -6100,34 +6110,45 @@ msgid ""
 "to [grapheme clusters](https://docs.rs/unicode-segmentation/latest/"
 "unicode_segmentation/struct.Graphemes.html)."
 msgstr ""
+"`String::chars` trả về một con trỏ chạy dọc theo các ký tự của chuỗi. Lưu ý "
+"rằng một `char` có thể khác với một \"ký tự\" thông thường bởi vì một khái "
+"niệm gọi là [grapheme clusters](https://docs.rs/unicode-segmentation/latest/"
+"unicode_segmentation/struct.Graphemes.html)."
 
 #: src/std-types/string.md
 msgid ""
 "When people refer to strings they could either be talking about `&str` or "
 "`String`."
-msgstr ""
+msgstr "Khi nói đến chuỗi, ta có thể sử dụng `&str` hoặc `String` đều được."
 
 #: src/std-types/string.md
 msgid ""
 "When a type implements `Deref<Target = T>`, the compiler will let you "
 "transparently call methods from `T`."
 msgstr ""
+"Khi một kiểu dữ liệu implement `Deref<Target = T>`, trình biên dịch sẽ cho "
+"phép gọi các hàm thuộc về `T` trên kiểu dữ liệu đó."
 
 #: src/std-types/string.md
 msgid ""
 "We haven't discussed the `Deref` trait yet, so at this point this mostly "
 "explains the structure of the sidebar in the documentation."
 msgstr ""
+"Chúng ta chưa đi sâu vào `Deref` trait, việc đề cập đến `Deref` ở đây chủ "
+"yếu là để giải thích cấu trúc của mục lục tài liệu."
 
 #: src/std-types/string.md
 msgid ""
 "`String` implements `Deref<Target = str>` which transparently gives it "
 "access to `str`'s methods."
 msgstr ""
+"`String` implement `Deref<Target = str>`, nên các hàm của `str` đều có thể "
+"được gọi trên các biến kiểu `String`."
 
 #: src/std-types/string.md
 msgid "Write and compare `let s3 = s1.deref();` and `let s3 = &*s1;`."
 msgstr ""
+"So sánh `let s3 = s1.deref();` và `let s3 = &*s1;`. Cái nào dễ hiểu hơn?"
 
 #: src/std-types/string.md
 msgid ""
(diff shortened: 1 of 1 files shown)


po/vi.po
```diff
@@ -6128,58 +6128,68 @@ msgstr ""
...
 msgid "String"
-msgstr ""
+msgstr "String"
 
 #: src/std-types/string.md
 msgid ""
 "[`String`](https://doc.rust-lang.org/std/string/struct.String.html) is the "
 "standard heap-allocated growable UTF-8 string buffer:"
 msgstr ""
+"[`String`](https://doc.rust-lang.org/std/string/struct.String.html) là kiểu "
```
qu-ngx, line 6138: String can be translated to Xâu ký tự. If you are unsure about the term translation, you can translate using the following format "English Term (Bản dịch tiếng Việt)" and proceed to go on to the part that can be easily translated. 

So it should be the following the format:
"[`String (Xâu ký tự)`] là kiểu (https://doc.rust-lang.org/std/string/struct.String.html) là kiểu " 
"dữ liệu chuẩn sử dụng để lưu trữ chuỗi UTF-8 trên bộ nhớ đệm xâu ký tự (buffer string) sử dụng cấu trúc chất đống (heap), có thể được mở rộng:"

po/vi.po
```diff
@@ -6188,57 +6198,75 @@ msgid ""
...
 "`String`."
-msgstr ""
+msgstr "Khi nói đến chuỗi, ta có thể sử dụng `&str` hoặc `String` đều được."
 
 #: src/std-types/string.md
 msgid ""
 "When a type implements `Deref<Target = T>`, the compiler will let you "
 "transparently call methods from `T`."
 msgstr ""
+"Khi một kiểu dữ liệu implement `Deref<Target = T>`, trình biên dịch sẽ cho "
```
qu-ngx, line 6217: "implement" not translated?

po/vi.po
```diff
@@ -6188,57 +6198,75 @@ msgid ""
...
+"phép gọi các hàm thuộc về `T` trên kiểu dữ liệu đó."
 
 #: src/std-types/string.md
 msgid ""
 "We haven't discussed the `Deref` trait yet, so at this point this mostly "
 "explains the structure of the sidebar in the documentation."
 msgstr ""
+"Chúng ta chưa đi sâu vào `Deref` trait, việc đề cập đến `Deref` ở đây chủ "
```
qu-ngx, line 6225: "trait" not translated?

po/vi.po
```diff
@@ -6188,57 +6198,75 @@ msgid ""
...
+"So sánh `let s3 = s1.deref();` và `let s3 = &*s1;`. Cái nào dễ hiểu hơn?"
 
 #: src/std-types/string.md
 msgid ""
 "`String` is implemented as a wrapper around a vector of bytes, many of the "
 "operations you see supported on vectors are also supported on `String`, but "
 "with some extra guarantees."
 msgstr ""
+"`String` được được viết dựa trên một vector gồm nhiều byte, nên những hàm "
+"gọi được trên vector cũng có thể được gọi trên `String`, nhưng đi kèm với "
```
qu-ngx, line 6248: remember to translate "String".

po/vi.po
```diff
@@ -6188,57 +6198,75 @@ msgid ""
...
 msgstr ""
+"Truy cập một ký tự bằng cách sử dụng `s3.chars().nth(i).unwrap()` với `i` "
+"nằm trong phạm vi của chuỗi, và với `i` nằm ngoài phạm vi của chuỗi."
 
 #: src/std-types/string.md
 msgid ""
 "To a substring by using `s3[0..4]`, where that slice is on character "
 "boundaries or not."
 msgstr ""
+"Truy cập một chuỗi con bằng cách sử dụng `s3[0..4]`, với phạm vi của chuỗi "
```
qu-ngx, line 6268: "xâu con" might be better than "chuỗi con" (just a suggestion). Instead of using "nằm trên" try out "nằm ngoài"? Cái chuỗi này nằm ngoài chuỗi kia sounds a bit more logical ?

po/vi.po
```diff
@@ -6248,6 +6276,11 @@ msgid ""
 "`Display`, so anything that can be formatted can also be converted to a "
 "string."
 msgstr ""
+"Nhiều kiểu dữ liệu có thể được chuyển đổi thành chuỗi bằng phương thức "
+"[`to_string`](https://doc.rust-lang.org/std/string/trait.ToString."
+"html#tymethod.to_string). Trait này đã được implement sẵn cho tất cả các "
```
qu-ngx, line 6281: "Trait" not translated

po/vi.po
```diff
@@ -6128,58 +6128,68 @@ msgstr ""
...
+msgstr "\"s3: độ dài = {}, số ký tự = {}\""
 
 #: src/std-types/string.md
 msgid ""
 "`String` implements [`Deref<Target = str>`](https://doc.rust-lang.org/std/"
 "string/struct.String.html#deref-methods-str), which means that you can call "
 "all `str` methods on a `String`."
 msgstr ""
+"`String` implement [`Deref<Target = str>`](https://doc.rust-lang.org/std/"
+"string/struct.String.html#deref-methods-str), có nghĩa tất cả các hàm của "
```
qu-ngx, line 6175: Translate String?

po/vi.po
```diff
@@ -6128,58 +6128,68 @@ msgstr ""
...
+"`str` đều có thể được gọi trên một biến kiểu `String`."
```
qu-ngx, line 6176: Translate "String"?

ghost (changes requested):
I have left some comments and feedbacks before you are able to merge. You should review and change your code before it is merged. Tks!

vohoanglong0107:
@qu-ngx I'm quite unsure about translating `implement` and `trait`, as there are no Vietnamese terms that accurately express them. The closest translation I could think of for `implement` is `kế thừa`, but it's more related to the OOP term `inherit` than `implement`.

qu-ngx:
> @qu-ngx I'm quite unsure about translating `implement` and `trait`, as there are no Vietnamese terms that accurately express them. The closest translation I could think of for `implement` is `kế thừa`, but it's more related to the OOP term `inherit` than `implement`.

Implement is also áp dụng, thi hành, trang bị, .... Trait aka đặc điểm, tính trạng, trạng thái. If you are not so sure, I think you can use some English dictionary (ie Meriam Webster) for the synonyms. English has many different shades so it is not tied to only one meaning at all.

vohoanglong0107:
@qu-ngx Also, when translating type name, such as `String`, I believe translate `String` the first time it appearse is good, but it shouldn't be translated in subsequent appearances, as it can have cause confusion. For example, in Rust `&str` and `String` can be translated as `Xâu` in Vietnamese, but using only `Xâu` in the translation will cause confusion, as the students might be unsure which type to use.

qu-ngx:
> @qu-ngx Also, when translating type name, such as `String`, I believe translate `String` the first time it appearse is good, but it shouldn't be translated in subsequent appearances, as it can have cause confusion. For example, in Rust `&str` and `String` can be translated as `Xâu` in Vietnamese, but using only `Xâu` in the translation will cause confusion, as the students might be unsure which type to use.

It still has translation in Vietnamese and also String translation is a well-known Vietnamese word. That is why when I say translate you put in the brackets. You can refer to the other successful PRs of other people as well if you want. It has no cost and no harm in doing it as not everyone learn it in English first time. That is why both is better than just one.

qu-ngx:
> @qu-ngx Also, when translating type name, such as `String`, I believe translate `String` the first time it appearse is good, but it shouldn't be translated in subsequent appearances, as it can have cause confusion. For example, in Rust `&str` and `String` can be translated as `Xâu` in Vietnamese, but using only `Xâu` in the translation will cause confusion, as the students might be unsure which type to use.

Also please remember you are doing the translation for everyone so clarity in the words and meanings have a significant role in why you are doing translations. This feedbacks are not set to target you, but to help you find a better way to make your translation as clear as possible to the other Vietnamese peers.

vohoanglong0107:
> Implement is also áp dụng, thi hành, trang bị, .... Trait aka đặc điểm, tính trạng, trạng thái. If you are not so sure, I think you can use some English dictionary (ie Meriam Webster) for the synonyms. English has many different shades so it is not tied to only one meaning at all.

@qu-ngx I could use these translations, but they wouldn't maintain the original meaning. For example, how would `Deserializer implement Sync Trait` be translated?
IMO, we shouldn't force a translation on a technical terms. [Rust Tiếng Việt](https://rust-tieng-viet.github.io/basic/trait/index.html) agree with me on this point, as they also don't translate `Trait` and `implement`

vohoanglong0107:
> Also, please remember you are doing the translation for everyone so clarity in the words and meanings have a significant role in why you are doing translations. These feedbacks are not set to target you, but to help you find a better way to make your translation as clear as possible to the other Vietnamese peers.

I understand, which is why I want to discuss with you the best way to translate this. I want to introduce this book to my Vietnamese friends, and I don't want to confuse them while reading the book.

vohoanglong0107:
@qu-ngx I have opened a new discussion to talk about this [here](https://github.com/google/comprehensive-rust/discussions/2062)

po/vi.po
qu-ngx: this implements is not a built-in in C++ btw. implements (some Vietnamese translation) should be good

ghost (changes requested):
I have removed some comments on translation for built-ins. Can you help me out with the translation? Tkiu so much!

vohoanglong0107:
@qu-ngx I think some previous comments are staled after https://github.com/google/comprehensive-rust/discussions/2062#discussioncomment-9335645. Could you give this another go?

qu-ngx:
> @qu-ngx I think some previous comments are staled after [#2062 (comment)](https://github.com/google/comprehensive-rust/discussions/2062#discussioncomment-9335645). Could you give this another go?

Oh yea, sorry I forgot to remove some on comments. Can you check it for me. And btw the other "implements" is an action word. It isn't the name of the built-in `implement`. Can you help translate that one?

po/vi.po
vohoanglong0107: `Deref<Target = T>` is a trait, so this `implement` should be a technical term in Rust

qu-ngx: "When a type implements `Deref<Target = T>`". What I understand here is  I think "implements" here is not a technical term in Rust since it has "s" in the word? Are you sure it's not? This is what I got from using a translator:
"Khi một kiểu biến triển khai `Deref<Target = T>`, trình biên dịch sẽ cho phép bạn .." 

The `implement` you referred to is without s. But in that sentence implement also go with "s". So is it a technical term or not?

vohoanglong0107:
@qu-ngx Are you referring to another `implement`?
Because `Deref<Target = T>` is a trait, you can learn more about this trait [here](https://doc.rust-lang.org/std/ops/trait.Deref.html), `implement` in  `implement Deref<Target = T>` isn't being used as English action verb. What `implement Deref<Target = T>` actually means is, any type that `implement Deref<Target = T>` will have a method called `deref`, which in many cases means that the type allows user to access the underlying type.
I hope this clear things up a bit.

ghost (commented):
The `implement` you referred to is without s. However, impelement in the sentence goes with "s". So is it a techincal term or not?

po/vi.po
vohoanglong0107: Haha, I agree with you that it's quite confused here. `implement` is being used here with a suffix of `s` because it was "following English grammar rule". However, I can assured you this is a technical term

vohoanglong0107: @qu-ngx If you are unsure, perharps this chapter in the Rust book (see [link](https://doc.rust-lang.org/book/ch10-02-traits.html#implementing-a-trait-on-a-type)) can help you make a decision.

vohoanglong0107: Ah, I should mention it, `implement` here is actually the expansion of `impl` keyword. That means `Vec implements Deref<Target = T>` is actually the same as `impl Deref<Target = T> for Vec`

qu-ngx: Yea, you better add it or put it somewhere in the brackets. And also if you don't translate the implements word then better keep the s with it as well so it should be implements instead of implement.

ghost (commented):
I agree with your comment in the code review. When u are done changing, just ping me.

po/vi.po
vohoanglong0107: @qu-ngx Ah, I can see where the confusion coming from. Rust string has this rather strange property that, unlike other language like C++, when you access a string with `s[i]` syntax, the returned value could be an invalid character. The reason is, Rust's string is made of unicode characters, so each character could span multiple index. Suppose you have the string `s = "Xin chào"`. All the `X`, `i`, `n`, ` `, `c`, `h`, `o` are ASCII characters, therefore it need only one byte, consuming one slot in the string. However, `à` is a unicode character, so it need 2 bytes, consuming 2 slot. What this means is that when you try to access "Xin chào", Rust allow you to access `s[0..2]`, which prints out `Xi`, `s[5..8]`, which equal to `hà`, but you can't access `s[5..7]`, since 7 is in the middle of `à` character. This is what the excerpt refer to as the `character boundaries`.
However, I acknowledge that my translation poorly conveys that idea. I will see if I can improve this.

vohoanglong0107:
@qu-ngx I think this should be ready for another go

tidefield (approved):
Sorry for the delay on reviewing the PRs. I've been on the road for the past few weeks.

ghost (approved):

PR #2053, opened 2024-05-05
Modularize Tokenizer's iterator for improved readability

snazzybytes:
While learning Rust and going through the course, I realized the exercise solution readability can be improved by adding Tokenizer helper methods to encapsulate the character collecting logic - which slims down iterator's next() method while enhancing clarity.

### DIFF
--- a/src/error-handling/exercise.rs
+++ b/src/error-handling/exercise.rs
@@ -57,28 +57,34 @@ enum TokenizerError {
 
 struct Tokenizer<'a>(Peekable<Chars<'a>>);
 
+impl<'a> Tokenizer<'a> {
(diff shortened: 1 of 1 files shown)


djmitche (changes requested):
Thanks for the PR!

This changes the solution, but that code also appears in the "problem" (the commented out part at the bottom of the file). If one changes, the other should as well. And, we still want to maintain the structure where the "problem" uses `panic!` and `expect`, and the solution replaces those with `Result`.

It's worth noting that both of the new utility methods you've added always return `Ok(..)`. If you change those so that they just return a Token, then they will not need to change from problem to solution, and can be included in both the problem and solution.

djmitche (approved):
Nice work!

PR #2064, opened 2024-05-07
Fix broken link to unsafe-traits.md

snazzybytes:
I came upon a broken link which takes you to (Document Not Found): https://google.github.io/comprehensive-rust/concurrency/unsafe/unsafe-traits.html

Verified locally with `mdbook serve` and now it links correctly


djmitche (approved):
Thanks! We don't have too many of these inter-page links so they do tend to get forgotten and break.

PR #2066, opened 2024-05-08
ignore nodejs deps for dprint

alix1383:


djmitche (commented):
What generates these files? I don't have them locally, although I do have dprint installed.

alix1383:
> What generates these files? I don't have them locally, although I do have dprint installed.

some people using NPM(node package manager) to install dprint 
[check npm section](https://dprint.dev/install/)

the packages located at `node_modules` and 
`package.json` store package name and versions

djmitche (approved):
got it - thanks!

PR #2074, opened 2024-05-10
vi: Control flow basics summary

HuyNVuong:
I prefer using "Lệnh" over "Biểu thức" because AFAIK Biểu thức refer to RHS expression things such as `a + b` or `100 if a == x else 200`.... So `Lệnh` seems like a better term in the header
Correct me if I'm wrong


po/vi.po
```diff
@@ -139,23 +139,23 @@ msgstr "Labels (nhãn)"
...
 
 #: src/SUMMARY.md src/control-flow-basics/macros.md
 msgid "Macros"
-msgstr ""
+msgstr "Macros"
 
 #: src/SUMMARY.md src/control-flow-basics/exercise.md
 msgid "Exercise: Collatz Sequence"
-msgstr ""
+msgstr "Bài tập: Chuỗi Collatz"
```
vohoanglong0107, line 158: I think this could be translated like this for consistency with other parts of the book:

```suggestion
msgstr "Thực hành: Chuỗi Collatz"
```

po/vi.po
```diff
@@ -111,11 +111,11 @@ msgstr ""
 
 #: src/SUMMARY.md src/control-flow-basics.md
 msgid "Control Flow Basics"
-msgstr ""
+msgstr "Điểu khiển luồng căn bản"
```
vohoanglong0107, line 114: I think `Control flow` can be used as is here. Reader can get a rough understanding in the `Glossary`

po/vi.po
```diff
@@ -139,23 +139,23 @@ msgstr "Labels (nhãn)"
 
 #: src/SUMMARY.md src/control-flow-basics/blocks-and-scopes.md
 msgid "Blocks and Scopes"
-msgstr ""
+msgstr "Blocks (khối lệnh) và scopes (phạm vi)"
```
vohoanglong0107, line 142: We have a rough translation of `block` and `scope` in the `Glossary`, so we can use those terms as is in my opinion.

vohoanglong0107 (commented):
I agree that `Lệnh` works better in this context.

po/vi.po
HuyNVuong: Ok, removed the Vietnamese translations

po/vi.po
HuyNVuong: Done

po/vi.po
HuyNVuong: I think Control flow seems like a hard word to spell for non-English speakers right? So I think keeping the translations here for instructor / students who have a hard time of spelling it refer to without the needs of referrring to the glossary is good. Maybe `Control flow (Điểu khiển luồng) căn bản`?

daivinhtran: Yeah `Control flow (Điểu khiển luồng) căn bản` is also a good idea.

HuyNVuong: Updated

qu-ngx (commented):
It doesn't look bad actually. Considering "Lệnh" and "Biểu thức" in Vietnamese can both represent "expressions", I think they could be used interchangably. But "cấu trúc" is also good for this one since "lệnh" can also be referred to the commands you type in the cmd.
What do you think @daivinhtran?

HuyNVuong:
> Collaborator



> It doesn't look bad actually. Considering "Lệnh" and "Biểu thức" in Vietnamese can both represent "expressions", I think they could be used interchangably. But "cấu trúc" is also good for this one since "lệnh" can also be referred to the commands you type in the cmd. What do you think @daivinhtran?

I think Cấu trúc might be a tad too far. Is it refers to English `if`? If there's a Vietnamese CS textbook or site that refers `if` as Cấu trúc I'm happy to update them

qu-ngx (approved):
SGTM! I just doubled check again. You are right. "Cấu trúc" seems to be a bit vague since it can also be referred to struct. This translation is already good enough I think.

Tks for your contributions!

PR #2078, opened 2024-05-13
vi: Translate From, Into and Comparison chapter from std-trait session

vohoanglong0107:
I think associated type deserves a slot in the Glossary


qu-ngx (approved):
SGTM!

mgeisler:
> I think associated type deserves a slot in the Glossary

Can you add "associated type" it to https://github.com/google/comprehensive-rust/blob/main/src/glossary.md? Then it will show up for translation the next time someone refreshes the Vietnamese translation.

PR #2083, opened 2024-05-17
Fix the debug exercise in the "Borrowing a Value" section

errge:
Trying to reproduce this RVO shows that actually the optimization that is used here is simple inlining.  Once inlining is disabled, the addresses change, even with `-O -C opt-level=3`.

The return of the values is also never changed to an "efficient memcpy", but instead was returned on eax+edx, although this result is of course specific to the calling convention of the platform ABI, but no memcpy on the most popular amd64 architecture.

I don't think it's educationally important to teach RVO here, so I didn't go into any length trying to force Rust to do a real RVO (passing Point address into the function on the assembly level, so the function can fill in the addition).  In my opinion the only important thing is, that if a student is actually clicking the Playground link and looks into the assembly, then our description should match the generated code.  That's why I just fixed the content instead of trying to fix the example to be RVO.

### DIFF
--- a/src/borrowing/shared.md
+++ b/src/borrowing/shared.md
@@ -35,14 +35,14 @@ slightly to include function arguments and return values.
 
 # More to Explore
 
-Notes on stack returns:
+Notes on stack returns and inlining:
 
 - Demonstrate that the return from `add` is cheap because the compiler can
-  eliminate the copy operation. Change the above code to print stack addresses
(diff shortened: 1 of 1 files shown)


djmitche (approved):
Looks good! I think the course should avoid getting _too_ into the weeds on code generation in general. Folks who are interested in such things can dig down on their own.

PR #2092, opened 2024-05-22
Move playground slide to overview section

randomPoison:
The playground slide seemed like it was in a weird place to me. I usually cover the playground in a preamble before starting in on the day 1 content, not sure if other teachers do the same though. Since we cover things like cargo in the overview section, it seems like the right place to mention the playground as well.


djmitche (commented):
Maybe we should rename "Using Cargo" to "Using Rust" and combine it with that first segment ("Hello, World")? That first segment is kind of weird anyway - all talk and no code. And, it would make sense to talk about installing and using cargo in the "actual" course and not just the preamble. WDYT?

randomPoison:
Hrm, yeah I like the idea of formalizing the little preamble that I do and fleshing out that first section of the slides. The things I normally cover are:

* Telling the students that course is meant to be interactive and that they should ask questions.
* Installing Rust via rustup.
* Creating a workspace with `cargo new`.
* Using VS Code and installing the rust-analyzer plugin.
* Finding the course materials online to follow along with.

Other than finding the course materials that all seems reasonable to cover in slides directly.

djmitche:
I like that! I think it's still reasonable to ask students to look at that before the class, since the install etc. might take some time. Do you want to move things about in this PR?

mgeisler:
> That first segment is kind of weird anyway - all talk and no code.

Yeah... it's very one-way and probably a little boring.


> I like that! I think it's still reasonable to ask students to look at that before the class, since the install etc. might take some time.

Yeah, I think so too — and we do actually ask people to ensure they have a working Rust setup.

But it would be good to have clear instructions on how they should do this: I'm one of those people who will setup my own editor for a new language, but I've noticed that there are many people who don't care. They just want clear instructions on how to setup the environment and then they'll work with it. So I think it would be good to have such instructions.

randomPoison:
I haven't touched this in a while so I'll close the PR and I can re-open if I look at this issue again.

PR #2093, opened 2024-05-22
Fix missing lifetime error in dangling reference example

randomPoison:
The example of returning a reference to a local variable doesn't compile due to a missing lifetime specifier, which isn't what we're trying to demonstrate with that example. I usually add the lifetime in manually in order to demonstrate the compiler error, but it occurs to me that if we make the argument a reference we can sneakily get the correct compiler error without having to introduce the lifetime syntax.


djmitche (approved):
Nice!

djmitche:
Merging is blocked on https://github.com/google/comprehensive-rust/issues/2094

Issue #2104, opened 2024-05-29
Page `Exercise: Iterator Method Chaining` rust code not rendering


peter-lockhart-pub:
When viewing this page in Chrome, I do not see the coding exercise https://google.github.io/comprehensive-rust/iterators/exercise.html

![Screenshot 2024-05-29 at 10 01 57](https://github.com/google/comprehensive-rust/assets/60918964/bfa8421d-087e-40eb-848a-07a3f162c191)

OSX
Chrome Version 125.0.6422.78 (Official Build) (x86_64)

djmitche:
Hm, I see it "flash" into existence and then disappear. I wonder if this has to do with the code-saving functionality?

fw-immunant:
I believe this is fixed now.

PR #2106, opened 2024-05-29
Revert "Saving the playground state in local storage (#1942)"

### LINKED ISSUE #2105: Most code snippets don't work anymore
As the title says, most code snippets don't work anymore

https://google.github.io/comprehensive-rust/concurrency/sync-exercises/solutions.html 
https://google.github.io/comprehensive-rust/iterators/exercise.html
https://google.github.io/comprehensive-rust/modules/exercise.html
etc...

djmitche:
This reverts commit e2cad7da8d85b3d7d783f064b6d4db5b6e7d9580.

Fixes #2105, #2104.

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -48,9 +48,6 @@ urlcolor = "red"
 curly-quotes = true
 additional-js = [
   "theme/speaker-notes.js",
-  "theme/redbox.js",
-  "theme/save-playgrounds.js",
-  "theme/instructor-menu.js",
 ]
 additional-css = [
   "theme/css/svgbob.css",
--- a/theme/instructor-menu.js
+++ b/theme/instructor-menu.js
@@ -24,10 +24,10 @@
     instructorMenu.title = "Utilities for course instructors";
     instructorMenu.innerHTML =
       '<i class="fa fa-ellipsis-v" aria-hidden="true"></i>';
-    redBoxButton.innerHTML = "Toggle aspect-ratio box";
+    redBoxButton.innerHTML = "aspect-ratio box";
     redBoxButton.title =
       "Outline the area that fits on one screen while teaching the course.";
-    playgroundStateButton.innerHTML = "Reset all playgrounds";
+    playgroundStateButton.innerHTML = "reset all playgrounds";
     playgroundStateButton.title =
       "Reset code in all playgrounds to its original value.";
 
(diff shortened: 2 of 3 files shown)


randomPoison (approved):
I ran into this in class today, happy to see that you already found the culprit and have a fix ready.

mani-chand:
> @mani-chand, I'm sorry we had to revert the change.

Any bug?

mani-chand:
> @mani-chand, I'm sorry we had to revert the change.

That's ok, I got it.

djmitche:
The bugs were linked above -- it seems that exercises no longer appeared. If you hit "refresh" they would flash into existence, then disappear.

I don't know why this specifically occurred with exercises. My guess would be that it's because they are longer than most code sampls.

I wonder if we should give up on this -- it seems there are just too many corner cases with strange interactions. And when it fails, it dramatically interferes with ongoing instruction. What do you think?

mani-chand:
> The bugs were linked above -- it seems that exercises no longer appeared. If you hit "refresh" they would flash into existence, then disappear.
> 

> I don't know why this specifically occurred with exercises. My guess would be that it's because they are longer than most code sampls.

I don't think it is because of longer codes but may be because of non-editable playground. 

> 
> I wonder if we should give up on this -- it seems there are just too many corner cases with strange interactions. And when it fails, it dramatically interferes with ongoing instruction. What do you think?

Yes, That's better.

djmitche:
Sorry about that! We really appreciate your work on it. Some projects just turn out to be, if not impossible, at least impracatical. And I think this was one of them. Maybe we could have gotten this to work by forking ACE to handle persistent state or forking mdBook to generate a single-page app or something else -- but all of that is too great a cost for the desired result.

mani-chand:
> Sorry about that! We really appreciate your work on it. Some projects just turn out to be, if not impossible, at least impracatical. And I think this was one of them. Maybe we could have gotten this to work by forking ACE to handle persistent state or forking mdBook to generate a single-page app or something else -- but all of that is too great a cost for the desired result.

There is nothing to say sorry, We did our best to make it work, Not everything work as expected. 

Second is that I got a job, It's been a hard time in job with high work pressure, I am getting less time to spend on this work.

mgeisler:
Yeah, I agree with Dustin: thanks for the work you put into this, @mani-chand!

I think we should leave it as it is now. Perhaps we can revisit this one day if we get a proper testing framework for our JavaScript code — something that runs in a headless browser and which we can use to stress test things.

PR #2108, opened 2024-05-30
Add a slide introducing `dyn Trait` in Generics

randomPoison:
I've been thinking it'd be simpler to introduce `dyn Trait` via `&dyn Trait` rather than waiting for the smart pointers section and `Box<dyn Trait>`. This PR adds a slide to the Generics section that introduces `&dyn Trait` and compares it to `&impl Trait`, juxtaposing monomorphization and static dispatch against type-erasure and dynamic dispatch. I've then updated the existing trait object slide to call back to the earlier introduction, and call out that using `Box<dyn Trait>` gives you an owned trait object rather than a borrowed one.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -100,6 +100,7 @@
   - [Generic Traits](generics/generic-traits.md)
   - [Trait Bounds](generics/trait-bounds.md)
   - [`impl Trait`](generics/impl-trait.md)
+  - [`dyn Trait`](generics/dyn-trait.md)
   - [Exercise: Generic `min`](generics/exercise.md)
     - [Solution](generics/solution.md)
 - [Standard Library Types](std-types.md)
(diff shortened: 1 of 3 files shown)


djmitche (approved):
Love it!  Looks like this needs a `dprint fmt` before merging though.

Looking at the course timings, do you think it makes sense to move generics into day-2 morning?

randomPoison:
> do you think it makes sense to move generics into day-2 morning?

Didn't it used to be in the morning? iirc I moved it to the afternoon because I was consistently only getting through the pattern matching and methods sections in the morning.

djmitche:
Oh, you're right. We should probably adjust the timings to match!

randomPoison:
> We should probably adjust the timings to match!

Ah yeah, i think we can just reduce the std types and traits to 1 hour each, that's closer to how much time I spend on each of those. Did you want me to do that in this PR?

djmitche:
Nope, I was just waiting for the `dprint fmt`, which I see has occurred. If you want to make another PR that'd be great!

PR #2109, opened 2024-05-31
Remove error handling from protobuf exercise

randomPoison:
The protobuf exercise is one of the more complicated exercises, and it comes before the section on error handling in Rust. I think that makes the error handling within the exercise more confusing for students, and adds to the cognitive burden for doing the exercise. As a potential way to address this, this PR removes the `Result`-based error handling in the exercise and replaces it with panics. This makes the code less idiomatic in Rust, but potentially more friendly to students because they don't need to worry about `Result` or error handling.

That said, I'm not entirely sure this is actually an improvement. The simpler thing might be to briefly describe the `?` operator to students so that they know they can use `?` to unwrap errors when calling functions like `parse_varint`. I'm not sure what the best approach is, so I'm open to suggestions here.

### DIFF
--- a/src/lifetimes/exercise.md
+++ b/src/lifetimes/exercise.md
@@ -42,7 +42,7 @@ a message into a series of calls to those callbacks.
 What remains for you is to implement the `parse_field` function and the
 `ProtoMessage` trait for `Person` and `PhoneNumber`.
 
-<!-- compile_fail because `mdbook test` does not allow use of `thiserror` -->
+<!-- compile_fail because the stubbed out code has type inference errors. -->
 
 ```rust,editable,compile_fail
 {{#include exercise.rs:preliminaries }}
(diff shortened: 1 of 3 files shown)


mgeisler:
Hi Nicole, thanks for catching this!

> The simpler thing might be to briefly describe the `?` operator to students so that they know they can use `?` to unwrap errors when calling functions like `parse_varint`. I'm not sure what the best approach is, so I'm open to suggestions here.

I think I like that idea: it fits with the spiral concept of visiting the same things multiple times. Here we could just show the expansion from [29.2. Try Operator](https://google.github.io/comprehensive-rust/error-handling/try.html) and tell people "we'll come back to this, think of it as an early return".

djmitche:
I suspect this was an ordering dependency we missed when re-shuffling the content. 

This particular exercise has a good bit more error-handling code than just `?` and `Result` -- I lean toward adopting @randomPoison's fix here. Maybe a comment pointing forward to the Error Handling segment would help defer any questions or misperceptions from students.

mgeisler:
> This particular exercise has a good bit more error-handling code than just `?` and `Result` -- I lean toward adopting @randomPoison's fix here. 

Good point, I'm happy with that too!

mgeisler (approved):
Thanks! Let's get this merged, preferably with a comment pointing to the error handling part like Dustin suggests.

djmitche (changes requested):
(re-flag me for review when the comment is added)

randomPoison:
@djmitche I added a speaker note calling ahead to the error handling section. Let me know if that's not what you had in mind though!

djmitche (approved):

PR #2114, opened 2024-06-03
Fix dead link in glossary

### LINKED ISSUE #2110: Dead link found in the glossary.md translations
See attached image here 
image

I think the current link obsolete and should be replaced. Please take a loook

djmitche:
Fixes #2110. I think `references/strings.html` is a better choice as it covers both String and &str.


mgeisler (approved):
Sounds good, thanks!

Issue #567, opened 2023-04-14
Add async examples that can't run in playground as cargo crates


djmitche:
From https://github.com/google/comprehensive-rust/pull/496/#discussion_r1166652299

mgeisler:
Right now, we have `src/exercises/Cargo.toml`, which in turn references a few Rust files. It brings in dependencies such as reqwest, scraper, and thiserror. This gives us some coverage via `cargo build` and `cargo test`.

You could directly extend this to include the dependencies for async examples — or you could create a different project if needed (I would aim for simplicity and just extend the current file).

For the Android day, we have fully functioning `Android.bp` files in the repository. They reference all the examples and ensures that we can build it when needed. I would like to build it in CI, but I haven't set that up yet.

mgeisler:
We should also take a look at mdbook-keeper (#175), which I believe is a testing tool which would allow us to include snippets with dependencies.

djmitche:
Wow, this has been sitting for a while. And is solved! There are now crates for each exercise, e.g., `src/concurrency/sync-exercises/Cargo.toml`.

PR #2118, opened 2024-06-05
Concurrency fixes

mgeisler:
Hi folks! I'm teaching the Concurrency class, so I went through it and made a bunch of changes.

I debated trying to split this up into smaller PRs — some of the commits here depend on the previous commits, so it's a bit annoying to split up.

If you're okay with looking at this via the individual commits, then it's probably easier to merge it like that. Otherwise I can try splitting things up after teaching my class tomorrow.


src/concurrency/async/async-await.md
```diff
@@ -31,9 +31,6 @@ Key points:
 - Note that this is a simplified example to show the syntax. There is no long
   running operation or any real concurrency in it!
 
-- What is the return type of an async call?
-  - Use `let future: () = async_main(10);` in `main` to see the type.
-
```
fw-immunant, line 36: I think it's still worthwhile to take the step of demonstrating that the return type of an async fn is not the written return type, but actually a future type.

src/concurrency/async-pitfalls/async-traits.md
```diff
@@ -4,22 +4,20 @@ minutes: 5
 
 # Async Traits
 
-Async methods in traits are were stabilized only recently, in the 1.75 release.
-This required support for using return-position `impl Trait` (RPIT) in traits,
-as the desugaring for `async fn` includes `-> impl Future<Output = ...>`.
+Async methods in traits are were stabilized in the 1.75 release. This required
```
fw-immunant, line 7: Maybe we should fix "are were" while we're at it.

fw-immunant (changes requested):
A few comments inline, but looks good otherwise.

src/concurrency/async/async-await.md
djmitche: I agree that the answer is now less detailed (and less confusing), but I agree with @fw-immunant) that this should stay:
 - it illustrates that this is not a magical exception from the type rules
 - in general, it's a good technique for finding the type of a value, and bears a bit of repeating

src/concurrency/async-pitfalls/async-traits.md
djmitche: I assume you mean
```suggestion
Async methods in traits are, were, and always will be stabilized in the 1.75 release. This required
```
;)

Cargo.toml
```diff
@@ -23,6 +23,9 @@ members = [
   "src/std-traits",
   "src/std-types",
   "src/testing",
+  "src/concurrency/async",
+  "src/concurrency/async-control-flow",
+  "src/concurrency/async-pitfalls",
```
djmitche, line 29: I think this fixes #567. Thanks!!

djmitche (approved):
Looks good!

src/concurrency/welcome.md
```diff
@@ -6,23 +6,29 @@ target_minutes: 180
...
 {{%session outline}}
 
-<details>
+## Target Audience
 
-- Rust lets us access OS concurrency toolkit: threads, sync. primitives, etc.
-- The type system gives us safety for concurrency without any special features.
-- The same tools that help with "concurrent" access in a single thread (e.g., a
-  called function that might mutate an argument or save references to it to read
-  later) save us from multi-threading issues.
```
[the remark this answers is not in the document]
mgeisler: Oh, thanks! I don't think I wanted to remove this part :smile: I wanted move it to somewhere else... I'll have another look!

PR #2123, opened 2024-06-05
Add slide about `Result` to error handling section

randomPoison:
I think it's worth covering `Result` again in detail in the error handling section, even though we also cover it the day before in the std library types section. When we first introduce it we only briefly cover it, but don't go into much detail about error handling in general. I think it'd be good to re-introduce `Result` and talk more broadly about how it fits into Rust's error handling story before going into the `?` operator.

### DIFF
--- a/src/error-handling/result.md
+++ b/src/error-handling/result.md
@@ -1,0 +1,77 @@
+---
+minutes: 5
+---
+
+# `Result`
+
+Our primary mechanism for error handling in Rust is the [`Result`] enum, which
+we briefly saw when discussing standard library types.
+
+```rust,editable
+use std::fs::File;
+use std::io::Read;
+
+fn main() {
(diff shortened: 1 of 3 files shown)


src/error-handling/result.md
```diff
@@ -0,0 +1,77 @@
+---
+minutes: 5
```
djmitche, line 2: Since this is review and doesn't go into `?` it probably needs less time than this (but it's just an estimate so _shrug_)

djmitche (approved):
Nice! I think this provides a good tie back to the previous segment.

src/error-handling/result.md
randomPoison: I think 5 minutes is pretty accurate. When I get to this point I don't just talk about `Result` but about Rust's error handling strategy more generally.

PR #2125, opened 2024-06-05
Adjust timing of standard library slides

randomPoison:
Reduce the timings of the standard library slides to make the timing more accurate to how much time is spent in classes. This was a lingering issue from moving the generics section on day 3 to the afternoon, so day 3 afternoon has been showing up in PRs as taking too long.


djmitche (approved):
Nice, thank you!

PR #2130, opened 2024-06-07
Clarify the suggested steps in the pin page

djmitche:
The speaker notes suggest an evolution of the code to support a periodic timer, but the last step was under-specified.

(As mentioned by @fw-immunant and referenced in #1536)


fw-immunant (approved):
Yep, `Box::pin` was the only way I was able to get this suggestion to work--trying to use stack pinning is not compatible with re-initializing in any formulation I could come up with.

djmitche:
Agreed -- maybe this will get better in future Rust versions!

Issue #2132, opened 2024-06-08
The language switching menu requires clicking on the text, not elsewhere in the menu item


zhuba-Ahhh:
The language switching button seems to need to be clicked twice to take effect

djmitche:
Could you say more? I just tried:

 * click the globe icon
 * click on a language

result:

 * page switched to that language

djmitche:
Oh, I did notice that if I click on the row in the menu, but not on the actual text, it does nothing. For example, clicking to the right of the word "English".

Is that what you're seeing?

zhuba-Ahhh:
> Oh, I did notice that if I click on the row in the menu, but not on the actual text, it does nothing. For example, clicking to the right of the word "English".
> 
> Is that what you're seeing?

Sorry, it seems my network is stuck.

zhuba-Ahhh:
I seem to have discovered that clicking on the blank part of the picture below will not work. You can try to reproduce it.
you Can confirm whether this needs adjustment
<img width="361" alt="image" src="https://github.com/google/comprehensive-rust/assets/84793349/00a373f5-f600-42b8-8d76-82a3a4070827">

djmitche:
Agreed, there is a bug in the target area there!

PR #2143, opened 2024-06-12
Fix: Click the language switch menu item to switch the language

zhuba-Ahhh:
close: https://github.com/google/comprehensive-rust/issues/2132


zhuba-Ahhh:
> close: #2132

PR #2146, opened 2024-06-15
Update broken "channels" link

snazzybytes:
Stumbled upon another broken link. Courtesy fix 🫡


djmitche (approved):
Appreciated! We've had some discussions about linting for these, but it's not clear how to do so easily.

PR #2154, opened 2024-06-20
Simplify the `select!` example

djmitche:
@mgeisler PTAL, as I believe you were the one to suggest this change in #1536.

This leaves some detail about `select!` out, but that might be just as well. Some of that detail is addressed in the `pin!` pitfall, later.


mgeisler (approved):
Thanks, I like this a lot! More to the point and much easier for students to keep in their head. Thanks!

PR #2164, opened 2024-06-24
Test that correct value is returned for height change

qwandor:


djmitche (commented):
Can we make a change like this to a third-party file?

mgeisler:
> Can we make a change like this to a third-party file?

Yes, my understanding is that this is okay. The `third_party` directory is there to mark the files as not originally authored by us, but the files don't have to be a pristine copy of upstream.

mgeisler:
An unrelated point is that I think we could replace this exercise with a similar exercise written from scratch. Fewer third-party files simplifies things a little bit in my opinion.

fw-immunant:
We already have various changes to this file: https://github.com/google/comprehensive-rust/commits/ccaa1d27280ea939561c057d6e5bc33f5819908a/third_party/rust-on-exercism/health-statistics.rs

djmitche (approved):
I suspect this will lead to a discussion of why `==` isn't defined for floats!

PR #2167, opened 2024-06-24
Update installation instructions for cargo-embed

qwandor:


djmitche (approved):
People are going to hate another `curl | sh` install method!

PR #2168, opened 2024-06-24
Use stable rustfmt

### LINKED ISSUE #2165: dprint fmt failing

djmitche:
This uses the `rustfmt` from the stable toolchain, instead of nightly. At the moment, nightly's `rustfmt` is nonfunctional, but in general we don't need bleeding-edge formatting (and in fact, the more often formatting changes, the more often we'll get spurious failures on PRs, as we are today).

Fixes #2165.

### DIFF
--- a/dprint.json
+++ b/dprint.json
@@ -9,7 +9,7 @@
       "command": "yapf3",
       "exts": ["py"]
     }, {
-      "command": "rustup run nightly rustfmt --edition 2021",
+      "command": "rustup run stable rustfmt --edition 2021",
       "exts": ["rs"]
     }, {
       "command": "msgcat -",


djmitche:
Once this is merged, I think we can rebase a bunch of other open PRs and land them.

qwandor (approved):

mgeisler:
> in general we don't need bleeding-edge formatting (and in fact, the more often formatting changes, the more often we'll get spurious failures on PRs, as we are today).

This is not 100% true: we use unstable features in our `rustfmt.toml` file:

```toml
# Please use a nightly rustfmt for these settings.
imports_granularity = "module"
wrap_comments = true
```

Luckily, both of these settings are no-ops with a stable `rustfmt`, as long as people wrap things nicely by hand :smile:

We could instead pin the nightly version used in CI: you're right that we don't want the "latest and greatest" when it comes to formatting, we just want a consistent result and some useful features as well.

PR #2171, opened 2024-06-25
Fix publish workflow

qwandor:
Need to `apt-get update` before `apt-get install`.

### DIFF
--- a/.github/workflows/install-mdbook/action.yml
+++ b/.github/workflows/install-mdbook/action.yml
@@ -18,6 +18,7 @@ runs:
     - name: Install mdbook-pandoc and related dependencies
       run: |
         cargo install mdbook-pandoc --locked --version 0.6.4
+        sudo apt-get update
         sudo apt-get install -y texlive texlive-luatex texlive-lang-cjk librsvg2-bin fonts-noto
         curl -LsSf https://github.com/jgm/pandoc/releases/download/3.2/pandoc-3.2-linux-amd64.tar.gz | tar zxf -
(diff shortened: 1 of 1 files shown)


djmitche (approved):

mgeisler:
> Need to `apt-get update` before `apt-get install`.

When I see this, I wonder: how did it work in the past?

The answer is that we've flip-flopped back and forth on this :smile: The `apt update` was added in the badly named #634, then taken out in #1490, and added back in #1584. I guess it was taken out since then since you're adding it back now.

This is a signal to me that there is something to learn here. Perhaps we need to run `apt update` in a small retry-loop? Perhaps we should only run it if `apt install` fails?

The three of us should avoid flip-flopping further on this and make a better fix next time we see problems here. That way we can move forward on this and hopefully make a robust and long-lasting fix.

djmitche:
Oh, good observation!

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/customizing-github-hosted-runners says "Always run sudo apt-get update before installing a package." My understanding is that it is necessary as the repo data cached in the docker image becomes out of date quite quickly, and you start getting 404's downloading packages that are no longer at that location.

I think the error in #1490 is not one that would succeed on retry. However, we have found that an many cases the `apt-get update` is not necessary, so perhaps we could make this step "soft-fail" -- that is, try to update but if that does not work continue on with the rest of the workflow. If there are downstream errors, they will be during `apt-get install` and it should be pretty quick to figure out what the root cause was.

PR #2182, opened 2024-07-01
Publish Japanese translation to language selector

henrif75:
#1463 #652

### DIFF
--- a/theme/index.hbs
+++ b/theme/index.hbs
@@ -189,6 +189,9 @@
                               <a id="zh-TW">Chinese Traditional (漢語)</a>
                           </button></li>
                           <li role="none"><button role="menuitem" class="theme">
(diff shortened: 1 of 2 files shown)


qwandor (approved):

theme/index.hbs
```diff
@@ -188,6 +188,9 @@
                           <li role="none"><button role="menuitem" class="theme">
                               <a id="zh-TW">Chinese Traditional (漢語)</a>
                           </button></li>
+                          <li role="none"><button role="menuitem" class="theme">
+                              <a id="ko">Japanese (日本語)</a>
```
qwandor, line 192: This seems to be linking to the Korean translation instead.

henrif75: Thanks for the catch, I'm sending out a fix right away.

PR #2186, opened 2024-07-04
Fix issues in std-types/result.md

JasonXJ:
The expression excercise is using the standard `Result` type instead of `Res`.


JasonXJ:
I think I just fixed the CLA issue (I am a googler btw)

djmitche (approved):
Thanks!

PR #2187, opened 2024-07-04
Remove lifetime trait bound on ProtoMessage

JasonXJ:
The lifetime trait bound is unncessary, and the code compiles without it.


fw-immunant (approved):
Agreed, as we don't use `'a` in `self` or the references to it there's no reason to bound the trait on it.

PR #2188, opened 2024-07-04
Run msgmerge for all languages.

qwandor:


mgeisler:
Hi @qwandor, is is generally wrong to mass-update the `.po` files. They must be synchronized with the version of the sources present at the time recorded in the POT-Creation-Date header in each file.

The header is updated automatically when translators refresh the translation: https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md#refreshing-an-existing-translation. Until it is updated, the translation remains fixed at the old version. This mechanism is what gives the translators a chance to work in peace on a given version of the course.

PR #2189, opened 2024-07-04
Publish all languages.

qwandor:


henrif75 (approved):

.github/workflows/publish.yml
```diff
@@ -19,7 +19,7 @@ concurrency:
 env:
   CARGO_TERM_COLOR: always
   # Update the language picker in index.hbs to link new languages.
-  LANGUAGES: bn da de es fa fr it ja ko pt-BR vi zh-CN zh-TW
+  LANGUAGES: bn da de el es fa fr id it ja ko pl pt-BR ro ru tr uk vi zh-CN zh-TW
```
mgeisler, line 22: If we're going to publish all languages, then we don't need this list :smile: 

The list here has been kept in rough alignment with the list of [Incomplete Translations](https://google.github.io/comprehensive-rust/running-the-course/translations.html#incomplete-translations), which is updated by hand as translators feel like they're ready to make their work available to the world.

I've been hesitant to generate and publish HTML for half-done translations since there is a risk that search engines get confused by the duplicate content. A half-done translation will result in lots and lots of pages with English content. We emit a correct `<html lang="xx">` tag, so I guess it should be fine, but it's something to be aware of.

PR #2190, opened 2024-07-04
Publish report of translation status.

qwandor:


djmitche (commented):
It doesn't look like this report is linked anywhere? What is it? I see https://github.com/google/mdbook-i18n-helpers/tree/main/i18n-report but there's no documentation about it.

qwandor:
I'll add links in a separate PR once I'm sure it's working properly. It's a report showing the status of each of the different language translations of the course. You can try generating it locally yourself with `i18n-report report.html po/*.po` if you want to see it before this PR is merged.

README.md
```diff
@@ -67,6 +67,7 @@ Then install these tools with:
 cargo install mdbook
 cargo install --locked mdbook-svgbob
 cargo install --locked mdbook-i18n-helpers
+cargo install --locked i18n-report
```
mgeisler, line 70: Technically, it's only the GitHub action that needs this tool, right? So we could avoid mentioning it in the README and thus save people a bit of time when setting up the tooling.

Since https://crates.io/i18n-report is currently empty, the command will actually fail right now.

@qwandor, would you be up for renaming the crate to something more specific to Gettext/PO files? Now that I see it in an install command, the name "i18n report" seems a bit too generic and all-encompassing :-)

Perhaps we could call it `gettext-report` or `po-stats` or similar? The nearest equivalent in the Gettext suite it [`msgfmt --stat`](https://www.gnu.org/software/gettext/manual/html_node/msgfmt-Invocation.html).

qwandor: > Technically, it's only the GitHub action that needs this tool, right? So we could avoid mentioning it in the README and thus save people a bit of time when setting up the tooling.
People might want to run it locally to see the status of translations after their changes, so I figured it was best to put the installation instructions here with the rest of the tools.

> Since https://crates.io/i18n-report is currently empty, the command will actually fail right now.
Do you mean https://crates.io/crates/i18n-report? I've sent https://github.com/google/mdbook-i18n-helpers/pull/213 to add a readme, but in any case it will still install fine.

> @qwandor, would you be up for renaming the crate to something more specific to Gettext/PO files? Now that I see it in an install command, the name "i18n report" seems a bit too generic and all-encompassing :-)
> 
> Perhaps we could call it `gettext-report` or `po-stats` or similar? The nearest equivalent in the Gettext suite it [`msgfmt --stat`](https://www.gnu.org/software/gettext/manual/html_node/msgfmt-Invocation.html).

Those do both seem like reasonable names, but I'm reluctant to rename it after already publishing it to crates.io, as it would be confusing to have two names for the same thing show up when people search for it there.

mgeisler: > > Since https://crates.io/i18n-report is currently empty, the command will actually fail right now.
> 
> Do you mean https://crates.io/crates/i18n-report? I've sent [google/mdbook-i18n-helpers#213](https://github.com/google/mdbook-i18n-helpers/pull/213) to add a readme, but in any case it will still install fine.

Oh, sorry, I meant the latter. I was too fast in my typing :smile: 

> > Perhaps we could call it `gettext-report` or `po-stats` or similar? The nearest equivalent in the Gettext suite it [`msgfmt --stat`](https://www.gnu.org/software/gettext/manual/html_node/msgfmt-Invocation.html).
> 
> Those do both seem like reasonable names, but I'm reluctant to rename it after already publishing it to crates.io, as it would be confusing to have two names for the same thing show up when people search for it there.

Yes, I see... I guess this name is okay then and it's not like there are a lot of demand for these crate names.

Issue #2173, opened 2024-06-25
Local `dprint` and `dprint` in CI want opposite things


djmitche:
locally, after installing the latest dprint and clearing ~/.cache/dprint,
```
from /usr/local/google/home/djmitche/p/comprehensive-rust/po/pl.po:
26052 26052| #~·"········let·Some(request_segment)·=·request_segments.next()·else·{\n"
26053 26053| #~·"············return·false;\n"
26054 26054| #~·"········};\n"
      26055|-#~·"········if·request_segment·!=·prefix_segment·&&·prefix_segment·!=·"
      26056|-#~·"\"*\"·{\n"
26055      |+#~·"········if·request_segment·!=·prefix_segment·&&·prefix_segment·!=·\"*\"·"
26056      |+#~·"{\n"
26057 26057| #~·"············return·false;\n"
26058 26058| #~·"········}\n"
26059 26059| #~·"····}\n"
```
but [fixing that in CI](https://github.com/google/comprehensive-rust/actions/runs/9663627985/job/26656256816?pr=2166) leads to
```
from /home/runner/work/comprehensive-rust/comprehensive-rust/po/pl.po:
26052 26052| #~·"········let·Some(request_segment)·=·request_segments.next()·else·{\n"
26053 26053| #~·"············return·false;\n"
26054 26054| #~·"········};\n"
      26055|-#~·"········if·request_segment·!=·prefix_segment·&&·prefix_segment·!=·\"*\"·"
      26056|-#~·"{\n"
26055      |+#~·"········if·request_segment·!=·prefix_segment·&&·prefix_segment·!=·"
[26](https://github.com/google/comprehensive-rust/actions/runs/9663627985/job/26656256816?pr=2166#step:5:29)056      |+#~·"\"*\"·{\n"
26057 26057| #~·"············return·false;\n"
26058 26058| #~·"········}\n"
26059 26059| #~·"····}\n"
```
(I'll save you squinting at those -- they are opposite diffs)

I presume this is due to different `msgcat` versions?
```
⸩ msgcat --version
msgcat (GNU gettext-tools) 0.21
```
Is anyone else seeing this? How can we make this stable? I have a `dprint check` pre-commit hook, but as it stands I always have to use `--no-verify` because it always fails.

djmitche:
Yes, someone else is seeing it :)

Shoudl we maybe disable formatting for .po files until this can be sorted out?

mgeisler:
[...]

PR #2192, opened 2024-07-05
Fix the format by `dprint fmt`

younies:
When running `dprint fmt`, the formatter automatically provided this PR fix.


djmitche:
It looks like this is the same as #2173.

PR #2205, opened 2024-07-08
es: Fix formatting error in svgbob figures

### LINKED ISSUE #1329: es: Formatting errors in svgbob diagrams

carreter:
This PR removes unsupported markdown styling from the svgbob figures in the Spanish translation.

Fixes #1329.


mgeisler (approved):
Gracias!

carreter:
De nada :)

Issue #2050, opened 2024-05-03
Verify and prune CODEOWNERS list for translations


henrif75:
Hi code owners (@suetfei, @wnghl, @anlunx, ...), can you please take a look at this one?

@henrif75, I think we should consider pruning the list of code owners here in case some of them have moved on — this way the contributors will have a more clear picture of what to expect when creating a PR. It's a lot of work to review these PRs, so a huge thank you goes to the many volunteers here!

_Originally posted by @mgeisler in https://github.com/google/comprehensive-rust/issues/1992#issuecomment-2093010247_

suetfei:
I am OK with pruning myself.

PR #2210, opened 2024-07-09
Prune CODEOWNERS file

henrif75:
Removed users without access to the repository.
#2050

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,17 +6,17 @@ po/de.po @ronaldfw @fechu
 po/el.po @Mandragorian
 po/es.po @deavid @vzz1x2
 po/fa.po @DannyRavi @hamidrezakp @moaminsharifi @mehrad77 @javad-jafari
-po/fr.po @sakex @lb034582341
-po/id.po @tjonganthony @nurlitadf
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -6,17 +6,17 @@ po/de.po @ronaldfw @fechu
 po/el.po @Mandragorian
 po/es.po @deavid @vzz1x2
 po/fa.po @DannyRavi @hamidrezakp @moaminsharifi @mehrad77 @javad-jafari
-po/fr.po @sakex @lb034582341
-po/id.po @tjonganthony @nurlitadf
-po/it.po @detro @nicomazz @bznein
+po/fr.po @sakex
+po/id.po
```
qwandor, line 10: Having no owners for this translation doesn't seem ideal, though if there really aren't any then should it be removed from the file entirely?

henrif75: I agree, there was only an initial effort to set it up and then nothing. I'll go ahead and remove it all.

mgeisler (approved):
Thanks @henrif75!

Could you add a comment on the issues for `id` to let people know that we're looking for a maintainer?

Issue #1911, opened 2024-03-10
Pandoc failure during local rendering


henrif75:
Following the build procedures listed in the [README.md](https://github.com/henrif75/comprehensive-rust/blob/main/README.md) page results in `pandoc` failure:
 
```
mdbook serve
2024-03-10 09:56:02 [INFO] (mdbook::book): Book building has started
2024-03-10 09:56:02 [INFO] (mdbook::book): Running the exerciser backend
2024-03-10 09:56:02 [INFO] (mdbook::renderer): Invoking the "exerciser" renderer
2024-03-10 09:56:03 [INFO] (mdbook::book): Running the html backend
2024-03-10 09:56:04 [INFO] (mdbook::book): Running the pandoc backend
2024-03-10 09:56:04 [INFO] (mdbook::renderer): Invoking the "pandoc" renderer
[2024-03-10T16:56:07Z INFO  mdbook_pandoc] Processing redirects in [output.html.redirect]
[2024-03-10T16:56:07Z WARN  mdbook_pandoc::preprocess] Failed to resolve redirect: why-rust/an-example-in-c.html => ../hello-world/example.html: Unable to normalize redirect destination: Unable to canonicalize path: /Users/henrif/Documents/comprehensive-rust/book/pandoc/pdf/src/why-rust/../hello-world/example.html: No such file or directory (os error 2)
[2024-03-10T16:56:07Z WARN  mdbook_pandoc::preprocess] Failed to resolve redirect: exercises/day-2/soluções-tarde.html => solutions-afternoon.html: Unable to normalize redirect destination: Unable to canonicalize path: /Users/henrif/Documents/comprehensive-rust/book/pandoc/pdf/src/exercises/day-2/solutions-afternoon.html: No such file or directory (os error 2)
[2024-03-10T16:56:07Z WARN  mdbook_pandoc::preprocess] Failed to resolve redirect: exercises/day-3/soluções-tarde.html => solutions-afternoon.html: Unable to normalize redirect destination: Unable to canonicalize path: /Users/henrif/Documents/comprehensive-rust/book/pandoc/pdf/src/exercises/day-3/solutions-afternoon.html: No such file or directory (os error 2)
[...]

PR #2212, opened 2024-07-10
Upgrade `mdbook-pandoc` and disable it except during CI workflows

max-heller:
Implements the approach summarized in https://github.com/google/mdbook-i18n-helpers/issues/200#issuecomment-2220035461 to disable running `mdbook-pandoc` even if it is available to avoid dependency-related issues during local rendering.

Should address https://github.com/google/comprehensive-rust/issues/1911 since `mdbook-pandoc` will no longer run locally.

### DIFF
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -36,6 +36,9 @@ else
     fi
 fi
 
+# Enable mdbook-pandoc to build PDF version of the course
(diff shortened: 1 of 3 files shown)


djmitche (approved):
Appreciated!

mgeisler:
Thanks a lot, @max-heller!

PR #2214, opened 2024-07-10
Diff translations on every PR

qwandor:
This uses a new mode of `i18n-report` to summarise the differences in translations as an action on each PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -162,3 +162,27 @@ jobs:
         env:
           MDBOOK_BOOK__LANGUAGE: ${{ matrix.language }}
         run: mdbook test
+
+  po-diff:
+    name: Translation diff
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup Rust cache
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -162,3 +162,27 @@ jobs:
...
+        uses: ./.github/workflows/install-mdbook
+
+      - name: Copy latest version of translations
+        run: cp -r po old-po
+
+      - name: Checkout base version
+        run: git checkout ${{ github.event.pull_request.base.sha }}
+
+      - name: Translation diff
+        run: i18n-report diff old-po/ po/ > $GITHUB_STEP_SUMMARY
```
kdarkhan, line 188: Just to confirm, in order to see the diff, the users will need to click to the executed job and see the summary from there? It won't be visible from the PR page directly?

There is [an existing action](https://github.com/google/comprehensive-rust/blob/d4e2e5d37ffdb0a9a7ffea51758d155c959024d8/.github/workflows/course-schedule-comment.yml) which leaves a comment within the PR. Will it be useful if a similar approach is used here?

kdarkhan: Sure, I can take look. Maybe it can be made generic enough such that the same action could be used for both.

kdarkhan (approved):

PR #2217, opened 2024-07-11
Update udev rule for microbit.

qwandor:
The subsystem seems to have changed with a newer version of udev, and using logindev and uaccess rather than plugdev is consistent with how other similar devices are handled.


adetaylor (approved):
I don't know anything about these udev or bare-metal parts of the course, so I'm not sure if it was intended that I review it, but LGTM as a generic keeping-up-to-date update

PR #2239, opened 2024-07-20
some template dont convert en to fa

javad-jafari:


po/fa.po
```diff
@@ -2713,13 +2713,13 @@ msgstr ""
 #: src/concurrency/shared-state.md src/concurrency/async.md
 #: src/concurrency/async-control-flow.md src/concurrency/async-pitfalls.md
 msgid "10 minutes"
-msgstr ""
+msgstr "10 دقیقه"
```
DannyRavi, line 2716: number should be in Farsi, like ۲ , not 2

DannyRavi (changes requested):
please Fix number error

DannyRavi (approved):
awesome

PR #2250, opened 2024-07-23
Set Rust caching key per language group

### LINKED ISSUE #2247: Link check runs on all languages
After merging #2195, I realized that `mdbook-linkcheck` runs on all languages. This is despite our best efforts in `build.yml` where we only install the tool on known-good languages:

https://github.com/google/comprehensive-rust/blob/ad9b41cd6cba0b87a2560a155fc5afff1f41713d/.github/workflows/build.yml#L150-L153

The problem is likely due to caching: while we only store the cache when building the English version, we use the cache for all languages.

jond01:
Fixes #2247 (a follow-up from #2195).

* Add a workflow dispatch manual trigger to the tests workflow.
* Move the link-checked languages JSON list to the build job's env.
* Set a caching key per language group to avoid extra `mdbook-linkcheck` installation on non-link-checked translations.
  For the meaning of this key, see:
  https://github.com/Swatinem/rust-cache?tab=readme-ov-file#example-usage
* Fix newly broken web link in 8a3ed21 ✅ 

I tested it on my fork, and it looks good:
https://github.com/jond01/comprehensive-rust/actions/workflows/build.yml
https://github.com/jond01/comprehensive-rust/actions/caches

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -132,6 +132,8 @@ jobs:
 
       - name: Setup Rust cache
         uses: ./.github/workflows/setup-rust-cache
+        with:
+          key: ${{ contains(fromJSON(env.LINK_CHECKED_LANGUAGES), matrix.language) }}
 
       - name: Install Gettext
         run: |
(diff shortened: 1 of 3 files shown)


qwandor (approved):

mgeisler:
Wow, thanks for fixing this so comprehensibly, @jond01!

PR #2259, opened 2024-08-02
fix(fa.po):Day1:Morning

javad-jafari:


DannyRavi (approved):
everything is alright.

PR #2268, opened 2024-08-08
Remove course-schedule workflows

djmitche:
These are complicated and not especially useful.


mgeisler (approved):
Thanks for simplifying our infrastructure!

PR #2276, opened 2024-08-12
Change let-else example to demonstrate undesirable nesting

### LINKED ISSUE #2070: `let`...`else` example should demonstrate denesting
Often when teaching people are underwhelmed by the utility of the `let`...`else` construct. I also don't think it makes a huge amount of sense as presented. The current example consists of several repetitions of the admittedly awkward pattern
```rust
let s = if let Some(s) = maybe_string {
	s
} else {
	return Err(String::from("got None"));
};
```
which is improved by `let`...`else`, but is also not a terribly common pattern on its own, and as I understand things isn't really what motivated this control flow construct.

My understanding is that the real utility of `let`...`else` is that it allows to de-nest what would otherwise be a rightward-drifting nesting of `if let`s:
```rust
if let Some(x) = obtain_x() {
	if let Some(y) = x.obtain_y() {
		if let z = compute_z(x, y) {
			// work with z
		}
	}
}
```
Which can instead become:
```rust
let Some(x) = obtain_x() else { return }
let Some(y) = x.obtain_y() else { return }
let Some(z) = compute_z(x, y) else { return }
// work with z
```

I think this makes the utility of the construct more clear.

fw-immunant:
Fixes #2070.

Previously we showed a forcibly de-nested version using both let and if-let. this is not a construction that new learners of Rust are likely to have seen or written, while nesting if-let is closer to patterns that appear in other languages and better motivates the de-nesting transformation to let-else

### DIFF
--- a/src/pattern-matching/let-control-flow.md
+++ b/src/pattern-matching/let-control-flow.md
@@ -42,22 +42,18 @@ off the end of the block).
 
 ```rust,editable
 fn hex_or_die_trying(maybe_string: Option<String>) -> Result<u32, String> {
-    let s = if let Some(s) = maybe_string {
-        s
+    if let Some(s) = maybe_string {
+        if let Some(first_byte_char) = s.chars().next() {
+            if let Some(digit) = first_byte_char.to_digit(16) {
+                Ok(digit)
+            } else {
+                return Err(String::from("not a hex digit"));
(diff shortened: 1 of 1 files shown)


mgeisler (approved):
Looks clearer to me, thanks!

PR #2278, opened 2024-08-12
interior-mutability.md: swap Cell/RefCell order and improve explanation

### LINKED ISSUE #2249: Swap order of Cell/RefCell
`Cell` is the simpler of these two so I always introduce it first. Do any other instructors object?

We also don't have much prose about how `RefCell` actually works in terms of the `Ref` and `RefMut` types and "runtime borrow checking"; it might be good to add more.

fw-immunant:
Fixes #2249.


mgeisler (approved):
I'm very happy to swap them!

With #2258 landing at some point, we'll soon be able to detect overly long slides — this one looks like one of them. So perhaps you could split up the page already now?

djmitche:
We can split in a follow-up, so merging now..

PR #2279, opened 2024-08-12
operators.md: mention Not trait's un-C-like behavior

### LINKED ISSUE #2072: Call out `Not` trait in speaker notes
We have a slide on the various operator traits in the standard library, which focuses on their overall design and how to implement these for our own types so we can use operator syntax with them. This is good, but there are a couple semantic surprises in how Rust operators behave, which we should note explicitly.

One is that the `Not` operator (`!`) does not "boolify" like the C/C++/JS incarnations of this operator; it instead negates each bit of an integer, so `!5u8` is 250.

Another is that the `&&`/`||` operators are not tied to traits because they are lazy in their right-hand sides, so they cannot be used with user-defined types.

fw-immunant:
Fixes #2072.

### DIFF
--- a/src/std-traits/operators.md
+++ b/src/std-traits/operators.md
@@ -44,6 +44,11 @@ Discussion points:
 - You could implement `Add` for two different types, e.g.
   `impl Add<(i32, i32)> for Point` would add a tuple to a `Point`.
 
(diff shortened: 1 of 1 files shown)


mgeisler (approved):
Thanks!

PR #2280, opened 2024-08-12
concurrency: avoid 1.. loops

### LINKED ISSUE #2060: concurrency: Avoid `1..ROUND_NUMBER` loops
Teaching the concurrency course it sticks out as strange that many of the example loops start at 1 and continue to (exclusively) a round number, such as `1..5` or `1..10`. This means we're comparing, for example, a loop in main that counts 1 2 3 4 with a loop in a background thread that counts 1 2 3 4 5 6 7 8 9: 4 and 9 iterations, respectively. I think it would be less distracting to loop for round numbers of iterations like 5 or 10, whether we start at 1 or 0. When I see a loop that starts 1 and has a non-inclusive end bound, my bug sense tingles, and I figure the same is probably true for a number of students; it's less distracting to do something more idiomatic as looping is not the focus here.

fw-immunant:
Fixes #2060. Note that this does not change the "blocking executor" example because on that slide it is worthwhile to sleep for 1 * 10ms on the first iteration and so on. But we shouldn't use one-indexed inclusive loops when the only significant feature of the loop is its iteration count.

### DIFF
--- a/src/concurrency/async/async-await.md
+++ b/src/concurrency/async/async-await.md
@@ -10,7 +10,7 @@ At a high level, async Rust code looks very much like "normal" sequential code:
 use futures::executor::block_on;
 
 async fn count_to(count: i32) {
-    for i in 1..=count {
+    for i in 0..count {
         println!("Count is: {i}!");
     }
 }
--- a/src/concurrency/async/runtimes/tokio.md
+++ b/src/concurrency/async/runtimes/tokio.md
@@ -10,7 +10,7 @@ Tokio provides:
 use tokio::time;
 
 async fn count_to(count: i32) {
-    for i in 1..=count {
+    for i in 0..count {
         println!("Count in task: {i}!");
         time::sleep(time::Duration::from_millis(5)).await;
     }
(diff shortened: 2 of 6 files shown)


mgeisler (approved):
Oh, I like this much better! The `..=` syntax has never really felt nice to me, though I understand why it's there (you cannot write `..256u8`).

src/concurrency/async/async-await.md
```diff
@@ -10,7 +10,7 @@ At a high level, async Rust code looks very much like "normal" sequential code:
 use futures::executor::block_on;
 
 async fn count_to(count: i32) {
-    for i in 1..=count {
+    for i in 0..count {
```
mgeisler, line 13: We will get an extra line of output now (here and everywhere else). You could consider reducing the input `count` to combat this.

I remember having changed some inputs from, say, 5 or 10 to just 3 when teaching precisely so that I can have everything on screen and reduce the vertical scrolling.

fw-immunant: I think there are few enough lines in most of these examples that this is reasonable.

PR #2281, opened 2024-08-12
from-and-into.md: mention losslessness and infallibility

### LINKED ISSUE #2069: `From` and `Into` slide should emphasize losslessness and infallibility
The standard library docs [mention this](https://doc.rust-lang.org/stable/std/convert/trait.From.html):
> The `From` trait is intended for perfect conversions. If the conversion can fail or is not perfect, use [`TryFrom`](https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html).

But it's easy to miss one line in the docs; I think we should emphasize this in the speaker notes. It also isn't necessarily obvious what "perfect" means, so I think "lossless" is a better way to describe the contract of these traits. This is worth underlining because the commonsense notion of being able to convert between types may also apply to things like C-style casts that truncate and are not appropriate for these traits.

fw-immunant:
Fixes #2069.

### DIFF
--- a/src/std-traits/from-and-into.md
+++ b/src/std-traits/from-and-into.md
@@ -4,7 +4,8 @@ minutes: 5
 
 # `From` and `Into`
 
-Types implement [`From`][1] and [`Into`][2] to facilitate type conversions:
+Types implement [`From`][1] and [`Into`][2] to facilitate type conversions.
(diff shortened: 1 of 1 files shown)


mgeisler (approved):
Nice!

Issue #63, opened 2022-12-26
Technical nits: Threads and Drop


QuineDot:
[This slide](https://google.github.io/comprehensive-rust/concurrency/threads.html) says that threads are all daemon threads, but this is untrue:
* For one, you're only talking about `thread::spawn` created threads, as per the very next slide
* For two, these threads are only detached [if you drop the `JoinHandle`](https://doc.rust-lang.org/std/thread/#spawning-a-thread)
* You also mention panicking and payloads, but you can only catch unwind payloads *on the current thread*; for panics of another `spawn`ed thread, you call `join` on the join handle to see if it panicked and to get the payload
  * I.e. you can only do this if the thread isn't detached

So you should talk about `JoinHandle` and detaching via `Drop`.

[On the scoped page](https://google.github.io/comprehensive-rust/concurrency/scoped-threads.html) you should talk about the different behavior:
* `ScopedJoinHandle` joins and blocks upon `Drop` instead of detaching
* Any dropped `ScopeJoinHandle` that witnesses a thread panic will cause `thread::scope` to panic
* But you can manually `join` and check for thread panics instead, similar to `JoinHandle`

Aside from the ability to borrow, this difference in handle drop behavior (and thus what may panic where) is the main difference between the two tools.

steffahn:
> * `ScopedJoinHandle` joins and blocks upon `Drop` instead of detaching

I don't think this is true. AFAIK, it only blocks upon leaving the `scope` the thread was spawned in.

mgeisler:
Hi @QuineDot,

Thanks for reading the slides so carefully :-)

> [This slide](https://google.github.io/comprehensive-rust/concurrency/threads.html) says that threads are all daemon threads, but this is untrue:
>
> * For one, you're only talking about `thread::spawn` created threads, as per the very next slide
> * For two, these threads are only detached [if you drop the `JoinHandle`](https://doc.rust-lang.org/std/thread/#spawning-a-thread)

Okay, I had not thought of it from that perspective. But let me first understand if we understand the same by _daemon thread_?

I first encountered the term in [Python](https://docs.python.org/3/library/threading.html#:~:text=A%20thread%20can%20be%20flagged,or%20the%20daemon%20constructor%20argument.) where a _daemon thread_ is a thread which will _not_ keep the program alive. That is all — non-daemon threads will keep the program alive, even if you exit from the main thread. When teaching the class, I was told that Java people also know the term with the same meaning.

[...]

PR #2282, opened 2024-08-12
threads: do not refer to "daemon" threads

fw-immunant:
This is the last remaining fix left from #63.


mgeisler (approved):
Thanks for taking out Python/Java specific terminology.

PR #2287, opened 2024-08-19
update break-continue.md

JasonXJ:
Move the speak note about "loop" into the page which we actually talk about "loop". Also update the wording to make it more accurate.


mgeisler (approved):
Thanks! This probably got out of sync in a reshuffle at some point :smile:

PR #2292, opened 2024-08-19
Update shared.md

JasonXJ:


fw-immunant (changes requested):
I think "responsibility" is too vague here, but "ownership" may not yet be clear (and prepositions should agree, "ownership of". I think it might be best to say "without taking responsibility for destroying the value".

JasonXJ:
Yeah, it should be "ownership of". IMO, using "ownership" is ok:

* It is the standard phrase we use for Rust.
* I think most software engineers do understand this concept
* The instructor of the course can explain more if anyone have questions

"without taking responsibility for destroying the value" is a bit more clear but it may imply that the owner needs to manually/explicitly destroy the value like in C.

fw-immunant (approved):

mgeisler:
Thanks, I also like that this uses the word "ownership" which you see Rust people throw around all the time.

JasonXJ:
(Just formatted the file to fix the pre-submit)

PR #2295, opened 2024-08-20
Update trait-bounds.md

JasonXJ:
We are going to talk about `impl Trait` in the very next slide. Let's don't duplicate it here and complicate the flow?


mgeisler (approved):
Nice, I agree this is shorter and flows better. I also like that the `T: Trait` syntax is how you talk about a type implementing a trait, which is, well, what this slide is about :smile:

Issue #1581, opened 2023-12-12
Exercise consistency: tests or main? and allow(unused)


djmitche:
See #1565 for context.

Also, put all text above the code sample!

mgeisler:
Good questions!

There are some practical reasons and constraints for some of this:

- Tests are not run in the embedded playgrounds — the playgrounds need a `main` function.
- On the other hand, we run tests via `cargo test` in CI!

So I like tests because they help keep the course code correct. I also think it's fine to show people simple unit tests from day 1: they're just small functions after all and don't look scary in my opnion.

We ask people to solve most or all exercises in https://play.rust-lang.org/ or locally using an editor with `rust-analyzer`. This means that they can execute tests just fine. However, having passing tests be the criteria for success takes away some of the free-form experimentation that you can do when you just have some code and a `main` function to call it from.

Related to this, it always bothered me how Rust doctests have to use `assert_eq!(foo, 123)` to express the expectations in your example. It feels dry and it means I cannot talk about things like `println!` or `dbg!` output in my example. I like the Python way much more where you show an interactive console session.

However, since we're not Python and don't have such interactive tools, I guess using small unit tests is okay. A bit boring, but probably also more professional :smile:

PR #2302, opened 2024-08-22
Added tests and implementation for I32(float) for Exercise: Protobuf Parsing

saharulik:
#  Update for Exercise: Protobuf Parsing
Some changes needed for better understanding of topic like tests and wire types.
## Changes 📋
- ### Add and change text and code: _exercise.md,_ _exercise.rs_ 📝
- ### Include tests 🧪
- ### `FieldValue::I32` to `f32` 
- ### Add field for message `PhoneNumber` 💰

## `FieldValue::I32` to `f32`  and add field for message `PhoneNumber`
Change implementation of proto message
Add to `PhoneNumber` message the field:
```proto
message PhoneNumber {
  optional string number = 1;
  optional string type = 2;
  + optional float balance = 3; 
}
```
The idea behind this decision is implement I32. It's used in code but not tested or used.
Firstly it's good idea to show how to works with  "I32" wire type of Protocol Buffers. And,
show how works with different types of data in proto and just in CS.

For the new field I choose _balance_  of number phone. The main idea is best solution for 
choosing between signed number and float type. Also change in main Person object decoding 
to work with new float type of field. 

## Tests
Tests is really helpful. It's show us what result we should except and just help to faster 
and better understand problem. Problem with implementation of code. 
For testing was chosen 4 test with idea of full explanation of working with bytes of 
protocol buffers.

> For some reasons then tried make equation between same objects the `[]` is not vector like 
> macro `println!` print it. For this problem in tests implemented ver![] for [] in field vector of 
> PhoneNumber. 

And, for working with equation added the derive **EquationEq**.
```rust
#[test] // Example
    fn test_id() {
        let person_id: Person = parse_message(&[
            0x10, 0x2a,
        ]);
        assert_eq!(person_id, Person { name: "", id: 42, phone: vec![], });
    } 
```

[^1]: My reference.


djmitche (changes requested):
Thanks for this!

I see that the comment in `exercise.rs` for `WireType::I32` is incorrect - it is precisely 32 bytes, but can contain any of `float`, `fixed32`, or `sfixed32`:
```
i32        := sfixed32 | fixed32 | float;
                encoded as 4-byte little-endian;
                memcpy of the equivalent C types (u?int32_t, float)
```
so the added sentence suggesting that I32 always contains a float is a bit misleading.

I also acknowledge that this type isn't actually used, aside from `as_i32`, which isn't used. I think this was once used in the exercise, but that usage was removed in the interest of shortening the duration of the exercise (hence the `#[allow(dead_code)]`).

In order to avoid the confusion over wire types and proto types, maybe it's best to just omit `I32` entirely, as has been done for `I64`.

To the question of adding tests: because this exercise is typically solved in the playground, we're a little limited as to what will run. With both a `main()` and `mod tests`, the playground chooses to run `main` and not the tests. So, they are a bit superfluous to include in `exercise.md`. However, it makes sense to include tests in `exercise.rs` to detect any bugs accidentally introduced to the exercise by later PRs.

So, I'd suggest:
 * removing use of I32 entirely
 * removing the `ANCHOR` from the tests, and the corresponding `{{#include}}`

saharulik (commented):
@djmitche Thanks for replying!
> so the added sentence suggesting that I32 always contains a float is a bit misleading.
> 
> I also acknowledge that this type isn't actually used, aside from `as_i32`, which isn't used. I think this was once used in the exercise, but that usage was removed in the interest of shortening the duration of the exercise (hence the `#[allow(dead_code)]`).
> 
> In order to avoid the confusion over wire types and proto types, maybe it's best to just omit `I32` entirely, as has been done for `I64`.

I really fully agree with you, but this exercise helps with better understanding of how encoding works in same format with different types. Meaning what it is not only for rust purpose, but either parsing, encoding, decoding and etc. This Wire type not as complicated as `LEN` type, and after, don't makes any different in terms of complexity and time, but even more - easier.  _I believe_ this simplification didn't resolve the problem of that people can had with this exercise, but change how we seeing this in terms of complexity of systems.

> To the question of adding tests: because this exercise is typically solved in the playground, we're a little limited as to what will run. With both a `main()` and `mod tests`, the playground chooses to run `main` and not the tests. 

In playground, we can choose options between build the code and run it or test it. It would be easier and faster with testing for student while writing code. I think it makes sense.

I change descriptions for I32, draw attention to working only with float type and instruct to test in playground website with corresponding option.
 
image

@djmitche, what do you think?

djmitche (commented):
> Meaning what it is not only for rust purpose, but either parsing, encoding, decoding and etc

If I'm understanding this sentence correctly, you're suggesting that this exercise is a way for students to learn about protobuf, as well as Rust. I disagree -- in fact, the opposite is true: students should spend most of their time on the exercise focused on Rust (and in particular on passing slices around) and minimal time understanding the protofbuf encoding.

This line in particular perplexes _me_, someone who is familiar with both Rust and protofbuf:
```
    I32(f32),
```
_I_ can read the rest of the code, and the encoding description, and understand that in this context the I32 wire type is only used to represent floats, so this is a kind of shortcut. But that will be baffling to students and generate questions that distract from the intended topic of handling slices. The line could be changed to `I32([u8;4])` and `as_float`, `as_fixed32`, `as_sfixed32` methods added, to at least clarify that this particular wire type can carry any of a number of proto types -- but then we're adding quite a bit of additional code for students to understand, and would also need to add corresponding fields so that those new types were relevant to the solution. And it would still lead to questions about why an integer type doesn't just use the VARINT wire type.

For context, a common student feedback for the course is that the exercises are too difficult to accomplish in the given time. In general, then, exercise revisions should simplify, and especially remove complexity that distracts from the topic of the course segment. Adding additional code related to `I32` does the opposite, as does exposing additional complexity -- however interesting -- related to protobuf.

As such, I remain convinced we should drop the `I32` type completely, as has been done for I64.

> In playground, we can choose options between build the code and run it or test it. It would be easier and faster with testing for student while writing code. I think it makes sense.

What I meant by "playground" was the embedded playground in the course materials:
image

That does not offer an option to run tests. Also, this exercise is on the afternoon of Day 3, and testing is introduced on Day 4, so it's a little early to be pressing students into understanding how tests work (although I admit they're pretty simple).

I think having the tests in `exercise.rs` is fine, as a tool for debugging the course content, but let's omit them from `exercise.md` -- the single example in `main` is sufficient to verify a student's solution.

saharulik:
Again, thanks for reply @djmitche!
> > Meaning what it is not only for rust purpose, but either parsing, encoding, decoding and etc
> 
> If I'm understanding this sentence correctly, you're suggesting that this exercise is a way for students to learn about protobuf, as well as Rust. I disagree -- in fact, the opposite is true: students should spend most of their time on the exercise focused on Rust (and in particular on passing slices around) and minimal time understanding the protofbuf encoding.

I agree with you. I bit misleading you, I was trying to say that exercises is the good example of working under real problems in Rust. Like previous exercises was about math, algorithms, events, data system. This exercise teach how to implement parser for reading data throw stream. I think so. It's still get good way to learn more about Rust in real world.    

> _I_ can read the rest of the code, and the encoding description, and understand that in this context the I32 wire type is only used to represent floats, so this is a kind of shortcut. But that will be baffling to students and generate questions that distract from the intended topic of handling slices. 

> And it would still lead to questions about why an integer type doesn't just use the VARINT wire type.

First of all, questions are good. Questions give people a chance to think about a problem and learn something new. Those who are truly interested will always find answers, even though it is not necessary. This is neither good nor bad.

> The line could be changed to `I32([u8;4])` and `as_float`, `as_fixed32`, `as_sfixed32` methods added, to at least clarify that this particular wire type can carry any of a number of proto types -- but then we're adding quite a bit of additional code for students to understand, and would also need to add corresponding fields so that those new types were relevant to the solution.

Changing type to I32([u8;4]) just add more complexity. Again refer to your words students should be closer to learning Rust and not taking much time on different plane. I would suggest to warn about implementation only for float type. 
**Like this:** 
https://github.com/google/comprehensive-rust/blob/650971de46b6ac68ef8f1f90a5b430748df60b02/src/lifetimes/exercise.rs#L24-L28

> For context, a common student feedback for the course is that the exercises are too difficult to accomplish in the given time. In general, then, exercise revisions should simplify, and especially remove complexity that distracts from the topic of the course segment. Adding additional code related to `I32` does the opposite, as does exposing additional complexity -- however interesting -- related to protobuf.
> 
> As such, I remain convinced we should drop the `I32` type completely, as has been done for I64.

I would suggest to give student opportunity to works with this type, but not necessarily. Like this:
https://github.com/google/comprehensive-rust/blob/0abf5168b2146008bba3a893b82d6783587e132b/src/lifetimes/exercise.md?plain=1#L45-L50

With this firstly where not need to rewrite code. Students can choose between working on that and learn more about decoding on Rust or continue learning fundamentals of Rust.  

> 
> > In playground, we can choose options between build the code and run it or test it. It would be easier and faster with testing for student while writing code. I think it makes sense.
> 
> What I meant by "playground" was the embedded playground in the course materials: image
> 
> That does not offer an option to run tests. Also, this exercise is on the afternoon of Day 3, and testing is introduced on Day 4, so it's a little early to be pressing students into understanding how tests work (although I admit they're pretty simple).
> 
> I think having the tests in `exercise.rs` is fine, as a tool for debugging the course content, but let's omit them from `exercise.md` -- the single example in `main` is sufficient to verify a student's solution.

In previous exercise we already used tests. The main reason why I decide to add tests because firstly it's make easier to debug code and secondary in previous exercise, at least 6, we had tests. It's not something new for students.
I would suggest to give opportunity to use this tests with additional instruction. Like this:

https://github.com/google/comprehensive-rust/blob/0abf5168b2146008bba3a893b82d6783587e132b/src/lifetimes/exercise.md?plain=1#L49-L52 

Your last word, and I rewrite as you wish 👍

djmitche:
I like your proposal to suggest that students implement I32 as a "more to explore" exercise. Especially at Google, students are likely familiar with using protobuf, so may be curious about how it is implemented and want to pursue this.

The issue of using tests vs. `main` in exercises is #1581. I expect the decision there will be to use `main` until the testing section, then switch to tests. But, you're right that it makes it easier to debug the implementation, so I think it's fine to keep them -- just don't show them in `exercise.md`.

saharulik:
@djmitche, Thanks for review! 🤝

saharulik (commented):
Need to review @djmitche

src/lifetimes/exercise.md
```diff
@@ -35,13 +36,21 @@ implemented as a "tag" followed by the value. The tag contains a field number
 how the payload should be determined from the byte stream.
 
 Integers, including the tag, are represented with a variable-length encoding
-called VARINT. Luckily, `parse_varint` is defined for you below. The given code
-also defines callbacks to handle `Person` and `PhoneNumber` fields, and to parse
-a message into a series of calls to those callbacks.
+called VARINT. Luckily, `parse_varint` is defined for you below. The encoding of
+I32 define precisely 4 bytes in little-endian order for types `fixed32`,
```
mgeisler, line 40:
```suggestion
`I32` define precisely 4 bytes in little-endian order for types `fixed32`,
```

mgeisler: Feel free to add it to `VARINT` above as well, though I fear this will require calling `dprint fmt` to fix the formatting :smile:

src/lifetimes/exercise.rs
```diff
@@ -33,7 +34,7 @@ enum FieldValue<'a> {
     Varint(u64),
     //I64(i64),  -- not needed for this exercise
     Len(&'a [u8]),
-    I32(i32),
+    I32(f32),
```
mgeisler, line 37: I don't think I understand why the PR changes this? From reading the discussion, I thought the PR was about giving students a chance to implement something more?

mgeisler:
> I like your proposal to suggest that students implement I32 as a "more to explore" exercise. Especially at Google, students are likely familiar with using protobuf, so may be curious about how it is implemented and want to pursue this.

Yes, I agree: it's super cool to have extra bonus material like this. @djmitche is also right that the exercises are too big already (same goes for some of the slides). We are unfortunately limited in time and we have to make some hard cuts here. @saharulik, I'm sure you can relate: everybody would like to add a minute here and there, but with 300+ slides, this can get out of hand :smile: Your addition is great, actually, and I'm sure some people will love tinkering with this extra step... most people just don't have time to wrap their heads around this in 20-25 minutes during class.

So I love to see some extra material in the speaker notes. The default exercise should be laser focused on letting people try out a new syntactical mechanism, such as pattern matching.

saharulik:
@djmitche, @mgeisler Thanks for reviews!

- Tests right now in Solution
- Hide PartialEq in Solution to avoid misconfusing https://github.com/google/comprehensive-rust/blob/ec508714a1b8c0737e395be6e05d04bccc0c58b8/src/lifetimes/exercise.rs#L165-L168
- Add descriptions for I32 and I64
- Comment I32, remove method and `match` arm

djmitche (approved):
I'm happy with this! Thanks for going through all those rounds of review :)

mgeisler:
> I'm happy with this! Thanks for going through all those rounds of review :)

Yeah, thanks a lot! There are many considerations in the course material, so thank you for working with us.

Thanks @djmitche for helping land this.

PR #2304, opened 2024-08-22
concurrency/sync-exercises translated to Fa

DannyRavi:
concurrency/sync-exercises translated to Farsi


javad-jafari (approved):
No problem

PR #1881, opened 2024-03-06
Updated static and const

mani-chand:
Updated the content for space issue .
#1464 issue.

PR #2305, opened 2024-08-22
Move const before static

tomfitzhenry:
Motivation:
* const usecases are likely more familiar than static usecases
* currently the static slide refers `const`, before it's introduced

Background: static and const were previously on the same slide, with const being introduced first (
https://github.com/google/comprehensive-rust/pull/1881 ) but when they were split, static was moved first.


fw-immunant (approved):
I agree, this makes more sense than the other order. Thanks!

PR #2317, opened 2024-08-26
Update approaches.md

JasonXJ:
I might be missing something, but RAII is generally associated with programming languages with manual memory management (prominently with C++, where it originated), so I think it is not very appropriate to have it mentioned here.


djmitche:
Hm, I'd tend to agree that I don't think of RAII as "full safety" memory management. @mgeisler what do you think?

mgeisler (approved):
Yeah, that is a fair point!

mgeisler:
Thanks both! I agree, RAII makes me think of C++ as well and that's probably not a good candidate for this list.

PR #2324, opened 2024-08-27
fa: bare-metal part 1/3

DannyRavi:
fa: bare-metal part 1/3


javad-jafari (approved):
Nice translation, keep it on

PR #2325, opened 2024-08-27
Misc. Android Improvements

randomPoison:
- Add missing package declaration on the AIDL interface slide.
- Make the parcelable example more self-contained. The existing code was referencing variables that weren't declared in the example code.
- Add a speaker note to the logging slide about explaining that the logger implementation is only needed in binaries, and that just the log facade is needed in libraries.


mgeisler (approved):
Thanks for improving this, I love the speaker note!

Please refactor the package line, if possible, before merging.

mgeisler:
I think we should merge this, possibly with the refactor I suggested above?

PR #2326, opened 2024-08-28
Update lifetimes/exercise.rs

JasonXJ:
Rename `as_string()` to `as_str()`. We are returning a `&str` not `String`. The latter name is also more idomatic and widely used.


mgeisler (approved):
Nice, this seems great to me!

PR #2331, opened 2024-08-30
Add "associated type" to glossary

vohoanglong0107:
# Why
For example, while reading [this line](https://github.com/google/comprehensive-rust/blob/3269cb9e4c2a2f1728851b60d1e516a3efca9876/src/std-traits/operators.md?plain=1#L39), a user may want to quickly look up what an "associated type" is.

This also helps with translations. In Vietnamese, there isn't a good translation for this term, so keeping the original term and providing a description of it in the glossary would be very helpful.


mgeisler (approved):
Thanks, looks good!

If you know the right place to link to, the  I would be very happy to get a follow-up PR with a link. That would make the glossary even more useful.

vohoanglong0107:
I see. Is [this](https://github.com/google/comprehensive-rust/blob/3269cb9e4c2a2f1728851b60d1e516a3efca9876/src/std-traits/operators.md?plain=1#L39) a good place to start? Or did you mean something else?

mgeisler:
> I see. Is [this](https://github.com/google/comprehensive-rust/blob/3269cb9e4c2a2f1728851b60d1e516a3efca9876/src/std-traits/operators.md?plain=1#L39) a good place to start? Or did you mean something else?

Yes, that would be a nice slide to link to!

PR #2332, opened 2024-08-30
error-handling: split thiserror into its own slide

### LINKED ISSUE #2027: Error-handling crates slide should be split
We used to (in v1) separately show using `thiserror` (to remove the tedium of manually writing boilerplate `From` and `Display` implementations for custom error enums) and `anyhow` (to flatten error types and allow adding context to errors). Now we show them together, and it makes it harder to understand the value proposition of `thiserror`. In particular, the example does not demonstrate the `#[from]` attribute anymore and loses the ability to dispatch on error enums because it uses the erased `anyhow::Result` type.

We should split this slide back into two slides again so that we can separately address these separate concerns. `thiserror` provides convenience without changing semantics, while `anyhow` changes error semantics in a way that is mostly suitable for applications but not libraries.

fw-immunant:
`thiserror` is best understood as a way to eliminate boilerplate on the patterns we've already seen, and then we can show it in conjunction with `anyhow` subsequently.

Fixes #2027.


src/error-handling/thiserror-and-anyhow.md
```diff
@@ -2,16 +2,14 @@
 minutes: 5
 ---
 
-# `thiserror` and `anyhow`
+# `anyhow`
 
-The [`thiserror`](https://docs.rs/thiserror/) and
-[`anyhow`](https://docs.rs/anyhow/) crates are widely used to simplify error
-handling.
+The [`anyhow`](https://docs.rs/anyhow/) crates provides a rich error type with
```
mgeisler, line 7:
```suggestion
The [`anyhow`](https://docs.rs/anyhow/) crate provides a rich error type with
```

src/SUMMARY.md
```diff
@@ -197,7 +197,8 @@
   - [Try Operator](error-handling/try.md)
   - [Try Conversions](error-handling/try-conversions.md)
   - [`Error` Trait](error-handling/error.md)
-  - [`thiserror` and `anyhow`](error-handling/thiserror-and-anyhow.md)
+  - [`thiserror`](error-handling/thiserror-and-anyhow.md)
```
mgeisler, line 200:
```suggestion
  - [`thiserror`](error-handling/thiserror.md)
```

mgeisler (commented):
I'm very happy to have the page split! I think you just need to rename the old combined page and setup a redirect (in `book.toml`).

src/error-handling/thiserror-and-anyhow.md
```diff
@@ -46,23 +44,16 @@ fn main() {
...
 - `anyhow::Error` is essentially a wrapper around `Box<dyn Error>`. As such it's
   again generally not a good choice for the public API of a library, but is
   widely used in applications.
 - `anyhow::Result<V>` is a type alias for `Result<V, anyhow::Error>`.
-- Actual error type inside of it can be extracted for examination if necessary.
-- Functionality provided by `anyhow::Result<T>` may be familiar to Go
-  developers, as it provides similar usage patterns and ergonomics to
-  `(T, error)` from Go.
+- Actual error type inside of it can be extracted for examination if necessary,
+  much like `std::any::Any`.
```
randomPoison, line 52: We never explain `Any` to students do we? Probably don't need to mention in the speaker notes, since it's an unnecessary tangent.

randomPoison (approved):
Excellent, I've been meaning to do this forever. Changes look good, modulo @mgeisler's outstanding requets.

src/error-handling/thiserror-and-anyhow.md
fw-immunant: I think it's worthwhile to mention because many students come from backgrounds in languages where upcasting/downcasting is more common. The fact that we can downcast if we happen to know the concrete underlying type here means we're not simply out of luck compared to using an error enum if we want to do something with one specific possible error, which could inform choices of which style to use.

djmitche: We have the option of `# More to Explore` in the speaker notes for things we don't intend to talk about but might be of interest. Maybe that helps here?

randomPoison: Yeah moving it to More to Explore seems reasonable to me.

PR #2333, opened 2024-08-30
clarify Fn traits slide

### LINKED ISSUE #2251: Fn traits slide is more confusing than necessary
The example code demonstrates several things at once:
- the applicable Fn* trait being dependent on how a closure interacts with the environment
- a `Fn` closure that closes over nothing (which is trivially `Copy`) being `Copy`
- the sub/supertrait relation that allows all of these to be passed to `apply_with_log` (which should probably be called `apply_and_log` instead)

Because the first closure (`add_3`) does not use the environment at all (even immutably), it doesn't clearly show the symmetry of environment use and the corresponding Fn* trait method receiver. Currently `add_3` is passed by value (not shared reference) to `apply_with_log`, which works because it is Copy, but does not perfectly correspond to the `&self` receiver of `Fn::call`. I think we should have all three closures access the environment (immutably, mutably, and by move) so as to make it as clear as possible what's happening here. We might also separately call out that closures capturing nothing, as well as functions, implement the `Fn` trait, and that closures implement `Copy` and `Clone` whenever possible, but these are all already mentioned in the speaker notes.

fw-immunant:
Improve naming and always capture something in our closures; explain other details/variations in speaker notes.

Fixes #2251.

### DIFF
--- a/src/std-traits/closures.md
+++ b/src/std-traits/closures.md
@@ -10,39 +10,41 @@ implement special [`Fn`](https://doc.rust-lang.org/std/ops/trait.Fn.html),
 [`FnOnce`](https://doc.rust-lang.org/std/ops/trait.FnOnce.html) traits:
 
 ```rust,editable
-fn apply_with_log(func: impl FnOnce(i32) -> i32, input: i32) -> i32 {
-    println!("Calling function on {input}");
-    func(input)
+fn apply_and_log(func: impl FnOnce(i32) -> i32, func_name: &str, input: i32) {
+    println!("Calling {func_name}({input}): {}", func(input))
 }
 
 fn main() {
-    let add_3 = |x| x + 3;
-    println!("add_3: {}", apply_with_log(add_3, 10));
-    println!("add_3: {}", apply_with_log(add_3, 20));
+    let n = 3;
+    let add_3 = |x| x + n;
+    apply_and_log(&add_3, "add_3", 10);
+    apply_and_log(&add_3, "add_3", 20);
 
     let mut v = Vec::new();
     let mut accumulate = |x: i32| {
         v.push(x);
         v.iter().sum::<i32>()
     };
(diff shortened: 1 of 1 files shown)


src/std-traits/closures.md
```diff
@@ -10,39 +10,41 @@ implement special [`Fn`](https://doc.rust-lang.org/std/ops/trait.Fn.html),
...
 
-An `Fn` (e.g. `add_3`) neither consumes nor mutates captured values, or perhaps
-captures nothing at all. It can be called multiple times concurrently.
+An `Fn` (e.g. `add_3`) neither consumes nor mutates captured values. It can be
+called needing only a shared reference to the closure, which means the closure
+can be executed repeatedly and even concurrently.
 
 An `FnMut` (e.g. `accumulate`) might mutate captured values. You can call it
-multiple times, but not concurrently.
+multiple times, but not concurrently (because doing so would require having
```
djmitche, line 43: for parallelism, mention that it is called with an exclusive reference to the closure

src/std-traits/closures.md
```diff
@@ -52,14 +54,16 @@ When you define a function that takes a closure, you should take `FnOnce` if you
...
 
 The compiler also infers `Copy` (e.g. for `add_3`) and `Clone` (e.g.
-`multiply_sum`), depending on what the closure captures.
+`multiply_sum`), depending on what the closure captures. Function pointers
+(references to `fn` items) implement `Copy` and `Fn`.
 
-By default, closures will capture by reference if they can. The `move` keyword
-makes them capture by value.
+By default, closures will capture each variable from an outer scope by the least
+demanding form of access they can. The `move` keyword forces capture by value.
```
djmitche, line 66: "demanding" might be a confusing term here -- perhaps a parenthetical to enumerate the options would help, especially in translations

mgeisler (approved):
Thanks, looks great!

PR #2349, opened 2024-09-05
fa: review til tuples-and-arrays/destructuring

javad-jafari:


DannyRavi (approved):
Please do more translations and then pull requested.

PR #2361, opened 2024-09-10
fa: chromium part 5/5 complete

DannyRavi:
fa: chromium part 5/5 complete


javad-jafari (approved):

javad-jafari (commented):
Ci step`s failed please check and fix it

mgeisler:
> Ci step`s failed please check and fix it

It is upset because it detected changes to both `msgid` and `msgstr` fields.

```
gh pr diff 2361 | bat -l patch
```

shows this:

```patch
diff --git a/po/fa.po b/po/fa.po
index 840d79599ed..1b21db68a21 100644
--- a/po/fa.po
+++ b/po/fa.po
@@ -16215,6 +16215,456 @@ msgstr ""
 "همچنین ممکن است لازم باشد`#include \"third_party/rust/cxx/v1/crate/include/"
 "cxx.h\"` را وارد کنید."
 
+msgid "Pass a C++ string from C++ into Rust."
+msgstr "یک رشته++C را از ++C به Rust منتقل کنید."
```

This means that the translation is both refreshed (the `msgid` fields change) and updated (the `msgstr` fields change).

It's not the end of the world, but it means that the diff is likely larger than it could be if the two kinds of changes were made in different PRs.

I suggest we override it for now and then things should sort themselves out when you follow [Refreshing an Existing Translation](https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md#refreshing-an-existing-translation) next.

PR #2363, opened 2024-09-11
fa: refreshed Translation For Farsi 

DannyRavi:
fa: refreshed Translation For Farsi 
I follow [Refreshing an Existing Translation](https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md#refreshing-an-existing-translation)


DannyRavi:
Hi @mgeisler I follow [Refreshing an Existing Translation](https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md#refreshing-an-existing-translation) and then I used to `dprint fmt ./po/fa.po` And there is no problem, but there is an error in the `Test / format` ‍.
What solution do you suggest?

DannyRavi:
> fa: refreshed Translation For Farsi

The problem is solved, and I think the reason for this is the use of `GitHub CodeSpaces`, which probably Git could not detect file changes quickly. Finally, when I did this `dprint fmt ./po/fa.po` process on my machine, the above problem was solved.

javad-jafari (approved):

PR #2364, opened 2024-09-12
fa: reform: generics + borrowing + lifetimes

DannyRavi:
fa: reform -> generics + borrowing + lifetimes


javad-jafari (approved):
There's no problem, Thanks Danny

PR #2369, opened 2024-09-17
Remove part about `Error` not being `no_std`

mgeisler:
This reverts #1005 now that Rust 1.81 has stabilized the `Error` trait.


randomPoison (approved):
Looks good. I think it's best to refer to `std::error::Error`, I think it's simpler to just tell students it's part of the standard library without going into the nuances of `core` vs `std`.

PR #2370, opened 2024-09-17
Miscellaneous minor improvements

randomPoison:
Various little changes that I've wanted when going through classes. Let me know if any of these seem problematic, I can pretty easily remove any commits that would be better to discuss separately.

### DIFF
--- a/src/control-flow-basics/break-continue/labels.md
+++ b/src/control-flow-basics/break-continue/labels.md
@@ -19,3 +19,15 @@ fn main() {
     print!("elements searched: {elements_searched}");
 }
 ```
+
+<details>
+
+- Labeled break also works on arbitrary blocks, e.g.
+  ```rust
+  'label: {
+      break 'label;
+      println!("This line gets skipped");
+  }
+  ```
+
+</details>
--- a/src/generics/generic-functions.md
+++ b/src/generics/generic-functions.md
@@ -19,7 +19,7 @@ fn pick<T>(n: i32, even: T, odd: T) -> T {
 
 fn main() {
     println!("picked a number: {:?}", pick(97, 222, 333));
-    println!("picked a tuple: {:?}", pick(28, ("dog", 1), ("cat", 2)));
+    println!("picked a tuple: {:?}", pick(28, "dog", "cat"));
 }
 ```
 
--- a/src/generics/generic-functions.md
+++ b/src/generics/generic-functions.md
@@ -19,13 +19,24 @@ fn pick<T>(n: i32, even: T, odd: T) -> T {
 
 fn main() {
     println!("picked a number: {:?}", pick(97, 222, 333));
-    println!("picked a tuple: {:?}", pick(28, "dog", "cat"));
+    println!("picked a string: {:?}", pick(28, "dog", "cat"));
 }
 ```
 
 <details>
 
 - Rust infers a type for T based on the types of the arguments and return value.
+
+- In this example we only use the primitive types `i32` and `&str` for `T`, but
+  we can use any type here, including user-defined types:
+
+  ```rust
+  struct Foo {
+      val: u8,
+  }
+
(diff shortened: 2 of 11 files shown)


src/generics/generic-functions.md
```diff
@@ -19,7 +19,7 @@ fn pick<T>(n: i32, even: T, odd: T) -> T {
 
 fn main() {
     println!("picked a number: {:?}", pick(97, 222, 333));
-    println!("picked a tuple: {:?}", pick(28, ("dog", 1), ("cat", 2)));
+    println!("picked a tuple: {:?}", pick(28, "dog", "cat"));
```
djmitche, line 22: These aren't tuples anymore. I guess that would be `"picked a string"`. But the idea here was to show that T doesn't just have to be a named type like `String` or `i32`, but can be any type.

src/methods-and-traits/exercise.rs
```diff
@@ -14,25 +14,18 @@
 
 // ANCHOR: solution
 // ANCHOR: setup
-use std::fmt::Display;
-
 pub trait Logger {
     /// Log a message at the given verbosity level.
-    fn log(&self, verbosity: u8, message: impl Display);
```
djmitche, line 21: Ah, `impl Trait` hasn't been covered yet -- 👍🏼

djmitche (approved):

src/generics/generic-functions.md
fw-immunant: I agree, I would add this as an additional example so the students see two "simple" types as well as a compound type.

src/pattern-matching/exercise.rs
```diff
@@ -44,11 +44,11 @@ fn eval(e: Expression) -> Result<i64, String> {
         Expression::Op { op, left, right } => {
             let left = match eval(*left) {
                 Ok(v) => v,
-                e @ Err(_) => return e,
+                Err(e) => return Err(e),
             };
             let right = match eval(*right) {
                 Ok(v) => v,
-                e @ Err(_) => return e,
+                Err(e) => return Err(e),
```
fw-immunant, line 51: I agree it's good to avoid introducing new syntax in a solution; this always surprises students. But do we have any other place where this syntax is introduced? I think it is worth mentioning.

src/control-flow-basics/break-continue/labels.md
```diff
@@ -19,3 +19,15 @@ fn main() {
...
+
+- Labeled break also works on arbitrary blocks, e.g.
+  ```rust
+  'label: {
+      break 'label;
+      println!("This line gets skipped");
+  }
+  ```
+
+</details>
```
fw-immunant, line 33: I did not know you could do this!

mgeisler (approved):

src/pattern-matching/exercise.rs
randomPoison: I don't think it's worth mentioning, honestly. It's a rarely-used syntax and something students can pretty easily look up if they run into it in the wild. At most maybe we could have a speaker note on the patterns page, but I wouldn't put it in the slides.

src/generics/generic-functions.md
randomPoison: Hrm, I get what you're saying about demonstrating that generic types can also be compound types, though I don't think it's all that important to emphasize. I think students can generalize from "integers and strings" to "any type". But saying it explicitly to students might be good, so would it be good enough to add a speaker note about it?

src/pattern-matching/exercise.rs
fw-immunant: A speaker note seems like the right place for something minor like this IMO.

randomPoison: Added.

fw-immunant (approved):

PR #2375, opened 2024-09-21
uk: Refresh translations

reta:
uk: Refresh translations

```
mdbook build
msgmerge --update po/uk.po book/xgettext/messages.pot
```

### DIFF
--- a/po/uk.po
+++ b/po/uk.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: Comprehensive Rust 🦀\n"
-"POT-Creation-Date: 2024-06-27T18:39:25-04:00\n"
+"POT-Creation-Date: 2024-09-21T12:49:47-04:00\n"
 "PO-Revision-Date: 2023-10-18 20:05+0300\n"
 "Last-Translator: Oleksandr Leoshko <sleo170504@gmail.com>\n"
 "Language-Team: \n"
@@ -197,9 +197,9 @@ msgstr "Спільні посилання"
 msgid "Exclusive References"
 msgstr "Ексклюзивні посилання"
 
-#: src/SUMMARY.md
-msgid "Slices: `&[T]`"
-msgstr "Зрізи: `&[T]`"
+#: src/SUMMARY.md src/references/slices.md:5
+msgid "Slices"
+msgstr "Зрізи"
 
 #: src/SUMMARY.md src/references/strings.md:9
 msgid "Strings"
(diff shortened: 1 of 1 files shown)


dyeroshenko (approved):

reta:
Thanks @dyeroshenko , I think the msgid check is expected to fail since it is merge with recent content. Thanks again!

mgeisler:
Yeah, thanks so much for these updates!

> I think the msgid check is expected to fail since it is merge with recent content. Thanks again!

Actually, the check is based on the POT-Creation-Date header: if this field is _unchanged_, then the `msgid` fields must all be unchanged too. The goal was to avoid having people edit the auto-generated `msgid` fields (since doing so would have no effect). The test is green now, so I think this was done correctly.

Running `dprint fmt` on my system fixes the formatting complains from the other CI check. It also produces changes in `pl.po` and `de.po`, which is weird... that is #2173, which I still don't know what to do about.

One small thing you can do at some point would be update version of mdbook-i18n-helpers used. I think you're using an old version since I see line numbers for the files outside of `SUMMARY.md`:

```patch
-#: src/SUMMARY.md
-msgid "Slices: `&[T]`"
-msgstr "Зрізи: `&[T]`"
+#: src/SUMMARY.md src/references/slices.md:5
+msgid "Slices"
+msgstr "Зрізи"
```

The `granularity` setting in `book.toml` ought to prevent this:

https://github.com/google/comprehensive-rust/blob/2713ea34757990f48c521c50ab471a836683c91f/book.toml#L25-L28

If not, then something is broken somewhere :smile:

reta:
> Actually, the check is based on the POT-Creation-Date header: if this field is _unchanged_, then the `msgid` fields must all be unchanged too.

Thanks @mgeisler this field is actually changed:

```
- "POT-Creation-Date: 2024-06-27T18:39:25-04:00\n"
+ "POT-Creation-Date: 2024-09-21T12:49:47-04:00\n"
```

mgeisler:
> > Actually, the check is based on the POT-Creation-Date header: if this field is _unchanged_, then the `msgid` fields must all be unchanged too.
> 
> Thanks @mgeisler this field is actually changed:
> 
> ```
> - "POT-Creation-Date: 2024-06-27T18:39:25-04:00\n"
> + "POT-Creation-Date: 2024-09-21T12:49:47-04:00\n"
> ```

Yeah, that must be why the "Prevent unintended msgid changes / check-msgid-changes" check passes. The "Test / format" should be gone now that @djmitche created #2381 — so this is ready to merge.

PR #2376, opened 2024-09-22
docs: update the spanish translation

Phosphorus-M:
Hi, recently (this weekend) I was working in made a huge update of the Spanish translation of this book.

This PR is related to the issue #1463 however it's been a while since the issue was created, and I noticed there were several contributions. It was quite well translated, with only a few texts to be fixed and some new texts to be added.


mgeisler:
Hi @Phosphorus-M, I'm sorry about the lack of reviews... can you try merging in the latest translations and update the PR?

I'm starting to use Gemini to review this kind of PRs, so we should be able to get a faster throughput going forward!

PR #2380, opened 2024-09-26
Fix thiserror slide

### LINKED ISSUE #2379: `thiserror` slide has compile errors
I think it just needs an `std::io` import.

djmitche:
Fixes #2379. This has `compile_fail` because `thiserror` isn't available from within `mdbook test`.


fw-immunant (approved):
Thanks!

Issue #2382, opened 2024-09-29
[Exercise][13. Methods and Traits] Better output of example solution


Alx-Lai:
In Section13.4, there's an [exercise code section](https://google.github.io/comprehensive-rust/methods-and-traits/solution.html) to write a `VerbosityFilter` of a `Logger`; however, the logger uses `eprintln!` macro to output the log, which output to `stderr`. Therefore, the run result in the website shows only `No output`.

Should we consider logging out messages from both `stdout` and `stderr`? 

Comparison
In Rust Playground:
```
=====Standard Error=====
   Compiling playground v0.0.1 (/playground)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.54s
     Running `target/debug/playground`
verbosity=2: Uhoh
=====Standard Output====
```

In Comprehensive Rust:
```
No output
```

djmitche:
Probably easiest to just change eprintln to println? Do you want to make up a PR?

Alx-Lai:
Hi @djmitche, just noticed #531, probably we can revert this change after #531 is fixed.

djmitche:
Thanks for spotting that! I think the change is fine with or without that fix.

PR #2383, opened 2024-09-30
Exercise: method and traits: change output

Alx-Lai:
In this exercise, the original StderrLogger output to stderr, which user cannot see the original log but "No output" in frontend". Change `eprintln!` to `println!` to make it see-able in the frontend. Also, due to the change, rename the logger struct.

Close: #2382


Alx-Lai:
Here's the result:

image

djmitche (approved):
Thanks for renaming the class, too -- that's the kind of detail I'd miss :)

Issue #2384, opened 2024-09-30
Add an "Implementation Details" segment


djmitche:
From https://github.com/google/comprehensive-rust/issues/1820#issuecomment-1948987570:

 * Implementation Details
    * Niche Optimization
    * Dynamic array layout (including NonNull::dangling)
    * Return Value Optimization
    * ??
    * Exercise

fw-immunant:
We might also mention how empty containers don't idiomatically allocate.

Issue #1820, opened 2024-02-15
Incorrect niche optimization explanation for the linked list in the box section


errge:
https://google.github.io/comprehensive-rust/smart-pointers/box.html#:~:text=A%20Box%20cannot%20be%20empty%2C%20so%20the%20pointer%20is%20always%20valid%20and%20non%2Dnull.%20This%20allows%20the%20compiler%20to%20optimize%20the%20memory%20layout

I think this is not correct.

I wrote this debugging code:

```
#[derive(Debug)]
enum List<T> {
    /// A non-empty list: first element and the rest of the list.
    Element(T, Box<List<T>>),
    /// An empty list.
    Nil,
}

fn main() {
    let list: List<i32> = List::Element(1, Box::new(List::Element(2, Box::new(List::Nil))));
    let list2: List<i32> = List::Element(1, Box::new(List::Element(2, Box::new(List::Nil))));
    let list3: List<i32> = List::Element(1, Box::new(List::Element(2, Box::new(List::Nil))));
    println!("{list:?}");
    unsafe {
        let first: u128 = std::mem::transmute(list);
        println!("{first:#034x}");

        let List::<i32>::Element(_, cdrbox) = list2 else {
            panic!("impossible");
        };
        let second: u128 = std::mem::transmute(*cdrbox);
        println!("{second:#034x}");

        let List::<i32>::Element(_, cdrbox3) = list3 else {
            panic!("impossible");
        };
        let List::<i32>::Element(_, cddrbox) = *cdrbox3 else {
            panic!("impossible");
        };
        let third: u128 = std::mem::transmute(*cddrbox);
        println!("{third:#034x}");
    }
}
```

What I'm trying to do here, is to get the representation out of Rust in all 3 steps of the linked list, and this is the output I have:

```
errge:~/rust $ ./target/release/box
Element(1, Element(2, Nil))
0x0000564fac517bc00000000100000000
0x0000564fac517be00000000200000000
0x0000564fac517be00000000200000001
errge:~/rust $ qemu-aarch64 ./target/aarch64-unknown-linux-musl/release/box
Element(1, Element(2, Nil))
0x00000055008068500000000100000000
0x00000055008068700000000200000000
0x00000055008068700000000200000001
errge:~/rust $ ./target/x86_64-unknown-linux-musl/release/box
Element(1, Element(2, Nil))
0x00007fa1c7b688500000000100000000
0x00007fa1c7b688700000000200000000
[...]

PR #2392, opened 2024-10-01
Add implementation-details segment

djmitche:
This is a work-in-progress. I will probably need some help with some of these!

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -162,6 +162,10 @@
   - [Struct Lifetimes](lifetimes/struct-lifetimes.md)
   - [Exercise: Protobuf Parsing](lifetimes/exercise.md)
     - [Solution](lifetimes/solution.md)
+- [Implementation Details](implementation-details.md)
+  - [Niche Optimization](implementation-details/niche-optimization.md)
+  - [Exercise: TBD](implementation-details/exercise.md)
+    - [Solution](implementation-details/solution.md)
 
 ---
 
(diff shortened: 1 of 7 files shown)


djmitche:
I think "dynamic array layout" refers to the differences in how C++ and Rust handle empty slices?

The other item in #2384 is "Return Value Optimization". What else should we include here?

djmitche:
@fw-immunant @errge @anforowicz -- I've implemented a derivative of the linked list that started #1820 as an exercise, but I'm not sure I like it.

For one thing, I was surprised to find that even with T=u64, the `List` enum still takes 16 bytes (8 for T, 8 for the box, with the NULL value of the box used to signify List::Nil). That's a little disappointing.

Also, this exercise ends up getting into the question of whether the head node is stored on the stack. With the `List` enum (top of the exercise), it is, while with the `List` struct ("solution"), it is not. The result is that the two arguably use the same space, and differ only in where they allocate a little bit of it.

I'm curious for your thoughts, or if you have any better ideas for an exercise related to the niche optimization. I have also added a slide on empty containers and `std::ptr::NonNull::dangle()`, but this segment comes before the unsafe segment so I don't think it's OK to ask students to wor with `dangle()`.

I'd also be open to other slides to add to this segment, especially if they lead naturally to an exercise.

mgeisler:
Sorry about that, Dustin! I would in principle like to see more advanced topics covered, but I also don't have bandwidth right now to help out here. Perhaps someone else will be able to pick it up in the future.

PR #2395, opened 2024-10-04
 fa: try to complete Farsi part 3

DannyRavi:
fa: try to complete Farsi part 3


DannyRavi:
Hi @mgeisler there is error when use `dprint fmt ./po/fa.po` that return something like this:

`No files found to format with the specified plugins at /home/Project/comprehensive-rust. You may want to try using `dprint output-file-paths` to see which files it's finding or run with `--allow-no-files`.`

I solved this with an older version of  `dprint.json`.

PR #2396, opened 2024-10-05
Generate PR suggestions when formatting is wrong

mgeisler:
This should make it much easier to do drive-by changes in the GitHub editor: the comment should contain a diff that can be committed directly from the online editor.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,14 +21,21 @@ jobs:
         run: |
           sudo apt update
           sudo apt install gettext yapf3
+          curl -fsSL https://dprint.dev/install.sh | sh
+          echo "/home/runner/.dprint/bin" >> $GITHUB_PATH
 
       - name: Install nightly rustfmt
         run: |
           rustup default nightly
           rustup component add rustfmt
 
-      - name: Check formatting
-        uses: dprint/check@v2.2
+      - name: Reformat
+        run: dprint fmt
+
+      - uses: parkerbxyz/suggest-changes@v1
+
+      - name: Check if there were any changes
+        run: git diff --exit-code
 
   typos:
     runs-on: ubuntu-latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,12 +29,13 @@ jobs:
           rustup default nightly
           rustup component add rustfmt
 
-      - name: Reformat
+      - name: Format the project
         run: dprint fmt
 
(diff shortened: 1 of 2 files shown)


mgeisler:
As noted in https://github.com/parkerbxyz/suggest-changes/issues/37, the action creates a new review every time the PR is modified. So if the formatting errors are not resolved right away, a new comment is added repeatedly.

I think we can live with that: people will hopefully quickly apply the formatting suggestion and so the comments should not be too burdensome.

I'm not 100% sure if this works then a PR comes from a fork: https://github.com/parkerbxyz/suggest-changes/issues/33

A final problem is the stickiness of the review left by the action. After all formatting problems have been resolved, the negative review will need to be dismissed by hand before the PR can be merged. The burden of this is on the PR reviewers, so this is probably acceptable?

mgeisler:
Oh no, there is another problem: the suggestions are attributed to the github-actions bot! This makes the CLA fail:

image

I think this would be the same, even if comments are used (https://github.com/parkerbxyz/suggest-changes/issues/36) instead of a review, but perhaps the email address used can be configured somehow?

djmitche (approved):

parkerbxyz:
> As noted in [parkerbxyz/suggest-changes#37](https://github.com/parkerbxyz/suggest-changes/issues/37), the action creates a new review every time the PR is modified. So if the formatting errors are not resolved right away, a new comment is added repeatedly.

This has been fixed in the latest release.

> After all formatting problems have been resolved, the negative review will need to be dismissed by hand before the PR can be merged.

I'd like to make the review action configurable so you can have it just leave comments without blocking merging. 🔜

mgeisler:
> This has been fixed in the latest release.

Thanks @parkerbxyz, that sounds great!

I will need to change the email address used for the suggestions (for the CLA check, as discussed in https://github.com/parkerbxyz/suggest-changes/issues/41). Then I think we can enable the workflow for our repository.

mgeisler:
@djmitche, we have documentation for setting up and registering a bot account here: go/github-docs/robots. I will probably go through this process to get a custom `GITHUB_TOKEN` we can use in the workflows.

mgeisler:
> @djmitche, we have documentation for setting up and registering a bot account here: go/github-docs/robots. I will probably go through this process to get a custom `GITHUB_TOKEN` we can use in the workflows.

I never go the time to do this, and I'm no longer with Google. So I'll close this for now. @gribozavr, this might be something to advertise to 20% contributors internally.

Issue #531, opened 2023-03-27
Show `stderr` from code blocks


mgeisler:
Today, mdbook doesn't show `stderr` when running Rust code. The problem is that it uses a Playground endpoint which doesn't deliver this information. However, the Playground itself does have this information via an internal endpoint. See https://github.com/rust-lang/mdBook/pull/1315 for details.

Fixing this would allow us to start using `dbg!(foo)` instead of `println!("foo: {foo:?}")` in our code samples and thus simplify a lot of code blocks.

With https://github.com/google/mdbook-i18n-helpers/pull/75 merged, many more code blocks would be automatically removed from the translation files, thus reducing churn for our translators.

osiris5498:
I'll try to work on it in the next two weeks.

mgeisler:
Thanks Emma! You might very likely have to make changes to the upstream `mdbook` — we should upstream as much as possible, both so it can benefit everybody and so we can avoid carrying patches (which makes it hard to update our version of `mdbook`).

Alx-Lai:
Hi @mgeisler, I applied rust-lang/mdBook#1315 in #2397. I think there's a room to discuss whether we should push it to upstream repo. My concern is that the original PR is put aside for a long time so we don't know if i'll be finally adopted or not; however, if we merge it first in downstream, the only possible effort comes iff it's finally adopted in upstream.

mgeisler:
With #2397 merged, we can finally start using `dbg!` throughout the course! I suggest we close this as finished now and then create a new issue to focus on this conversion.

mgeisler:
I created #2478 to track converting print statements to `dbg!`.

mgeisler:
The change in #2397 did somehow not work when deployed — I don't know why, but I saw 500 Server Errors on every request. I reverted the change for now, but I hope we can fix it soon.

mgeisler:
This was added again #2503, so this is now fixed! Thanks a lot, @Alx-Lai!

PR #2397, opened 2024-10-06
Execution: Add stderr block

Alx-Lai:
Add a stderr block.
Pros:
  - we can make use of dbg macro in our code Cons:
  - there's a limitation that the compile message also shows

Applies patches from rust-lang/mdBook#1315 since the original change was not merged by rust-lang.

Issue: #531


mgeisler:
Hi @Alx-Lai, oh wow, this patch looks great! Thanks a lot for looking into this, I would very much like to have this feature implemented.

I gave the PR a quick test run locally and this is how it looks:

https://github.com/user-attachments/assets/af54a9f2-b589-4666-b3c7-ee1ac6c5df9e

This is nearly perfect. A few comments:

- Could you match and remove the three lines that appear when things compile correctly? The lines which say:

    ```
        Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.60s
         Running `target/debug/playground`
    ```

    By removing them in the happy case, we end up with the same output as today, which helps limit the amount of vertical space used. This is important for a live class where it's very disturbing to scroll up and down :)

- Could you avoid showing the `stderr` block immediately when compilation starts? Ideally, there would be a single output with the combination of `stdout` and `stderr`, but I understand that the Playground delivers the data in two separate fields.
- Similarly, the block with the `stderr` output should be hidden/cleared when the code is recompiled.

I looked briefly at the full Playground and discovered that they use a WebSocket now to stream the output line by line! They include markers to show if this is a `stdout` or `stderr`:

image

I would be happy to merge this PR as a first step — I don't know how difficult it will be to adapt the `mdbook` code to use a WebSocket as well.

Alx-Lai:
Ok, I'll spend time looking into it. Thanks for the feedback!

Alx-Lai:
Updated the commit.
- Remove compile message with a not-that-elegant way, the compile error message would not be replaced.
- hide and clear stderr block before compiling the code.

mgeisler:
Hi @Alx-Lai, thanks for the update! I haven't forgotten about it, I'm just slow to review the change.

I'll try asking internally if someone else can help review this.

gotoextreme:
I just tested the updated version, and it works great!

[Screencast from 2024-11-20 16-12-30.webm](https://github.com/user-attachments/assets/ede52213-f361-4914-a94c-bb824eb5a149)

mgeisler:
Thanks a lot, @gotoextreme! That's super helpful :) I think we can merge this now and then we'll soon have proper testing support (#2462) to ensure that things say stable as we refactor and extend this further.

mgeisler (approved):
Thanks a lot, this is an exciting step forward! It will allow us to use `dbg!` throughout the course, which I've wanted for a long time.

mgeisler:
Hey @Alx-Lai, I saw this working when testing locally, but now that it's deployed to https://google.github.io/comprehensive-rust/, I see


```
POST https://play.rust-lang.org/execute 500 (Internal Server Error)
```

every time I try running some code. Can you check if you see the same? If so, then we'll have to revert this for now and figure out what is happening.

mgeisler:
Alright, @Alx-Lai confirmed that he sees the same error, so I reverted this in #2479. I would still love to see this feature, so I hope we can figure out why it didn't work on the published version.

mgeisler:
I think we can add this again, please see #2503.

PR #2402, opened 2024-10-14
Bump dprint plugins to their latest versions

mgeisler:


djmitche:
@mgeisler I think this just needed a reformat (and now maybe an update?)

PR #2406, opened 2024-10-15
Split large unsafe function slide

mgeisler:
The old slice was doing several things at the same time: demonstrating
both external functions as well as unsafe Rust functions.

We now treat those two topics separately. In addition, the “Calling
Unsafe Functions” heading has become its own slide with a non-crashing
example that shows what can go wrong if an argument is misunderstood
in a call to an unsafe function. The old example didn’t actually
illustrate the danger clearly: it would produce mangled UTF-8 output,
which the Playground server refuses to print.

Part of #2445.


src/SUMMARY.md
```diff
@@ -207,6 +207,9 @@
   - [Mutable Static Variables](unsafe-rust/mutable-static.md)
   - [Unions](unsafe-rust/unions.md)
   - [Unsafe Functions](unsafe-rust/unsafe-functions.md)
+    - [Unsafe External Functions](unsafe-rust/unsafe-extern-c-functions.md)
+    - [Unsafe Rust Functions](unsafe-rust/unsafe-rust-functions.md)
+    - [Calling Unsafe Functions](unsafe-rust/calling-unsafe-functions.md)
```
djmitche, line 212: For subslides, we've used a subdirectory to contain the subslides. Also, only the top-level slide has the timing information. See src/methods-and-traits/traits for an example.

src/unsafe-rust/unsafe-rust-functions.md
```diff
@@ -0,0 +1,44 @@
...
+
+You can mark your own functions as `unsafe` if they require particular
+conditions to avoid undefined behaviour.
+
+```rust,editable
+/// Swaps the values pointed to by the given pointers.
+///
+/// # Safety
+///
+/// The pointers must be valid and properly aligned.
```
djmitche, line 15: ..and must not overlap!

src/unsafe-rust/unsafe-rust-functions.md
```diff
@@ -0,0 +1,44 @@
...
+```
+
+<details>
+
+We wouldn't actually use pointers for a `swap` function --- it can be done
+safely with references.
+
+Note that unsafe code is allowed within an unsafe function without an `unsafe`
+block. We can prohibit this with `#[deny(unsafe_op_in_unsafe_fn)]`. Try adding
+it and see what happens. This will likely change in a future Rust edition.
```
djmitche, line 42: https://github.com/rust-lang/rust/issues/120535 in the 2024 edition, in fact!

src/unsafe-rust/unsafe-extern-c-functions.md
```diff
@@ -0,0 +1,33 @@
...
+    }
+}
+```
+
+<details>
+
+`abs` is unsafe because it is an external function (FFI). Calling external
+functions is usually only a problem when those functions do things with pointers
+which might violate Rust's memory model, but in general any C function might
+have undefined behaviour under any arbitrary circumstances.
```
djmitche, line 28: Is it worth mentioning here that there is no automatic checking of the Rust function signature against that of the target function? For example, in this case the C function might take `uint64_t` and this would cause UB.

src/unsafe-rust/calling-unsafe-functions.md
```diff
@@ -0,0 +1,47 @@
...
+
+<details>
+
+Key points:
+
+- The second argument to `slice::from_raw_parts` is the number of _elements_,
+  not bytes! This example demonstrate undefined behavior by reading past the end
+  of one array and into another.
+
+- The standard library contain many low-level unsafe functions. Prefer the safe
```
djmitche, line 41:
```suggestion
- The standard library contains many low-level unsafe functions. Prefer the safe
```

src/unsafe-rust/unsafe-rust-functions.md
djmitche: oh, because these are u8 it's actually ok to overlap

src/unsafe-rust/unsafe-functions/extern-c.md
```diff
@@ -0,0 +1,32 @@
+# Unsafe External Functions
+
+All functions implemented in a foreign language are considered unsafe in Rust:
```
qwandor, line 3: This is no longer true as of 1.82. We should talk about `unsafe extern` here.

src/unsafe-rust/unsafe-functions.md
```diff
@@ -1,100 +1,19 @@
...
-
-    // SAFETY: ...
-    unsafe {
-        swap(&mut a, &mut b);
-    }
-
-    println!("a = {}, b = {}", a, b);
-}
-```
+- Foreign functions in `extern "C"` blocks.
```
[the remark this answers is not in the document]
djmitche: This PR was from before Rust 1.82, so perhaps we can land it and update, unless you think that splitting the slides doesn't make sense anymore?

qwandor (approved):

djmitche (approved):

djmitche:
Let's merge it before Rust 1.83.0 :)

PR #2407, opened 2024-10-15
Reduce vertical space slightly in UART examples

mgeisler:


src/bare-metal/aps/examples/src/pl011.rs
```diff
@@ -122,10 +122,7 @@ impl Uart {
 
         // SAFETY: We know that self.registers points to the control registers
         // of a PL011 device which is appropriately mapped.
-        unsafe {
-            // Write to the TX buffer.
```
qwandor, line 126: I'd like to keep this comment, as it's not immediately obvious what this line does. How about moving it up above the safety comment?

src/bare-metal/aps/examples/src/pl011_minimal.rs
```diff
@@ -43,10 +43,7 @@ impl Uart {
 
         // SAFETY: We know that the base address points to the control
         // registers of a PL011 device which is appropriately mapped.
-        unsafe {
-            // Write to the TX buffer.
```
qwandor, line 47: Ditto.

djmitche:
@qwandor I applied the requested reverts!

qwandor (approved):

PR #2410, opened 2024-10-15
Apply `clippy::semicolon_if_nothing_returned`

mgeisler:
This is for consistency: if a block returns `()`, and the last
expression of the block also returns `()`, then the final `;` can be
left out.

However, this is a little confusing (I was asked about this in a class
today) and it is inconsistent.

See
https://rust-lang.github.io/rust-clippy/master/index.html#/semicolon_if_nothing_returned
for details.


michael-kerscher (approved):
looks good to me

PR #2413, opened 2024-10-15
Consistently inline formatting arguments

mgeisler:


michael-kerscher (commented):
I checked with this regex (which is trying to find println! calls with a comma that is not followed by brackets or a dot or a star (I give you rather the command than individual results so you can have a deeper look if you want to)
grep "println!(\".*\",[^().+*]*)" . -R --include="*.html" --include="*.rs"

The current change looks good to me, but I wonder if more instances of this can be fixed for consistency.

mgeisler:
> what we should do is to enable Clippy

This brings up a question: `dprint` can run the code blocks through `rustfmt` for us, even code blocks embedded in Markdown files. This is awesome and was a big motivation for introducing `dprint`. Can we create the same integration with `dprint` and `clippy`?

michael-kerscher (approved):
I don't want to block this - the current changes look good to me.
Using clippy with dprint and `uninlined_format_args` sounds like a good plan to catch the rest

michael-kerscher:
Thinking about this again, after writing my comment in #2025 I don't think dprint is a good choice for executing clippy.

dprint tries itself to figure out all relevant files on individually tries to check (or fix) them. clippy needs some context on the entire crate though to be able to check some things. That's why I would keep clippy out of dprint in a separate step.

PR #2415, opened 2024-10-15
Format filenames consistently

mgeisler:
The style used in the rest of the course is _italic_ filenames, so we
should be careful to not suddenly use **bold** names.

The style is of course up for debate, but we need to ensure
consistency first.


fw-immunant:
Presumably the colons shouldn't be italicized. Or does that look particularly bad?

mgeisler:
> Presumably the colons shouldn't be italicized. Or does that look particularly bad?

Oh, nice catch! You're right that the colons are not italic on our other pages! Let me fix it!

michael-kerscher (approved):
looks good to me

PR #2422, opened 2024-10-15
Call out that `self` receivers are covered later

mgeisler:
We have not formally introduced ownership at this point, so the
exercise can be confusing for students.

Alternatively, we could consider removing the `finish` method since
it’s not essential for a first encounter with methods.


djmitche (approved):
I like the callout better -- it hints to instructors that they don't need to go into detail yet, even if a student asks a question about it.

PR #2430, opened 2024-10-15
Wrap assembly comments consistently

mgeisler:
Some lines were much longer than others. They are now consistently
wrapped at 70 columns. Why so short, why not 80 or 90? Because we
sometimes need to browse these files during class, and there we don’t
have a lot of horizontal space when using large fonts.

### DIFF
--- a/src/bare-metal/aps/examples/entry.S
+++ b/src/bare-metal/aps/examples/entry.S
@@ -34,18 +34,22 @@
 .set .L_TCR_TG0_4KB, 0x0 << 14
 /* 4 KiB granule size for TTBR1_EL1. */
 .set .L_TCR_TG1_4KB, 0x2 << 30
-/* Disable translation table walk for TTBR1_EL1, generating a translation fault instead. */
+
+/*
+ * Disable translation table walk for TTBR1_EL1, generating a
+ * translation fault instead.
+ */
 .set .L_TCR_EPD1, 0x1 << 23
 /* Translation table walks for TTBR0_EL1 are inner sharable. */
 .set .L_TCR_SH_INNER, 0x3 << 12
 /*
- * Translation table walks for TTBR0_EL1 are outer write-back read-allocate write-allocate
- * cacheable.
+ * Translation table walks for TTBR0_EL1 are outer write-back
+ * read-allocate write-allocate cacheable.
  */
 .set .L_TCR_RGN_OWB, 0x1 << 10
 /*
- * Translation table walks for TTBR0_EL1 are inner write-back read-allocate write-allocate
- * cacheable.
+ * Translation table walks for TTBR0_EL1 are inner write-back
+ * read-allocate write-allocate cacheable.
  */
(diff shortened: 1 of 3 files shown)


src/bare-metal/aps/examples/exceptions.S
```diff
@@ -64,7 +65,9 @@
 	ldr x18, [sp, #8 * 18]
 	ldp x29, x30, [sp, #8 * 20]
 
-	/* Restore registers elr_el1 & spsr_el1, using x0 & x1 as scratch. */
+	/*
+     * Restore registers elr_el1 & spsr_el1, using x0 & x1 as scratch.
```
michael-kerscher, line 69: this comment seems to have an incorrect indentation

michael-kerscher: it looks like tabs vs. spaces

src/bare-metal/aps/examples/exceptions.S
```diff
@@ -64,7 +65,9 @@
...
+     */
```
mgeisler, line 70:
```suggestion
	/*
	 * Restore registers elr_el1 & spsr_el1, using x0 & x1 as scratch.
	 */
```

src/bare-metal/aps/examples/image.ld
```diff
@@ -67,7 +67,10 @@ SECTIONS
 		data_end = .;
 	} >image
 
-	/* Everything beyond this point will not be included in the binary. */
+	/*
+     * Everything beyond this point will not be included in the
+     * binary.
+     */
```
mgeisler, line 73:
```suggestion
	/*
	 * Everything beyond this point will not be included in the
	 * binary.
	 */
```

src/bare-metal/aps/examples/exceptions.S
mgeisler: Oh my, thanks for spotting this. I don't know why the comment was indented using a tab originally — I've now indented the other comments with tabs too.

src/bare-metal/aps/examples/entry.S
```diff
@@ -72,18 +78,20 @@
...
+ * these may contain boot parameters.
  */
 // ANCHOR: entry
 .section .init.entry, "ax"
 .global entry
 entry:
     /*
-     * Load and apply the memory management configuration, ready to enable MMU and
-     * caches.
+     * Load and apply the memory management configuration, ready to
```
michael-kerscher, line 93: these are spaces to indent, files below have tabs. Should we try to have one style over all files?
This could also be a separate commit

michael-kerscher (approved):
The commit looks good to me. There are differences in the indentation style (tabs vs spaces) though.

src/bare-metal/aps/examples/entry.S
mgeisler: Yeah, having a consistent style would be good! Perhaps we could teach `dprint` to do this, but it might not be worth is since we touch these files very very rarely.

PR #2432, opened 2024-10-15
Remove unnecessary null check in FFI exercise

mgeisler:
We only assign `self.dir` once and we only assign it if the pointer is
non-null. We can therefore simplify the drop implementation a little.


fw-immunant (approved):
This has bugged me for a while. Thanks for the fix.

PR #2434, opened 2024-10-15
Update unsafe trait example to zerocopy version 0.8

semihbkgr:
Zerocopy crate version 0.8 introduced changes to its API, which caused the example code to break.
https://github.com/google/zerocopy/discussions/1680
> AsBytes -> [IntoBytes](https://docs.rs/zerocopy/0.8.*/zerocopy/trait.IntoBytes.html)


src/unsafe-rust/unsafe-traits.md
```diff
@@ -8,36 +8,35 @@ Like with functions, you can mark a trait as `unsafe` if the implementation must
 guarantee particular conditions to avoid undefined behaviour.
 
 For example, the `zerocopy` crate has an unsafe trait that looks
-[something like this](https://docs.rs/zerocopy/latest/zerocopy/trait.AsBytes.html):
+[something like this](https://docs.rs/zerocopy/latest/zerocopy/trait.IntoBytes.html):
 
 ```rust,editable
-use std::mem::size_of_val;
+use std::mem;
 use std::slice;
```
mgeisler, line 15: Nice change and now we can import both with one line (vertical space is a precious resource on our slides):

```suggestion
use std::{mem, slice};
```

mgeisler:
Hi @semihbkgr, please run `dprint fmt` to fix the formatting (or apply it by hand from the output of the failing test). Then we can merge this!

mgeisler (approved):

mgeisler:
Thanks for the help, this is super useful!

src/unsafe-rust/unsafe-traits.md
```diff
@@ -8,36 +8,35 @@ Like with functions, you can mark a trait as `unsafe` if the implementation must
...
+        let len = mem::size_of_val(self);
+        let slf: *const Self = self;
+        unsafe { slice::from_raw_parts(slf.cast::<u8>(), len) }
     }
 }
 
-// SAFETY: `u32` has a defined representation and no padding.
-unsafe impl AsBytes for u32 {}
+/// Types which are free from interior mutability.
+pub unsafe trait Immutable {}
```
semihbkgr, line 31: @mgeisler, I think it had to be added to make the tests pass.

fw-immunant: Shouldn't we still have a `SAFETY` comment here, given the `unsafe impl`?

mgeisler: > EDIT: Oh, I misread this change. It would be ideal for this slide to still actually demonstrate an `unsafe impl` as that is the unsafe capability being discussed here.

Yeah, I agree we should keep the `unsafe impl`.

@semihbkgr, could you simplify the `IntoBytes` trait and remove the `where Self: Immutable` bound? The code has to be super simple and just demonstrate the concept of an unsafe trait (it should not be 100% like what zerocopy has).

src/unsafe-rust/unsafe-traits.md
```diff
@@ -8,36 +8,32 @@ Like with functions, you can mark a trait as `unsafe` if the implementation must
...
+pub unsafe trait IntoBytes {
     fn as_bytes(&self) -> &[u8] {
-        unsafe {
-            slice::from_raw_parts(
-                self as *const Self as *const u8,
-                size_of_val(self),
-            )
-        }
+        let len = mem::size_of_val(self);
+        let slf: *const Self = self;
```
mgeisler, line 22: Can this be expressed more simply now that Rust 1.82 is out? I just saw #2449 from @qwandor which uses the `&raw const` syntax to create a `*const` pointer.

src/unsafe-rust/unsafe-traits.md
```diff
@@ -8,36 +8,32 @@ Like with functions, you can mark a trait as `unsafe` if the implementation must
...
+        unsafe { slice::from_raw_parts(slf.cast::<u8>(), len) }
     }
 }
 
-// SAFETY: `u32` has a defined representation and no padding.
-unsafe impl AsBytes for u32 {}
+/// SAFETY: `u32` has no padding bytes.
```
qwandor, line 27: The defined representation is also important, and safety comments aren't usually doc comments.

qwandor (approved):

mgeisler:
Now that the example uses `&raw`, it's breaking because the GH runners don't have the very latest stable version of Rust. That will be fixed by #2449.

qwandor (approved):

PR #2447, opened 2024-10-20
vi: translate Default trait

vohoanglong0107:
Translate the `Default` chapter of the `std-trait` section to Vietnamese

### DIFF
--- a/po/vi.po
+++ b/po/vi.po
@@ -6911,65 +6911,76 @@ msgstr "\"Dữ liệu đã được ghi: {:?}\""
 
 #: src/std-traits/default.md
 msgid "The `Default` Trait"
-msgstr ""
+msgstr "Trait `Default`"
 
 #: src/std-traits/default.md
 msgid ""
 "[`Default`](https://doc.rust-lang.org/std/default/trait.Default.html) trait "
 "produces a default value for a type."
 msgstr ""
+"Kiểu dữ liệu implement trait [`Default`](https://doc.rust-lang.org/std/"
+"default/trait.Default.html) sẽ sở hữu một giá trị mặc định."
 
 #: src/std-traits/default.md
 msgid "\"John Smith\""
-msgstr ""
+msgstr "\"John Smith\""
 
 #: src/std-traits/default.md
 msgid "\"{default_struct:#?}\""
-msgstr ""
+msgstr "\"{default_struct:#?}\""
 
 #: src/std-traits/default.md
 msgid "\"Y is set!\""
-msgstr ""
+msgstr "\"Biến Y đã được gán!\""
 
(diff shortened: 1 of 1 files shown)


qu-ngx (commented):
I am running some tests on your PR first. I will get back to this PR later when they are done.

po/vi.po
```diff
@@ -6911,72 +6911,86 @@ msgstr "\"Dữ liệu đã được ghi: {:?}\""
...
 #: src/lifetimes/solution.md
 msgid "\"{:#?}\""
-msgstr ""
+msgstr "\"{:#?}\""
 
 #: src/std-traits/default.md
 msgid ""
 "It can be implemented directly or it can be derived via `#[derive(Default)]`."
 msgstr ""
+"Người dùng có thể trực tiếp implement trait này, hoặc derive trait này bằng "
```
qu-ngx, line 6949: I know this trans holds "trait" as a type but since "implement" type isn't here for this one can we use "ứng dụng" or "áp dụng" instead?  Which one do you prefer better and why?

po/vi.po
```diff
@@ -6911,72 +6911,86 @@ msgstr "\"Dữ liệu đã được ghi: {:?}\""
...
+"`#[derive(Default)]`."
 
 #: src/std-traits/default.md
 msgid ""
 "A derived implementation will produce a value where all fields are set to "
 "their default values."
 msgstr ""
+"Khi derive trait này, từng miền của giá trị mặc định được tạo ra sẽ được gán "
```
qu-ngx, line 6957: "derive" will be resolved as well according to the above "implement".

qu-ngx (commented):
I left some comments. Please feel free to correct me if I'm wrong anywhere:)) Tks!

po/vi.po
vohoanglong0107: In my opinion, when discussing traits, it's better to use the raw term "implement" than the its translated version, such as "ứng dụng" or "áp dụng". This is because using implement implies that the type implementing the trait will have all the necessary method and type signatures. The same doesn't apply for "ứng dụng" or "áp dụng", as we ofter use them in a more generic ways.
I believe the same argument can be used for "inheritance", but unlike "implement trait", Vietnamese literature has used "kế thừa" for a long enough time that Vietnamese programmers understand "kế thừa" to mean "inherit every single method and variable".
I admit that this is not optimal, but until we find a better translation for this word, or the Vietnamese Rust user base grows large enough to accept a canonical translation for it, I think it's best to be explicit about what we mean.

qu-ngx: Thanks:) I resolved it

qu-ngx (approved):
LGTM! Thanks for your contributions!

PR #2449, opened 2024-10-25
Updates for Rust 1.82

qwandor:
Rust 1.82 adds `&raw` expressions, and marks some attributes as unsafe.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,6 +55,9 @@ jobs:
       - name: Setup Rust cache
         uses: ./.github/workflows/setup-rust-cache
 
+      - name: Update Rust
+        run: rustup update
+
       - name: Build Rust code
         run: cargo build
 
@@ -89,7 +92,9 @@ jobs:
           sudo apt install gcc-aarch64-linux-gnu
 
       - name: Install toolchain
-        run: rustup target add ${{ matrix.target }}
+        run: |
+          rustup update
+          rustup target add ${{ matrix.target }}
 
       - name: Build Rust code
         working-directory: ${{ matrix.directory }}
@@ -139,6 +144,9 @@ jobs:
         run: |
           sudo apt update
           sudo apt install gettext
+
+      - name: Update Rust
+        run: rustup update
 
       - name: Install mdbook
         uses: ./.github/workflows/install-mdbook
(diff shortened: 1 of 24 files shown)


src/android/interoperability/with-c.md
```diff
@@ -6,14 +6,13 @@ Similarly, you can export Rust functions and call them from C.
 You can do it by hand if you want:
 
 ```rust
-extern "C" {
-    fn abs(x: i32) -> i32;
+unsafe extern "C" {
+    safe fn abs(x: i32) -> i32;
```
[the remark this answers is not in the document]
qwandor: I don't think it makes sense to have a safety comment here. Arguably unsafe extern functions should document their safety requirements, the same as any other unsafe functions, but the fact that this is a `safe fn` already says that it doesn't have any safety requirements, there is nothing to fulfil.

.github/workflows/build.yml
```diff
@@ -55,6 +55,9 @@ jobs:
       - name: Setup Rust cache
         uses: ./.github/workflows/setup-rust-cache
 
+      - name: Update Rust
+        run: rustup update
```
mgeisler, line 59: I'm skeptical of such unpinned updates, but then again, I realize that the build machines on GH also using a floating version — probably tracking stable with some lag?

Perhaps it's okay since the Rust tooling is generally pretty conservative about adding breaking changes.

mgeisler (approved):
I think this is good and the safety comments also look very nice to me.

.github/workflows/build.yml
mgeisler: Ah, I just spotted a small problem with this: we need to update Rust _before_ setting up the cache (just above. The cache is keyed on the Rust version active at the time the action is loaded.

Let me crate a PR for that.

mgeisler: @qwandor, do you know how this update plays together with the tools already installed?

I see

```
  stable-x86_64-unknown-linux-gnu updated - rustc 1.82.0 (f6e511eec 2024-10-15) (from rustc 1.81.0 (eeb90cda1 2024-09-04))
warning: tool `rust-analyzer` is already installed, remove it from `/home/runner/.cargo/bin`, then run `rustup update` to have rustup manage this tool.
warning: tool `rustfmt` is already installed, remove it from `/home/runner/.cargo/bin`, then run `rustup update` to have rustup manage this tool.
warning: tool `cargo-fmt` is already installed, remove it from `/home/runner/.cargo/bin`, then run `rustup update` to have rustup manage this tool.
```

in the latest run on `main`. It would be a shame if we now end up with `rustfmt` out of sync with `rustc`.

PR #2460, opened 2024-11-01
Update Rust before setting up the cache

mgeisler:
The cache is keyed on the Rust version used, so we should update Rust before setting up the cache.

This is a small update to #2449.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,11 +52,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Setup Rust cache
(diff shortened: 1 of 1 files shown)


mgeisler:
Looking a bit more at `rustup`, I wonder if we should start using a toolchain file: https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file.

I've never had to do this before, but it seems like it could simplify our setup in a few places? It could perhaps make

https://github.com/google/comprehensive-rust/blob/f8882190f332061d0dc71d68019f8bd5485db553/.github/workflows/build.yml#L70-L81

unnecessary (by moving this into separate toolchain files).

qwandor (approved):

PR #2463, opened 2024-11-18
Change tests for pattern-matching exercise

mo271:
to catch wrong `right == 0` checks.

To change the number like this catches for example a wrong implementation only looking if `right_result` is zero, but not checking if we are doing a division.
````
fn eval(e: Expression) -> Result<i64, String> {
    match e {
        Expression::Op { op, left, right } => {
            let left_result = eval(*left)?;
            let right_result = eval(*right)?;
            // Correct line commented out. (Or even better match inside the match below)
            //if (right_result == 0) && (matches!(op, Operation::Div)) {
            if right_result == 0 {
                return Err(String::from("division by zero"));
            }
            Ok(match op {
                Operation::Add => left_result + right_result,
                Operation::Sub => left_result - right_result,
                Operation::Mul => left_result * right_result,
                Operation::Div => left_result / right_result
                })
        }
        Expression::Value(value) => Ok(value),
    }
}
```

### DIFF
--- a/src/pattern-matching/exercise.rs
+++ b/src/pattern-matching/exercise.rs
@@ -97,7 +97,7 @@ fn test_recursion() {
         left: Box::new(Expression::Op {
             op: Operation::Sub,
             left: Box::new(Expression::Value(3)),
-            right: Box::new(Expression::Value(4)),
+            right: Box::new(Expression::Value(0)),
         }),
         right: Box::new(Expression::Value(5)),
     };
@@ -107,7 +107,7 @@ fn test_recursion() {
             left: Box::new(term1),
             right: Box::new(term2),
         }),
-        Ok(85)
+        Ok(105)
     );
 }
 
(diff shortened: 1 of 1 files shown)


src/pattern-matching/exercise.rs
```diff
@@ -97,7 +97,7 @@ fn test_recursion() {
         left: Box::new(Expression::Op {
             op: Operation::Sub,
             left: Box::new(Expression::Value(3)),
-            right: Box::new(Expression::Value(4)),
+            right: Box::new(Expression::Value(0)),
```
djmitche, line 100: 3-0 = 3, so this isn't a very interesting case of subtraction.

I guess the idea here is to uncover the bug where _any_ expression with 0 in the RHS is considered an error? I'm not convinced that's worth covering in an exercise like this, but maybe it could be done in another test case, rather than modifying this one.

mo271: Yeah, that was the idea. I'm also not sure if we needs this tests, but it has come up during an real world session...
Added as separate test, but if that makes this file too long, perhaps it is not worth adding them.
In that case: no hard feelings, just close this pr...

djmitche (approved):
Looks good! I figured this actually came up :)

If we get feedback that this is annoying long, we can remove it.

PR #2468, opened 2024-11-23
Add section about Ownership before References

Mandragorian:
Introduction of references was a bit unmotivated without a brief introduction to Ownership. The read might not understand why they are important in rust.

### DIFF
--- a/src/references/ownership.md
+++ b/src/references/ownership.md
@@ -4,11 +4,17 @@ minutes: 5
 
 # Ownership
 
-We already mentioned that one of the main selling points of rust is compile time memory safety, _without_ garbage collection. In order to achieve this, rust imposes some very important rules on how values can be used. These rules are collectively known as the concept of _Ownership_.
+We already mentioned that one of the main selling points of rust is compile time
+memory safety, _without_ garbage collection. In order to achieve this, rust
+imposes some very important rules on how values can be used. These rules are
+collectively known as the concept of _Ownership_.
 
-We will briefly discuss this here, so that you know about. But don't worry if you don't immediately feel comfortable with it! It is going to come up again multiple times throughout the course.
+We will briefly discuss this here, so that you know about. But don't worry if
+you don't immediately feel comfortable with it! It is going to come up again
+multiple times throughout the course.
 
-With the above disclaimer, let's see an example where the concept of Ownership comes into play:
+With the above disclaimer, let's see an example where the concept of Ownership
+comes into play:
 
 ```rust,editable,compile_fail
 fn main() {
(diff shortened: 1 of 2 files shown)


src/references/ownership.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+If you run the code above, you will see that it fails to compile. Most likely
+the error messages will seem quite cryptic to you at this point. That is O.K.!
+Just focus on the fact that rust won't let you do this.
+
+And this is because of Ownership: Initially `s` is the owner of the `String` we
+created. With `let _y = s;`, we moved the ownership of the `String` from `s` to
+`_y`. Then, when we try to use `s`, rust won't let us, since it doesn't own a
+value to be used any more.
```
djmitche, line 34: Most of this slide is disclaimer, with this being the key explanatory bit. And, it uses "owner" and "moved" in an undefined way! At this point, we don't actually need to talk about moving or dropping, two things that are closely related to ownership.

In a chat conversation, Dmytro suggested describing references as a way to not copy data, and that seems like a simpler way to introduce this, particularly since we just covered tuples and arrays. Perhaps an update to the "shared references" slide would do better:
 * Use a large value like `[0u8; 1024]` as an argument to a function
 * Demonstrate that dangling references are forbidden by returning `&[0u8;1024]` referencing a local variable

The next slide, on exclusive references, can then do a similar thing but use the passed array as an output buffer. And then the next slide could do a similar thing with `&mut[u8]`.

Finally, the slide on strings currently mentions "owned", but that could be softened to just a non-copyable type, because Rust doesn't want to automatically copy what might be very large values.

Mandragorian: Well, any definition will have to rely on other "undefined" terms. Even in the book, the definition of ownership is that (among others) "every value has an owner" without defining what that "owner" means. I could say that the owner "changed" instead of "moved" to not introduce a lot of terminology.

I don't believe we need to talk about moving concretely. But I think it is good to make a small reference to the fact that something like this happens. This way it can be in the back of the mind of participants, and it could help to process the course  from that point of view.

I feel like describing references as a way to avoid copying might lead to confusion. Not only because in reality it is not just that, but also it might lead participants to think that "moving = copying", which is not always true.

At the same time I agree that maybe the shared references slide could do with a better example with a function call, as it feels like a more natural use of references.

randomPoison:
I'm not in favor of this change, personally. I don't think it'd be helpful to introduce ownership here. I think we can pretty easily talk about references without needing to mention ownership. Usually I frame it in terms of pointers/references in other languages, since students are already going to be familiar with the concept of a reference (though with different semantics depending on what languages they know). I also think @djmitche's suggestion to frame it in terms of "pass data around without copying" is a good idea, since it provides some motivation for references while not needing to bring up ownership. We also have an example of showing how a dangling reference can't be created/returned, which I think is a good way to hint at ownership without having to fully introduce it.

djmitche:
I just taught the course again, and we used the "pass data without copying it" phrase without causing any confusion. So I'll close this, in agreement with @randomPoison.

PR #2469, opened 2024-11-24
use makefile to install, test and serve materials

AbdulfatahMohammedSheikh:
The process of installing the required crates in manual. So instead I added a make file to simplify the process

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,14 @@
+install:
+	cargo install mdbook
+	cargo install --locked mdbook-svgbob
+	cargo install --locked mdbook-i18n-helpers
+	cargo install --locked i18n-report
+	cargo install --locked mdbook-linkcheck
+	cargo install --locked --path mdbook-exerciser
+	cargo install --locked --path mdbook-course
+
+test: 
+	mdbook test
+
+serve:
+	mdbook serve
--- a/README.md
+++ b/README.md
@@ -68,25 +68,19 @@ cd comprehensive-rust
 Then install these tools with:
 
 ```shell
-cargo install mdbook
-cargo install --locked mdbook-svgbob
-cargo install --locked mdbook-i18n-helpers
-cargo install --locked i18n-report
-cargo install --locked mdbook-linkcheck
-cargo install --locked --path mdbook-exerciser
-cargo install --locked --path mdbook-course
+make install
 ```
 
 Run
 
 ```shell
-mdbook test
+make test
 ```
 
 to test all included Rust snippets. Run
 
 ```shell
-mdbook serve
+make serve
 ```
 
 to start a web server with the course. You'll find the content on


mgeisler:
Thanks for putting this up, it's something I hadn't considered before...

I'm also not a super fan of `make` these days: I am afraid we will get questions from people on Windows about how to install it and how to run it?

Introducing a (soft) dependency on new non-Rust software sounds like a step back, so I don't think we should do this.

But, this gives me an idea... would it be possible for us to create a project in this repository which depends on everything we need? Let's say it's named `install-comprehensive-rust-tools`, then we could ask people to do:

```
cargo install --locked --path install-comprehensive-rust-tools
```

I'm not sure this can work since I don't think binaries from dependencies will be installed.

AbdulfatahMohammedSheikh:
> Thanks for putting this up, it's something I hadn't considered before...
> 
> I'm also not a super fan of `make` these days: I am afraid we will get questions from people on Windows about how to install it and how to run it?
> 
> Introducing a (soft) dependency on new non-Rust software sounds like a step back, so I don't think we should do this.
> 
> But, this gives me an idea... would it be possible for us to create a project in this repository which depends on everything we need? Let's say it's named `install-comprehensive-rust-tools`, then we could ask people to do:
> 
> ```
> cargo install --locked --path install-comprehensive-rust-tools
> ```
> 
> I'm not sure this can work since I don't think binaries from dependencies will be installed.

Interesting, it is cool idea. As for the Makefile i don't like it either but what i don't like more than a makefile is long list of dependencies to install.

djmitche:
I've used [cargo xtask](https://github.com/matklad/cargo-xtask) in some other situations as a way to automate things in a repo. With some work, that could mean that just running `cargo xtask install-deps` would perform the same function as the Makefile.

mgeisler:
> I've used [cargo xtask](https://github.com/matklad/cargo-xtask) in some other situations as a way to automate things in a repo. With some work, that could mean that just running `cargo xtask install-deps` would perform the same function as the Makefile.

Yes, I would be happier with the `cargo xtask` trick which I've seen a few places. However, now that I've actually read how it works, I must say that I had hoped it would bring along some infrastructure for running simple commands :smile: I believe that is left as an exercise for the reader, right?

mgeisler:
> Interesting, it is cool idea. As for the Makefile i don't like it either but what i don't like more than a makefile is long list of dependencies to install.

Yeah, I agree with you! It gets worse: the commands used in CI are not synchronized with that we show in the README. So real bugs can creep in where CI does one thing, but people do something else. It would be nice to align this more carefully with a nice cross-platform setup.

mgeisler:
I think we should close this and then open an issue for this instead. We should do something very simple to make Dustin's `cargo xtask install-deps` work. It could be as simple as a

```rust
fn main() -> Result<(), Box<dyn std::error::Error>> {
  Command::new(env!("CARGO")).arg("install").arg("mdbook").status()?;
  Command::new(env!("CARGO")).arg("install").arg("mdbook-i18n-helpers").status()?;
}
```

Put that in a new workspace member, hook up the Cargo alias and we should be in business :smile:  To show that it works, make the [GitHub CI code](https://github.com/google/comprehensive-rust/blob/main/.github/workflows/install-mdbook/action.yml) that install the dependencies use the new task.

mgeisler:
I copied the above comment to #2509, let's continue the discussion there!

PR #2470, opened 2024-11-25
ja: refresh po file

HidenoriKobayashi:
Let me refresh the po file, since I plan to work on some of the fuzzy translation stuff this week.

This is basically output of only doing: `msgmerge --update po/ja.po book/xgettext/messages.pot` and removing a few fuzzy tags for obviously correct translations.


keiichiw (approved):
Thanks!

PR #2471, opened 2024-11-25
Add a test framework to test the JS part of the course

### LINKED ISSUE #2462: JavaScript tests
Today, we don't have any tests for the JavaScript code found in

- https://github.com/google/comprehensive-rust/blob/main/theme/book.js: executes the Rust Playground code (comes from the upstream `mdbook` project).
- https://github.com/google/comprehensive-rust/blob/main/theme/speaker-notes.js: our own home-grown speaker note system.
- https://github.com/google/comprehensive-rust/blob/main/theme/save-playgrounds.js: not used currently, could save the state of a playground between page loads.

This has caused bugs in the past (the playgrounds disappeared at some point) and it makes it hard to upgrade `mdbook` (because we need to  merge in changes to `book.js` manually).

It would be very helpful to have tests for these files. This should be something we can run on the command line in GitHub CI. I think we need something which 

- We can run on pre-rendered HTML files. I envision us running `mdbook build` (for each language).
- A headless browser would then be needed to inspect the pages.
- We would need a unit testing framework of some sort.
  - I used https://www.protractortest.org/ in the past, but there are many others.

michael-kerscher:
This is a first draft for a test framework for testing the JS part of the book that is discussed in #2462

It is using [webdriverIO](https://webdriver.io/) and the webdriverIO [Expect API](https://webdriver.io/docs/api/expect-webdriverio/) in combination with [Mocha](https://mochajs.org/). WebdriverIO is taking care of accessing the webpage with a real browser and can access the state of the page so behavior can be asserted.

Currently only a small test for the speaker-notes implementation demos the functionality.

The [Static Server Service](https://webdriver.io/docs/static-server-service/) is used to serve the book in a way that the test runner can access it.

A CI integration can look like https://webdriver.io/docs/githubactions/ and is implemented with a headless setup. In CI it uses the language variable to set environment variable that configures where the built book should be mounted from

Fixes #2462.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,4 +1,4 @@
-# Testing the comprehensive rust course book
+# Testing Comprehensive Rust
 
 The course material contains JS code that can break and needs to be checked for
 functionality. Examples are `theme/speaker-notes.js` or `theme/book.js`.
@@ -11,7 +11,7 @@ so behavior can be asserted.
 
 The [Static Server Service](https://webdriver.io/docs/static-server-service/) is
 used mainly in the [CI](../github/workflows/build.yml) to serve the book on port
-localhost:8080 such that the test runner can access it. This mode is used when
+`localhost:8080` such that the test runner can access it. This mode is used when
 `npm start` or `npm test` is executed.
 
 For local testing and quick iterations it is possible to use `mdbook serve`
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,0 +1,20 @@
+# Testing the comprehensive rust course book
+
+The course material contains JS code that can break and needs to be checked for
+functionality. Examples are `theme/speaker-notes.js` or `theme/book.js`.
+
+Comprehensive Rust is using [webdriverIO](https://webdriver.io/) and the
+[webdriverIO Expect API](https://webdriver.io/docs/api/expect-webdriverio/) in
+combination with [Mocha] (https://mochajs.org/). WebdriverIO is taking care of
+accessing the webpage with a real browser and can access the state of the page
+so behavior can be asserted.
+
+The [Static Server Service](https://webdriver.io/docs/static-server-service/) is
+used mainly in the [CI](../github/workflows/build.yml) to serve the book on port
+localhost:8080 such that the test runner can access it. This mode is used when
+`npm start` or `npm test` is executed.
+
+For local testing and quick iterations it is possible to use `mdbook serve`
+which creates a small HTTP server on port 3000 by default. There is a special
+config that is invoked with `npm run test-mdbook` that uses
+`http://localhost:3000`
--- a/tests/wdio.conf.ts
+++ b/tests/wdio.conf.ts
@@ -52,7 +52,11 @@ export const config: WebdriverIO.Config = {
   capabilities: [
     {
       // capabilities for local browser web tests
-      browserName: "chrome", // or "firefox", "microsoftedge", "safari"
+      browserName: "chrome",
+      "goog:chromeOptions": {
+        // comment this to see the evaluation
+        args: ["headless", "disable-gpu"],
+      },
     },
   ],
 
--- a/tests/wdio.conf.ts
+++ b/tests/wdio.conf.ts
@@ -68,7 +68,7 @@ export const config: WebdriverIO.Config = {
   // Define all options that are relevant for the WebdriverIO instance here
   //
   // Level of logging verbosity: trace | debug | info | warn | error | silent
-  logLevel: "info",
+  logLevel: "warn",
   //
   // Set specific log levels per logger
   // loggers:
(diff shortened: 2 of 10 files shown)


tests/wdio.conf.ts
```diff
@@ -0,0 +1,310 @@
...
+  ],
+
+  //
+  // ===================
+  // Test Configurations
+  // ===================
+  // Define all options that are relevant for the WebdriverIO instance here
+  //
+  // Level of logging verbosity: trace | debug | info | warn | error | silent
+  logLevel: "info",
```
mgeisler, line 71: I tested this locally and when I change this to `warn`, I get very useful output:

```suggestion
  logLevel: "warn",
```

Introducing a small error results in output like this:

```
% TEST_BOOK_DIR=../book/en/html npm run wdio

> wdio
> wdio run ./wdio.conf.ts


Execution of 2 workers started at 2024-11-28T15:58:10.041Z

[0-0] RUNNING in chrome - file:///src/basic.test.ts
[0-1] RUNNING in chrome - file:///src/speaker-notes.test.ts
[0-0] Error in "Basic test.should have the default_theme light"
Error: expect(received).toBe(expected) // Object.is equality

Expected: "xlight"
Received: "light"
    at Context.<anonymous> (.../src/comprehensive-rust/tests/src/basic.test.ts:7:63)
[0-0] FAILED in chrome - file:///src/basic.test.ts
[0-1] 2024-11-28T15:58:16.536Z WARN webdriver: WebDriverError: javascript error: {"status":60,"value":"[object HTMLButtonElement] has no size and location"}
[0-1]   (Session info: chrome=131.0.6778.85) when running "actions" with method "POST" and args "{"actions":[{"id":"action1","type":"wheel","parameters":{},"actions":[{"type":"scroll","x":-751,"y":1079,"deltaX":0,"deltaY":0,"duration":0,"origin":{"element-6066-11e4-a52e-4f735466cecf":"f.82DFDD24A66D5F65A979257D82375039.d.48683DE2EE1A6FB6FAD0CDAB4CA1C1B3.e.44"}}]}]}"
[0-1] PASSED in chrome - file:///src/speaker-notes.test.ts

Spec Files:	 1 passed, 1 failed, 2 total (100% completed) in 00:00:06  
```

mgeisler (approved):
This looks really cool from a quick local test run!

mgeisler:
Small things to add here or in a later PR:

- a README which briefly explains the setup and in particular has links to documentation for the
  - the Mocha unit test framework and
  - the WebdriverIO test runner (just so I don't have to look it up myself when I want to add more tests)
- when we have a README in the directory, we should then also add a link to to other relevant documentation so people know to update the tests
- perhaps a default command for `npm` so that `npm run` does the right thing by default? If I remember correctly from my web developer days, that's possible :smile:
- would it be possible to have a setup for local development where the tests run against the server on http://localhost:3000 started by `mdbook serve`? Using the HTML on disk is great for GitHub CI, but I think it would be cool to iterate with `mdbook serve` locally

Feel free to do all this here or in later PRs, the existing setup is already a great improvement on what we have right now.

tests/wdio.conf.ts
michael-kerscher: sounds good to me!

michael-kerscher:
Thinking about mdbook serve, yes there is a benefit to this I have not thought of by not having to recreate the book.
This looks a bit more difficult to support two different setups though. Either I'm using mdbook serve in CI as well or I modify the config so it can be configured by the CI by setting a environment variable

michael-kerscher:
I included a second config, deriving from the regular config that uses `http://localhost:3000` (mdbook serve). The CI still uses the static file service to test exactly what is used in production.
A Readme is now also available.
I'm open to discuss if `npm start` should be developer friendly and use the `npm test-mdbook` or `npm test`. Currently it is testing with the static service

tests/README.md
```diff
@@ -0,0 +1,20 @@
+# Testing the comprehensive rust course book
```
mgeisler, line 1:
```suggestion
# Testing Comprehensive Rust
```

tests/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+Comprehensive Rust is using [webdriverIO](https://webdriver.io/) and the
+[webdriverIO Expect API](https://webdriver.io/docs/api/expect-webdriverio/) in
+combination with [Mocha] (https://mochajs.org/). WebdriverIO is taking care of
+accessing the webpage with a real browser and can access the state of the page
+so behavior can be asserted.
+
+The [Static Server Service](https://webdriver.io/docs/static-server-service/) is
+used mainly in the [CI](../github/workflows/build.yml) to serve the book on port
+localhost:8080 such that the test runner can access it. This mode is used when
```
mgeisler, line 14:
```suggestion
`localhost:8080` such that the test runner can access it. This mode is used when
```

PR #2474, opened 2024-11-26
ja: update summary translations

HidenoriKobayashi:
Update translations from SUMMARY.md.

Unfortunately, poedit changes the wrapping length of the po files and I could not get it back by changing the poedit setting for line length. I believe we won't get any more of these noisy diffs for my further changes this week.

`Cancellation` is the last item I modified in this PR.

### DIFF
--- a/po/ja.po
+++ b/po/ja.po
@@ -190,7 +190,7 @@ msgstr "配列のイテレート"
 #: src/SUMMARY.md src/tuples-and-arrays.md
 #: src/tuples-and-arrays/destructuring.md
 msgid "Patterns and Destructuring"
-msgstr "パターンと分配"
+msgstr "パターンとデストラクト"
 
 #: src/SUMMARY.md src/tuples-and-arrays.md src/tuples-and-arrays/exercise.md
 msgid "Exercise: Nested Arrays"
@@ -272,11 +272,11 @@ msgstr ""
 
 #: src/SUMMARY.md src/pattern-matching.md
 msgid "Destructuring Structs"
-msgstr "構造体編"
+msgstr "構造体のデストラクト"
 
 #: src/SUMMARY.md src/pattern-matching.md
 msgid "Destructuring Enums"
-msgstr "列挙型の分配"
+msgstr "列挙型のデストラクト"
 
 #: src/SUMMARY.md src/pattern-matching.md
 #: src/pattern-matching/let-control-flow.md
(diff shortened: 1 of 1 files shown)


po/ja.po
```diff
@@ -525,14 +506,12 @@ msgid "Borrowing"
...
-#, fuzzy
 msgid "Borrowing a Value"
-msgstr "借用"
+msgstr "値の借用"
 
 #: src/SUMMARY.md src/borrowing.md src/borrowing/borrowck.md
-#, fuzzy
 msgid "Borrow Checking"
-msgstr "借用"
+msgstr "借用に関するチェック"
```
keiichiw, line 514:
```suggestion
msgstr "借用チェック"
```

As "borrow checker" is usually translated to "借用チェッカー", "借用チェック" would be enough and simple

po/ja.po
```diff
@@ -507,9 +489,8 @@ msgid "`Rc`"
 msgstr "`Rc`"
 
 #: src/SUMMARY.md src/smart-pointers.md src/smart-pointers/trait-objects.md
-#, fuzzy
 msgid "Owned Trait Objects"
-msgstr "トレイトオブジェクト"
+msgstr "所有トレイトオブジェクト"
```
keiichiw, line 493:
```suggestion
msgstr "所有されたトレイトオブジェクト"
```

In my understanding, this is "Owned" + "Trait objects".

po/ja.po
```diff
@@ -739,9 +708,8 @@ msgid "AIDL"
 msgstr "AIDL（Androidインターフェイス定義言語）"
 
 #: src/SUMMARY.md src/android/aidl/birthday-service.md
-#, fuzzy
 msgid "Birthday Service Tutorial"
-msgstr "/** 誕生日サービスのインターフェース。*/"
+msgstr "誕生日サービスのインターフェース"
```
keiichiw, line 712: ”チュートリアル” instead of "インターフェイス"?

po/ja.po
```diff
@@ -773,33 +740,28 @@ msgid "Changing API"
...
-#, fuzzy
 msgid "Sending Objects"
-msgstr "トレイトオブジェクト"
+msgstr "オブジェクトの送信"
 
 #: src/SUMMARY.md src/android/aidl/types/parcelables.md
-#, fuzzy
 msgid "Parcelables"
-msgstr "変数"
+msgstr "パーセラブル"
```
keiichiw, line 764:
```suggestion
msgstr "Parcelables"
```

I think this is a technical term that we don't have no Japanese word. So, I guess keeping it in English would be more helpful
https://developer.android.com/reference/android/os/Parcelable

keiichiw (commented):
Thank you for working on this!

HidenoriKobayashi:
Thanks for the review and suggestions! Please take another look.

HidenoriKobayashi:
@keiichiw san, could help review this again please?

keiichiw (approved):
Sorry for the late review. LGTM. Approved!

PR #2475, opened 2024-11-26
Add more description of proto to the proto exercise

### LINKED ISSUE #2466: Clarify Protobuf exercise (lifetimes)
Without knowing the Protobuf wire format, the exercise description and example code is hard to understand.
It would have helped to add an additional sentence describing that submessages (like `PhoneNumber` as part of `Person`) are encoded as `Len` type of the encoded submessage (1. encode `PhoneNumber` 2. store it as `Len` type field in the `Person` wire format).

Also the name `Len` for basically a length annotated byte array is a bit unintuitive.

Furthermore, it would have helped to have the tests already be presented as part of the exercise and not only as part of the solution.

djmitche:
This also moves the test cases into `main` so they are visible.

Fixes #2466.

### DIFF
--- a/src/lifetimes/exercise.md
+++ b/src/lifetimes/exercise.md
@@ -29,15 +29,36 @@ message Person {
 }
 ```
 
+## Messages
+
 A proto message is encoded as a series of fields, one after the next. Each is
 implemented as a "tag" followed by the value. The tag contains a field number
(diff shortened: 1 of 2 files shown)


onihilist (commented):
seems nice :)

mgeisler (approved):
Thanks!

PR #2479, opened 2024-11-28
Revert "Execution: Add stderr block"

mgeisler:
Reverts google/comprehensive-rust#2397

I'm seeing

```
POST https://play.rust-lang.org/execute 500 (Internal Server Error)
```

on every request when trying to run code in the Playground. I'm not sure why, but we need to revert this ASAP if others see the same.

### DIFF
--- a/theme/book.js
+++ b/theme/book.js
@@ -110,17 +110,10 @@ function playground_text(playground, hidden = true) {
     }
 
     function run_rust_code(code_block) {
-        var result_stderr_block = code_block.querySelector(".result.stderr");
-        if (!result_stderr_block) {
-            result_stderr_block = document.createElement('code');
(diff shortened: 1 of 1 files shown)


mgeisler:
@michael-kerscher, another reason why we need to merge and start using your unit testing code :smile:

mgeisler:
> @michael-kerscher, another reason why we need to merge and start using your unit testing code 😄

Actually, I now think we ran into a [Playground incident](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/playground.20incident.202024-12-03) where the Playground was slow for everyone.

I would like to get @Alx-Lai's change back in, so I'll set up a revert for this.

Issue #1083, opened 2023-08-15
Add speaker notes to every page


mgeisler:
We currently have speaker notes on 165 out of 265 pages.

- We should aim to add speaker notes on all new pages
- We should also add notes to all existing pages.

Notes are very helpful for both instructors and people who read the course on their own, so we want to be as complete as possible here.

michael-kerscher:
Is there already an easy way to distinguish pages without speaker details from ones that already have them? The only way to find them currently is to actively click through "random" pages and try to see if speaker notes are available.

If there is no good mechanism yet that I just missed, what do you think about some automation and/or dashboard to easily get an overview on missing metadata?

mgeisler:
> Is there already an easy way to distinguish pages without speaker details from ones that already have them? The only way to find them currently is to actively click through "random" pages and try to see if speaker notes are available.

I think `grep` or `rg` is the only structured way. A command like this could be a start:

```
rg -g '*.md' -g '!{SUMMARY,README,exercise,solution*,welcome*}.md' --files-without-match '<details>' src | sort
```

This gives me this list right now:

```
src/android/aidl/birthday-service.md
src/android/aidl/example-service/deploy.md
src/android/aidl/example-service/implementation.md
src/android/aidl.md
src/android/aidl/types.md
src/android/aidl/types/parcelables.md
src/android/aidl/types/primitives.md
src/android/build-rules/binary.md
src/android/build-rules/library.md
src/android/interoperability/cpp/android-build-rust.md
src/android/interoperability/cpp/generated-cpp.md
src/android/interoperability/cpp.md
src/android/interoperability/java.md
src/android/interoperability.md
src/android/interoperability/with-c/bindgen/c-library.md
src/android/interoperability/with-c/bindgen.md
src/android/interoperability/with-c/calling-rust.md
src/android/interoperability/with-c/hand-written.md
src/android/interoperability/with-c.md
src/android/logging.md
src/android/testing.md
src/bare-metal/android.md
src/bare-metal.md
src/bare-metal/setup.md
src/bare-metal/useful-crates.md
src/borrowing.md
src/chromium/adding-third-party-crates/checking-in.md
src/chromium/adding-third-party-crates/configuring-cargo-toml.md
src/chromium/adding-third-party-crates/configuring-gnrt-config-toml.md
src/chromium/adding-third-party-crates/depending-on-a-crate.md
src/chromium/adding-third-party-crates/downloading-crates.md
[...]

PR #2498, opened 2024-12-04
Add speaker notes to flow control basics

mgeisler:
Part of #1083.

### DIFF
--- a/src/control-flow-basics.md
+++ b/src/control-flow-basics.md
@@ -1,3 +1,12 @@
 # Control Flow Basics
 
 {{%segment outline}}
+
+<details>
+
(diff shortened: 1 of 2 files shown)


djmitche (approved):
I guess the intent here is to have _something_ in the speaker notes, even if it's not especially substantive?

mgeisler:
> I guess the intent here is to have _something_ in the speaker notes, even if it's not especially substantive?

Good and very fair question! I think my take on the notes is that they can provide a running commentary to the reader. Perhaps I'm drifting away from them being for the _speaker_ and instead being for both _speaker_ and people taking the class.

PR #2503, opened 2024-12-06
Add "Execution: Add stderr block" again

mgeisler:
Reverts google/comprehensive-rust#2479, which is a revert of #2397.

I think the problem was not related to @Alx-Lai's change, the [Playground was slow](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/playground.20incident.202024-12-03) for everyone.

### DIFF
--- a/theme/book.js
+++ b/theme/book.js
@@ -110,10 +110,17 @@ function playground_text(playground, hidden = true) {
     }
 
     function run_rust_code(code_block) {
-        var result_block = code_block.querySelector(".result");
+        var result_stderr_block = code_block.querySelector(".result.stderr");
+        if (!result_stderr_block) {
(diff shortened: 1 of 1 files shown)


qwandor (approved):

Alx-Lai:
I tested it locally and it's good from my side. Because of the unstable problem, @mgeisler can you also test again in your personal build? Just in case we need to rollback it again.

mgeisler:
> I tested it locally and it's good from my side. Because of the unstable problem, @mgeisler can you also test again in your personal build? Just in case we need to rollback it again.

Yes, I tested it here and it works fine! Thanks for testing it too. With #2462 solved, we should add a little JavaScript test next to ensure the functionality doesn't regress unexpectedly.

PR #2506, opened 2024-12-06
uk: Fix failing test case

reta:
Fix one failing test case introduced by https://github.com/google/comprehensive-rust/pull/2490

PR #2507, opened 2024-12-09
uk: avoid translating fixed strings in protobufs

mgeisler:
The protobuf exercise asks people to decode a few fixed byte slices:

  https://google.github.io/comprehensive-rust/uk/lifetimes/exercise.html

Some of the messages in the slices are fixed English strings, such as “home”. They should thus not be translated into other languages.

This should fix the CI. I guess it wasn’t caught before merging because of a parallel update.

### DIFF
--- a/po/uk.po
+++ b/po/uk.po
@@ -9950,11 +9950,11 @@ msgstr "\"\""
 
 #: src/lifetimes/solution.md
 msgid "\"beautiful name\""
-msgstr "\"красиве ім'я\""
+msgstr "\"beautiful name\""
 
 #: src/lifetimes/solution.md
 msgid "\"Evan\""
-msgstr "\"Еван\""
+msgstr "\"Evan\""
 
 #: src/lifetimes/solution.md
 msgid "\"+1234-777-9090\""
@@ -9962,7 +9962,7 @@ msgstr ""
 
 #: src/lifetimes/solution.md
 msgid "\"home\""
-msgstr "\"дім\""
+msgstr "\"home\""
 
 #: src/welcome-day-4.md
 msgid "Welcome to Day 4"


mgeisler:
Made obsolete by #2506.

@reta, in case you weren't aware, the best way to indicate that a string should remain untranslated is to copy the `msgid` to the `msgstr`. That way, it won't [show up as untranslated](https://google.github.io/comprehensive-rust/translation-report.html).

reta:
> @reta, in case you weren't aware, the best way to indicate that a string should remain untranslated is to copy the `msgid` to the `msgstr`. That way, it won't [show up as untranslated](https://google.github.io/comprehensive-rust/translation-report.html).

Thanks @mgeisler , yes I am aware of that but it makes a bit difficult to track new messages that require translation (they are always having empty `msgstr`) vs the ones where translation is not required. Thanks for the hint anyway!

mgeisler:
> > @reta, in case you weren't aware, the best way to indicate that a string should remain untranslated is to copy the `msgid` to the `msgstr`. That way, it won't [show up as untranslated](https://google.github.io/comprehensive-rust/translation-report.html).
> 
> Thanks @mgeisler , yes I am aware of that but it makes a bit difficult to track new messages that require translation (they are always having empty `msgstr`) vs the ones where translation is not required.

Okay, though I don't understand how it makes difficult to track new messages? You will have

```po
msgid "untranslated"
msgstr ""

msgid "no-op translation"
msgstr "no-op translation"
```

in the `uk.po` file. So new messages are those with an empty `msgstr`, the already translated messages (no-op or not) will have a non-empty `msgstr`.

If a no-op translation changes, you end up with something like

```po
#, fuzzy
msgid "changed no-op translation"
msgstr "no-op translation"
```

The fuzzy entry will be ignored when building the book, thus resulting in "changed no-op translation" showing up, which is what you'll want.

reta:
> The fuzzy entry will be ignored when building the book, thus resulting in "changed no-op translation" showing up, which is what you'll want.

Thanks @mgeisler , this is correct, I have difficulties to track this case:

```
msgid "untranslated-does-not-need-to"
msgstr ""
```

```
msgid "new-translation-that-need-to-be-translated"
msgstr ""
```

May this is a flaw in my workflow only, but thanks a lot for sharing your thoughts over it.

reta:
> @reta, in case you weren't aware, the best way to indicate that a string should remain untranslated is to copy the `msgid` to the `msgstr`. That way, it won't [show up as untranslated](https://google.github.io/comprehensive-rust/translation-report.html).

@mgeisler my sincere apologies, I completely misread your initial suggestion, this is exactly the way we (I) should be doing it, sorry about that.

PR #2510, opened 2024-12-11
Changes to do bare-metal from chromebook

djmitche:
On my chromebook:
 * I needed build-essentials to install cc
 * I needed to enable sharing the USB device with linux. This instruction was from a support.google.com article so well-known but not obvious.

### DIFF
--- a/src/bare-metal.md
+++ b/src/bare-metal.md
@@ -29,7 +29,7 @@ To get started, install some tools we'll need later. On gLinux or Debian:
 <!-- mdbook-xgettext: skip -->
 
 ```bash
-sudo apt install gdb-multiarch libudev-dev picocom pkg-config qemu-system-arm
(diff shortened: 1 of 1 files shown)


src/bare-metal.md
```diff
@@ -47,6 +47,11 @@ echo 'SUBSYSTEM=="hidraw", ATTRS{idVendor}=="0d28", MODE="0660", GROUP="logindev
 sudo udevadm control --reload-rules
 ```
 
+You should see "NXP ARM mbed" in the output of `lsusb` if the device is
+available. If you are using a Linux environment on a Chromebook, you will need
+to share the USB device with linux, via
```
mgeisler, line 52: Tiny, tiny fix:

```suggestion
to share the USB device with Linux, via
```

mgeisler (approved):
Thanks!

PR #2511, opened 2024-12-12
Fix example from zerocopy.

### LINKED ISSUE #2472: Unsafe trait implementation for `IntoBytes` is incorrect (and UB)
The [`IntoBytes` implementation](https://google.github.io/comprehensive-rust/unsafe-rust/unsafe-traits.html) is incorrect, as shown when running this program under Miri. The correct syntax is `&raw const *self` (which returns a `*const Self`) rather than `&raw const self` (which returns a `*const &Self`).

```rust
use std::{mem, slice};

/// ...
/// # Safety
/// The type must have a defined representation and no padding.
pub unsafe trait IntoBytes {
    fn as_bytes(&self) -> &[u8] {
        let len = mem::size_of_val(self);
        unsafe { slice::from_raw_parts((&raw const self).cast::<u8>(), len) }
    }
}

// SAFETY: `u32` has a defined representation and no padding.
unsafe impl IntoBytes for u32 {}

fn main() {
    println!("{:?}", 42.as_bytes());
}
```

```
error: Undefined Behavior: out-of-bounds pointer use: alloc811 has been freed, so this pointer is dangling
  --> src/main.rs:17:22
   |
17 |     println!("{:?}", 42.as_bytes());
   |                      ^^^^^^^^^^^^^ out-of-bounds pointer use: alloc811 has been freed, so this pointer is dangling
   |
   = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
   = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
help: alloc811 was allocated here:
  --> src/main.rs:7:17
   |
7  |     fn as_bytes(&self) -> &[u8] {
   |                 ^^^^^
help: alloc811 was deallocated here:
  --> src/main.rs:10:6
   |
10 |     }
   |      ^
   = note: BACKTRACE (of the first span):
   = note: inside `main` at src/main.rs:17:22: 17:35

note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace
```

qwandor:
This was changed incorrectly in #2434.

Fixes #2472.

### DIFF
--- a/src/unsafe-rust/unsafe-traits.md
+++ b/src/unsafe-rust/unsafe-traits.md
@@ -19,7 +19,8 @@ use std::{mem, slice};
 pub unsafe trait IntoBytes {
     fn as_bytes(&self) -> &[u8] {
         let len = mem::size_of_val(self);
-        unsafe { slice::from_raw_parts((&raw const self).cast::<u8>(), len) }
+        let slf: *const Self = self;
+        unsafe { slice::from_raw_parts(slf.cast::<u8>(), len) }
     }
 }
 


src/unsafe-rust/unsafe-traits.md
```diff
@@ -19,7 +19,8 @@ use std::{mem, slice};
 pub unsafe trait IntoBytes {
     fn as_bytes(&self) -> &[u8] {
         let len = mem::size_of_val(self);
-        unsafe { slice::from_raw_parts((&raw const self).cast::<u8>(), len) }
+        let slf: *const Self = self;
```
gendx, line 22: Nit: call it `ptr`?

gendx (changes requested):

src/unsafe-rust/unsafe-traits.md
qwandor: This example is a shortened version of the actual upstream `zerocopy` code, so I think it's best to avoid unnecessary changes.

djmitche (approved):

gendx (approved):

Issue #1390, opened 2023-10-19
Make "Speaker Notes" translatable


mgeisler:
The "Speaker Notes" string is currently hard-coded in the JavaScript code that injects it:

https://github.com/google/comprehensive-rust/blob/e9c61b91c19190612f87db8082e70f7a84bdd9ef/theme/speaker-notes.js#L100

There are several other strings in that file that need translation too.

Sweetdevil144:
Hello @mgeisler !!
Is the issue still open?
If so then I would like to work on it!! 🙂

mgeisler:
Hi @Sweetdevil144, it is open! Please go ahead and propose a plan for how we can do this.

Sweetdevil144:
Hello @mgeisler ,
I think a practical solution for making "Speaker Notes" translatable would be to implement a language file that houses translations for all relevant strings within the code. Based on the user's selected language preference, we can dynamically load the corresponding translation from the file.

mgeisler:
> I think a practical solution for making "Speaker Notes" translatable would be to implement a language file that houses translations for all relevant strings within the code.

Yeah, I think we should find a way to inject the string "Speaker Notes" into the PO files we already use to translate the course. We have a translation pipeline set up for the course and we should try to build on it.

The translation work by

- Extracting text from the pages using a custom `mdbook` backend
- Translating text when publishing using a custom `mdbook` preprocessor

See [TRANSLATIONS.md](https://github.com/google/comprehensive-rust/blob/main/TRANSLATIONS.md) for the details and see [mdbook-i18n-helpers](https://github.com/google/mdbook-i18n-helpers) for the tooling.

Now, how do speaker notes work? They work purely using JavaScript, see [speaker-notes.js](https://github.com/google/comprehensive-rust/blob/main/theme/speaker-notes.js). The code looks for a `details` element and transforms what is inside. The `Speaker Notes` string come from there.

To make something appear in the PO files, it needs to be part of the pages when the `mdbook-xgettext` backend is running. We can do this by hand: we could turn every `<details> ...` into `## Speaker Notes` heading. The JavaScript would then need to be updated to find this heading instead of finding the `details` element — which becomes a little tricky when the heading is translated for some langauges (we could perhaps add an ID with a [header attribute](https://rust-lang.github.io/mdBook/format/markdown.html#heading-attributes)?).

[...]

PR #2513, opened 2024-12-13
tests: hello world playground run with success and failure tests

michael-kerscher:
Implement two tests for the rust playground:
- successful run of the hello world example with the hello world message in stdout and a hidden stderr
- on purpose compilation error is shown in stderr and "No output" in stdout

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -189,6 +189,7 @@ jobs:
         run: npm install
         working-directory: ./tests
       - name: Test Javascript
+        if: matrix.language == 'en'
         run: npm test
         env:
           TEST_BOOK_DIR: ../book/comprehensive-rust-${{ matrix.language }}/html
--- a/tests/package-lock.json
+++ b/tests/package-lock.json
@@ -7,10 +7,10 @@
       "name": "test",
       "devDependencies": {
         "@types/mocha": "^10.0.10",
-        "@wdio/cli": "^9.3.0",
-        "@wdio/local-runner": "^9.3.0",
-        "@wdio/mocha-framework": "^9.2.8",
-        "@wdio/static-server-service": "^9.2.2"
+        "@wdio/cli": "^9.5.3",
+        "@wdio/local-runner": "^9.5.3",
+        "@wdio/mocha-framework": "^9.5.0",
+        "@wdio/static-server-service": "^9.5.0"
       }
     },
     "node_modules/@babel/code-frame": {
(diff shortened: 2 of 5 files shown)


michael-kerscher:
Tests are failing because the Hello world message is localized

michael-kerscher:
Without localization issues another error class shows.
There are additional issues that don't show in my local runs and also not for all languages:

```
[0-2] 2024-12-13T15:25:07.756Z WARN webdriver: WebDriverError: javascript error: {"status":60,"value":"[object HTMLButtonElement] has no size and location"}
[0-2]   (Session info: chrome=131.0.6778.139) when running "actions" with method "POST" and args "{"actions":[{"id":"action1","type":"wheel","parameters":{},"actions":[{"type":"scroll","x":-751,"y":978,"deltaX":0,"deltaY":0,"duration":0,"origin":{"element-6066-11e4-a52e-4f735466cecf":"f.173C819DA147237EEEF427DC48FFF96F.d.A21CD0374576D9BD3747C260A5497CB1.e.45"}}]}]}"
[0-1] Error in "Playground.executes the hello world code and prints the hello message"
Error: Expect $(`.ace_content`) to exist
```

michael-kerscher:
I looked again into this and it seems I used a not ideal directory for the hello-world page as the "types-and-values" directory does not have this file in e.g. pl translation

$ find comprehensive-rust-* -name "hello-world.html"
comprehensive-rust-en/html/hello-world/hello-world.html
comprehensive-rust-en/html/hello-world.html
comprehensive-rust-en/html/types-and-values/hello-world.html
comprehensive-rust-en/pandoc/pdf/src/hello-world/hello-world.html
comprehensive-rust-pl/html/hello-world/hello-world.html
comprehensive-rust-pl/html/hello-world.html

using the hello-world directory redirection
https://github.com/google/comprehensive-rust/blob/main/book.toml#L193

mgeisler:
Happy New Year!

> I looked again into this and it seems I used a not ideal directory for the hello-world page as the "types-and-values" directory does not have this file in e.g. pl translation

Perhaps a good first step would be to limit these tests to the English text?

The `mdbook test` run will already execute the translated source code for each language, thus ensuring that the translations don't accidentally break the Rust code (this happened in the past).

The JavaScript tests are not there to ensure that translations compile, their goal is to ensure that the JS code works. I think it's sufficient to test with just the English base version. One day (#1390), we might do some localization with the JS code, but not today :smile:

michael-kerscher:
Yes, I agree, testing the English base version is the best approach. If there is something that depends on something language specific, this could be added later.
To document it here as well: the issue with translations that differ in directory structure is likely be caused by translations being up2date regarding the theme, js and such things but the Markdown is on the state of the last change to the translation change (this avoids having out-of-date translations that have a lot of English text). #2541 should document this behavior

mgeisler (approved):
Nice, this looks great!

mgeisler:
When we have 2-3 of these tests, I guess we can factor out the common setup code needed to interact with the Playgrounds (the clicking and entering keys could be replaced with a `runPlaygroundCode` function that handled it all).

Let's merge this now and see where people take it.

PR #2514, opened 2024-12-14
Better demonstrate imports in modules exercise

randomPoison:
The solution to the modules exercise changes how `Button`, `Window`, and `Label` are referenced, adding a `widgets::` prefix to them. This seems weird to me because the more idiomatic thing to do would be to import those types at the top of the file. Unless I'm missing a reason why the solutions is written the way it is, I think this is a good simplification.


djmitche (approved):
Good fix! Maybe that was just an oversight?

PR #2515, opened 2024-12-14
Add a slide on `match` to control flow section

randomPoison:
I find that it's hard to introduce `match` and the concept of pattern matching in the same slide (which we currently do). It tends to make the first pattern matching slide to take a very long time to get through. I think it'd be good to introduce `match` without pattern matching in the day 1 control flow section. This will make it easier to focus on pattern matching when we get to it on day 2.

Timing is a bit tight day 1 morning so I'm a little worried this just worsens the problem, but I think the amount of time it'd save on day 2 (which also is very tight when I teach) is a worthwhile tradeoff.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -30,6 +30,7 @@
     - [Solution](types-and-values/solution.md)
 - [Control Flow Basics](control-flow-basics.md)
   - [`if` Expressions](control-flow-basics/if.md)
+  - [`match` Expressions](control-flow-basics/match.md)
   - [Loops](control-flow-basics/loops.md)
     - [`for`](control-flow-basics/loops/for.md)
     - [`loop`](control-flow-basics/loops/loop.md)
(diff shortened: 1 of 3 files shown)


djmitche (commented):
Nice work on the "more to explore"!

The "Pattern Matching" slide where `match` is (re-)introduced could be slimmed down with this change, as well. At least, mark it as not requiring 10 minutes, and add some "callbacks" to the new slide on day 1.

randomPoison:
Good point! I've cleaned up the redundant info on that slide, let me know if you think it should be tweaked further.

randomPoison:
Oh, also I want to leave the later `match` slide at 10 minutes. I always takes way longer than 10 mins because I'm explaining both `match` and `patterns`. I think with this change that 10 mins estimate is more accurate.

djmitche (approved):
Keeping it at 10m makes sense, then!

PR #2517, opened 2024-12-14
Remove error handling from expression evaluation exercise

randomPoison:
I think it would be good to simplify the expression evaluation exercise by removing the error handling around the divide-by-zero case. I think it overcomplicates the exercise and and adds confusion since at this point we haven't introduced `Result` (or at least not in any detail). This allows the students to just focus on writing the pattern matches on `Expression` and `Op`, and allows the exercise to be shorter (and I think we need to free up some time where we can, my classes often run long and cut into how much time students have for exercises).

### DIFF
--- a/src/pattern-matching/exercise.md
+++ b/src/pattern-matching/exercise.md
@@ -1,5 +1,5 @@
 ---
-minutes: 30
+minutes: 20
 ---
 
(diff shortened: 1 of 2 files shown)


djmitche (approved):
This change makes a lot of sense -- thanks!

Issue #2477, opened 2024-11-28
Timeouts cause `Playground Communication: Cannot read properties of undefined (reading 'trim')`


mgeisler:
If a Playground program times out (typically due to an infinite loop), we see a strange error message:

> _Playground Communication: Cannot read properties of undefined (reading 'trim')_

We should see if we can handle this better.

Alx-Lai:
Sure! I can take a look tomorrow

Alx-Lai:
The behavior in 1. https://play.rust-lang.org/ vs 2. comprehensive-rust with following program.

```rust
fn main() {
    loop {}
}
```

1. https://play.rust-lang.org/
There's a textbox showing over runtime.
![Image](https://github.com/user-attachments/assets/73d70117-a1c5-44f3-b350-68bd52de1475)


2. comprehensive-rust
```
Playground Communication: Cannot read properties of undefined (reading 'trim')
```

mgeisler:
Hey @Alx-Lai, thanks for testing this so nicely!

I've noticed that the official Playground uses a WebSocket based system now — and the endpoint we use is a simpler one-shot approach. I would like to see us migrate to the WebSocket eventually, but for now, I think we should just wrap the code in a bit of error handling to avoid showing the ugly `Cannot read properties of undefined (reading 'trim')` message.

PR #2520, opened 2024-12-14
book.js: early return if response error

Alx-Lai:
It used to have a ugly error message shows that "Playground Communication: Cannot read properties of undefined (reading 'trim')", which is due to the fact that the response lacks stdout attr.

Issue: #2477


Alx-Lai:
Testing Results:

Normal:
image

Compiled error:
image

Timeout:
image

Connection error:
image

PR #2521, opened 2024-12-14
Rework the error handling exercise to be based on the expression evaluator exercise

randomPoison:
This PR follows up from https://github.com/google/comprehensive-rust/pull/2517, taking the error handling that was removed in that PR and reintroducing it as the error handling exercise. The core goal of the exercise is the same as before: Take some code that panics and make it return a `Result`, but I think this change is an improvement in a few ways:

* The new exercise is simpler, focusing on a single function with a single error case.
* It still requires that students both create an error and propagate it with `?`.
* It removes the need for `thiserror` as an external crate, simplifying the solution and allowing it to be tested with `mdbook test`.
* It demonstrates that an error type can be as simple as an empty struct.
* Reusing the solution from a previous exercise should reduce the upfront cognitive burden for students, hopefully meaning they can focus more on solving the solution instead of trying to wrap their head around a parser implementation.

### DIFF
--- a/src/error-handling/exercise.rs
+++ b/src/error-handling/exercise.rs
@@ -12,212 +12,98 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// ANCHOR: solution
-use thiserror::Error;
+#![allow(dead_code)]
 // ANCHOR: types
-use std::iter::Peekable;
-use std::str::Chars;
-
-/// An arithmetic operator.
-#[derive(Debug, PartialEq, Clone, Copy)]
-enum Op {
+/// An operation to perform on two subexpressions.
+#[derive(Debug)]
+enum Operation {
     Add,
     Sub,
+    Mul,
+    Div,
 }
 
-/// A token in the expression language.
-#[derive(Debug, PartialEq)]
-enum Token {
-    Number(String),
-    Identifier(String),
-    Operator(Op),
-}
-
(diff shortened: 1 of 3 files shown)


src/error-handling/exercise.rs
```diff
@@ -12,212 +12,101 @@
...
+fn eval(e: Expression) -> i64 {
+    match e {
+        Expression::Op { op, left, right } => {
+            let left = eval(*left);
+            let right = eval(*right);
+            match op {
+                Operation::Add => left + right,
+                Operation::Sub => left - right,
+                Operation::Mul => left * right,
+                Operation::Div => if right != 0 {
```
djmitche, line 53: `src/pattern-matching/exercise.rs` does not contain this conditional (it just relies on `/` to panic).

I think that's OK, but maybe claiming it's "the original" in the comment above is inaccurate enough to bother folks (like me)

djmitche (approved):
This looks great, maybe just update the comments as noted.

src/error-handling/exercise.rs
randomPoison: Yeah true it's not exactly the same as the solution to the earlier exercise, I deliberately added in the panic to show students where exactly the error case is. Personally I don't think that'd be too confusing for students, but let me add a speaker note calling that out to teachers so they can mention it if students are confused.

PR #2522, opened 2024-12-14
Add a slide about struct visibility and encapsulation

randomPoison:
The current slide on visibility introduces the idea that modules are the encapsulation boundary, but it doesn't explicitly doesn't talk about struct fields/methods. I think it's worth talking explicitly about structs and encapsulation because most students are going to be coming from a background where types (instead of modules) are the encapsulation boundary. It's worth talking explicitly about how encapsulation works, even if we're kind of repeating information from the previous slide.


djmitche (approved):
Nice!

PR #2523, opened 2024-12-16
Rework iterator section

randomPoison:
This PR attempts to improve clarity and flow in the section on iterators. I don't think the current materials do a good job of motivating iterators or putting them in a context where students can understand them. This PR does the following to try to improve this:

- **Add motivation slide**. This slide uses the example of using a C-style `for` loop to iterate over the contents of an array. This helps contextualize the concept of iteration in a way that most students will understand; Many languages have C-style `for` loops, so students are likely going to be familiar with the syntax and will have seen `for (i = 0; i < len; i++)` many times.
- **Replace fibonacci example with slice iter example**. This continues the idea from the previous slide, demonstrating how to implement the same concept (iterating over the contents of an array) using `Iterator`. I also think this example is easier to understand than the fibonacci one, I found the logic there kind of confusing/hard to explain to students. It also demonstrates an iterator with a termination case, whereas the fibonacci example never terminates.
- **Add slide for iterator helper methods**. I think splitting helper methods into its own slide gives them a bit more room to breathe. This also allows the previous slide to be more focused on implementing `Iterator`.
- **Reorder slides to talk about `collect` before `IntoIterator`**. I think this flows better after the slide on helper methods since this is just calling out a specific helper method.
- **Add speaker notes and reword things for clarity/focus**.

### DIFF
--- a/src/iterators/helpers.md
+++ b/src/iterators/helpers.md
@@ -9,7 +9,7 @@ In addition to the `next` method that defines how an iterator behaves, the
 customized iterators.
 
 ```rust,editable
-let result: i32 = (1..10) // Create a range from 1 to 9
+let result: i32 = (1..=10) // Create a range from 1 to 10
     .filter(|&x| x % 2 == 0) // Keep only even numbers
     .map(|x| x * x) // Square each number
     .sum(); // Sum up all the squared numbers
--- a/src/iterators/helpers.md
+++ b/src/iterators/helpers.md
@@ -23,7 +23,7 @@ println!("The sum of squares of even numbers from 1 to 9 is: {}", result);
   over collections (e.g. `map`, `filter`, `reduce`, etc). This is the trait
   where you can find all the documentation about them.
 
-- Many of these helper methods take the original iterator and produces a new
+- Many of these helper methods take the original iterator and produce a new
   iterator with different behavior. These are know as "iterator adapter
   methods".
 
--- a/src/iterators/helpers.md
+++ b/src/iterators/helpers.md
@@ -24,7 +24,8 @@ println!("The sum of squares of even numbers from 1 to 9 is: {}", result);
   where you can find all the documentation about them.
 
 - Many of these helper methods take the original iterator and produces a new
-  iterator with different behavior. These are know as "iterator adapter methods".
+  iterator with different behavior. These are know as "iterator adapter
+  methods".
 
 - Some methods, like `sum` and `count`, consume the iterator and pull all of the
   elements out of it.
--- a/src/iterators/helpers.md
+++ b/src/iterators/helpers.md
@@ -14,7 +14,7 @@ let result: i32 = (1..=10) // Create a range from 1 to 10
     .map(|x| x * x) // Square each number
     .sum(); // Sum up all the squared numbers
 
-println!("The sum of squares of even numbers from 1 to 9 is: {}", result);
+println!("The sum of squares of even numbers from 1 to 10 is: {}", result);
 ```
 
 <details>
--- a/src/iterators/helpers.md
+++ b/src/iterators/helpers.md
@@ -24,7 +24,7 @@ println!("The sum of squares of even numbers from 1 to 9 is: {}", result);
   where you can find all the documentation about them.
 
 - Many of these helper methods take the original iterator and produces a new
-  iterator with different behavior. These are know as "iterator adapter methods.
+  iterator with different behavior. These are know as "iterator adapter methods".
 
 - Some methods, like `sum` and `count`, consume the iterator and pull all of the
   elements out of it.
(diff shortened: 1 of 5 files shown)


src/iterators/motivation.md
```diff
@@ -0,0 +1,48 @@
...
+- A condition to determine when iteration is done.
+- Logic for updating the state of iteration each loop.
+- Logic for fetching each element using that iteration state.
+
+In a C-style for loop you declare these things directly:
+
+```c,editable
+for (int i = 0; i < array_len; i += 1) {
+    int elem = array[i];
+}
```
randomPoison, line 20: As I put this PR up it occurs to me that we don't need to use C-style `for` to demonstrate this concept, we could show the same thing in pure Rust:
```rust
let mut i = 0;
while i < array.len() {
    let elem = array[i];
    i += 1;
}
```
Might still be good to mention C-style `for` in the speaker notes, since it's still a common point of reference.

randomPoison: I kinda still feel like the C version is going to be clearer and more familiar for most students, tho

src/iterators/iterator.md
```diff
@@ -2,49 +2,63 @@
...
-    next: u32,
+struct SliceIter<'a> {
+    slice: &'a [i32],
+    i: usize,
 }
 
-impl Iterator for Fibonacci {
-    type Item = u32;
+impl<'a> Iterator for SliceIter<'a> {
+    type Item = &'a i32;
```
djmitche, line 18: Minor detail, but I'd like to get away from the idea that all lifetimes are named `'a` -- maybe use `'sl` instead?

src/iterators/iterator.md
```diff
@@ -2,49 +2,63 @@
...
+- The `SliceIter` example implements the same logic as the C-style `for` loop
+  demonstrated on the last slide.
 
-- `IntoIterator` is the trait that makes for loops work. It is implemented by
-  collection types such as `Vec<T>` and references to them such as `&Vec<T>` and
-  `&[T]`. Ranges also implement it. This is why you can iterate over a vector
-  with `for i in some_vec { .. }` but `some_vec.next()` doesn't exist.
+- Point out to the students that iterators are lazy: Creating the iterator just
+  initializes the struct but does not otherwise do any work. No work happens
+  until the `next` method is called.
```
djmitche, line 47: Worth mentioning here that iterators need not be finite.

src/iterators/helpers.md
```diff
@@ -0,0 +1,42 @@
...
+`Iterator` trait provides 70+ helper methods that can be used to build
+customized iterators.
+
+```rust,editable
+let result: i32 = (1..10) // Create a range from 1 to 9
+    .filter(|&x| x % 2 == 0) // Keep only even numbers
+    .map(|x| x * x) // Square each number
+    .sum(); // Sum up all the squared numbers
+
+println!("The sum of squares of even numbers from 1 to 9 is: {}", result);
```
djmitche, line 17:
```suggestion
println!("The sum of squares of even numbers from 1 to 10 is: {}", result);
```

src/iterators/helpers.md
```diff
@@ -0,0 +1,42 @@
...
+---
+
+# `Iterator` Helper Methods
+
+In addition to the `next` method that defines how an iterator behaves, the
+`Iterator` trait provides 70+ helper methods that can be used to build
+customized iterators.
+
+```rust,editable
+let result: i32 = (1..10) // Create a range from 1 to 9
```
djmitche, line 12:
```suggestion
let result: i32 = (1..=10) // Create a range from 1 to 10
```

djmitche (approved):
This is a great improvement to this segment!

src/iterators/iterator.md
randomPoison: How about `'s`?

src/iterators/iterator.md
randomPoison: Good point, added a speaker note about it.

src/iterators/motivation.md
randomPoison: @djmitche Do you have an opinion here? I'm leaning towards leaving this as a c-style `for` for readability and familiarity

randomPoison: You know what, I'm going to leave it as a C-style `for` but I'm going to add the rust equivalent in the speaker notes for easy reference. Let me know if you think that's a bad idea for whatever reason, though.

src/iterators/iterator.md
djmitche: `'s` is good!

src/iterators/motivation.md
djmitche: I think the C-style is fine.

PR #2531, opened 2025-01-01
Upgrade `mdbook-pandoc` to fix missing images in PDFs

max-heller:
`mdbook-pandoc` 0.8 now supports `` elements in raw HTML, fixing a few missing images in the PDF versions of the course.

### DIFF
--- a/.github/workflows/install-mdbook/action.yml
+++ b/.github/workflows/install-mdbook/action.yml
@@ -17,11 +17,11 @@ runs:
 
     - name: Install mdbook-pandoc and related dependencies
       run: |
-        cargo install mdbook-pandoc --locked --version 0.7.0
+        cargo install mdbook-pandoc --locked --version 0.8.0
         sudo apt-get update
         sudo apt-get install -y texlive texlive-luatex texlive-lang-cjk librsvg2-bin fonts-noto
-        curl -LsSf https://github.com/jgm/pandoc/releases/download/3.2.1/pandoc-3.2.1-linux-amd64.tar.gz | tar zxf -
-        echo "$PWD/pandoc-3.2.1/bin" >> $GITHUB_PATH
+        curl -LsSf https://github.com/jgm/pandoc/releases/download/3.6.1/pandoc-3.6.1-linux-amd64.tar.gz | tar zxf -
+        echo "$PWD/pandoc-3.6.1/bin" >> $GITHUB_PATH
       shell: bash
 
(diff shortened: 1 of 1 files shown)


mgeisler (approved):
Thanks so much, happy new year!

mgeisler:
Oh, the Persian translation seems to have a problem with this version. The check says:

```
  [2025-01-02T08:50:02Z INFO  mdbook_pandoc::pandoc::renderer] Running pandoc
  Error producing PDF.
  ! Undefined control sequence.
  l.279 ...اوقات زبان Rust را با C و \LR
  
  pandoc exited unsuccessfully
  Error: -02 08:50:31 [ERROR] (mdbook::renderer): Renderer exited with non-zero return code.
  Error: -02 08:50:31 [ERROR] (mdbook::utils): Error: Rendering failed
  Error: -02 08:50:31 [ERROR] (mdbook::utils): 	Caused By: The "pandoc" renderer failed
```

mgeisler:
> Oh, the Persian translation seems to have a problem

Hmm... I wonder why this could merge? Perhaps we're not blocking on that check like I hoped?

mgeisler:
> > Oh, the Persian translation seems to have a problem
> 
> Hmm... I wonder why this could merge? Perhaps we're not blocking on that check like I hoped?

Indeed... I have to add each language individually to the branch protection rules :smile: I guess we could restructure this with a CI job that waits on every translation and only succeeds when they all succeed.

For now, I guess we can back this out and then fix this later.

max-heller:
> > > Oh, the Persian translation seems to have a problem
> > 
> > 
> > Hmm... I wonder why this could merge? Perhaps we're not blocking on that check like I hoped?
> 
> Indeed... I have to add each language individually to the branch protection rules 😄 I guess we could restructure this with a CI job that waits on every translation and only succeeds when they all succeed.
> 
> For now, I guess we can back this out and then fix this later.

Ah sorry, didn't realize this had merged! I'm in the midst of a pretty big rework so I probably won't get around to fixing whatever is up here until that's complete, reverting for now makes sense.

mgeisler:
> Ah sorry, didn't realize this had merged!

No problem at all, I had also not noticed it until today :smile:

> I'm in the midst of a pretty big rework so I probably won't get around to fixing whatever is up here until that's complete, reverting for now makes sense.

Alright, that sounds very cool! I'm of course happy to wait for this!

PR #2537, opened 2025-01-03
Use clamp in bare-metal compass solution

hurryabit:
The custom function `cap` does the same as `Ord::clamp`, which was introduced in Rust 1.50. Let's use the latter instead.

I've flashed the new program onto my microbit and can confirm it still works as intended.


djmitche (approved):

mgeisler:
The build error was fixed by #2540.

PR #2540, opened 2025-01-04
Revert "Upgrade `mdbook-pandoc` to fix missing images in PDFs"

mgeisler:
Reverts google/comprehensive-rust#2531

This accidentally broke the build because the Persian translation doesn't build with this version.

### DIFF
--- a/.github/workflows/install-mdbook/action.yml
+++ b/.github/workflows/install-mdbook/action.yml
@@ -17,11 +17,11 @@ runs:
 
     - name: Install mdbook-pandoc and related dependencies
       run: |
-        cargo install mdbook-pandoc --locked --version 0.8.0
+        cargo install mdbook-pandoc --locked --version 0.7.0
         sudo apt-get update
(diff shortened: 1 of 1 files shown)


mgeisler:
Cc @max-heller. I invited you to the repository in case you want to help review PRs like this :smile:

max-heller:
> Reverts #2531
> 
> This accidentally broke the build because the Persian translation doesn't build with this version.

Given that I upgraded both `mdbook-pandoc` and `pandoc`, I'm curious which it is that caused the breakage.

mgeisler:
> Given that I upgraded both `mdbook-pandoc` and `pandoc`, I'm curious which it is that caused the breakage.

Yeah, good question! I must admit that I didn't investigate this further :) If you approve the PR, then it should merge right away since the tests all pass here.

max-heller (approved):

mgeisler:
Thanks!

PR #2544, opened 2025-01-12
Fix missing images and Arabic fonts in course PDFs

max-heller:
Let's give #2531 another try
- Fixes missing images in PDFs (those specified as  elements)
- Configures Arabic fonts

[The fix](https://github.com/max-heller/mdbook-pandoc/pull/144/files#diff-b17b0f99aa22ee1ccf99778ca74c3c7a27af87f41544194cc95f82c012e8a507) for the Persian translation issue, if you're curious.


mgeisler (approved):
Amazing, thanks for fixing this!

PR #2547, opened 2025-01-14
Fix a few accessibility violations

henrif75:
* Add chrome.css and general.css to override default css file. This is necessary to fix links without underline (hyperlinks relying only on color).
* Fix pop-out button without id and wrong ARIA-ROLE
* Speaker's notes now at correct heading sequence (H3)

### DIFF
--- a/theme/css/general.css
+++ b/theme/css/general.css
@@ -1,59 +1,80 @@
 /* Base styles and content styles */
 
 :root {
-    /* Browser default font-size is 16px, this way 1 rem = 10px */
-    font-size: 62.5%;
-    color-scheme: var(--color-scheme);
+  /* Browser default font-size is 16px, this way 1 rem = 10px */
+  font-size: 62.5%;
+  color-scheme: var(--color-scheme);
 }
 
 html {
-    font-family: "Open Sans", sans-serif;
-    color: var(--fg);
-    background-color: var(--bg);
-    text-size-adjust: none;
-    -webkit-text-size-adjust: none;
+  font-family: "Open Sans", sans-serif;
+  color: var(--fg);
+  background-color: var(--bg);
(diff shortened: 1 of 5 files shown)


djmitche (approved):
I was surprised to find that general.css and chrome.css are already referenced in index.hbs.

theme/css/general.css
```diff
@@ -0,0 +1,279 @@
+/* Base styles and content styles */
```
mgeisler, line 1: It's strange that we've been missing it all along :smile: 

My understanding is that `mdbook build` will pull in theme files from https://github.com/rust-lang/mdBook/tree/master/src/theme/css, unless the local theme has a file with the same name.

But perhaps I've misunderstood this all along?

theme/index.hbs
```diff
@@ -289,7 +289,6 @@
 
                 <!-- Apply ARIA attributes after the sidebar and the sidebar toggle button are added to the DOM -->
                 <script>
-                    document.getElementById('sidebar-toggle').setAttribute('aria-expanded', sidebar === 'visible');
```
mgeisler, line 292: Does this change perhaps belong in https://github.com/rust-lang/mdBook/blob/master/src/theme/index.hbs?

mgeisler:
Thanks so much for looking at this, Henri!

henrif75:
> > I was surprised to find that general.css and chrome.css are already referenced in index.hbs.
> 
> I thought missing files were pulled in automatically from the basic theme. Loading the page with the inspector open shows
> 
> image
> 
> So I think the files were copied correctly already.

FWIW, if any of the CSS files (e.g., `general.css` , `chrome.css`) is not available at the source when `mdbook` runs, it will use the default files and copy them to the target folder.

Issue #2549, opened 2025-01-14
JS tests might be flaky


mgeisler:
Oh, this fails because of an unrelated timeout:

```
[0-1] Error in "Playground.executes the hello world code and prints the hello message"
Error: Expect  to have text

Expected: StringContaining "🌍"
Received: "The operation timed out: deadline has elapsed"
    at Context.<anonymous> (/home/runner/work/comprehensive-rust/comprehensive-rust/tests/src/playground.test.ts:[30](https://github.com/google/comprehensive-rust/actions/runs/12767680980/job/35586546795?pr=2548#step:15:31):37)
[0-1] Error in "Playground.shows error messages in stderr if the code is broken"
Error: Expect $(`code.result.stderr`) to be displayed
```

The new JS test is from #2513 (cc @michael-kerscher). We might need to disable it for now if it turns out to be flaky.

_Originally posted by @mgeisler in https://github.com/google/comprehensive-rust/issues/2548#issuecomment-2590451225_

max-heller:
[Getting the same error](https://github.com/google/comprehensive-rust/actions/runs/12771966158/job/35601489041#step:15:33) on my latest commit merged to main

mgeisler:
Thanks @max-heller, let's disable this test for now!

michael-kerscher:
I'm not entirely sure yet how to proceed from here. I would like to test the playground, as this is an important bit of the course book - but I also don't want to have flaky tests that don't have anything to do with regular commits.
There would be some options though that I wanted to document and bring up for discussion:
- increase timeouts for tests and/or some retry behavior (in production not test code)
- detect if relevant files have been changed (e.g. filter for templates and javascript only). A change in the markdown part or translations should not modify the behavior of the html files. This limits impact of broken tests to relevant changes only.
- mock the playground server https://webdriver.io/docs/mocksandspies/ in such a way that we don't need to rely on external dependencies
Example on how to do this:
```
    // mock the response so this test does not rely on the playground backend to be working
    const playground_mock = await browser.mock(
      "https://play.rust-lang.org/execute",
      { method: "post" }
    );

    // be aware that there is a still a preflight OPTIONS request
    playground_mock.respond(
      {
        success: true,
        exitDetail: "Exited with status 0",
        stdout: "Hello 🌍!\n",
        stderr:
[...]

PR #2551, opened 2025-01-15
Remove flaky JS test

mgeisler:
This should stop the test failures, but of course it won’t fix the underlying problem.

Reverts part of #2513. Related to #2549.


michael-kerscher (approved):
If tests are flaky, they should not be used in this configuration. I'll have another look on how we can improve this before reintroducing that test again

PR #2553, opened 2025-01-16
Update timings for first 'references' segment

djmitche:
Based on a course session just completed. The instruction itself took less time, but the exercise took longer.


mgeisler (approved):
Thanks for tracking this!

PR #2555, opened 2025-01-16
Remove speaker note on "runtime evaluated constants"

djmitche:
It's unclear what this would mean! It was introduced in 89ddb2c19.


mgeisler (approved):
Sure, happy to trim this. I think it refers to a constant with a `OnceCell`.

djmitche:
Oh, I will make a note to add some content on that topic.

PR #2557, opened 2025-01-16
Make const slide less silly

djmitche:
A constant named ZERO that does not contain zero seems pretty silly!

This also shows an example of a const fn.


mgeisler (approved):

mgeisler:
Silly indeed! :smile:

PR #2563, opened 2025-01-17
Fix minor inconsistencies and naming issues

hurryabit:


djmitche (approved):
Thanks!

I think that health-statistics exercise needs to go, but this is strictly better.

PR #2565, opened 2025-01-18
Fix incorrectly hidden lines in Rust code blocks in PDFs

max-heller:
Upgrades `mdbook-pandoc` to pull in a fix relating to hidden lines in Rust code blocks. The bug was causing lines like `#[test]` to be hidden when they should not have been.

## HTML
Screenshot 2025-01-18 at 12 50 26 PM

## PDF (before)
Screenshot 2025-01-18 at 12 51 25 PM

## PDF (after)
Screenshot 2025-01-18 at 12 52 25 PM

### DIFF
--- a/.github/workflows/install-mdbook/action.yml
+++ b/.github/workflows/install-mdbook/action.yml
@@ -17,7 +17,7 @@ runs:
 
     - name: Install mdbook-pandoc and related dependencies
       run: |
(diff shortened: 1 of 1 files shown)


djmitche:
Thank you!

djmitche (approved):

PR #2567, opened 2025-01-19
Split let control flow into mutliple sub-slides

djmitche:
There are three kinds of syntax here, making for a very long and hard-to-navigate slide. Splitting it up helps!


src/pattern-matching/let-control-flow/let-else.md
```diff
@@ -0,0 +1,57 @@
...
+
+```rust,editable
+fn hex_or_die_trying(maybe_string: Option<String>) -> Result<u32, String> {
+    // TODO: The structure of this code is difficult to follow -- rewrite it with let-else!
+    if let Some(s) = maybe_string {
+        if let Some(first_byte_char) = s.chars().next() {
+            if let Some(digit) = first_byte_char.to_digit(16) {
+                Ok(digit)
+            } else {
+                return Err(String::from("not a hex digit"));
```
hurryabit, line 16: While we're at this: Is the `return` here very idiomatic? It's basically in tail position, so just having `Err(...)` would seem better to me.

hurryabit (changes requested):
Yeah, that's a good idea.

src/pattern-matching/let-control-flow/let-else.md
andriyDev: While I agree this was not idiomatic, the code on this section is intended to be rewritten to use let-else. Therefore it might actually be desirable to keep the returns so that presenters can just copy-paste the returns into the let-else rather than having to add them manually.

djmitche: I think that's why they existed. @hurryabit what are your thoughts?

hurryabit: IMO, there's more value if the attendees get to compare two idiomatic versions. In fact, having to write a little more slows down things a bit and gives people more time to digest what we're teaching them. I would think that the slowdown introduced by having to write `return` three times is very reasonable.

All that said, I'd prefer to start with the idiomatic version.

djmitche: I think that's what the current version of the PR is -- do you want to mark it approved?

hurryabit (approved):
I thought I already had approved. Here we go.

djmitche:
danke!

PR #2568, opened 2025-01-19
Be clear that the methods-and-traits exercise does not require generics

djmitche:
When teaching the course, I got a little tripped up thinking students would need to make the `VerbosityFilter` generic over `Logger`. Let's be clearer that this is not required, and will be described later.

This also updates the generic-types slide to repeat the exercise, completing that thought.


hurryabit (approved):
I like how we build on things learnt earlier with this approach. Nice!

PR #2569, opened 2025-01-19
Add empty structs

djmitche:
This should be quick, but introduces the syntax and the concept of a ZST.


hurryabit (commented):
I'm not convinced this is super useful. I think the concept is so simple that we can explain it in passing when we need it for the first time. Adding an additional slide here costs more time and probably increases increments mental load without much benefit.

hurryabit (approved):
That works for me. Thanks for being so understanding. :)

PR #2570, opened 2025-01-19
Include the From trait in the generic traits slide

djmitche:
This saves a bunch of tabbing back and forth from the docs to the slide.


hurryabit (approved):
This should make our life easier. Thank you.

PR #2571, opened 2025-01-19
Allow the 'unused' category of lints

djmitche:
These sort of warnings can be distracting when commenting out a few lines of code or demonstrating some other concept. They can be re-enabled for a code block with `warnunused`.

I filed https://github.com/rust-lang/mdBook/issues/2527 to get behavior like this upstream.

### DIFF
--- a/theme/book.js
+++ b/theme/book.js
@@ -127,6 +127,11 @@ function playground_text(playground, hidden = true) {
 
         let text = playground_text(code_block);
         let classes = code_block.querySelector('code').classList;
+        // Unless the code block has `warnunused`, allow all "unused" lints to avoid cluttering
+        // the output.
+        if(!classes.contains("warnunused")) {
+            text = '#![allow(unused)]\n' + text;
+        }
         let edition = "2015";
         if(classes.contains("edition2018")) {
             edition = "2018";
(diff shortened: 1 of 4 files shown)


hurryabit (approved):
Good idea. Thanks.

theme/book.js
```diff
@@ -127,6 +127,11 @@ function playground_text(playground, hidden = true) {
 
         let text = playground_text(code_block);
         let classes = code_block.querySelector('code').classList;
+        // Unless the code block has `warnunused`, allow all "unused" lints to avoid cluttering
+        // the output.
+        if(!classes.contains("warnunused")) {
+            text = '#![allow(unused)]\n' + text;
+        }
```
mgeisler, line 134: The idea here is cool, and I like @max-heller's comment a lot on https://github.com/rust-lang/mdBook/issues/2527#issuecomment-2602904546.

However, it should be noted that every modification to `book.js` required work when we update `mdbook.` There is unfortunately a coupling between versions of `mdbook` and the default theme, which is what we've customized here :grimacing: I now created #2924 to record this problem (which has been with us for ages).

Cc @gribozavr for reference.

PR #2573, opened 2025-01-19
Split interior mutability, mention OnceCell/OnceLock

djmitche:
These types are really only useful as a static or in a user-defined type, neither of which are covered at this point.


hurryabit (approved):
Makes sense. Thank you.

PR #2574, opened 2025-01-19
Break closures into its own segment

djmitche:
In teaching the course last week, we broke here, partly due to time constraints, but partly because this is a pretty mind-bending topic to tackle at the end of an information-dense day. A break helps, and spreading the content over a few slides helps as well.

By the timings in the course, this leaves day 2 looking like

*Fundamentals // Day 2 Morning*
_1 hour and 55 minutes: (1 hour and 10 minutes short)_

* Welcome - _3 minutes_
* Pattern Matching - _45 minutes_
* Methods and Traits - _50 minutes_

*Fundamentals // Day 2 Afternoon*
_3 hours and 30 minutes (⏰ *30 minutes too long*)_

* Welcome - _0 minutes_
* Generics - _45 minutes_
* Standard Library Types - _1 hour_
* Standard Library Traits - _1 hour_
* Closures - _20 minutes_

Maybe we should move generics to the morning session?

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -92,10 +92,6 @@
   - [Deriving](methods-and-traits/deriving.md)
   - [Exercise: Generic Logger](methods-and-traits/exercise.md)
     - [Solution](methods-and-traits/solution.md)
-
-# Day 2: Afternoon
-
-- [Welcome](welcome-day-2-afternoon.md)
 - [Generics](generics.md)
   - [Generic Functions](generics/generic-functions.md)
   - [Generic Data Types](generics/generic-data.md)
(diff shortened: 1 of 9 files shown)


hurryabit (commented):
Should we move generics into the morning session?

djmitche:
Uh, I asked first :)

fw-immunant:
As a data point, when teaching I usually end up breaking after standard library types (which is quite a slog, because I do think it worthwhile to dig into the particular APIs provided by collection types, but it can get a bit dry) and before standard library traits.

hurryabit (approved):
Generics in the morning session look nice. Thanks.

djmitche:
> As a data point, when teaching I usually end up breaking after standard library types (which is quite a slog, because I do think it worthwhile to dig into the particular APIs provided by collection types, but it can get a bit dry) and before standard library traits.

The design is to break between each segment, so that is an expected break :)

PR #2576, opened 2025-01-20
Add return statements to the Fibonacci exercise.

andriyDev:
At this point in the course, we have not explained return statements. Better to have it set up to avoid questions!


djmitche (approved):
Good thinking -- I think this is more idiomatic anyway!

PR #2577, opened 2025-01-21
Add a "while let" title to separate "let else" from "while let" stuff.

andriyDev:


djmitche:
I split these into sub-slides in https://github.com/google/comprehensive-rust/pull/2567 -- is that a better solution?

andriyDev:
Yup that's looks better!

Issue #2541, opened 2025-01-07
Document translation workflow


michael-kerscher:
The translation workflow needs documentation. Example: What is the commit that is used for translations (markdown content vs. other things like themes, JS)

PR #2579, opened 2025-01-21
Document translation workflow

mgeisler:
This is a first draft of explaining the documentation workflow. Let's discuss this a bit before merging.

### DIFF
--- a/TRANSLATIONS.md
+++ b/TRANSLATIONS.md
@@ -215,6 +215,68 @@ the hard work, even if it is incomplete.
 
 [CODEOWNERS]: https://github.com/google/comprehensive-rust/blob/main/.github/CODEOWNERS
 
+## Publication Workflow
+
+> This section is for the developers of Comprehensive Rust, but it might give
+> you valuable background information on how the translations are published.
+
+When a change is made to the `main` branch, the [`publish.yml`] GitHub CI
+workflow starts.
+
+The `publish` job in this workflow will:
+
+- Install dependencies as described in [`CONTRIBUTING`](CONTRIBUTING.md).
+
+- Build every translation of the course, including the original English, using
+  [`build.sh`]. The English HTML ends up in `book/html/`, the HTML for the `xx`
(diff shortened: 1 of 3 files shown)


TRANSLATIONS.md
```diff
@@ -215,6 +215,54 @@ the hard work, even if it is incomplete.
...
+
+- Build every translation of the course, including the original English, using
+  [`build.sh`](https://github.com/google/comprehensive-rust/blob/main/.github/workflows/build.sh).
+  The English HTML ends up in `book/html/`, the HTML for the `xx` language ends
+  up in `book/xx/html/`.
+
+- Publish the entire `book/html/` directory to
+  https://google.github.io/comprehensive-rust/.
+
+### `build.sh`
```
michael-kerscher, line 239: I think we should mention when this script is being called (publish.yml and build.yml) and that both times there are artifacts that can be inspected if necessary.

TRANSLATIONS.md
```diff
@@ -215,6 +215,54 @@ the hard work, even if it is incomplete.
...
+The `build.sh` script is used both when testing code from a PR and when
+publishing the finished book.
+
+The job of the script is to call `mdbook build`, but with a few extra steps:
+
+- It will enable the PDF output using `mdbook-pandoc`. This is disabled by
+  default to make it easier for people to run `mdbook build` without having to
+  configure LaTeX.
+
+#### Restoring Translations
```
michael-kerscher, line 250: We should include that the English translation does not involve the git restore step and always builds from HEAD

michael-kerscher:
This PR is relevant for #2541

michael-kerscher (changes requested):
This looks good overall but I would like to have the two comments addressed before merging this

TRANSLATIONS.md
mgeisler: Great idea, done!

michael-kerscher (approved):

PR #2581, opened 2025-01-22
Briefly touch on match ergonomics

djmitche:
This is done in the speaker notes as it's a relatively minor point, but one that students should have in the back of their mind when they wonder, "hey, how does a `&Foo` match against `Foo` patterns??"


hurryabit (approved):
I like that!

PR #2582, opened 2025-01-22
Improve tuple destructuring

djmitche:
This slide had two code samples, neither of which had a `main` and thus neither of which would run. This removes the first (which is redundant to one a few slides earlier), adds a `main`, and expands the second to use a 3-tuple.


hurryabit (approved):
Good idea. Thank you.

fw-immunant (approved):
LGTM as well.

PR #2583, opened 2025-01-22
Add a picture to Rc

djmitche:
I suppose my `svgbob` skills leave a bit to be desired, but I think the meaning is clear:

image

Now that I look through the `Rc` implementation, there's a weak count for every strong count, so the `weak: 0` here is inaccurate. But, maybe this is too much of an implementation detail? Should I just concentrate on strong refs? I suppose I could put a `...` in that upper-right box, to suggest there's more going on here?

### DIFF
--- a/src/smart-pointers/rc.md
+++ b/src/smart-pointers/rc.md
@@ -19,6 +19,22 @@ fn main() {
 }
 ```
 
+Each `Rc` points to the same shared data structure, containing strong and weak
+pointers and the value:
+
+```bob
+ Stack                     Heap
+.- - - - - - - -.     .- - - - - - - - - - - - - - - - -.
+:               :     :                                 :
+:     +-----+   :     :   +-----------+-------------+   :
+:  a: | o---|---:--+--:-->|  count: 2 |  value: 10  |   :
+:     +-----+   :  |  :   +-----------+-------------+   :
+:  b: | o---|---:--+  :                                 :
+:     +-----+   :     `- - - - - - - - - - - - - - - - -'
+:               :     
+`- - - - - - - -'
+```
+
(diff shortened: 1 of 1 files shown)


src/smart-pointers/rc.md
```diff
@@ -19,6 +19,22 @@ fn main() {
...
 
+Each `Rc` points to the same shared data structure, containing strong and weak
+pointers and the value:
+
+```bob
+ Stack                     Heap
+.- - - - - - - -.     .- - - - - - - - - - - - - - - - -.
+:               :     :                                 :
+:     +-----+   :     :   +-------------+-----------+   :
+:  a: | o---|---:-----:-->|  strong: 2  |  weak: 0  |   :
```
hurryabit, line 30: Why not make it `weak: 2` and say that every strong reference is also considered a weak reference. Or just leave this slight "lie". I don't think people taking this course need this level of detail to be correct.

hurryabit (approved):
Very nice. Thank you.

src/smart-pointers/rc.md
djmitche: I just omitted the "weak" bit -- that can remain an implementation detail :)

src/smart-pointers/rc.md
```diff
@@ -19,6 +19,22 @@ fn main() {
...
+:     +-----+   :     :   +-------------+-----------+   :
```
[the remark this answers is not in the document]
djmitche: Good idea - that is an ugly cross! But I tweaked it a bit -- WDYT?

image

fw-immunant:
Thanks for this change--I was teaching this slide recently and wished I had a diagram instead of describing the memory layout of `Rc` only verbally.

hurryabit (approved):
I think this is a good level of detail. Thank you.

PR #2584, opened 2025-01-22
Improve explanation of lifetimes

djmitche:
This approach seems to balance formalism with understanding. That is, it doesn't mention contravariance, but suggests that lifetime annotations in parameters and return values mean "opposite" things. It also leverages the understanding that types must be specified in function signatures, and are used to check types in the function body and at call sites.

### DIFF
--- a/src/lifetimes/lifetime-annotations.md
+++ b/src/lifetimes/lifetime-annotations.md
@@ -12,12 +12,9 @@ also be explicit: `&'a Point`, `&'document str`. Lifetimes start with `'` and
 `'a` is a typical default name. Read `&'a Point` as "a borrowed `Point` which is
 valid for at least the lifetime `a`".
 
-Lifetimes are always inferred by the compiler: you cannot assign a lifetime
-yourself. Explicit lifetime annotations create constraints where there is
-ambiguity; the compiler verifies that there is a valid solution.
-
-Lifetimes become more complicated when considering passing values to and
-returning values from functions.
+Explicit lifetime annotations, like types, are required on function signatures
+(but can be elided in common cases). These provide information for inference at
+callsites and within the function body.
 
 <!-- The multi-line formatting by rustfmt in left_most is apparently
      intentional: https://github.com/rust-lang/rustfmt/issues/1908 -->
@@ -56,9 +53,11 @@ Add `'a` appropriately to `left_most`:
 fn left_most<'a>(p1: &'a Point, p2: &'a Point) -> &'a Point {
 ```
 
-This says, "given p1 and p2 which both outlive `'a`, the return value lives for
-at least `'a`.
+This says there is some lifetime `'a` which both `p1` and `p2` outlive, and
+which outlives the return value. The borrow checker verifies this within the
+function body, and uses this information in `main` to determine a lifetime for
+`p3`.
 
-In common cases, lifetimes can be elided, as described on the next slide.
+Try dropping `p2` in `main` before printing `p3`.
 
 </details>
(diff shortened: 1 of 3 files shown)


src/lifetimes/struct-lifetimes.md
```diff
@@ -2,7 +2,7 @@
 minutes: 5
 ---
 
-# Lifetimes in Data Structures
+# Struct Lifetimes
```
hurryabit, line 5: Is this the right title? This also works on enums. Maybe something like "Lifetimes in User Defined Types"?

hurryabit: Actually, during the course I thought that using a struct with a field named `document` (rather than a newtype) would make the explanations somewhat more natural. What do you think?

hurryabit (changes requested):
I like that. Some ideas for making it even better.

src/lifetimes/struct-lifetimes.md
djmitche: > Is this the right title? This also works on enums. Maybe something like "Lifetimes in User Defined Types"?

I'll change it to "Lifetimes in Data Structures" which just barely fits in the sidebar.

> Actually, during the course I thought that using a struct with a field named `document` (rather than a newtype) would make the explanations somewhat more natural. What do you think?

That's fixed in https://github.com/google/comprehensive-rust/pull/2585.

src/lifetimes/lifetime-annotations.md
```diff
@@ -12,12 +12,9 @@ also be explicit: `&'a Point`, `&'document str`. Lifetimes start with `'` and
 `'a` is a typical default name. Read `&'a Point` as "a borrowed `Point` which is
 valid for at least the lifetime `a`".
 
-Lifetimes are always inferred by the compiler: you cannot assign a lifetime
-yourself. Explicit lifetime annotations create constraints where there is
-ambiguity; the compiler verifies that there is a valid solution.
-
-Lifetimes become more complicated when considering passing values to and
-returning values from functions.
```
fw-immunant, line 20: I think this removed text is load-bearing in that it counteracts a common misconception that lifetimes annotations can alter the runtime behavior of programs (in terms of how long values exist and when they are destroyed). We don't need to keep this exact phrasing but in addition to describing lifetimes as constraints we should have some explicit indication that only ownership, not lifetime annotations, control when objects are destroyed and determine the concrete lifetime of a given value.

fw-immunant (changes requested):
See inline comment.

src/lifetimes/lifetime-annotations.md
djmitche: Thanks! What do you think of the update?

fw-immunant: "The borrow checker just ensures this is done safely." implies that the borrow checker is validating ownership rather than borrowing; I think it would be better to say that "The borrow checker just validates that borrows never extend beyond the concrete lifetime of the value."

I also think it may be better to say "values" rather than "objects". While "objects" is perfectly correct, many students come from languages which make a distinction between objects and other kinds of values and may wrongly infer that the borrow checker only applies to non-primitives.

djmitche:
@fw-immunant or @hurryabit PTAL

fw-immunant:
I'd still like to see [this comment](https://github.com/google/comprehensive-rust/pull/2584#discussion_r1931010740) addressed.

djmitche:
Addressed -- sorry about that!

fw-immunant (approved):
LGTM now.

PR #2585, opened 2025-01-22
Clarify struct-lifetimes slide

djmitche:
In teaching the course, the verbal distinction between "doc" and "dog" was not clear, so this PR moves away from those symbols.

This also makes the Highlight struct a little more substantial, and replaces `erase` with a simple call to `drop` to keep the example short.


fw-immunant:
I would suggest `document` rather than `text` as the lifetime name just to demonstrate that it does not need to match the variable, and to keep the intuition that the lifetime refers to the entire source string rather than just the portion within the slice. But this is really a bikeshed.

hurryabit (approved):
I like this. Thank you!

PR #2589, opened 2025-01-26
Move the "Trait Bounds" slide right after "Generic Functions".

andriyDev:
The "Generic Data Types" slide now uses trait bounds, which makes it very confusing to explain. (e.g., "ignore this trait bound stuff while I explain why we need to stutter to describe the generic args on an impl block").

Also in the "Generic Functions" slide, the speaker notes talks about how we essentially need to treat the generic args as black boxes - this is probably more important to address then talking about all the ways you can make things generic.


randomPoison (approved):
I think this is a good change. Even before the changes to the generic data types slide I had been thinking it'd be better to cover trait bounds first and then go over generic data types and generic trait. I think this is a good change for the flow of the slides.

fw-immunant:
Agreed; students are often champing at the bit to ask how they can limit generics to only certain sets of types, and usually I have to tell them to wait a few slides before we get there.

PR #2592, opened 2025-01-30
Change values of example array in for loop slide

randomPoison:
The original example was mildly confusing since both the range-based and array-based loops were printing out the same sequence of numbers. I figure it'd be a bit more clear that the array-based loop is iterating over the elements of the array if the values are different.


djmitche (approved):
Good idea :)

PR #2594, opened 2025-01-30
Replace `unimplemented` with `todo`

randomPoison:
At some point we started using `todo` instead of `unimplemented`, but a couple places still use the old macro. May as well normalize on `todo` everywhere, since `unimplemented` is baaaaaaasically deprecated.

### DIFF
--- a/src/iterators/exercise.md
+++ b/src/iterators/exercise.md
@@ -13,7 +13,7 @@ return value.
 
 ```rust
 {{#include exercise.rs:offset_differences}}
-    unimplemented!()
+    todo!()
 }
 
 {{#include exercise.rs:unit-tests}}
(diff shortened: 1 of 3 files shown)


fw-immunant:
It's fine to make things consistent everywhere, but since when is `unimplemented!` on the way out? I thought it was for functionality that isn't and shouldn't be implemented.

randomPoison:
Oh fair point, I didn't think about `unimplemented` meaning "deliberately unimplemented"

> The difference between `unimplemented!` and [`todo!`](https://doc.rust-lang.org/stable/std/macro.todo.html) is that while todo! conveys an intent of implementing the functionality later and the message is “not yet implemented”, unimplemented! makes no such claims. Its message is “not implemented”.

So `todo!` still probably makes more sense for us to use in the exercise code, but `unimplemented` isn't deprecated in the way I thought.

djmitche (approved):
Consistency is good! Both macros are mentioned in the macros page, so I don't think this is losing anything.

randomPoison:
> Both macros are mentioned in the macros page

tbh I kinda want to remove `unimplemented!` from the macros slide as well since we never end up using it in the course. I guess I'll throw up a PR for that so we can discuss.

PR #2595, opened 2025-01-30
Minor whitespace changes

randomPoison:
* Add extra lines in a couple places to make code examples more readable. In both these cases I'm trying to break the code into more logical groups, generally some statement followed by a `println` showing the result of that operation.
* Split the first example of struct initialization across a couple of lines for readability. I personally struggle to read that statement when it's all one line, so I expect it'd be hard for students, and I always break it up like this in class.

### DIFF
--- a/src/references/shared.md
+++ b/src/references/shared.md
@@ -14,8 +14,10 @@ the referenced data cannot change.
 fn main() {
     let a = 'A';
     let b = 'B';
+
     let mut r: &char = &a;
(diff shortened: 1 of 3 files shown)


randomPoison:
oh wait, this formatting is enforced by the formatter. How do I tell it to ignore this block?

randomPoison:
Ah, found it

djmitche (approved):

PR #2596, opened 2025-01-30
Remove slide on shadowing

randomPoison:
I don't think we need to cover shadowing. I wouldn't say it's an *advanced* concept, but I think it's somewhat niche. It's a useful thing to know about the language, but it's also used fairly infrequently (in my experience). I think it just takes too long to explain for how rarely it is used, since I also have to explain why it's useful (students tend to assume this is an anti-pattern). I don't think we lose much by dropping this, and I think the time we gain back is more valuable.

The shadowing slide was technically the first place we talk about scoping in Rust, so I've moved that to the previous slide on blocks. I think it fits there a bit more naturally. We do briefly cover the topic in the slide on ownership, but I think it's also good to mention it earlier.


djmitche (approved):
I do see shadowing quite frequently IRL, and in fact it's such a habit I often try to do it in C++! But I agree that this is a good bit to trim.

randomPoison:
Yeah that's fair, it's not _that_ rare, but it's unwieldly to explain in my experience, and not terribly important in my opinion. I figure students can grok it when they see it happening in the wild.

PR #2598, opened 2025-01-30
Move slide on blocks to beginning of control flow section

randomPoison:
I think it'd be good to move the slides on blocks and scopes to the beginning of the control flow section. This allows us to introduce a couple of core concepts that are common to all control flow concepts (returning values from blocks and scoping rules). I think this'll then make it easier for students to understand how to return values from the other control flow constructs. I think the way we currently introduce it with `if`/`else` is harder for students to understand, and I think the slides on blocks demonstrates it better.

This PR currently looks like I'm moving two slides to the beginning of the section, but I also have https://github.com/google/comprehensive-rust/pull/2596 up which proposes to remove the shadowing slide, so my goal is to have just a single slide on blocks at the beginning of the control flow section.


djmitche (approved):
Good idea!

PR #2603, opened 2025-01-30
Remove trait bound on `VerbosityFilter` in the generic data types slide

randomPoison:
It's generally more idiomatic in Rust to not have trait bounds on the data type itself. I think we better demonstrate how trait bounds are used in impl blocks with generic data types in the impl block below the struct definition. I've also added a speaker note to call this out if students ask.


djmitche (approved):
I don't agree with this idiom, but you're right that it's idiomatic :)

PR #2604, opened 2025-01-31
Add "More to Explore" section to struct patterns slide

randomPoison:
Talking about matching on references and interactions with constants are useful to note but not super important imo. Breaking those notes into the "More to Explore" section would at least help me remember to only go into that if students ask or there's enough extra time to cover them.

### DIFF
--- a/src/pattern-matching/destructuring-structs.md
+++ b/src/pattern-matching/destructuring-structs.md
@@ -14,6 +14,9 @@ Like tuples, Struct can also be destructured by matching:
 
 - Change the literal values in `foo` to match with the other patterns.
 - Add a new field to `Foo` and make changes to the pattern as needed.
+
+## More to Explore
+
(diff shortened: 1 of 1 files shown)


djmitche (approved):
I agree about the constants, but it seems like match ergonomics is a subtle syntactic thing that can be confusing if you're not aware of it:
```
match ref_to_a_foo {
  Foo(x) => ..
}
```
it's not at all obvious that `x` is a reference in this context.

randomPoison:
I agree match ergonomics is a bit subtle and could definitely be a point of confusion for new programmers, but I think it's also easier to understand once you better understand ownership and references. I find explaining it here to be distracting from the more core point about how patterns work, since that's the main thing students are trying to grok at this point.

djmitche:
Good point!

PR #2605, opened 2025-01-31
Add division example to expression exercise

randomPoison:
Students pointed out that the solution to the expression evaluation exercise was producing a warning because the `Div` variant was never constructed. This fixes the warning.


djmitche (approved):

djmitche:
(apologies for the delay reviewing all these PRs -- they went into my to-do list and took a while to get to the top)

randomPoison:
No worries about the delay, none of these are urgent changes :)

PR #2607, opened 2025-01-31
Add speaker note with manual derive example

randomPoison:
I often end up typing this example out manually to help motivate using `derive` instead of a manual implementation. I don't think it needs to be in the main slide, but having the example at hand would save a minute or two of awkward typing.


djmitche (approved):
Awesome!

PR #2608, opened 2025-01-31
Add slide reviewing irrefutable patterns

randomPoison:
We already talk about irrefutable patterns in day 1, but only briefly. I always review the concept at the beginning of the pattern matching section to refresh students' memories, so this is just formalizing that review into a slide at the beginning of the section. I also use this opportunity to introduce `_` and `..`, which should help students understand that patterns can do more than just directly destructure values.

### DIFF
--- a/src/pattern-matching/infallible.md
+++ b/src/pattern-matching/infallible.md
@@ -34,6 +34,9 @@ fn main() {
 - All of the demonstrated patterns are _irrefutable_, meaning that they will
   always match the value on the right hand side.
 
+- Patterns are type-specific, including irrefutable patterns. Try adding or
+  removing an element to the tuple and look at the resulting compiler errors.
+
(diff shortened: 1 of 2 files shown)


src/pattern-matching/infallible.md
```diff
@@ -0,0 +1,63 @@
...
+
+fn main() {
+    takes_tuple(('a', 777, true));
+}
+```
+
+<details>
+
+- All of the demonstrated patterns are _irrefutable_, meaning that they will
+  always match the value on the right hand side.
```
fw-immunant, line 35: I would say something here about how irrefutable patterns are still type-specific, e.g. `(_, _)` only matches pairs and `()` only matches the unit type.

randomPoison: Good point, I added a note about that.

djmitche (approved):

PR #2610, opened 2025-01-31
update to mdbook-0.4.44 and sync book.js and index.hbs

michael-kerscher:
mdbook in CI pipeline is updated to mdbook-0.4.44
  - reran mdbook init --theme
  - keep playground improvements
  - keep language selector and suggest edit / edit to translation button text
  - improvement: toc is in separate js file, makes html file way smaller
  - additionally updated to mdbook-i18n-helpers-0.3.5

relevant upstream changes that are used:
- https://github.com/rust-lang/mdBook/pull/2414
- https://github.com/rust-lang/mdBook/pull/2421
- https://github.com/rust-lang/mdBook/pull/2454
- https://github.com/rust-lang/mdBook/pull/2463

### DIFF
--- a/install-mdbook.sh
+++ b/install-mdbook.sh
@@ -1,10 +1,10 @@
 #!/bin/bash
 # The --locked flag is important for reproducible builds. It also
 # avoids breakage due to skews between mdbook and mdbook-svgbob.
-cargo install mdbook --locked --version 0.4.37
+cargo install mdbook --locked --version 0.4.44
 cargo install mdbook-svgbob --locked --version 0.2.1
 cargo install mdbook-pandoc --locked --version 0.9.3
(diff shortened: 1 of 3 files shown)


michael-kerscher:
This is more difficult and needs some additional work. Cargo install --locked does not use the Cargo.lock from this repository but the one of the lastest version of the crate - so this is not what we want (for now).

There will be a separate work item for creating an install script with locked versions to replicate what we have in the CI environment so developers can use the same versions. Then we can update the install script and update everything.

michael-kerscher:
The output of the updated versions looks good to me and is ready for a review.

djmitche:
Tested this myself and it seems to be in good shape!

This doen't fix the issue of the speaker notes "reverting" to show the slides themselves, but we can address that separately.

djmitche (approved):

PR #2623, opened 2025-02-03
use Cargo.lock version for mdbook in the CI

### LINKED ISSUE #2620: Gitlab CI and Cargo.lock get out of sync
The Gitlab CI workflows use specific versions for mdbook tools that are different from what developers are using. E.g. mdbook is currently using 0.4.37:
https://github.com/google/comprehensive-rust/blob/4218c95e8dfa996a21069e7cb8c30d64f530ba5a/.github/workflows/install-mdbook/action.yml#L11

but Cargo.lock would use 0.4.44 - this is regularly bumped via dependabot like in https://github.com/google/comprehensive-rust/commit/010bd291c00c42c133b0043f6caa93e3e7c1a622
https://github.com/google/comprehensive-rust/blob/4218c95e8dfa996a21069e7cb8c30d64f530ba5a/Cargo.lock#L1640-L1641

and the developers are even instructed to use a completely different version that might not be one of those as it is not even locked (other tools are locked and might not be compatible with the "random" (most up-to-date) version of mdbook
https://github.com/google/comprehensive-rust/blob/4218c95e8dfa996a21069e7cb8c30d64f530ba5a/README.md?plain=1#L74-L80

We should use the same versions for developers and what the CI uses.

This is especially interesting for people developing javascript/theme code as mdbook generates/copies lots of files into the target folder if they are not overriden by us.

For example our [book.js](https://github.com/google/comprehensive-rust/blob/4218c95e8dfa996a21069e7cb8c30d64f530ba5a/theme/book.js) is copied but https://google.github.io/comprehensive-rust/css/general.css is not in our repository [not existing general.css](https://github.com/google/comprehensive-rust/blob/4218c95e8dfa996a21069e7cb8c30d64f530ba5a/theme/css/general.css)

Two options (second is my prefered proposal)
- Keep manual overrides for the Gitlab CI: Dev and CI are not in sync and might cause issues when someone changes the javascript code or relies on its functionality (when using mdbook serve - as in https://github.com/google/comprehensive-rust/issues/2588
- Fix the instruction to install the --locked version of mdbook and use this version in the Gitlab CI as well. Caveat: Dependabot updates need to be reviewed carefully and as we override some things (e.g. [index.hbs](https://github.com/google/comprehensive-rust/blob/4218c95e8dfa996a21069e7cb8c30d64f530ba5a/theme/index.hbs)) we need to carefully track updates to the official files with each update.
  - To make this update process reliable, write tests for the things we rely on in [tests/src](https://github.com/google/comprehensive-rust/tree/main/tests/src) so things like missing theme buttons or search buttons are visible immediately.

michael-kerscher:
This fixes #2620 and forces usage of the versions specified in Cargo.lock

The benefit would be to have the same versions in the CI that the developers use on their systems

What are the reasons for the manually locked versions in the CI?
Discussion might be necessary for how to best solve the issue for why this was locked.
We could configure dependabot to ignore mdbook related updates so we can update these versions manually.

### DIFF
--- a/tests/src/objects/playground.ts
+++ b/tests/src/objects/playground.ts
@@ -1,0 +1,106 @@
+import { $, browser } from "@wdio/globals";
+
+export default class Playground {
+  /**
+   * stores a playground mock
+   */
+  mock: WebdriverIO.Mock;
+
+  get area() {
+    return $(".ace_content");
+  }
+  get start_button() {
+    return $("button.play-button");
+  }
+  get stderr() {
+    return $("code.result.stderr");
+  }
+  get stdout() {
+    return $("code.result.stdout");
+  }
+
+  /**
+   * activate the playground by clicking into it
+   */
+  async activate() {
+    // clicking into the content is necessary for the button to be displayed
+    await this.area.waitForClickable();
+    await this.area.click();
+  }
+
+  /**
+   *  run the code in the playground editor
+   */
+  async run() {
+    // clicking the button triggers action
+    await this.start_button.waitForClickable();
+    // await browser.debug();
+    await this.start_button.click();
+  }
+
+  /**
+   * generate a new mock for the playground that overrides playground requests
+   */
+  async mock_new() {
+    return browser.mock("https://play.rust-lang.org/execute", {
+      method: "post",
+    });
+  }
+
+  /**
+   * reset the mock object
+   */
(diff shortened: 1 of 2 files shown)


michael-kerscher:
This is not working in the current way. The issue #2620 needs some more discussion on the best approach

PR #2626, opened 2025-02-05
Create install-mdbook.sh and lock to the current versions used in the CI

### LINKED ISSUE #2588: Incorrect behavior with `mdbook serve` compared to published version.
Using `mdbook serve` locally (following the instructions in the README) is quite different from the version on `https://google.github.io/comprehensive-rust`.

1. There's no theme or search button.
2. Clicking on a lesson (e.g., Control Flow Basics > break and continue) does not "expand" the appropriate section in the side bar.
3. Links don't "reset" correctly. For example, viewing Control Flow Basics > break and continue, then clicking Tuples and Arrays in the sidebar results in a Document not found page (it is trying to access `localhost:3000/control-flow-basics/tuples-and-arrays.html`).

This was on Chrome. I also tried it on Firefox (with no extensions) to the same end result. I am serving from a Linux laptop.

michael-kerscher:
Move mdbook installation into a script and use exact versions from the CI.
Update README.md to instruct developers to use the same versions as the CI to sync both environments.
This is related to #2620 and it fixes #2588

### DIFF
--- a/.github/workflows/install-mdbook/action.yml
+++ b/.github/workflows/install-mdbook/action.yml
@@ -5,37 +5,14 @@ description: Install mdbook with the dependencies we need.
 runs:
   using: composite
   steps:
-    # The --locked flag is important for reproducible builds. It also
-    # avoids breakage due to skews between mdbook and mdbook-svgbob.
     - name: Install mdbook
-      run: cargo install mdbook --locked --version 0.4.37
+      run: ./install-mdbook.sh
       shell: bash
 
-    - name: Install mdbook-svgbob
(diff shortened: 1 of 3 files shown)


michael-kerscher:
Should I move the install script into some subfolder to have the root more clean? I cannot find a suitable existing folder for this. I could create a scripts/ folder but this would for now only contain this file

djmitche:
> Should I move the install script into some subfolder to have the root more clean? I cannot find a suitable existing folder for this. I could create a scripts/ folder but this would for now only contain this file

I don't think this is necessary!

djmitche (approved):
Looks good!

PR #2628, opened 2025-02-06
Remove `unreachable!` from macros slide

randomPoison:
After #2594 we no longer use `unimplemented!` in the course, so I don't think we need to mention it in the macros slide.

### DIFF
--- a/src/control-flow-basics/macros.md
+++ b/src/control-flow-basics/macros.md
@@ -44,4 +44,19 @@ especially critical.
 The course does not cover defining macros, but a later section will describe use
 of derive macros.
 
+## More To Explore
+
+There are a number of other useful macros provided by the standard library. Some
+other examples you can share with students if they want to know more:
+
+- [`assert!`] and related macros can be used to add assertions to your code.
+  These are used heavily in writing tests.
+- [`unreachable!`] is used to mark a branch of control flow that should never be
+  hit.
+- [`eprintln!`] allows you to print to stderr.
+
 </details>
+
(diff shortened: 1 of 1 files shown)


djmitche:
I do see it frequently in the wild, but I suppose it's easy enough for a student to look it up in the docs when they encounter it. What do you think @fw-immunant?

fw-immunant:
I think `unreachable!` is important to teach because Rust behaves differently here (by panicking safely) than C++'s [`unreachable`](https://en.cppreference.com/w/c/program/unreachable) which is an optimization hint and UB to execute.

randomPoison:
> Is this about `unreachable!` or `unimplemented!`?

Oh, I completely spaced and thought I was removing `unimplemented!`, but it is, in fact, `unreachable!` 😅. I'm still inclined to remove it since I don't think it adds much useful information for the students and, as @djmitche noted, it's easy to look up when encountered in the wild.

@fw-immunant would it be okay to move this into "More to Explore" in the speaker notes? There are other macros in std that could also be worth adding there as extra things to cover, time allowing.

mgeisler:
> @fw-immunant would it be okay to move this into "More to Explore" in the speaker notes? There are other macros in std that could also be worth adding there as extra things to cover, time allowing.

I like this idea: my gut feeling is that we have too much stuff overall, so anything non-essential that we can trim is a win. Adding a paragraph saying that there are more macros available would be useful.

randomPoison:
I've added that note to the speaker notes along with a couple other macros that can be brought up in class if students want to know more.

djmitche (approved):
Looks good - now it just focuses on the macros used in the course!

PR #2629, opened 2025-02-07
Add note that borrock error can be triggered by direct mutation

randomPoison:
This is a minor thing that I usually point out to students on this slide.


djmitche (approved):
I assume `borrock` is a typo of `borrowchecker`?

randomPoison:
Lol that was supposed to be "borrowck" which is (I think) a shortening of "borrow checker" that people use.

PR #2630, opened 2025-02-07
Add `main` method to code snippet

randomPoison:
The code snippet wouldn't compile/run directly because of a missing `main` method.

Also remove unnecessary `&`.


michael-kerscher (approved):
lgtm

PR #2631, opened 2025-02-07
Move `use` statements to beginning of `widgets.rs`

randomPoison:
I think the convention is to put `use` statements (including `pub use`) towards the top of the file under the `mod` statements.

### DIFF
--- a/src/modules/solution.md
+++ b/src/modules/solution.md
@@ -28,13 +28,13 @@ src
 
 ```rust,ignore
 // ---- src/widgets.rs ----
-mod button;
-mod label;
-mod window;
-
 pub use button::Button;
 pub use label::Label;
 pub use window::Window;
+
+mod button;
+mod label;
+mod window;
 
 pub trait Widget {
     /// Natural width of `self`.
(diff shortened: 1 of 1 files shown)


michael-kerscher:
I usually put `mod` after `use` so I tried to find some official guidance here and found this in the rust style guide.
https://doc.rust-lang.org/style-guide/items.html#:~:text=Put%20imports%20before%20module%20declarations
> Put imports before module declarations

randomPoison:
Oh you're absolutely right. I spaced on that because they're `pub use` statements, but they also go at the top generally.

michael-kerscher:
I noticed that we have other slides showing what you first attempted - and probably we should follow one style in this.

michael-kerscher:
This example uses use then mod:
- https://doc.rust-lang.org/reference/items/use-declarations.html#r-items.use.visibility.unambiguous
- https://doc.rust-lang.org/reference/items/use-declarations.html#r-items.use.as

These examples use mod then use:
- https://doc.rust-lang.org/reference/items/use-declarations.html#r-items.use.path.namespace

michael-kerscher (approved):
I'm going with the official rust style guide here and approve it in this order

randomPoison:
Yeah I think sticking with the style guide is the right move 👍

PR #2633, opened 2025-02-07
Add link to `GridRefIter` example

randomPoison:
One of the speaker notes on the `IntoIterator` slide suggests implementing a version of `GridIter` that iterates by reference. I find that this is a bit too cumbersome to do live in class, so I think linking to a complete solution in the playground is a better way to walk through the final version.


djmitche (approved):
Oh, I like the pattern of linking to a playground from speaker notes!

Issue #2004, opened 2024-04-18
Moving too quickly between pages on the popped-out "speaker notes" window resets to non-notes view


fw-immunant:
To reproduce, tap left and right a couple times while the speaker notes window is focused after popping it out into its own window. It will revert to showing a second copy of the course material, as if the speaker notes window had been closed.

mgeisler:
Thanks for logging this! The speaker notes system is quite rudimentary and it could benefit from someone with better JavaScript skills than me giving it an overhaul.

randomPoison:
I've found that navigating at all in the speaker notes immediately breaks the connection between the speaker notes and the main window. I'm assuming this is the same issue that @fw-immunant ran into here.

djmitche:
I've seen that as well.

michael-kerscher:
I can also reproduce this reliably by clicking into the notes and go to the right. The speaker notes view is now turning into the regular slide view".
Moving left does break the connection to the regular slides but stays in the speaker notes view - it is independent from the regular slides though

Interesting behavior: moving left or right has a different effect

michael-kerscher:
I found the issue with the "next" navigation and created a pull request.

There is an additional issue causing the break:
https://github.com/google/comprehensive-rust/blob/c05f0b6f023bf39b8d7f96bfa63e449d010eeb51/theme/speaker-notes.js#L119-L123

This triggers on any unload of the page - which means prev+next navigation events.

There is a workaround that tries to avoid acting on this, but depending on the the sequence and timing of events this still might break https://github.com/google/comprehensive-rust/blob/c05f0b6f023bf39b8d7f96bfa63e449d010eeb51/theme/speaker-notes.js#L209-L226

michael-kerscher:
To add some thoughts on this: We need some mechanism for the slides to discover that the speaker notes are closed. Currently this is done by signaling back to the slides when the speaker notes close (with the `beforeunload` event - that triggers for any unload, be it window close or some full navigation from the current page).

Options that might be viable and need some discussion:
[...]

PR #2634, opened 2025-02-07
Fix bug in speaker notes that broke navigating to next slide

michael-kerscher:
Change this selector to use the ~= selector to test if a white space separated word "prev" or "next" is contained

Fixes a speaker notes bug that did not allow going to the next slide in the speaker notes.
The reason for that is that the "rel" attribute contained "prev" respective "next prefetch".

See: https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors#attrvalue_2

This fixes part of #2004 when going to the right (containing "prefetch").

### DIFF
--- a/theme/speaker-notes.js
+++ b/theme/speaker-notes.js
@@ -189,7 +189,7 @@
 
     // Update prev/next buttons to keep speaker note state.
     document
-      .querySelectorAll('a[rel="prev"], a[rel="next"]')
+      .querySelectorAll('a[rel~="prev"], a[rel~="next"]')
       .forEach((elem) => {
         elem.href += "#speaker-notes-open";
       });


djmitche (approved):
Nice work!!

randomPoison:
What behavior exactly is this fixing? I tried it locally and navigating to the right still breaks the speaker notes (speaker notes become disconnected from the main slides, and the speaker note section re-appears at the bottom of the regular notes). Was there a different issue that this fixes?

michael-kerscher:
I'm aware of the second issue and that is some more difficult communication issue between the windows - I described the root cause that I see in https://github.com/google/comprehensive-rust/issues/2004#issuecomment-2643325284

michael-kerscher:
@randomPoison This change is fixing only the issue that the hash "#speaker-notes-open" was not appended to the urls when navigating to "next" thus the next speaker note slide did not know it was a speaker note window

PR #2637, opened 2025-02-11
Use `jni_libs` for JNI dependency

randomPoison:
It looks like the module type used for adding a JNI dependency to a Java binary has changed to `jni_libs`. This gets our various Android examples building again, since currently the incorrect dependency causes Soong errors even if you're trying to build a different module.


mgeisler (approved):
Thanks for fixing this! I don't know what changed, but it's super nice that you update this regularly as you teach the course.

PR #2638, opened 2025-02-12
Fix vendored crates path

randomPoison:
In Android vendored crates now live under `external/rust/android-crates-io/crates`.


mgeisler (approved):
Awesome!

PR #2639, opened 2025-02-12
Update AIDL intermediates path

randomPoison:
The path to the generated Rust code for AIDL definitions seems to have changed, this updates the slide to reflect the current location. This also assumes that the CompRust repo is checked out at the root of an AOSP checkout.

### DIFF
--- a/src/android/aidl/example-service/service-bindings.md
+++ b/src/android/aidl/example-service/service-bindings.md
@@ -9,7 +9,7 @@ _birthday_service/aidl/com/example/birthdayservice/IBirthdayService.aidl_:
 }
 ```
 
-_out/soong/.intermediates/.../birthdayservice/IBirthdayService.rs_:
+_out/soong/.intermediates/comprehensive-rust/.../com_example_birthdayservice.rs_:
 
 <!-- The example below is a cleaned up and simplified version of the real code. -->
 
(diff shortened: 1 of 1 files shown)


mgeisler:
> This also assumes that the CompRust repo is checked out at the root of an AOSP checkout.

Yeah, I left that part out in the `...` — since the path is long and ugly, I figured we should not try to show it in full.

src/android/aidl/example-service/service-bindings.md
```diff
@@ -9,7 +9,7 @@ _birthday_service/aidl/com/example/birthdayservice/IBirthdayService.aidl_:
 }
 ```
 
-_out/soong/.intermediates/.../birthdayservice/IBirthdayService.rs_:
+_out/soong/.intermediates/comprehensive-rust/.../com_example_birthdayservice.rs_:
```
mgeisler, line 12: Can we do this to hint that "you find `com_example_birthdayservice.rs` somewhere in the `.intermediates` directory"?

```suggestion
_out/soong/.intermediates/.../com_example_birthdayservice.rs_:
```

mgeisler (approved):
Thanks for updating the path — my instinct would be show as little as possible.

When teaching this, I've used `fd IBirthdayService.rs soong/.intermediates` a few times to quickly find the file. That is basically why the path is shown (plus to be consistent with the other code blocks that show file content).

PR #2640, opened 2025-02-12
Add missing `static_libs` dep for libbirthday example

randomPoison:
Turns out it's not enough to just depend on the bindgen module, we also need a direct `static_libs` dependency on the C module.


mgeisler (approved):
Alright, thanks for testing this!

PR #2641, opened 2025-02-12
Add note about AIDL methods taking `&self`

randomPoison:
This is useful context that I always try to explain to students, so having a speaker note would be helpful.


mgeisler (approved):
Excellent!

PR #2642, opened 2025-02-12
Reorder CXX build steps to show genrules first

randomPoison:
The genrules are inputs to the `cc_library_static`, which we're currently showing first. I think showing the genrules first, then showing how they're used in `cc_library_static`, makes more sense (and that's how I always teach it in class).


mgeisler (approved):
Sure!

Issue #2398, opened 2024-10-07
Android section is untested


mgeisler:
We need to figure out a way to test the code in the Android section. I'm teaching the class right now, and I've discovered several places where the code has drifted from AOSP in such a way that it no longer compiles.

mgeisler:
While fixing this, I realized that we should change how we store the Android examples: if we could include them somewhere in AOSP, we would be forced to keep them current as the platform evolves.

We would then pull in the relevant pages and examples as part of a build process — or perhaps we could simply include them via a Git sub-module? I don't have a lot of experience with those, but since AOSP is a large collection of Git repositories, it seems reasonable to assume that we could pull in _one_ of these repositories as a sub-module.

PR #2643, opened 2025-02-13
Add `Android.bp` entries for googletest and mockall example tests

randomPoison:
This makes the googletest and mockall examples buildable and runnable in an AOSP checkout. To make this work I had to slightly modify the examples to fix linter warnings that were resulting in errors in the build.

### DIFF
--- a/src/android/testing/Android.bp
+++ b/src/android/testing/Android.bp
@@ -11,3 +11,19 @@ rust_test {
     host_supported: true,
     test_suites: ["general-tests"],
 }
+
+rust_test {
+    name: "libgoogletest_example",
+    crate_name: "googletest_example",
+    srcs: ["googletest.rs"],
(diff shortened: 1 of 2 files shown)


mgeisler (approved):
Thank you!

Do you think you could spend a few afternoons on #2398? I'm basically suggesting moving the code here into AOSP itself. It should be possible to put it somewhere near the other Android Rust documentation. We then need to check this bit out when building teh book, possibly via a Git submodule.

randomPoison:
> Do you think you could spend a few afternoons on #2398?

Yup, I plan on doing that. Putting the code in AOSP and then pulling it in via a submodule sounds like a reasonable approach, I've just got to look into how we get a new repo added to AOSP and then pull out the code. Not sure when I'll get to it but hopefully soon 🤞

PR #2644, opened 2025-02-14
Be more consistent about tests vs. main

### LINKED ISSUE #1581: Exercise consistency: tests or main? and allow(unused)

djmitche:
The content slides all use `fn main`, with the exception of the testing segment. But with this change, where it makes sense exercises use tests instead, and not both tests and `fn main`.

A small change in `book.js` supports running tests when a code sample does not have `fn main` but does have `#[test]`, so these work naturally. 

Fixes #1581.

### DIFF
--- a/src/control-flow-basics/exercise.rs
+++ b/src/control-flow-basics/exercise.rs
@@ -25,15 +25,14 @@ fn collatz_length(mut n: i32) -> u32 {
     len
 }
 
-// ANCHOR: tests
+// ANCHOR: main
+fn main() {
+    println!("Length: {}", collatz_length(11)); // should be 15
+}
+// ANCHOR_END: main
+// ANCHOR_END: solution
+
 #[test]
 fn test_collatz_length() {
     assert_eq!(collatz_length(11), 15);
 }
-// ANCHOR_END: tests
-
-// ANCHOR: main
-fn main() {
-    println!("Length: {}", collatz_length(11));
-}
-// ANCHOR_END: main
--- a/src/error-handling/exercise.md
+++ b/src/error-handling/exercise.md
@@ -10,7 +10,7 @@ Rewrite `eval` to instead use idiomatic error handling to handle this error case
 and return an error when it occurs. We provide a simple `DivideByZeroError` type
 to use as the error type for `eval`.
 
-```rust
+```rust,editable
 {{#include exercise.rs:types}}
 
 {{#include exercise.rs:eval}}
--- a/src/error-handling/exercise.md
+++ b/src/error-handling/exercise.md
@@ -10,7 +10,7 @@ Rewrite `eval` to instead use idiomatic error handling to handle this error case
 and return an error when it occurs. We provide a simple `DivideByZeroError` type
 to use as the error type for `eval`.
 
-```rust,editable
+```rust
 {{#include exercise.rs:types}}
 
 {{#include exercise.rs:eval}}
--- a/src/error-handling/exercise.rs
+++ b/src/error-handling/exercise.rs
@@ -88,25 +88,30 @@ fn eval(e: Expression) -> Result<i64, DivideByZeroError> {
 // ANCHOR_END: solution
 
 // ANCHOR: tests
-#[test]
-fn test_error() {
-    assert_eq!(
-        eval(Expression::Op {
-            op: Operation::Div,
-            left: Box::new(Expression::Value(99)),
-            right: Box::new(Expression::Value(0)),
-        }),
-        Err(DivideByZeroError)
-    );
-}
+#[cfg(test)]
+mod test {
+    use super::*;
 
-fn main() {
-    let expr = Expression::Op {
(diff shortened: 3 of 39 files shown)


src/control-flow-basics/exercise.rs
```diff
@@ -25,15 +25,14 @@ fn collatz_length(mut n: i32) -> u32 {
...
-    println!("Length: {}", collatz_length(11));
+    println!("Length: {}", collatz_length(11)); // should be 15
 }
 // ANCHOR_END: main
+// ANCHOR_END: solution
+
+#[test]
+fn test_collatz_length() {
+    assert_eq!(collatz_length(11), 15);
+}
```
randomPoison, line 38: This is excluding the test from the starting code and solution, was that intended? I thought the goal was to normalize tests on using tests instead of `main`.

src/error-handling/exercise.rs
```diff
@@ -88,25 +88,30 @@ fn eval(e: Expression) -> Result<i64, DivideByZeroError> {
...
-        eval(Expression::Op {
-            op: Operation::Div,
-            left: Box::new(Expression::Value(99)),
-            right: Box::new(Expression::Value(0)),
-        }),
-        Err(DivideByZeroError)
-    );
-}
+#[cfg(test)]
+mod test {
```
randomPoison, line 92: I don't think we need to setup the full `mod test` for the exercise tests. It's a best practice in actual Rust projects, but I think it just adds visual noise that is distracting for students. I think it's cleaner to just see the list of tests in the same module. Also we don't bother setting up a `mod test` for most of the other exercises, so this is a bit inconsistent with our other tests.

src/error-handling/exercise.md
```diff
@@ -10,7 +10,7 @@ Rewrite `eval` to instead use idiomatic error handling to handle this error case
 and return an error when it occurs. We provide a simple `DivideByZeroError` type
 to use as the error type for `eval`.
 
-```rust,editable
```
randomPoison, line 13: I'd prefer that we don't remove `editable` from the exercise sources, it can be useful to make small edits to them in place to demonstrate things to students.

What's the reason for removing these (both here and for the other exercises where you made the same change)? Are we just trying to prevent students from using the exercise slide to do the exercise (i.e. because if students navigate away from the slide they'll loose their changes)?

src/tuples-and-arrays/exercise.md
```diff
@@ -27,14 +27,9 @@ Copy the code below to <https://play.rust-lang.org/> and implement the function.
...
 
 ```rust,should_panic
-
 {{#include exercise.rs:transpose}}
     todo!()
 }
 
-{{#include exercise.rs:tests}}
```
randomPoison, line 37: Why do we not want to include the test here?

src/tuples-and-arrays/exercise.rs
```diff
@@ -25,7 +25,21 @@ fn transpose(matrix: [[i32; 3]; 3]) -> [[i32; 3]; 3] {
...
+
+    println!("matrix: {:#?}", matrix);
+    let transposed = transpose(matrix);
+    println!("transposed: {:#?}", transposed);
+}
+// ANCHOR_END: main
+// ANCHOR_END: solution
+
 #[test]
 fn test_transpose() {
```
randomPoison, line 44: Why are we excluding the test from the solution?

randomPoison (changes requested):

src/error-handling/exercise.rs
djmitche: I didn't in most exercises, but this is the segment after the "testing" segment (and the second-to-last segment, the last being "Unsafe Rust" which isn't amenable to tests), so it seems a good place to continue to follow the best practice. So, I'd like to leave this as-is.

src/error-handling/exercise.md
djmitche: Yeah, it's to prevent students from editing in-place, but your motivation is good.

djmitche:
OK, I've reviewed all of the exercises, and made a few changes. I've also added a section in CONTRIBUTING.md regarding the choice of main or tests.

Here's the summary:

 * Types and Values - main
 * Control Flow Basics - main
 * Tuples and Arrays - main
 * References - main
 * User-Defined Types - main
 * Pattern Matching - tests (with brief introduction)
 * Methods and Traits - main (uses stdout)
 * Generics - tests
 * Standard Library Types - main (uses stdout)
 * Standard Library Traits - tests
 * Memory Management - main (asserting properites of the resulting object wouldn't be very useful)
 * Smart Pointers - tests
 * Borrowing - tests
 * Lifetimes - tests
 * Iterators - tests
 * Modules - main (challenge is not testing-related)
 * Testing - tests
 * Error Handling - tests
 * Unsafe Rust - main (makes sense)

randomPoison (approved):

Issue #2445, opened 2024-10-18
Overlong slides should be split up or reworked into a smaller slide


michael-kerscher:
The original issue https://github.com/google/comprehensive-rust/issues/1464 provides arguments on why scrolling on slides should be avoided.
To find slides that are interesting to rework I created https://github.com/google/comprehensive-rust/issues/2234 and wrote a small tool to evaluate the slides.

I now did a first run of the new tool and these are the longest pages where the main content element is bigger than 1333 pixels (16:9 ratio if 750 pixels is the assumed width)

https://google.github.io/comprehensive-rust/running-the-course/day-4.html 3508
https://google.github.io/comprehensive-rust/android/interoperability/with-c/bindgen.html 2117
https://google.github.io/comprehensive-rust/unsafe/writing-unsafe-functions.html 1445
https://google.github.io/comprehensive-rust/unsafe/unsafe-functions.html 1445
https://google.github.io/comprehensive-rust/unsafe-rust/unsafe-functions.html 1445
https://google.github.io/comprehensive-rust/unsafe/extern-functions.html 1445
https://google.github.io/comprehensive-rust/unsafe/calling-unsafe-functions.html 1445
https://google.github.io/comprehensive-rust/bare-metal/aps/entry-point.html 1392
https://google.github.io/comprehensive-rust/pattern-matching/let-control-flow.html 1384
https://google.github.io/comprehensive-rust/control-flow/while-let-expressions.html 1384
https://google.github.io/comprehensive-rust/android/interoperability/java.html 1376
https://google.github.io/comprehensive-rust/android/interoperability/with-c/rust.html 1358
https://google.github.io/comprehensive-rust/index.html 1339

This is how I created the result. Filter the output for exercises, print summary, etc. only output MaxHeight violations, only output height in pixel, sort by height descending and create a link

```
$ python ./.github/workflows/run-evaluator.py book/comprehensive-rust-en/ violations-en.csv
$ grep -Po "(?<=html/).*MaxHeight" violations-en.csv | grep -v -e print.html -e exercises/ -e exercise.html -e solution.html -e glossary.html -e structure.html | sort -h -t, -k 2 -r | sed -e "s|^|https://google.github.io/comprehensive-rust/|" | cut -d, -f 1,3 | tr ',' ' '
```

PR #2645, opened 2025-02-14
Split C interop slides into smaller slides.

egithinji:
Part of #2445.

This splits up the slides in the interoperability with C section into smaller slides. 

For calling C from Rust, instead of the long slide containing the sample C lib, explaining bindgen, and how to run the Rust code, these are now presented on separate slides. 

Similarly for calling Rust from C, the simple Rust library is on its own slide, followed by a slide on how to then call and run the Rust library from C.


src/android/interoperability/with-c/c-library.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+Create a wrapper header file for the library (not strictly needed in this
+example):
+
+_interoperability/bindgen/libbirthday_wrapper.h_:
+
+```c
+{{#include bindgen/libbirthday_wrapper.h:include}}
+```
```
randomPoison, line 32: This should be moved to the next slide about using bindgen. This is part of the bindgen setup, not the setup for the C library.

randomPoison (changes requested):
Thanks for breaking up these big slides! This should help the flow when going through this in class. I just have one request to move content around a little, but otherwise these changes look good.

src/android/interoperability/with-c/c-library.md
egithinji: Thanks for reviewing! I've made the adjustment.

randomPoison (approved):

randomPoison:
Everything looks good now, thanks for the contribution! ❤️

Issue #2478, opened 2024-11-28
Use `dbg!` instead of `println!` for simple debug prints


mgeisler:
Originally, we did not show `stderr` output from the Playground programs. This was fixed in #2397 and so we can finally start converting the trivial debug print statements to use `dbg!`.

Specifically, I'm thinking of the many print statements that look like this:

```rust
println!("a: {a:?}");
```

They can simply become

```rust
dbg!(a);
```

When doing this conversion, pay attention to any `<!-- i18n:skip -->` lines before the code block. They were added to skip translation of the `"a: {a:?}"` string, but now that the string disappears, the skip directive should be cleaned up too.

egithinji:
Hi @mgeisler , I can look into this.

PR #2652, opened 2025-02-21
Use dbg! instead of println! in Day 1 mng session

egithinji:
As mentioned in #2478, this cleans up the code blocks when all that is needed is a trivial debug print statement.
Only making changes to Day 1 morning session so that I can get feedback before proceeding with the rest of the course.


egithinji:
> Looks like it needs a `dprint fmt` run before it can be merged..

Done. Thank you for the review and also for the line numbering fix!

PR #2653, opened 2025-02-21
Preserve line numbers by not injecting a newline

djmitche:
The `dbg!` macro, being deployed more widely in #2478, shows the line number. But if we inject `#![allow(..)]\n` then the printed numbers do not match those in the textarea. It turns out that `\n` is not required!

cc @egithinji


randomPoison (approved):
Ah thanks for fixing this, students have gotten confused by this before.

PR #2655, opened 2025-02-23
Use chopsticks and limit philosophers.

scubed2:
Use chopstick to explain why 2 are needed to eat.
Limit async to 2 philosophers so they can deadlock in tokio.
(Tested with [3, 4, 5] philosophers and they all were able to run without deadlock
with lock ordering disabled.)


djmitche (approved):
Thanks for this!

djmitche:
I admit I always sort of wondered why they needed two forks ;)

PR #2657, opened 2025-02-25
Use dbg! instead of println! in Day 2. 

egithinji:
Part of #2478 to clean up code blocks when all that is needed is a trivial debug print statement.

As mentioned in previous related PRs, in some places I've opted to retain the use of println! because dbg! makes it less readable.


djmitche (approved):
Thanks!!

djmitche:
I appreciate the careful thought :P)

PR #2661, opened 2025-02-27
Add speaker note about returning values from `loop`

randomPoison:
I like to mention this because it's a fun curiosity. It's probably not worth putting in the main slide, but I think a speaker note would be nice.


djmitche (commented):
This is already mentioned in https://google.github.io/comprehensive-rust/control-flow-basics/break-continue.html

randomPoison:
Oh good point I missed that.

PR #2663, opened 2025-02-27
Add note that slices can't grow

randomPoison:
This came up in class recently, and I've gotten questions about it a few times, so seems like it deserves a speaker note.


djmitche (approved):
Good thinking -- especially for people coming from Go and the arcane horrors of `append()`, this might be confusing!

PR #2665, opened 2025-02-27
Add speaker note about `for` and array refs

randomPoison:
Students have pointed this out before so I figure a speaker note would be a good reminder to cover it.

### DIFF
--- a/src/references/solution.md
+++ b/src/references/solution.md
@@ -3,3 +3,11 @@
 ```rust,editable
 {{#include exercise.rs:solution}}
 ```
+
+<details>
+
+- Note that in `normalize` we were able to do `*item /= mag` to modify each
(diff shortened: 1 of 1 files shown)


src/references/solution.md
```diff
@@ -3,3 +3,11 @@
 ```rust,editable
 {{#include exercise.rs:solution}}
 ```
+
+<details>
+
+- Note that in `normalize` we were able to do `*item /= mag` to modify each
+  element. This is because we're iterating using a reference to an array, which
```
djmitche, line 10: Critically, it's an exclusive reference.

randomPoison: Ah good point, I updated to call that out explicitly.

djmitche (approved):

PR #2669, opened 2025-02-27
Use dbg! instead of println! in day 3 & 4.

egithinji:
Part of #2478 to clean up code blocks when all that is needed is a trivial debug print statement.


djmitche (approved):
Nice, thanks!

PR #2671, opened 2025-02-28
Rework generic fn examples to show monomorphized versions

randomPoison:
Something I always do when covering generic fns is I like to show the monomorphized versions of `pick` to make it clear to students what generics are doing behind the scenes. In my most recent class I tried going the other way around, showing the monomorphized versions first to more clearly motivate what generics are used for, and I liked the way it went. I think motivating generics by first showing code duplication and then showing how generics allow us to de-duplicate makes for a good teaching flow, and I think it also helps make things clearer to students coming from more dynamic languages that don't have an equivalent to generics.

I also changed the `pick` fns to take a `bool` as the first argument because I think that makes things slightly clearer/cleaner, but I'm not married to that change either.

### DIFF
--- a/src/generics/generic-functions.md
+++ b/src/generics/generic-functions.md
@@ -24,7 +24,6 @@ fn pick_char(cond: bool, left: char, right: char) -> char {
     }
 }
 
-/// Pick `left` or `right` depending on the value of `n`.
 fn pick<T>(cond: bool, left: T, right: T) -> T {
     if cond {
         left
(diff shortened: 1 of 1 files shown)


djmitche (commented):
I like the change, but it probably deserves a speaker note regarding how to use the slide.

mgeisler:
The idea of showing the result of monomorphization is great, but I'm afraid the slide become too tall now?

I would suggest putting this into the speaker notes instead: keep the `pick<T>` function in the slide and let the speaker notes explain that it expands to `pick_i32` and `pick_char` depending on how it's called.

randomPoison:
Both of your comments make sense to me, so I've moved the monomorphized versions into the speaker notes and added a note explaining that they can be used to both motivate the generic `pick` and demonstrate monomorphization. Let me know how it looks!

djmitche (approved):

PR #2672, opened 2025-02-28
Add speaker note about UFCS

randomPoison:
I like to note that methods can also be called like associated functions because this fact comes up when talking about traits (usually from questions students ask).


mgeisler (approved):
Thanks!

PR #2673, opened 2025-02-28
Add division example to expression exercise

randomPoison:
[I'm having deja vu](https://github.com/google/comprehensive-rust/pull/2605). Fixes a warning about an unused enum variant in the expression evaluation exercise.


djmitche (approved):
Oops, sorry if I was responsible for that!

PR #2675, opened 2025-02-28
Fix bug where speaker notes are not connected to regular window

michael-kerscher:
Fixes bug #2004.
Refactored the communication between the speaker notes window and the regular window by using a Broadcast channel - this is now self-recovering(!) even if speaker notes are closed and manually re-opened!

For better readability and maintainability refactored some string-based states into enum style code and refactored detection of the type of windows (print, speaker note, regular window)

Manually tested the new code and the speaker notes window does not disconnect from the regular window anymore.
This now works way more reliable, even if there are (still) some UI glitches that have been there before already.

### DIFF
--- a/theme/speaker-notes.js
+++ b/theme/speaker-notes.js
@@ -25,7 +25,6 @@
     Regular: "regular",
     RegularWithSpeakerNotes: "regular-speaker-notes",
     SpeakerNotes: "speaker-notes",
-    SpeakerNotesDefunct: "speaker-notes-defunct",
     PrintPage: "print-page",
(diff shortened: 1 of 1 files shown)


djmitche (approved):
I tested this locally and it seems to work quite well -- thank you!!

I'm not an expert on plain JS. I don't see anything obviously wrong here, and there's basically no risk to landing this, but if you'd like a more skilled review of the implementation feel free to add someone else before merging.

djmitche:
lol, or auto-merge :)

PR #2677, opened 2025-02-28
Remove `pub` from traits exercise

randomPoison:
We haven't talked about visibility at this point and the new keyword sometimes confuses students.


djmitche (approved):
I think I added this to avoid unused warnings, which aren't an issue anymore..

PR #2687, opened 2025-03-01
Correct minor typos in slide notes

ColinPitrat:
- There's a `NotCloneable` in the code, no `NonCloneable`
 - Point<StderrLogger> doesn't make sense in this context

### DIFF
--- a/src/closures/traits.md
+++ b/src/closures/traits.md
@@ -2,7 +2,7 @@
 minutes: 10
 ---
 
-# Closures
+# Closure traits
 
 Closures or lambda expressions have types which cannot be named. However, they
 implement special [`Fn`](https://doc.rust-lang.org/std/ops/trait.Fn.html),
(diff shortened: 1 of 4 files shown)


djmitche:
It looks like this needs a run of `dprint fmt` before it can be merged.

By the way, per https://docs.github.com/en/get-started/using-github/github-flow it's a good idea to make a branch before working on a PR, instead of making a PR from your `main` branch. It's fine for this PR but just an FYI for next time.

ColinPitrat:
Ran `dprint fmt` but I hit an issue:
```
Error formatting (...)/comprehensive-rust/.github/workflows/check-msgid-changes.py. Message: Cannot start formatter process: No such file or directory (os error 2)
```

I worked around it by excluding `.github` from dprint, but I'm wondering if that's expected, if something needs to be fixed in dprint.json or if it's just that my version of dprint is not the right one.

michael-kerscher (approved):

michael-kerscher:
> Ran `dprint fmt` but I hit an issue:
> 
> ```
> Error formatting (...)/comprehensive-rust/.github/workflows/check-msgid-changes.py. Message: Cannot start formatter process: No such file or directory (os error 2)
> ```
> 
> I worked around it by excluding `.github` from dprint, but I'm wondering if that's expected, if something needs to be fixed in dprint.json or if it's just that my version of dprint is not the right one.

From the error message this looks like the `yapf3` command is missing on your system (see https://github.com/google/comprehensive-rust/blob/main/CONTRIBUTING.md#linux)

Issue #2509, opened 2024-12-11
Cross-platform command to install dependencies


mgeisler:
We should do something very simple to make Dustin's `cargo xtask install-deps` work. It could be as simple as a

```rust
fn main() -> Result<(), Box<dyn std::error::Error>> {
  Command::new(env!("CARGO")).arg("install").arg("mdbook").status()?;
  Command::new(env!("CARGO")).arg("install").arg("mdbook-i18n-helpers").status()?;
}
```

Put that in a new workspace member, hook up the Cargo alias and we should be in business :smile:  To show that it works, make the [GitHub CI code](https://github.com/google/comprehensive-rust/blob/main/.github/workflows/install-mdbook/action.yml) that install the dependencies use the new task.

_Originally posted by @mgeisler in https://github.com/google/comprehensive-rust/issues/2469#issuecomment-2536995146_

michael-kerscher:
completed with #2688. Even some other tasks are already done with xtasks

PR #2688, opened 2025-03-03
Xtask

egithinji:
### Background
Comprehensive Rust requires a number of tools to be installed (e.g. mdbook and mdbook-course). As mentioned in #2509 (and discussed in #2469) it would be nice to have a cross platform command for installing these dependencies. Currently these are installed using a shell script (`install-mdbook.sh`) but this isn't truly cross platform e.g. for Windows users.

### Xtask

[xtask](https://github.com/matklad/cargo-xtask) outlines an approach for automating tasks in a Rust project. It involves using cargo's aliasing feature to allow us to run commands like `cargo xtask <some task>` to perform adhoc tasks via a Rust binary that we might otherwise need a shell script for.

In this PR we add support for a `cargo xtask install-tools` command that will replace the `install-mdbook.sh` script and install the dependent tools. We can potentially extend it to support for other tasks e.g. `cargo xtask fmt`.


Cargo.lock
```diff
@@ -3703,6 +3703,10 @@ version = "0.5.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1e9df38ee2d2c3c5948ea468a8406ff0db0b29ae1ffde1bcf20ef305bcc95c51"
 
+[[package]]
+name = "xtask"
+version = "0.1.0"
```
michael-kerscher, line 3708: version 0.1.0 looks pretty new to me. When looking up the website https://github.com/matklad/cargo-xtask I'm not sure how well this will be maintained. I have seen https://github.com/matklad/cargo-xtask?tab=readme-ov-file#status though and if cargo is using this in their build process, this might also fit this repository.

xtask/src/main.rs
```diff
@@ -0,0 +1,70 @@
...
+    Ok(())
+}
+
+fn install_tools() -> Result<(), DynError> {
+    println!("Installing project tools...");
+    // The --locked flag is important for reproducible builds. It also
+    // avoids breakage due to skews between mdbook and mdbook-svgbob.
+    Command::new(env!("CARGO"))
+        .arg("install")
+        .args(["mdbook", "--locked", "--version", "0.4.44"])
```
michael-kerscher, line 27: Would you mind putting these args into some kind of array that you loop over and execute these very similar commands by generating the individual commands? While this already looks good, it might be easier to spot the installed tools and versions that way.

michael-kerscher:
Hi @egithinji. Thank you very much for drafting this pull request! I really like the idea of having this cross platform task execution tool!
I can already think of an additional use case for this. We have a test framework that is testing the generated website and currently needs to be executed manually in the tests subdirectory. Maybe it makes sense to include such tasks as well (not in this PR though, this is a idea for future work)

xtask/src/main.rs
egithinji: Yes I agree. I've extracted them into an array of arrays. What do you think of how it is now?

egithinji:
> Hi @egithinji. Thank you very much for drafting this pull request! I really like the idea of having this cross platform task execution tool! I can already think of an additional use case for this. We have a test framework that is testing the generated website and currently needs to be executed manually in the tests subdirectory. Maybe it makes sense to include such tasks as well (not in this PR though, this is a idea for future work)

Thanks @michael-kerscher for reviewing! Yes that sounds like a good candidate for an additional task. Happy to collaborate on this further.

Cargo.lock
michael-kerscher: Thank you for pointing this out that this was just a guideline!
Yesterday when I looked at this I was under the impression this is a dependency, really should have seen that.
Just later down I saw actual code with no dependency.

xtask/src/main.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+fn install_tools() -> Result<(), DynError> {
+    println!("Installing project tools...");
+
+    for args in INSTALLATION_ARGS.iter() {
+        Command::new(env!("CARGO"))
+            .arg("install")
+            .args(args.iter().filter(|a| **a != ""))
+            .status()?;
```
michael-kerscher, line 41: the `status()` result is currently not used. This provides a Status that can be any value, e.g. 101.
You could check this status with e.g. `.success()`

xtask/src/main.rs
```diff
@@ -0,0 +1,60 @@
...
+        eprintln!("{e}");
+        std::process::exit(-1);
+    }
+}
+
+fn execute_task() -> Result<(), DynError> {
+    let task = env::args().nth(1);
+    match task.as_deref() {
+        Some("install-tools") => install_tools()?,
+        _ => print_help(task.as_deref()),
```
michael-kerscher, line 29: It might make sense to return a non-successful error code here to indicate an issue.
We want the CI script to break if an invalid command, e.g. with a typo like ìnstall_tools` is used. I'm not sure about what to return for an actual explicit call like "cargo xtask help"

xtask/src/main.rs
```diff
@@ -0,0 +1,60 @@
...
+            .args(args.iter().filter(|a| **a != ""))
+            .status()?;
+    }
+
+    Ok(())
+}
+
+fn print_help(task: Option<&str>) {
+    if let Some(t) = task {
+        eprintln!(
```
michael-kerscher, line 49: this is already the *ln* variant of the command. I would remove the last new line of this help string

xtask/src/main.rs
egithinji: Done

xtask/src/main.rs
egithinji: Adjusted to surface the specific exit code in CI (and to exit when any of the tools fails).

xtask/src/main.rs
egithinji: Adjusted so that invalid command exits with exit code (-1).

xtask/src/main.rs

xtask/src/main.rs
```diff
@@ -26,35 +13,59 @@ fn execute_task() -> Result<(), DynError> {
...
+fn get_help_string(task: Option<&str>) -> String {
     if let Some(t) = task {
-        eprintln!(
+        format!(
             "Unrecognized task '{t}'. Available tasks:
 
 install-tools            Installs the tools the project depends on.
-
-Run with `cargo xtask [task]`.
 "
```
michael-kerscher, line 66: This variant still has a manual new line at the end that is printed with an additional new line if the error is returned.

xtask/src/main.rs
```diff
@@ -26,35 +13,59 @@ fn execute_task() -> Result<(), DynError> {
...
-            .args(args.iter().filter(|a| **a != ""))
-            .status()?;
+            .args(args)
+            .status()
+            .expect("Failed to execute cargo install");
+
+        if !status.success() {
+            let error_message = format!(
+                "cargo install {} {} exited with status code: {}",
+                args.get(0).unwrap(),
```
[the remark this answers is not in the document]

xtask/src/main.rs

michael-kerscher (approved):
looks good to me now

Cargo.lock
mgeisler: > Yes thanks for pointing this out and just to add: IIUC https://github.com/matklad/cargo-xtask is just a guideline on how to structure

and importantly, the `xtask` mentioned here is _not_ the same thing as this repository (`xtask` is a local dependency).

xtask/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "xtask"
+version = "0.1.0"
+edition = "2021"
```
mgeisler, line 4: Please add a `publish = false` line to avoid accidentally publishing this:

```suggestion
edition = "2021"
publish = false
```

mgeisler: You're also very welcome to add a one-line description so people can find out what this is.

xtask/src/main.rs
```diff
@@ -0,0 +1,69 @@
+use std::{env, process::Command};
```
mgeisler, line 1: I would appreciate a small (3-5 lines?) docstring here which explains what this binary is for.

xtask/src/main.rs
```diff
@@ -0,0 +1,69 @@
...
+
+fn main() {
+    if let Err(e) = execute_task() {
+        eprintln!("{e}");
+        std::process::exit(-1);
+    }
+}
+
+fn execute_task() -> Result<(), DynError> {
+    let task = env::args().nth(1);
```
mgeisler, line 13: We already depend on Clap, so you could use it here (feel free to do this in a later PR!).

xtask/src/main.rs
mgeisler: Feel free to do this in a later PR!

xtask/src/main.rs
```diff
@@ -0,0 +1,69 @@
...
+            return Err(Box::from(get_help_string(task.as_deref())));
+        }
+    }
+    Ok(())
+}
+
+fn install_tools() -> Result<(), DynError> {
+    println!("Installing project tools...");
+
+    let install_args: Vec<Vec<&str>> = vec![
```
mgeisler, line 26: I would drop the explicit type annotation, it's probably not necessary and all the `vec![...]` calls make it fairly easy to see at a glance.

xtask/src/main.rs
```diff
@@ -0,0 +1,69 @@
...
+        vec!["--path", "mdbook-exerciser", "--locked"],
+        vec!["--path", "mdbook-course", "--locked"],
+    ];
+
+    for args in &install_args {
+        let status = Command::new(env!("CARGO"))
+            .arg("install")
+            .args(args)
+            .status()
+            .expect("Failed to execute cargo install");
```
mgeisler, line 44: Just curious: have you tested how the errors look if you provoke an error? Like by installing something which doesn't exist?

xtask/src/main.rs
```diff
@@ -0,0 +1,69 @@
+use std::{env, process::Command};
+
+type DynError = Box<dyn std::error::Error>;
```
mgeisler, line 3: Feel free to refactor this to use anyhow later! It makes error handling very streamlined in small scripts like this one.

mgeisler (approved):
This is awesome, thanks a lot for taking this up!

Please make sure to update the two issues linked (#2509 and #2469) so people know about this new system.

xtask/src/main.rs
egithinji: Yes I tested manually the various scenarios (missing task, unkown task, and internal error e.g. typo 'mdbok instead of mdbook' in the install-tools logic). We get appropriate error messages and behaviour in each case.

xtask/src/main.rs
egithinji: Have pushed a new PR (#2707)  for this and the other suggestions.

.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[alias]
```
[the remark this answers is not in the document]
egithinji: Have pushed a new PR (https://github.com/google/comprehensive-rust/pull/2707) for this and the other suggestions.

xtask/Cargo.toml
egithinji: Have pushed a new PR (https://github.com/google/comprehensive-rust/pull/2707) for this and the other suggestions.

xtask/src/main.rs
egithinji: Have pushed a new PR (https://github.com/google/comprehensive-rust/pull/2707) for this and the other suggestions.

xtask/src/main.rs
egithinji: Have pushed a new PR (https://github.com/google/comprehensive-rust/pull/2707) for this and the other suggestions.

xtask/src/main.rs
egithinji: Have pushed a new PR (https://github.com/google/comprehensive-rust/pull/2707) for this and the other suggestions.

PR #2691, opened 2025-03-06
Add the weak count to the RC diagram

ColinPitrat:
This adds the weak count in the RC diagram, and rename `count` to `strong` making it more accurate.


djmitche (changes requested):
I considered this when updating the diagram, but this, too, is actually incorrect -- the weak count is 2 in this case (if I remember correctly from that PR). This gets into more implementation detail than we really want to address in this slide, so I chose to approximate the truth and stick to the topic at hand.

The text above this diagram is incorrect, though, in that it mentions strong and weak _pointers_ when of course they are counters.

Maybe the best approach is removing mention of weak from the slide itself (and replacing "pointers" with "counters"), and adding a speaker note indicating that the diagram isn't telling the full story, perhaps with a link to the Rc implementation.

Issue #2234, opened 2024-07-19
Automated slide size analysis with a new slide scoring tool


michael-kerscher:
The comprehensive rust training slides vary in size and some contain too much content and are getting too big to be used in classes as scrolling starts to be necessary. Once you have to scroll a slide this throws people off during class and this should be avoided therefore.

#1464 tries to give an overview on why slides that are too big are not useful and how good slides could look like. Unfortunately this is a very manual task to look over all the slides to find violations against the the rule to not need to rely on scrolling.

I propose a small tool to create automated statistics on the size of slides, find violations against a to be defined policy and even score them to sort them to be able to focus and target on the slides that need most work.

Such a tool would enable to easily change themes or other large scale changes and see how this affects the slides.
Additionally this can be used by contributors to easily judge locally if the slide is getting too big and later on this could also be used in the CI.

The scoring tool could use [selenium](https://www.selenium.dev/) via [webdriver](https://www.selenium.dev/documentation/webdriver/) with the crate [fantoccini](https://crates.io/crates/fantoccini) that offers methods like [Element.rectangle()](https://docs.rs/fantoccini/0.19.3/fantoccini/elements/struct.Element.html#method.rectangle).

mgeisler:
Thanks Michael, that would be a nice tool indeed!

michael-kerscher:
I'm not sure yet how to an integration in the workflow should look like as this is a quite expensive operation (3mins for the English translation).
Different translations don't necessarily have the same length but probably use a similar amount of pixels on the screen so it might be sufficient to only have this for the English translation in the beginning.
Later stage could be detecting what languages have been changed and only render these.

mgeisler:
> I'm not sure yet how to an integration in the workflow should look like as this is a quite expensive operation (3mins for the English translation).

Yeah, we should be able to just run this in another parallel job.

To start with, I guess we need to fine-tune the tool a bit to make it give us a report that we can file issues from?

We have some CSS and JS for drawing the big red rectangle you see in #1464: [`theme/redbox.js`](https://github.com/google/comprehensive-rust/blob/main/theme/redbox.js). I forget how we trigger this, but it should be easy enough to find.

That box has been my rule of thumb for when slides become too tall. I guess you can use that as a first step?

michael-kerscher:
[...]

PR #2693, opened 2025-03-08
Evaluate Slide size and block if they grow above a certain treshold (with exemption mechanism)

michael-kerscher:
This enables a test for the width and height of slides (excluding some special cases completely).
The mechanism has an exemption mechanism to temporarily exempt slides from the rules.
Even exempted slides are checked for the rule violation and once the slides are compliant they must be removed from the exemption list to avoid future regression (the check fails in the CI if compliant slides are exempted!)

This also provides a good opportunity to always have an up-to-date list of overlong slides in [slide-exemptions.list.ts](tests/src/slides/slide-exemptions.list.ts) that can be worked on.

On the CLI it can be locally used with `npm run test -- --spec=src/slide-size.test.ts` (after creating the list with `./src/slides/create-slide.list.sh ../book/html/`).
The CI environment specifies the env var `TEST_BOOK_DIR` that is used to specifiy the html directory so it can create the list of slides on-the-fly, check against hardcoded exemptions and evaluate.

This is a new solution for #1464 within the new test framework. This is related to #2234 and makes the mdbook-slide-evaluator from #2258 obsolete and should be removed as this is a not so powerful nor flexible framework.


michael-kerscher:
before merging this, this needs an additional polish so this does not just break the `npm test` call. Currently the slides list needs to be created by manually calling `create-slide.list.sh`

mgeisler (approved):
This looks very nice, thank you!

.github/workflows/build.yml
```diff
@@ -190,7 +190,7 @@ jobs:
         working-directory: ./tests
       - name: Test Javascript
         if: matrix.language == 'en'
-        run: npm test
+        run: ./src/slides/create-slide.list.sh; npm test
```
mgeisler, line 193: Can we use the multi-line YAML syntax here?

mgeisler: The

```yaml
run: |
  ./src/slides/create-slide.list.sh
  npm test
```

syntax from a quick look at https://yaml-multiline.info/.

michael-kerscher: ack

michael-kerscher:
@mgeisler I added an empty version of the slides.list.ts to avoid issues with running tests locally. Otherwise the test suite crashes with:

```
[0-3] Spec file(s): file:///home/<user>/workspace/comprehensive-rust/tests/src/slide-size.test.ts
[0-3] Error: Error: Cannot find module './slides/slides.list'
```

PR #2696, opened 2025-03-19
Improve tests for BinaryTree exercise

bartoszkp:
Current tests still pass if `len` is implemented to calculate the height of the tree (i.e. max(left.len(), right.len()) + 1 for each node). It seems this is quite a common misunderstanding when doing this course.

With the new assert height implementation will fail, which hints towards implementing `len` as a total number of nodes.


djmitche (approved):
Nice fix - thanks!

PR #2707, opened 2025-04-06
Xtask improvements

egithinji:
Some minor improvements to an already-merged PR (#2688) on the task automation via xtask. Main ones being:

- Adding more explanatory comments about what the xtask package is and what it does
- Using Clap for CLI arg parsing
- Using Anyhow for error handling

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -20,6 +20,7 @@
 //! the tools.
 
 use std::{env, process::Command};
+use clap::Parser;
 
 type DynError = Box<dyn std::error::Error>;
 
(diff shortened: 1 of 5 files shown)


xtask/src/main.rs
```diff
@@ -1,29 +1,60 @@
...
+    task: String,
+}
+
+fn execute_task() -> Result<()> {
+    let task = Args::parse().task;
+    match task.as_str() {
+        "install-tools" => install_tools()?,
         _ => {
-            return Err(Box::from(get_help_string(task.as_deref())));
+            return Err(anyhow!(unrecognized_task_string(task.as_str())));
```
mgeisler, line 48: I'm surprised it should be necessary with this case? I would have expected Clap to be able to handle it for us?

We can merge this PR now, but could you look into using a Rust enum to represent the available commands? I'm pretty sure Clap can be told that a command line argument must be a valid enum variant — and so it can return an error earlier if people use an unrecognized name.

I think there is an example of what I mean here: https://docs.rs/clap/latest/clap/_cookbook/cargo_example_derive/index.html

We want each xtask (such as `install-tools`) to become a sub-command.

mgeisler (approved):
Thanks @egithinji! Let's merge this now since it's a nice improvement to what we had before.

PR #2710, opened 2025-04-09
remove the multilingual field from book.toml

szabgab:
See https://github.com/szabgab/public-mdbooks/issues/10


michael-kerscher (approved):
this was never used and is now removed upstream in mdbook

PR #2711, opened 2025-04-16
Remove type annotations in move.md

randomPoison:
At this point I don't think we need the type annotation, and I think the example is a bit clearer without them, so I usually delete these as I'm explaining in class.


djmitche (approved):
Good catch -- I think this slide used to be earlier in the course, when we were still including type annotations here-and-there.

michael-kerscher (approved):
looks good to me

PR #2712, opened 2025-04-16
Use Vec::new instead of vec!

randomPoison:
Using the `vec!` macro to create an empty `Vec` is a bit weird imo, generally I only see the macro used when you actually want to initialize the `Vec` with some values (like is done in the examples for this exercise). Students are more likely to use `Vec::new`, and I think that's the more idiomatic approach, so I think using `Vec::new` here would be better.

### DIFF
--- a/src/memory-management/exercise.rs
+++ b/src/memory-management/exercise.rs
@@ -65,8 +65,8 @@ impl PackageBuilder {
         Self(Package {
             name: name.into(),
             version: "0.1".into(),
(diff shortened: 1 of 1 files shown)


djmitche (approved):

djmitche:
I tend to use `vec![]` just because it's fewer characters, but you're right that doing so is unusual.

Also, bummer, I appear to not be able to merge things anymore.

PR #2713, opened 2025-04-16
Remove extra credit from binary tree exercise

randomPoison:
At this point we haven't talked about iterators, so students are only going to be confused by this suggestion.


fw-immunant (approved):
I agree, this isn't a helpful suggestion at this point in the course.

PR #2719, opened 2025-04-19
Update link to macros in the book

bjornjorgensen:
image


deavid (approved):

src/index.md
```diff
@@ -49,7 +49,7 @@ Rust is a large language and we won't be able to cover all of it in a few days.
 Some non-goals of this course are:
 
 - Learning how to develop macros: please see
-  [Chapter 19.5 in the Rust Book](https://doc.rust-lang.org/book/ch19-06-macros.html)
```
djmitche, line 52: Rather than try to track an unstable version number here, let's just refer to the Rust Book as a whole.

djmitche (commented):
We generally don't make changes to the translation directly; rather, those are updated as and when the per-language translators have capacity. So you can limit this change to just `src/index.md` (and remove all of the reviewers)

src/index.md
bjornjorgensen: ok, so you will have "_Learning how to develop macros: please see the Rust Book and [Rust by Example](https://doc.rust-lang.org/rust-by-example/macros.html) instead._" 
or?

bjornjorgensen:
..(and remove all of the reviewers)
Those are added by a bot.. I cant find a way to remove those now. Do you know how I can remove those?

djmitche (approved):
Patch looks better. I no longer have permissions on the repo, unfortunately, but hopefully someone who does can fix the reviewers and/or just merge this.

michael-kerscher (approved):
lgtm

PR #2720, opened 2025-04-21
Update generic notes to match example

### LINKED ISSUE #2716: Outdated instructions in generic functions slide
https://github.com/google/comprehensive-rust/blob/52114364461650b55619cef1dd94d827a42b48cb/src/generics/generic-functions.md?plain=1#L65C1-L67C57

I think it's referencing an earlier version of the pick function. Now, there's no n, even, or odd parameters to modify, so these instructions are nonsensical

djmitche:
Fixes #2716.

### DIFF
--- a/src/generics/generic-functions.md
+++ b/src/generics/generic-functions.md
@@ -57,14 +57,14 @@ fn main() {
       val: u8,
   }
 
-  pick(123, Foo { val: 7 }, Foo { val: 456 });
+  pick(false, Foo { val: 7 }, Foo { val: 99 });
   ```
 
 - This is similar to C++ templates, but Rust partially compiles the generic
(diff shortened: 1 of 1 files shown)


src/generics/generic-functions.md
michael-kerscher, line 1: It seems that the expected function signature on line 60 is also wrong. first argument of pick should be a boolean condition, but 123 is used. Can you fix this as well?

michael-kerscher: additionally… 456 does not fit into u8

michael-kerscher (changes requested):
the change looks good, but the notes still has related issues

michael-kerscher (approved):
looks good to me

PR #2725, opened 2025-04-30
Make install-tools command work from any directory in the workspace.

### LINKED ISSUE #2708: xtask install-tools command requires being in root directory of project
Via #2688 we can now install the dependent project tools with one cross platform command `cargo xtask install-tools`. The command fails to install `mdbook-exerciser` and `mdbook-course` however unless you are in the root of the repository. This is because these two dependencies are located in the local filesystem.

One of the benefits of the cargo xtask approach is we can run commands from anywhere so it would be nice if this could be fixed somehow.

egithinji:
This fixes #2708 by creating a CARGO_WORKSPACE_DIR env variable to act as an anchor path, allowing the installation of mdbook-exerciser and mdbook-course to succeed from any directory within the repository. Based on the approach mentioned here: https://github.com/rust-lang/cargo/issues/3946#issuecomment-973132993

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,3 +2,8 @@
 # We use this alias for task automation in the project.
 # See README in xtask directory.
 xtask = "run --package xtask --"
+
+[env]
+# To provide an anchor to the root of the workspace when working with paths.
+# See https://github.com/rust-lang/cargo/issues/3946#issuecomment-973132993
+CARGO_WORKSPACE_DIR = { value = "", relative = true }
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -54,6 +54,11 @@ fn execute_task() -> Result<()> {
 fn install_tools() -> Result<()> {
     println!("Installing project tools...");
 
+    let path_to_mdbook_exerciser =
+        format!("{}mdbook-exerciser", env!("CARGO_WORKSPACE_DIR"));
+    let path_to_mdbook_course =
+        format!("{}mdbook-course", env!("CARGO_WORKSPACE_DIR"));
+
     let install_args = vec![
         // The --locked flag is important for reproducible builds. It also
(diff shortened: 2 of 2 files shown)


xtask/src/main.rs
```diff
@@ -54,6 +54,11 @@ fn execute_task() -> Result<()> {
 fn install_tools() -> Result<()> {
     println!("Installing project tools...");
 
+    let path_to_mdbook_exerciser =
+        format!("{}mdbook-exerciser", env!("CARGO_WORKSPACE_DIR"));
```
michael-kerscher, line 58: I would like to wrap the workspace dir from the environment in a  [Path](https://doc.rust-lang.org/std/path/struct.Path.html) and join here with the directory name to get a proper path that can be easily converted to a string. That way you don't get into the trouble of formatting with the right dir separators.

.cargo/config.toml
```diff
@@ -2,3 +2,8 @@
 # We use this alias for task automation in the project.
 # See README in xtask directory.
 xtask = "run --package xtask --"
+
+[env]
+# To provide an anchor to the root of the workspace when working with paths.
+# See https://github.com/rust-lang/cargo/issues/3946#issuecomment-973132993
+CARGO_WORKSPACE_DIR = { value = "", relative = true }
```
michael-kerscher, line 9: That is an interesting solution and makes this independent of the location of the xtask directory!
If I read the linked comment correctly this is the path in which the central .cargo directory lives?

michael-kerscher (changes requested):

michael-kerscher:
regarding the failed `vi` build I created https://github.com/google/comprehensive-rust/pull/2726 which should get rid of some spurious fails from the test framework

xtask/src/main.rs
egithinji: Yes true thanks, done.

michael-kerscher (approved):

PR #2742, opened 2025-05-07
mdbook-exerciser: ensure all code block bytes of the examples are written to the file

michael-kerscher:
- write() attempts to write the entire buffer does not guarantee this.
  Not writing all bytes is not considered an error.
  It just returns the number of bytes.
- write_all() ensures that everything is written or throws an error


djmitche (approved):
I'm curious if this ever caused problems? I think `write` for actual files never does a short write (but I suppose if the output is piped this might do a partial write).

michael-kerscher:
@djmitche Very likely not. I don't think with our setups and example file size we have ever seen this. I just fixed this preemptive as I have noticed this while looking through some code.

PR #2743, opened 2025-05-07
Add xtask command for running web-tests.

egithinji:
Currently, to run the tests that are located in the `tests` directory (the js tests), one has to navigate to the directory and run `npm test` or `npm start`. We now have a way of automating such task execution using the binary in the `xtask` directory. This pr makes use of this by introducing a new command `cargo xtask web-tests` that can be run from anywhere in the repo to run the tests in the `tests` directory.


michael-kerscher (approved):
That new command is great!

PR #2745, opened 2025-05-12
tests: refactor slide style guide tests and add playground size checks

michael-kerscher:
The previous slide-size.test.ts was refactored and a slide class was implemented for convenience.
The test now checks the playground size (if available on that slide) by checking if scrollbars are visible.
To make this test more efficient, the mocha before() hook is used to load the slide once and check with all relevant tests before loading the next slide.


michael-kerscher:
This now checks for #2012 and provides a list of violations in the playground-size-exemptions in [tests/src/slides/slide-exemptions.list.ts](https://github.com/google/comprehensive-rust/blob/main/tests/src/slides/slide-exemptions.list.ts). This can act as a TODO list

tests/src/objects/slide.ts
```diff
@@ -0,0 +1,52 @@
+import { $, browser } from "@wdio/globals";
+
+export default class Slide {
+  /**
+   * conviencene functions for interacting with a slide
```
randomPoison, line 5: Typo: should be "convenience".

randomPoison (approved):
Looks good, my only comment is for a typo.

tests/src/objects/slide.ts
michael-kerscher: thanks for spotting this!

PR #2746, opened 2025-05-14
Remove explicit derefs from shared ref slide

randomPoison:
When teaching I generally remove these. I emphasize that a reference can generally be used as if you have the referenced value directly, since in most cases you don't have to explicitly dereference a reference. On the next slide we show mutable references, and we need to use a deref when writing through a mutable reference, so I think that's the better place to point out the deref operator.


djmitche (approved):
makes sense to me!

PR #2751, opened 2025-05-16
Use aarch64-rt in bare metal AP section and exercise

qwandor:
Also use `global_asm!` rather than `cc::Build` to build assembly files. This avoids a dependency and some build scripts.


djmitche (approved):
This simplifies the exercise nicely!

PR #2752, opened 2025-05-16
Use safe-mmio crate in PL011 UART driver example and RTC exercise solution

qwandor:


mgeisler:
> Use safe-mmio crate in PL011 UART driver example and RTC exercise solution

A question that comes to mind is _why_?

I see that you're replacing some raw pointers and hand-written structs with something from a new crate. What is the benefit?

Put differently, before, I could reason about the code directly. Now, I need to know something about the two crates that are in use. So it almost seems like a step back from a pedagogical point of view?

mgeisler (approved):
Feel free to merge since I understand that you would like to show people how safe and nice the API can be. But consider the learning opportunity of hand-writing the more low-level approach as well.

qwandor:
> Feel free to merge since I understand that you would like to show people how safe and nice the API can be. But consider the learning opportunity of hand-writing the more low-level approach as well.

True, using `safe-mmio` obscures a bit what is actually going on, which makes it more difficult to explain. It's also an opinionated choice. On the other hand, I want to make the final driver we end up with an example of good practice, because I've seen that people tend to copy and paste the examples from this course as a starting point for writing their own code.

I think this is a tricky balance we have in a lot of the course, between teaching the fundamentals of how the Rust language works, and showing the practical side of how to write good idiomatic Rust code. I'm inclined to have more of the latter, as people are going to go away from this course to work on new projects in Rust, sometimes without an experienced Rust developer on their team, and so I think it's important to set them up to write good Rust code right away.

One option I considered was to keep the old version as a step along the way, and then have another slide where we go through and refactor it to an improved version using `safe-mmio`. I'm worried that makes for too many steps though, as we already have the first version in `pl011_minimal.rs` using a single raw pointer.

qwandor:
I'll see how this goes when I teach the course again next week and go from there.

PR #2753, opened 2025-05-17
Simplify project commands using `cargo xtask`

### LINKED ISSUE #2741: Use Clap to handle unrecognized commands in xtask
Could you look into using a Rust enum to represent the available commands? I'm pretty sure Clap can be told that a command line argument must be a valid enum variant — and so it can return an error earlier if people use an unrecognized name.

I think there is an example of what I mean here: https://docs.rs/clap/latest/clap/_cookbook/cargo_example_derive/index.html

We want each xtask (such as `install-tools`) to become a sub-command.

_Originally posted by @mgeisler in https://github.com/google/comprehensive-rust/pull/2707#discussion_r2030935143_

egithinji:
We are already using `cargo xtask install-tools` to install the project's tools, and `cargo xtask web-tests` to run the js tests. In this PR we provide support for the various `mdbook` commands (`test`,`serve`, and `build`)  with `cargo xtask` commands. This provides a uniform interface for running tasks in the project. Additionally it allows these commands to work from within any dirrectory (previously you'd need to navigate to the workspace root in order to run say `mdbook build`).

Additionally we're improving the xtask code by making use of `Clap` enums to handle validation of the possible tasks to run via xtask (this closes #2741 ).


xtask/src/main.rs
```diff
@@ -69,6 +84,7 @@ fn install_tools() -> Result<()> {
         vec!["mdbook-pandoc", "--locked", "--version", "0.9.3"],
         vec!["mdbook-i18n-helpers", "--locked", "--version", "0.3.5"],
         vec!["i18n-report", "--locked", "--version", "0.2.0"],
+        vec!["mdbook-linkcheck", "--locked", "--version", "0.7.7"],
```
egithinji, line 87: There is a github action that installs this manually [here](https://github.com/google/comprehensive-rust/blob/dbd1a09031714bda7be972ea14686f44d8cbe451/.github/workflows/build.yml#L166). Can that action be updated to instead run `cargo xtask install-tools` which will then ensure linkcheck is installed? I'm concerned with having two places where we define this version, but don't have much context about that github action.

egithinji: I have reverted installing linkcheck as part of install-tools because it results in build failures of some languages in the CI.

README.md
```diff
@@ -74,25 +71,17 @@ Then install these tools with:
...
+| `cargo xtask web-tests`     | Run the web driver tests in the tests directory.                                                                                                                                                                                                                                                                                                                                           |
+| `cargo xtask build`         | Create a static version of the course in the `book/` directory. Note that you have to separately build and zip exercises and add them to book/html. To build any of the translated versions of the course, run MDBOOK_BOOK__LANGUAGE=xx mdbook build -d book/xx where xx is the ISO 639 language code (e.g. da for the Danish translation). TRANSLATIONS.md contains further instructions. |
```
michael-kerscher, line 84: please keep the link to TRANSLATIONS.md

michael-kerscher (approved):

PR #2755, opened 2025-05-22
Simplify type cast and remove error handling in protobuf exercise

randomPoison:
While using `.try_into().expect()` is the more robust solution, I think it's more complex than we want here. We don't really care about error handling in this exercise, so I think a simple `as` cast is clearer here. I think we should also remove the logic for checking if there are enough bytes left, since it simplifies things and makes the solution easier to read.

### DIFF
--- a/src/lifetimes/exercise.rs
+++ b/src/lifetimes/exercise.rs
@@ -127,10 +127,7 @@ fn parse_field(data: &[u8]) -> (Field, &[u8]) {
         }
         WireType::Len => {
             let (len, remainder) = parse_varint(remainder);
-            let len: usize = len.try_into().expect("len not a valid `usize`");
-            if remainder.len() < len {
-                panic!("Unexpected EOF");
-            }
+            let len = len as usize;
             let (value, remainder) = remainder.split_at(len);
             (FieldValue::Len(value), remainder)
         }


djmitche (commented):
In other cases where we've removed something that would be likely in "real" Rust, we've gotten more than one PR trying to put it back. I think it would help to add some commentary here to indicate this is omitted deliberately. Maybe that's as simple as "// TODO: Handle invalid lengths"? That would be enough to fend off any PRs, and also provide a jumping-off point for instructors to talk about what might go wrong here (e.g., `split_at` might panic) and how to handle it.

randomPoison:
Hm yeah I've noticed that sometimes things I remove for simplicity end up back in the slides. That's kind of frustrating but I get why it's happening. 

> Maybe that's as simple as "// TODO: Handle invalid lengths"?

I think if we put a TODO in there we're just going to get PRs trying to fix the TODO 😅. Maybe it would make sense to add something to the contribution docs stating that we want to air on the side of simplicity and clarity in the code examples and exercises, even when that makes the code less idiomatic. I don't that that'd stop the PRs, but it'd give us something to point to when somebody tries to make the code more "idiomatic" in a way that distracts from the point of the example/exercise.

mgeisler:
> I think it's more complex than we want here. We don't really care about error handling in this exercise

Would it work to change `parse_varint` to return `usize` instead of `u64`?

djmitche:
I think that would just move the `as`. into parse_varint, and a varint is defined to contain a 64-bit integer, so that would be a bit odd. I've added a comment that doesn't say "TODO", maybe that will work?

randomPoison:
@djmitche I like your comment, I think calling out that we're doing the simplest thing and not the most robust thing is a reasonable approach. Do you think this is good to merge now?

djmitche:
Yes, I'm happy with this (sorry for the delay)

PR #2765, opened 2025-06-02
Std traits exercise: Add link to playground

PaulXiCao:


src/std-traits/exercise.md
```diff
@@ -6,7 +6,8 @@ minutes: 30
 
 In this example, you will implement the classic
 ["ROT13" cipher](https://en.wikipedia.org/wiki/ROT13). Copy this code to the
-playground, and implement the missing bits. Only rotate ASCII alphabetic
+[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d3ad23a195b8c8f56a007cc6b1ab8eb8)
+, and implement the missing bits. Only rotate ASCII alphabetic
```
djmitche, line 10: This encodes the _current_ content of this exercise in the link, but that would not be automatically updated if/when the exercise changes. If there's a way to address that and ensure that the link always includes precisely the text below it on the slide, that would be great. Otherwise, I think either a link to the playground homepage or no link at all would be best.

A quick search shows that we have linked to specific playground gists in a few other places -- we should probably remove those as well!

PaulXiCao: Found [this crate](https://docs.rs/playground-api/latest/playground_api/struct.Client.html#method.gist_create). Maybe we can build something that automatically gets a playground link to the current code example. Possibly as a separate ci job?

djmitche: Oo, that's a good idea! Do you want to make an issue to track it?

Issue #2767, opened 2025-06-03
Speed up publish CI pipeline


michael-kerscher:
Currently a run of the [publish pipeline](https://github.com/google/comprehensive-rust/actions/workflows/publish.yml) is taking around 30 minutes.

Most of the time is spent in the "build all translation" step.

https://github.com/google/comprehensive-rust/blob/c1773a7fdbfc493f29005af6f3713745c6674d5d/.github/workflows/publish.yml#L55-L60

This is the main loop for building the translations. This is currently writing files into the `book/<language>/html` folder, then moved into `book/html/<language>`. The build.sh script is heavily modifying the current state of the repository so this is not trivially parallelized in the same step.

Options:
- separate steps by using Githubs [job matrix](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#using-a-matrix-strategy) to run each translation as a separate job. Note: this requires all other steps in that job to run multiple times which is wasting resources. Installing mdbook in that environment should be cached properly. The English translation should then also be handled like other languages as this is currently a special case.
- Alternative: Split the translation step into two steps (likely more efficient)
  - (for each language) restore the entire book source as currently and copy all necessary files into a new `source/<language>/` directory
  - With all sources in each `source/<language>` folder, this can be easily parallelized after modifying the build script.

https://github.com/google/comprehensive-rust/blob/c1773a7fdbfc493f29005af6f3713745c6674d5d/.github/workflows/build.sh#L27-L28

![Image](https://github.com/user-attachments/assets/9901ef1b-b77f-4ba0-a06c-b0d7c2383c2c)

michael-kerscher:
Current effort in refactoring the build.sh script https://github.com/google/mdbook-i18n-helpers/pull/268 and the issue https://github.com/google/mdbook-i18n-helpers/pull/268.

From my perspective the new tool could be an opportunity to create such an independent copy of all sources.

djmitche:
Do you have any idea what is bounding that step? CPU? Disk bandwidth? I don't know how many cores the runners have, but either of those might mean parallelization within the step will not reduce overall runtime very much.

michael-kerscher:
[...]

PR #2772, opened 2025-06-03
ci: parallelize the mdbook build <language> process

michael-kerscher:
Make the mdbook build process parallel. This is a simple approach by just cloning the entire repository into new directories and running the process in parallel. This is a demonstration for #2767.

The process appears to be mostly CPU bound and not memory or disk heavy so the amount of cores on the CI runner matter.
mdbook build does not seem to use more than 1 core and a typical github CI machine [has 4 cores](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories). So the expected speedup would be 4x.

Currently 30 minutes are used for publish, 25 minutes of this is build process, this would be reduced to 5 + 6.25 = ~12 minutes, thus more than halving the process.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -53,12 +53,10 @@ jobs:
         run: .github/workflows/build.sh en book
 
       - name: Build all translations
+        # note: build.sh here uses the given book path relative to source/{language}
         run: |
-          for po_lang in ${{ env.LANGUAGES }}; do
-              .github/workflows/build.sh $po_lang book/$po_lang
-              mv book/$po_lang/html book/html/$po_lang
-          done
-
+          parallel -i .github/workflows/build.sh {} book/{} -- ${{ env.LANGUAGES }}
+          parallel -i mv source/{}/book/{}/html book/html/{} -- ${{ env.LANGUAGES }}
       - name: Build translation report
         run: i18n-report report book/html/translation-report.html po/*.po
 
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -15,6 +15,19 @@ set -Eeuo pipefail
 book_lang=${1:?"Usage: $0 <book-lang> <dest-dir>"}
 dest_dir=${2:?"Usage: $0 <book-lang> <dest-dir>"}
 
+# ensure source dir exists
+mkdir -p source
+# clean previous build artifacts
+SOURCE_DIR="source/${book_lang}"
+rm -rf "$SOURCE_DIR"
+mkdir "$SOURCE_DIR"
+
+# clone the current state into the source directory
+git clone . "$SOURCE_DIR"
+
+# now work from that new directory as the base directory
+cd "$SOURCE_DIR"
+
 if [ "$book_lang" = "en" ]; then
     echo "::group::Building English course"
 else


michael-kerscher:
Down from 25 minutes to 11 minutes. Less than expected, but also less than half the previous run time

djmitche:
Nice!!

michael-kerscher:
This now finished (even with missing rust caches!) in 21 minutes. https://github.com/google/comprehensive-rust/actions/runs/15612132409 With caching this would get down quite a lot.

@mgeisler can you review the [comprehensive-rust-all](https://github.com/google/comprehensive-rust/actions/runs/15612132409/artifacts/3314829885) artifact if that is the correct structure? This would show that uploading the intermediate steps and artifact downloading with merge works as expected.

The workflow still needs to be refactored a little bit as there is quite some duplication now but this only serves as a POC as of now

michael-kerscher:
@djmitche @mgeisler any thoughts on the latest approach?

djmitche (commented):
I like this approach, but I admit I don't know much about GitHub actions so if there are any subtle gotcha's I would not be the person to detect them.

.github/workflows/publish.yml
```diff
@@ -50,14 +64,49 @@ jobs:
...
+      - name: Setup Rust cache
+        uses: ./.github/workflows/setup-rust-cache
+
+      - name: Install Gettext
+        run: |
+          sudo apt update
+          sudo apt install gettext
+
+      - name: Install mdbook
+        uses: ./.github/workflows/install-mdbook
```
mgeisler: Small question: do we need to install Rust, `mdbook`, and the Gettext tools here? I would guess not since everything is built already, no? We're just putting the pieces together and uploading it all?

.github/workflows/publish.yml
```diff
@@ -69,14 +118,8 @@ jobs:
...
-      - name: Upload artifact
-        uses: actions/upload-pages-artifact@v3
+      - name: Upload all translations (for verification)
+        uses: actions/upload-artifact@v4
         with:
-          path: book/html
-
-      - name: Deploy to GitHub Pages
-        id: deployment
-        uses: actions/deploy-pages@v4
```
mgeisler, line 82: I'm surprised this went away? Is this not the most important part, so to speak? :smile:

.github/workflows/publish.yml
```diff
@@ -3,31 +3,45 @@ name: Publish
...
-      name: github-pages
-      url: ${{ steps.deployment.outputs.page_url }}
+  create-translation:
+    strategy:
+      matrix:
+        language:
+          [
+            "en",
+            "ar",
+            "bn",
```
mgeisler: From what I understand of YAML, this can be simplified to

```suggestion
        language:
          - "en"
          - "ar"
          - "bn"
```

and so on, which seems ever so slightly more high level.

mgeisler:
> I like this approach, but I admit I don't know much about GitHub actions so if there are any subtle gotcha's I would not be the person to detect them.

I think this is a great approach, well done, @michael-kerscher! Originaly, we only had a simple `mdbook build`, which is very fast. But now, I guess the PDF generation is what slows things down dramatically? So it makes sense to do this in parallel.

As for the caching, if building `mdboo` (and other binaries) takes a long time, we could build them just _once_, cache them, and reuse them in each job going forward. I forget if we're able to run 20+ jobs in paralle, or if they run, say 4 or 8 at a time? If it's the latter, then building `mdbook` once could help here

.github/workflows/publish.yml
michael-kerscher: yes, it is! This was removed for development and testing purposes to not accidentally deploy stuff

.github/workflows/publish.yml
michael-kerscher: No... but we do ;)

We actually only need i18n-report, that is installed in the install-mdbook step https://github.com/google/comprehensive-rust/pull/2772/commits/cd922c46b4a37ba4b185d30c5cc5c5aa07391ad7

michael-kerscher: The code changed over time a bit but essentially  `cargo xtask install-tools` installs the necessary tool at this moment. For simplification and POC this is just installed as is. This needs to be refactored

michael-kerscher: I don't know exactly if i18n-report requires gettext.

I could upload the .cargo/bin/i18n-report binary in the previous job then download it in all jobs to use use that. This has the potential to be a bit more obscure in the future. But it probably is already pretty obscure where this is from at this point.

michael-kerscher: while doing this and already making this a bit more obscure, I could create a tool-building phase that uploads all rust tools so we can upload the entire .cargo/bin/ directory and just use this in every translation

mgeisler: > I don't know exactly if i18n-report requires gettext

The way I remember it is that it used the Rust library directly, but @qwandor will have the details 🙂

I agree with you that this can and should be further refactored and untangled.

mgeisler: > uploads all rust tools so we can upload the entire .cargo/bin/ directory and just use this in every translation

Yes, that would work! Alternatively and perhaps better: we don't need to build things like `mdbook`, we can download a binary from https://github.com/rust-lang/mdBook/releases.

There is `cargo-binstall` (https://github.com/cargo-bins/cargo-binstall) which I think can help?

The cool way would then be to publish have a little GitHub action so that it becomes very convenient to use `mdbook`. We can publish binaries for `mdbook-i18n-helpers` as well and install them the same way.

michael-kerscher: I noticed we need gettext because msgmerge is used in line 69 https://github.com/google/comprehensive-rust/blob/2af48cff2eb225f5937853f8a3b5366e51c3666c/.github/workflows/publish.yml#L65-L70

mgeisler: I see, thanks for the research!

.github/workflows/publish.yml
michael-kerscher: I added the original code back with comments so I don't forget to re-add this once this is ready

.github/workflows/publish.yml
```diff
@@ -3,31 +3,43 @@ name: Publish
 # See also TRANSLATIONS.md.
 
 on:
+  pull_request:
```
michael-kerscher, line 6: this needs to be removed before merging

.github/workflows/publish.yml
```diff
@@ -3,31 +3,43 @@ name: Publish
...
   push:
     branches:
       - main
   workflow_dispatch:
 
-permissions:
```
michael-kerscher, line 11: this needs to be re-added before merging

michael-kerscher:
With #2915 merged, I was able to remove the extra build step, creating and downloading artifacts again. The change is now reduced to the split into 
- create-translation: produces a language artifact with the build.sh
- publish: loads all language artifacts, does the report and publishing of everything.

the previous sequential build of the language is now using individual jobs to do this in parallel and it looks way better.

Current state: For maximum efficiency we still (according to the current builds) need binaries for
- mdbook-svgbob
- i18n-report

- mdbook-linkcheck2 requires some additional thinking: e.g. build a binary in this repository so it is available for cargo binstall.

and we could shave some minutes of the process by caching the installation that is pretty slow e.g.  with https://github.com/marketplace/actions/install-and-cache-apt-tools

mgeisler:
> and we could shave some minutes of the process by caching the installation that is pretty slow e.g. with https://github.com/marketplace/actions/install-and-cache-apt-tools

Ah, fun, I had a similar idea today and put #2916 up for review. Looking over [the code](https://github.com/tecolicom/actions-use-apt-tools/blob/main/action.yml), I don't see how this can be quicker than downloading things from the (internal) Apt mirrors GitHub maintains.

My thinking with regards to caching is that it should be used when the thing being cached takes a significant amount of time to _create_. Unpacking binaries from the GitHub cache or from an Apt mirror ought to be comparable in time.

In #2916, I found that most time was spent on updating the man page database... So telling `dpkg` to not install documentation helped a ton. My guess is that this could be why the GH Action you link to exists.

.github/workflows/publish.yml
```diff
@@ -3,31 +3,43 @@ name: Publish
...
+          - "ko"
+          - "pl"
+          - "pt-BR"
+          - "ro"
+          - "ru"
+          - "tr"
+          - "uk"
+          - "vi"
+          - "zh-CN"
+          - "zh-TW"
```
mgeisler, line 42: This is fine (and easy to sort!), but just in case you hadn't thought of it,

```yaml
language: ["en", "ar", ...]`
```

would be equivalent.

.github/workflows/publish.yml
mgeisler: Thanks, makese sense! :smile:

.github/workflows/publish.yml
mgeisler: Seeing this, I would guess we could get a speed-up if `i18n-report` was moved to a job running in parallel with all the translations.

The final publish job would then become truly simple: it would gather all the HTML from the translations, plus the HTML from `i18n-report`, and publish it all.

michael-kerscher:
> > and we could shave some minutes of the process by caching the installation that is pretty slow e.g. with https://github.com/marketplace/actions/install-and-cache-apt-tools
> 
> Ah, fun, I had a similar idea today and put #2916 up for review. Looking over [the code](https://github.com/tecolicom/actions-use-apt-tools/blob/main/action.yml), I don't see how this can be quicker than downloading things from the (internal) Apt mirrors GitHub maintains.
> 
> My thinking with regards to caching is that it should be used when the thing being cached takes a significant amount of time to _create_. Unpacking binaries from the GitHub cache or from an Apt mirror ought to be comparable in time.
> 
> In #2916, I found that most time was spent on updating the man page database... So telling `dpkg` to not install documentation helped a ton. My guess is that this could be why the GH Action you link to exists.

Exactly that was my thought, installing the packages did something with the content of the packages and processing the documentation. You approach is even better as there is no use for documentation in the CI environment (at least not yet, AI is not embedded deep enough that it requires that documentation :D)

.github/workflows/publish.yml
```diff
@@ -50,14 +62,46 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Update Rust
+        run: rustup update
+
+      - name: Install Gettext
+        run: |
+          sudo apt update
+          sudo apt install gettext
```
mgeisler, line 94: This should now use the new `apt-get-install` action 🎉

.github/workflows/publish.yml
mgeisler: Feel free to resolve this and then we can potentially move `i18n-report` later!

.github/workflows/publish.yml
```diff
@@ -69,14 +113,21 @@ jobs:
           for file in synced-po/*.po; do msgmerge --update $file synced-po/messages.pot ; done
           i18n-report report book/html/synced-translation-report.html synced-po/*.po
 
-      - name: Setup Pages
-        uses: actions/configure-pages@v5
-
-      - name: Upload artifact
-        uses: actions/upload-pages-artifact@v4
+      # TODO: remove this before merging into main. This is just for verification of the build result while developing this workflow.
```
mgeisler, line 116: Just a comment here so we don't forget.

.github/workflows/publish.yml
```diff
@@ -69,14 +113,21 @@ jobs:
...
+        uses: actions/upload-artifact@v4
         with:
-          path: book/html
-
-      - name: Deploy to GitHub Pages
-        id: deployment
-        uses: actions/deploy-pages@v4
+          name: comprehensive-rust-all
+          path: book/
+      # TODO: uncomment the following lines before merging into main. This is removed for development purposes.
```
mgeisler, line 122: Another comment to ensure we don't forget.

.github/workflows/publish.yml
michael-kerscher: I think it was dprint that forced me to this solution

PR #2774, opened 2025-06-04
Add xtask support for refresh slide list.

### LINKED ISSUE #2744: Implement parameter for cargo xtask web-tests to refresh slide list to test in the web tests
with #2743 implementing a cargo xtask web-tests command there is opportunity for a parameter of this command to make life easier.

There is a script [create-slide.list.sh](https://github.com/google/comprehensive-rust/blob/ce9d74742dd6c343b46237ab09d363c223a5a004/tests/src/slides/create-slide.list.sh) that is used to create the list of available slides on the system.

It requires either the TEST_BOOK_DIR env variable (defined in the CI script) or the directory as the first parameter (wdio.conf.ts defaults to ../book/html if the environment is not set).

That could be a parameter of this command to run the script. Something along the lines of "--refresh-slide-list" or shorter.

egithinji:
Adds support for a `refresh-slide-list` argument when running the command `cargo xtask web-tests`. Allows one to also specify an optional book html directory **if** one uses the `refresh-slide-list` argument. Fixes #2744 .

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -45,18 +45,15 @@ struct Cli {
 #[derive(Subcommand)]
 enum Task {
     /// Installs the tools the project depends on.
-    InstallTools {},
+    InstallTools,
     /// Runs the web driver tests in the tests directory.
     WebTests {
-        /// (Re)generate the list of slides that are tested by the slide size test.
-        #[arg(short, long, group = "refresh")]
-        refresh_slide_list: bool,
-        /// Optional book html directory
-        #[arg(short, long, requires = "refresh")]
+        /// Optional 'book html' directory - if set, will also refresh the list of slides used by slide size test.
+        #[arg(short, long)]
         dir: Option<String>,
     },
     /// Tests all included Rust snippets.
-    RustTests {},
+    RustTests,
     /// Starts a web server with the course.
     Serve {},
     /// Create a static version of the course in the `book/` directory.
@@ -67,9 +64,7 @@ fn execute_task() -> Result<()> {
     let cli = Cli::parse();
     match cli.task {
         Task::InstallTools {} => install_tools()?,
-        Task::WebTests { refresh_slide_list, dir } => {
-            run_web_tests(refresh_slide_list, dir)?
-        }
+        Task::WebTests { dir } => run_web_tests(dir)?,
         Task::RustTests {} => run_rust_tests()?,
         Task::Serve {} => start_web_server()?,
         Task::Build {} => build()?,
@@ -121,23 +116,19 @@ fn install_tools() -> Result<()> {
     Ok(())
 }
 
-fn run_web_tests(refresh_slide_list: bool, dir: Option<String>) -> Result<()> {
+fn run_web_tests(dir: Option<String>) -> Result<()> {
     println!("Running web tests...");
 
     let path_to_tests_dir = Path::new(env!("CARGO_WORKSPACE_DIR")).join("tests");
     let refresh_slides_script = "./src/slides/create-slide.list.sh".to_string();
 
-    if refresh_slide_list {
+    if let Some(d) = &dir {
         println!("Refreshing slide lists...");
-        let mut command = Command::new(refresh_slides_script);
-        command.current_dir(path_to_tests_dir.to_str().unwrap());
-        // refresh_slide_list optionally takes a directory as a parameter
-        if let Some(d) = dir {
-            command.arg(d);
-        }
-
-        let status =
-            command.status().expect("Failed to execute create-slide.list.sh");
+        let status = Command::new(refresh_slides_script)
+            .current_dir(path_to_tests_dir.to_str().unwrap())
+            .arg(d)
+            .status()
+            .expect("Failed to execute create-slide.list.sh");
 
         if !status.success() {
             let error_message = format!(
@@ -148,11 +139,14 @@ fn run_web_tests(refresh_slide_list: bool, dir: Option<String>) -> Result<()> {
         }
     }
 
-    let status = Command::new("npm")
-        .current_dir(path_to_tests_dir.to_str().unwrap())
-        .arg("test")
-        .status()
-        .expect("Failed to execute npm test");
+    let mut command = Command::new("npm");
+    command.current_dir(path_to_tests_dir.to_str().unwrap());
+    command.arg("test");
+
+    if let Some(d) = dir {
+        command.env("TEST_BOOK_DIR", d);
+    }
+    let status = command.status().expect("Failed to execute npm test");
 
     if !status.success() {
         let error_message = format!(
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -21,7 +21,7 @@
 
 use anyhow::{Ok, Result, anyhow};
 use clap::{Parser, Subcommand};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::{env, process::Command};
 
 fn main() -> Result<()> {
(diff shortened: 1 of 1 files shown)


xtask/src/main.rs
```diff
@@ -38,32 +38,41 @@ fn main() -> Result<()> {
...
+    #[command(subcommand)]
     task: Task,
 }
 
-#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, ValueEnum)]
+#[derive(Subcommand)]
 enum Task {
     /// Installs the tools the project depends on.
-    InstallTools,
+    InstallTools {},
```
michael-kerscher: empty {} are not necessary. enum variants can have different structures. please remove the empty ones

xtask/src/main.rs
```diff
@@ -38,32 +38,41 @@ fn main() -> Result<()> {
...
     /// Runs the web driver tests in the tests directory.
-    WebTests,
+    WebTests {
+        /// (Re)generate the list of slides that are tested by the slide size test.
+        #[arg(short, long, group = "refresh")]
```
michael-kerscher: I thought about this for a bit. In the current build process, we use the environment variable that is set to specify the directory, so the refresh script does not require this as a commandline argument.
Reason for this was that initially I did not have a good way to provide the directory to the wdio environment so I used environment variables and reused this then for the script.

Now that we have a command that is calling both the refresh and the npm test in the same command, we could set the environment variable `TEST_BOOK_DIR` for the npm process directly in this command and always require the directory to be provided to the xtask subcommand

Then we could merge the --refresh_slide_list that is just a boolean with the --dir that requires the first one into one argument. The user needs to specify this anyways.

What do you think?

egithinji: Just to make sure I understand correctly, are you suggesting we do the following:
- do away with the --refresh_slide_list flag
- always refresh the slides when web-tests is called
- make the --dir arg mandatory
- assign the value passed via --dir to a`TEST_BOOK_DIR` env variable that is used by both create-slide.list.sh and npm?

michael-kerscher: > do away with the --refresh_slide_list flag
- yes, but rename --dir to something that is recognizable that this will refresh the slide list
> always refresh the slides when web-tests is called
- no, refresh if --dir is provided
> make the --dir arg mandatory
- no, if not provided, do not refresh
> assign the value passed via --dir to aTEST_BOOK_DIR env variable that is used by both create-slide.list.sh and npm?
- I don't particularly care about how the create-slide-list.sh is called exactly, either is fine for the script, but as npm needs the TEST_BOOK_DIR anyways, I would like to see the environment variable set (and overwrite if already set, you don't need to check if it was set). So create-slide.list.sh does not need this explicitly set as an command line argument as it also reads that environment variable.

xtask/src/main.rs

xtask/src/main.rs
egithinji: Please check now. Left the arg named --dir but explained its purpose in the help. If --dir is set, `create-slide-list.sh` will run and the directory specified by --dir will also be set as the env variable for `npm test`.

xtask/src/main.rs
```diff
@@ -38,32 +38,36 @@ fn main() -> Result<()> {
...
+    WebTests {
+        /// Optional 'book html' directory - if set, will also refresh the list of slides used by slide size test.
+        #[arg(short, long)]
+        dir: Option<String>,
+    },
     /// Tests all included Rust snippets.
     RustTests,
     /// Starts a web server with the course.
-    Serve,
+    Serve {},
```
michael-kerscher: this empty struct could be just a plain enum variant (without {})

xtask/src/main.rs
```diff
@@ -38,32 +38,36 @@ fn main() -> Result<()> {
...
     /// Create a static version of the course in the `book/` directory.
-    Build,
+    Build {},
```
michael-kerscher: remove {}

xtask/src/main.rs
```diff
@@ -38,32 +38,36 @@ fn main() -> Result<()> {
...
 
 fn execute_task() -> Result<()> {
     let cli = Cli::parse();
     match cli.task {
-        Task::InstallTools => install_tools()?,
-        Task::WebTests => run_web_tests()?,
-        Task::RustTests => run_rust_tests()?,
-        Task::Serve => start_web_server()?,
-        Task::Build => build()?,
+        Task::InstallTools {} => install_tools()?,
```
michael-kerscher: also remove the empty {} in this match

xtask/src/main.rs
```diff
@@ -112,16 +116,37 @@ fn install_tools() -> Result<()> {
     Ok(())
 }
 
-fn run_web_tests() -> Result<()> {
+fn run_web_tests(dir: Option<String>) -> Result<()> {
     println!("Running web tests...");
 
     let path_to_tests_dir = Path::new(env!("CARGO_WORKSPACE_DIR")).join("tests");
+    let refresh_slides_script = "./src/slides/create-slide.list.sh".to_string();
```
michael-kerscher: please use a path here

michael-kerscher:
I'm sorry for the huge delay. I already prepared my review but somehow did not click the submit button...

xtask/src/main.rs

xtask/src/main.rs

xtask/src/main.rs

xtask/src/main.rs

egithinji:
Hi @michael-kerscher . I noticed a difference in behavior when passing a directory using the full path (e.g. `cargo xtask web-tests -d /usr/.../comprehensive-rust/book/html`) vs from relative the workspace directory (e.g. `cargo xtask web-tests -d book/html`). 

When using the full path, the slide list gets recreated but it seems like the actual testing of the slide dimensions doesn't happen. When using the shorter path from the workspace directory, it seems to work as expected. Is there a difference in how these paths are treated from within the web tests?

michael-kerscher:
Hmm, if that is an issue for the actual test runner, this would be some interaction with npm or something inside wdio. I need to look at this

xtask/src/main.rs
```diff
@@ -112,16 +116,39 @@ fn install_tools() -> Result<()> {
...
+    command.current_dir(path_to_tests_dir.to_str().unwrap());
+    command.arg("test");
 
-    let status = Command::new("npm")
-        .current_dir(path_to_tests_dir.to_str().unwrap())
-        .arg("test")
-        .status()
-        .expect("Failed to execute npm test");
+    if let Some(d) = dir {
+        command.env("TEST_BOOK_DIR", d);
```
michael-kerscher, line 151: this is taking the directory provided by the user in direct way and stored in TEST_BOOK_DIR.
This is no issue with absolute paths, as an absolute path always resolves correctly, but if a relative path e.g. "book/html" is being used this causes issues as this is relative to the folder of the working directory (here the CARGO_WORKSPACE_DIR/tests, leading to a path like ..../tests/book/html. The user would need to specify ../book/html to correctly specify this directory. But this conflicts with the above usage (line 130) where the path is relative to the CARGO_WORKSPACE_DIR directly.

We can make our life much more simple by providing a canonical representation (absolute path) to **both** the above script and the npm command. Path (PathBuf) supports a `canonicalize()` function that provides the canonic absolute representation of the path. Please use that for the above script and this call. This should resolve your issue

```
let absolute_dir = dir.map(|d| d.canonicalize()).transpose()?;
```

egithinji: Yes indeed that was the issue. I've changed to convert the dir argument to an absolute path and the behavior is now consistent whether user passes a relative or absolute directory. Also we get an error if they pass an invalid directory.

PR #2775, opened 2025-06-11
Updates to bare metal day after teaching it again

qwandor:
This fixes some issues running some of the examples, and various other minor issues I found along the way.


src/exercises/bare-metal/rtc/memory.ld
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+MEMORY
+{
+	image : ORIGIN = 0x40080000, LENGTH = 2M
+}
```
djmitche, line 20: This is a lot simpler!!

qwandor: Right, the rest of the linker script comes from `aarch64-rt` now.

PR #2776, opened 2025-06-16
Support setting language and output directory in cargo xtask

egithinji:
In addition to simplifying building locally (no need to set an environment variable), this makes it possible to use the `cargo xtask build` command in the CI and specify any output location, rather than relying on the build.sh script.


xtask/src/main.rs
```diff
@@ -51,9 +51,25 @@ enum Task {
...
+    },
+    /// Create a static version of the course.
+    Build {
+        /// ISO 639 language code (e.g. da for the Danish translation).
+        #[arg(short, long)]
+        language: Option<String>,
+
+        /// Directory to place the build. If not provided, defaults to the book/ directory (or the book/xx directory if a language is provided).
+        #[arg(short, long)]
+        output: Option<String>,
```
michael-kerscher: `Option<PathBuf>`

xtask/src/main.rs
```diff
@@ -51,9 +51,25 @@ enum Task {
...
-    /// Create a static version of the course in the `book/` directory.
-    Build,
+    Serve {
+        /// ISO 639 language code (e.g. da for the Danish translation).
+        #[arg(short, long)]
+        language: Option<String>,
+
+        /// Directory to place the build. If not provided, defaults to the book/ directory (or the book/xx directory if a language is provided).
+        #[arg(short, long)]
+        output: Option<String>,
```
michael-kerscher: You could use an `Option<PathBuf>` for better semantics. If we want to do anything with the provided path we would get a lot of methods for free if we have the correct semantic type

xtask/src/main.rs
```diff
@@ -156,15 +170,29 @@ fn run_rust_tests() -> Result<()> {
...
+        command.env("MDBOOK_BOOK__LANGUAGE", &language);
+    }
+
+    // If the 'output' arg is specified by the caller, use that, otherwise output to the 'book/' directory
+    // (or the 'book/xx' directory if a language was specified).
+    command.arg("-d");
+    if let Some(d) = output {
+        command.arg(d);
+    } else {
+        command.arg(format!("book/{}", language.unwrap_or("".to_string())));
```
michael-kerscher: This is a good candidate for a `Path` otherwise we have directory separator issues e.g. with Windows

xtask/src/main.rs
```diff
@@ -176,15 +204,29 @@ fn start_web_server() -> Result<()> {
     Ok(())
 }
 
-fn build() -> Result<()> {
+fn build(language: Option<String>, output: Option<String>) -> Result<()> {
```
michael-kerscher: PathBuf for consistency with the commandline args

xtask/src/main.rs
```diff
@@ -156,15 +170,29 @@ fn run_rust_tests() -> Result<()> {
     Ok(())
 }
 
-fn start_web_server() -> Result<()> {
+fn start_web_server(language: Option<String>, output: Option<String>) -> Result<()> {
```
michael-kerscher: `output: Option<Pathbuf>` for consistency with the PathBuf in the arguments

xtask/src/main.rs
```diff
@@ -176,15 +204,29 @@ fn start_web_server() -> Result<()> {
...
-    let status = Command::new("mdbook")
-        .current_dir(path_to_workspace_root.to_str().unwrap())
-        .arg("build")
-        .status()
-        .expect("Failed to execute mdbook build");
+    let mut command = Command::new("mdbook");
+    command.current_dir(path_to_workspace_root.to_str().unwrap());
+    command.arg("build");
+
+    if let Some(language) = &language {
```
michael-kerscher, line 212: no change but idea for future: we could even check if the relevant `<language>.po` file exists

michael-kerscher (changes requested):

xtask/src/main.rs
egithinji: Done. Thanks didn't know Clap could accept PathBuf! This is much better.

xtask/src/main.rs

xtask/src/main.rs

xtask/src/main.rs
egithinji: Noted.

xtask/src/main.rs

michael-kerscher (approved):
looks good to me

PR #2780, opened 2025-06-19
Update cargo.md

mgeisler:
Now that there is a `rustup` package, we can take away the scary language and ask people to install that. This has the advantage that people get a trusted tool to download the latest Rust compiler.

### DIFF
--- a/src/cargo.md
+++ b/src/cargo.md
@@ -22,12 +22,10 @@ auto-completion and jump-to-definition functionality for [VS Code], [Emacs],
 
 <details>
 
-- On Debian/Ubuntu, you can also install Cargo, the Rust source and the
-  [Rust formatter] via `apt`. However, this gets you an outdated Rust version
-  and may lead to unexpected behavior. The command would be:
+- On Debian/Ubuntu, you can install `rustup` via `apt`:
 
   ```shell
-  sudo apt install cargo rust-src rustfmt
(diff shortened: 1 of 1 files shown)


src/cargo.md
```diff
@@ -22,12 +22,10 @@ auto-completion and jump-to-definition functionality for [VS Code], [Emacs],
 
 <details>
 
-- On Debian/Ubuntu, you can also install Cargo, the Rust source and the
-  [Rust formatter] via `apt`. However, this gets you an outdated Rust version
-  and may lead to unexpected behavior. The command would be:
+- On Debian/Ubuntu, you can install `rustup` via `apt`:
```
djmitche, line 25: Is there a step after this, to run `rustup`, or does that occur automatically?

michael-kerscher: once rustup is installed, this does not yet have a default toolchain installed. You could append `rustup default stable` to achieve that goal. And maybe link to https://wiki.debian.org/Rust (Using crates from the Internet (with rustup))

michael-kerscher: additional hint: if you installed via some other method into your home directory already you might want to use `rustup self uninstall` to avoid using the manually installed version - otherwise this might be unexpected

michael-kerscher:
Oh, I missed the enabled autosubmit.

PR #2781, opened 2025-06-22
Update speaker notes for arrays.md with clarification on runtime check optimization

daivinhtran:
According to the error I got from trying out-of-bound index, array accesses seem to be checked at compile time.

```
error: this operation will panic at runtime
 --> src/main.rs:3:5
  |
3 |     a[7] = 0;
  |     ^^^^ index out of bounds: the length is 5 but the index is 7
  |
  = note: `#[deny(unconditional_panic)]` on by default

error: could not compile `playground` (bin "playground") due to 1 previous error
```

### DIFF
--- a/src/tuples-and-arrays/arrays.md
+++ b/src/tuples-and-arrays/arrays.md
@@ -27,7 +27,7 @@ fn main() {
   later.
 
 - Try accessing an out-of-bounds array element. Array accesses are checked at
-  runtime. Rust can usually optimize these checks away, and they can be avoided
+  compile time. Rust can usually optimize these checks away, and they can be avoided
   using unsafe Rust.
 
 - We can use literals to assign values to arrays.
--- a/src/tuples-and-arrays/arrays.md
+++ b/src/tuples-and-arrays/arrays.md
@@ -27,8 +27,8 @@ fn main() {
   later.
 
 - Try accessing an out-of-bounds array element. Array accesses are checked at
-  compile time. Rust can usually optimize these checks away, and they can be avoided
-  using unsafe Rust.
+  compile time. Rust can usually optimize these checks away, and they can be
+  avoided using unsafe Rust.
 
 - We can use literals to assign values to arrays.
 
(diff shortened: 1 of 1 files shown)


src/tuples-and-arrays/arrays.md
```diff
@@ -27,8 +27,8 @@ fn main() {
   later.
 
 - Try accessing an out-of-bounds array element. Array accesses are checked at
-  runtime. Rust can usually optimize these checks away, and they can be avoided
-  using unsafe Rust.
+  compile time. Rust can usually optimize these checks away, and they can be
```
djmitche: The original meaning was correct, and important.  In more sophisticated cases than this simple example, Rust may very well leave the runtime bounds-checks in place, and these can result in a substantial performance overhead above unchecked languages like C++.

It's difficult to demonstrate this in a short code sample because the checks are often optimized away at compile time -- in this case to an unconditional panic.  I suspect [`black_box`](https://doc.rust-lang.org/stable/std/hint/fn.black_box.html) could be useful to avoid this optimization. Do you want to try adjusting the example so that it shows the runtime bounds check?

djmitche (changes requested):

src/tuples-and-arrays/arrays.md
tidefield: Ah I see what you mean. I tried using `black_box` but it seems to not prevent the compile-time safety checks. I changed the out-of-bound index to be returned from a function and add that example to below the note. PTAL. Let me know what you think.

djmitche: Nice! The speaker note still says "compile time" though. It is also a good idea to add a sentence about how the optimization is so good that it's hard to give an example of runtime checks failing, and here requires a function to generate the index.

tidefield: Oops. PTAL again.

djmitche (approved):
Awesome, thanks!!

djmitche:
(just needs formatting fixed)

PR #2786, opened 2025-06-25
Make pattern matching exercise code editable

randomPoison:
I swear I already did this one...

@djmitche it looks like you had removed this `editable` after I last added it, was there some reason for that?


djmitche (approved):
The idea was to have > @djmitche it looks like you had removed this `editable` after I last added it, was there some reason for that?

I removed `editable` in all of the exercises to encourage students to use the playground instead of the slide (where an errant left or right arrow can delete their work). But this is an "intro" code block where a student wouldn't write their solution anyway, so I think it's fine to keep `editable` here.

PR #2788, opened 2025-06-25
Add note about trait bound in closures exercise

randomPoison:
I've tried removing this trait bound thinking it's unnecessary only to then be surprised that it breaks type inference. I think that's worth calling out to students, so I think a speaker note here would be appropriate.


djmitche (approved):
Nice, this is the kind of quality stuff the students love :)

PR #2789, opened 2025-06-25
Move closures section to before libstd sections

randomPoison:
A lot of stuff in the standard library makes use of closures, so I think it makes sense to cover those first so that we can more easily reference libstd functions like `map` and `filter`.


fw-immunant:
This makes sense as explained, but as performed it moves the closures section into the middle of the standard library discussion, between types and traits. Since many methods of common types rely on closures, I think closures should be covered prior to the types content as well.

randomPoison:
@fw-immunant oh good catch. I meant to move it to before the std types as well, idk how I messed that up. I'll throw up another PR to fix that.

Issue #2732, opened 2025-05-05
Binary tree exercise is a binary *search* tree


fw-immunant:
The exercise should be explicit that the trees involved are binary *search* trees, that is, they are binary trees with the additional invariant that all values in the left subtree of a node compare less than the value stored at that node and the same for values that compare greater on the right. We rely on this property for the implementations of the various methods, so we should be clear about it.

PR #2790, opened 2025-06-25
Add note to binary-tree exercise about duplicate values

randomPoison:
Just a small note to clarify a minor point of confusion with the exercise.


fw-immunant:
This is because it isn't simply a binary tree, it's a binary search tree (see #2732). I think saying so would make this clearer, though it doesn't hurt to also explicitly mention that values are only present at most once.

djmitche (approved):
I agree with @fw-immunant's comments too

gribozavr (approved):

PR #2792, opened 2025-06-27
Add section about UART driver without safe-mmio back in

qwandor:
Having taught the course again since the change to use safe-mmio, it seems worth keeping this as an intermediate step rather than jumping straight from pointer arithmetic to safe-mmio's abstractions.


src/bare-metal/aps/safemmio/using.md
```diff
@@ -0,0 +1,15 @@
+# Using It
```
djmitche, line 1: It looks like `better-uart/using.md` needs removed?

src/bare-metal/aps/better-uart/driver.md
```diff
@@ -3,22 +3,15 @@
...
-  to work around a bug where the compiler can emit instructions that prevent
-  MMIO virtualisation.
-- The `field!` and `field_shared!` macros internally use `&raw mut` and
-  `&raw const` to get pointers to individual fields without creating an
-  intermediate reference, which would be unsound.
+- Note the use of `&raw const` / `&raw mut` to get pointers to individual fields
+  without creating an intermediate reference, which would be unsound.
+- The example isn't included in the slides because it is very similar to the
+  `safe-mmio` example which comes next. You can run it in QEMU with `make qemu`
+  under `src/bare-metal/aps/examples` if you need to.
```
djmitche, line 15: I'm not sure what this comment means -- this slide is, indeed, included..

djmitche (approved):
This looks good!

src/bare-metal/aps/safemmio/using.md
qwandor: You're right, thanks. Done.

src/bare-metal/aps/better-uart/driver.md
qwandor: Now it's gone.

djmitche (approved):

PR #2800, opened 2025-07-01
Introduce 'Idiomatic Rust' learning module

LukeMathWalker:
This PR introduces:

- A new section for the "Idiomatic Rust" learning module
- (The beginning of) the section on newtype patterns

### DIFF
--- a/src/idiomatic/leveraging-the-type-system.md
+++ b/src/idiomatic/leveraging-the-type-system.md
@@ -26,7 +26,7 @@ Additional items speaker may mention:
   types. ["Domain Modeling Made Functional"][1] is a great resource on the
   topic, with examples written in F#.
 
-- Despite its functional roots, functional design patterns don't translate as-is
+- Despite Rust's functional roots, functional design patterns don't translate as-is
   to Rust. For instance, extensive use of higher-kinded functions and types can
   result in code that is harder to read and maintain. Design patterns in Rust
   must take into account (and leverage!) the granular control over mutability
--- a/src/idiomatic/leveraging-the-type-system.md
+++ b/src/idiomatic/leveraging-the-type-system.md
@@ -35,8 +35,8 @@ Additional items speaker may mention:
   types in Rust.
 
   Evaluate, on a case-by-case basis, whether a more imperative approach may be
-  easier to implement. Consider using in-place mutation, relying on Rust's borrow-checker and type system to
-  control what can be mutated, and where.
+  easier to implement. Consider using in-place mutation, relying on Rust's
+  borrow-checker and type system to control what can be mutated, and where.
 
 - The same caution should be applied to object-oriented design patterns. Rust
   doesn't support inheritance, and object decomposition should take into account
--- a/src/idiomatic/leveraging-the-type-system.md
+++ b/src/idiomatic/leveraging-the-type-system.md
@@ -27,11 +27,16 @@ Additional items speaker may mention:
   types. ["Domain Modeling Made Functional"][1] is a great resource on the
   topic, with examples written in F#.
 
-- Despite Rust's functional roots, functional design patterns don't translate
-  as-is to Rust. For instance, extensive use of higher-order functions and
-  higher-kinded types can result in code that is harder to read and maintain.
-  Design patterns in Rust must take into account (and leverage!) the granular
-  control over mutability that comes with its borrow checker.
+- Despite Rust's functional roots, not all functional design patterns can be
+  easily translated to Rust.
+
+  For example, you must have a solid grasp on a broad selection of advanced
+  topics to design APIs that leverage higher-order functions and higher-kinded
+  types in Rust.
+
+  Evaluate, on a case-by-case basis, whether a more imperative approach may be
+  easier to implement, relying on Rust's borrow-checker and type system to
+  control what can be mutated, and where.
 
 - The same caution should be applied to object-oriented design patterns. Rust
   doesn't support inheritance, and object decomposition should take into account
--- a/src/idiomatic/leveraging-the-type-system.md
+++ b/src/idiomatic/leveraging-the-type-system.md
@@ -26,15 +26,15 @@ Additional items speaker may mention:
   types. ["Domain Modeling Made Functional"][1] is a great resource on the
   topic, with examples written in F#.
 
-- Despite Rust's functional roots, functional design patterns don't translate as-is
-  to Rust. For instance, extensive use of higher-order functions and higher-kinded types can
-  result in code that is harder to read and maintain. Design patterns in Rust
-  must take into account (and leverage!) the granular control over mutability
-  that comes with its borrow checker.
+- Despite Rust's functional roots, functional design patterns don't translate
+  as-is to Rust. For instance, extensive use of higher-order functions and
+  higher-kinded types can result in code that is harder to read and maintain.
+  Design patterns in Rust must take into account (and leverage!) the granular
+  control over mutability that comes with its borrow checker.
 
 - The same caution should be applied to object-oriented design patterns. Rust
-  doesn't support inheritance, and object decomposition should take into account the
-  constraints introduced by the borrow checker.
+  doesn't support inheritance, and object decomposition should take into account
+  the constraints introduced by the borrow checker.
 
 - Mention that type-level programming can be often used to create "zero-cost
   abstractions", although the label can be misleading: the impact on compile
--- a/src/idiomatic/leveraging-the-type-system.md
+++ b/src/idiomatic/leveraging-the-type-system.md
@@ -19,7 +19,8 @@ entire codebase.
 Additional items speaker may mention:
 
 - Rust's type system borrows a lot of ideas from functional programming
-  languages.\
+  languages.
+
   For example, Rust's enums are known as "algebraic data types" in languages
   like Haskell and OCaml. You can take inspiration from learning material geared
   towards functional languages when looking for guidance on how to design with
--- a/src/idiomatic/leveraging-the-type-system.md
+++ b/src/idiomatic/leveraging-the-type-system.md
@@ -35,7 +35,7 @@ Additional items speaker may mention:
   types in Rust.
 
   Evaluate, on a case-by-case basis, whether a more imperative approach may be
-  easier to implement, relying on Rust's borrow-checker and type system to
+  easier to implement. Consider using in-place mutation, relying on Rust's borrow-checker and type system to
   control what can be mutated, and where.
 
 - The same caution should be applied to object-oriented design patterns. Rust
--- a/src/idiomatic/leveraging-the-type-system.md
+++ b/src/idiomatic/leveraging-the-type-system.md
@@ -1,0 +1,47 @@
+---
+minutes: 5
+---
+
+# Leveraging the Type System
+
+Rust's type system is _expressive_.\
+We can use types and traits to build abstractions that make our code harder to
+misuse. In some cases, we can even go as far as enforcing correctness at
+_compile-time_. Quite often, these abstractions have no runtime
+overhead[^zero-cost].
+
+The type system can also be used to model concepts and constraints from your
+business domain. By designing our types carefully, we can improve the clarity
+and maintainability of the entire codebase.
+
+<details>
+
+Additional items speaker may mention:
+
+- Rust's type system borrows a lot of ideas from functional programming
+  languages.\
+  For example, Rust's enums are known as "algebraic data types" in languages
+  like Haskell and OCaml. You can take inspiration from learning material geared
+  towards functional languages when looking for guidance on how to design with
+  types. ["Domain Modeling Made Functional"][1] is a great resource on the
+  topic, with examples written in F#.
+
+- Despite its functional roots, functional design patterns don't translate as-is
+  to Rust. For instance, extensive use of higher-kinded functions and types can
+  result in code that is harder to read and maintain. Design patterns in Rust
+  must take into account (and leverage!) the granular control over mutability
+  that comes with its borrow-checker.
+
+- The same caution should be applied to object-oriented design patterns. Rust
+  doesn't support inheritance, and object boundaries must be mindful of the
+  constraints introduced by the borrow-checker.
+
+</details>
+
+{{%segment outline}}
+
+[1]: https://pragprog.com/titles/swdddf/domain-modeling-made-functional/
+
+[^zero-cost]: They often referred to as "zero-cost abstractions", although the
+    label can be misleading: the impact on compile times and code complexity may
+    be significant.
--- a/src/idiomatic/leveraging-the-type-system.md
+++ b/src/idiomatic/leveraging-the-type-system.md
@@ -4,15 +4,15 @@ minutes: 5
 
 # Leveraging the Type System
 
-Rust's type system is _expressive_.\
-We can use types and traits to build abstractions that make our code harder to
-misuse. In some cases, we can even go as far as enforcing correctness at
-_compile-time_. Quite often, these abstractions have no runtime
-overhead[^zero-cost].
+Rust's type system is _expressive_: you can use types and traits to build
+abstractions that make your code harder to misuse.
 
-The type system can also be used to model concepts and constraints from your
-business domain. By designing our types carefully, we can improve the clarity
-and maintainability of the entire codebase.
+In some cases, you can go as far as enforcing correctness at _compile-time_,
+with no runtime overhead.
+
+Types and traits can model concepts and constraints from your business domain.
+With careful design, you can improve the clarity and maintainability of the
+entire codebase.
 
 <details>
 
@@ -36,12 +36,12 @@ Additional items speaker may mention:
   doesn't support inheritance, and object boundaries must be mindful of the
   constraints introduced by the borrow-checker.
 
+- Mention that type-level abstractions are often referred to as "zero-cost
+  abstractions", although the label can be misleading: the impact on compile
+  times and code complexity may be significant.
+
 </details>
 
 {{%segment outline}}
 
 [1]: https://pragprog.com/titles/swdddf/domain-modeling-made-functional/
-
-[^zero-cost]: They often referred to as "zero-cost abstractions", although the
-    label can be misleading: the impact on compile times and code complexity may
-    be significant.
--- a/src/idiomatic/leveraging-the-type-system.md
+++ b/src/idiomatic/leveraging-the-type-system.md
@@ -27,7 +27,7 @@ Additional items speaker may mention:
   topic, with examples written in F#.
 
 - Despite Rust's functional roots, functional design patterns don't translate as-is
-  to Rust. For instance, extensive use of higher-kinded functions and types can
+  to Rust. For instance, extensive use of higher-order functions and higher-kinded types can
   result in code that is harder to read and maintain. Design patterns in Rust
   must take into account (and leverage!) the granular control over mutability
   that comes with its borrow-checker.
--- a/src/idiomatic/leveraging-the-type-system.md
+++ b/src/idiomatic/leveraging-the-type-system.md
@@ -33,7 +33,7 @@ Additional items speaker may mention:
   that comes with its borrow checker.
 
 - The same caution should be applied to object-oriented design patterns. Rust
-  doesn't support inheritance, and object boundaries must be mindful of the
+  doesn't support inheritance, and object decomposition should take into account the
   constraints introduced by the borrow-checker.
 
 - Mention that type-level abstractions are often referred to as "zero-cost
--- a/src/idiomatic/leveraging-the-type-system.md
+++ b/src/idiomatic/leveraging-the-type-system.md
@@ -34,7 +34,7 @@ Additional items speaker may mention:
 
 - The same caution should be applied to object-oriented design patterns. Rust
   doesn't support inheritance, and object decomposition should take into account the
-  constraints introduced by the borrow-checker.
+  constraints introduced by the borrow checker.
 
 - Mention that type-level abstractions are often referred to as "zero-cost
   abstractions", although the label can be misleading: the impact on compile
--- a/src/idiomatic/leveraging-the-type-system.md
+++ b/src/idiomatic/leveraging-the-type-system.md
@@ -36,7 +36,7 @@ Additional items speaker may mention:
   doesn't support inheritance, and object decomposition should take into account the
   constraints introduced by the borrow checker.
 
-- Mention that type-level abstractions are often referred to as "zero-cost
+- Mention that type-level programming can be often used to create "zero-cost
   abstractions", although the label can be misleading: the impact on compile
   times and code complexity may be significant.
 
--- a/src/idiomatic/leveraging-the-type-system.md
+++ b/src/idiomatic/leveraging-the-type-system.md
@@ -30,7 +30,7 @@ Additional items speaker may mention:
   to Rust. For instance, extensive use of higher-order functions and higher-kinded types can
   result in code that is harder to read and maintain. Design patterns in Rust
   must take into account (and leverage!) the granular control over mutability
-  that comes with its borrow-checker.
+  that comes with its borrow checker.
 
 - The same caution should be applied to object-oriented design patterns. Rust
   doesn't support inheritance, and object boundaries must be mindful of the
--- a/src/idiomatic/leveraging-the-type-system/newtype-pattern.md
+++ b/src/idiomatic/leveraging-the-type-system/newtype-pattern.md
@@ -34,6 +34,10 @@ assert_ne!(UserId(1), UserId(2));
 
 <details>
 
+- Students should have encountered the newtype pattern in the "Fundamentals"
+  course, when they learned about
+  [tuple structs](../../user-defined-types/tuple-structs.md).
+
 - Run the example to show students the error message from the compiler.
 
 - Modify the example to use a typealias instead of a newtype, such as
--- a/src/idiomatic/leveraging-the-type-system/newtype-pattern.md
+++ b/src/idiomatic/leveraging-the-type-system/newtype-pattern.md
@@ -19,8 +19,7 @@ fn double(n: u64) -> u64 {
     n * 2
 }
 
-// This doesn't compile ❌
-double(UserId(1));
+double(UserId(1)); // 🛠️❌
 ```
 
 The Rust compiler won't implicitly convert to (or from) the underlying type.\
@@ -28,8 +27,7 @@ It won't let you use methods or operators defined on the underlying type either:
 
 ```rust,compile_fail
 # pub struct UserId(u64);
-// This doesn't compile ❌
-assert_ne!(UserId(1), UserId(2));
+assert_ne!(UserId(1), UserId(2)); // 🛠️❌
 ```
 
 <details>
--- a/src/idiomatic/leveraging-the-type-system/newtype-pattern.md
+++ b/src/idiomatic/leveraging-the-type-system/newtype-pattern.md
@@ -22,8 +22,8 @@ fn double(n: u64) -> u64 {
 double(UserId(1)); // 🛠️❌
 ```
 
-The Rust compiler won't implicitly convert to (or from) the underlying type.\
-It won't let you use methods or operators defined on the underlying type either:
+The Rust compiler won't let you use methods or operators defined on the
+underlying type either:
 
 ```rust,compile_fail
 # pub struct UserId(u64);
--- a/src/idiomatic/leveraging-the-type-system/newtype-pattern.md
+++ b/src/idiomatic/leveraging-the-type-system/newtype-pattern.md
@@ -1,0 +1,49 @@
+---
+minutes: 5
+---
+
+# Newtype Pattern
+
+A _newtype_ is a wrapper around an existing type, often a primitive:
+
+```rust
+/// A unique user identifier, implemented as a newtype around `u64`.
+pub struct UserId(u64);
+```
+
+Unlike type aliases, newtypes aren't interchangeable with the wrapped type:
+
+```rust,compile_fail
+# pub struct UserId(u64);
+fn double(n: u64) -> u64 {
+    n * 2
+}
+
+// This doesn't compile ❌
+double(UserId(1));
+```
+
+The Rust compiler won't implicitly convert to (or from) the underlying type.\
+It won't let you use methods or operators defined on the underlying type either:
+
+```rust,compile_fail
+# pub struct UserId(u64);
+// This doesn't compile ❌
+assert_ne!(UserId(1), UserId(2));
+```
+
+<details>
+
+- Run the example to show students the error message from the compiler.
+
+- Modify the example to use a typealias instead of a newtype, such as
+  `type MessageId = u64`. The modified example should compile, thus highlighting
+  the differences between the two approaches.
+
+- Stress that newtypes, out of the box, have no behaviour attached to them. You
+  need to be intentional about which methods and operators you are willing to
+  forward from the underlying type. In our `UserId` example, it is reasonable to
+  allow comparisons between `UserId`s, but it wouldn't make sense to allow
+  arithmetic operations like addition or subtraction.
+
+</details>
--- a/src/idiomatic/leveraging-the-type-system/newtype-pattern/is-it-encapsulated.md
+++ b/src/idiomatic/leveraging-the-type-system/newtype-pattern/is-it-encapsulated.md
@@ -19,12 +19,12 @@ impl Username {
     }
 }
 
-impl DerefMut for Username {
+impl std::ops::DerefMut for Username { // ‼️
     fn deref_mut(&mut self) -> &mut Self::Target {
         &mut self.0
     }
 }
-# impl Deref for Username {
+# impl std::ops::Deref for Username {
 #     type Target = str;
 #
 #     fn deref(&self) -> &Self::Target {
--- a/src/idiomatic/leveraging-the-type-system/newtype-pattern/is-it-encapsulated.md
+++ b/src/idiomatic/leveraging-the-type-system/newtype-pattern/is-it-encapsulated.md
@@ -36,7 +36,7 @@ impl DerefMut for Username {
 
 <details>
 
-- `DerefMut` allows users to get a mutable reference to the wrapped type.
+- `DerefMut` allows users to get a mutable reference to the wrapped value.
 
   The mutable reference can be used to modify the underlying data in ways that
   may violate the invariants enforced by `Username::new`!
--- a/src/idiomatic/leveraging-the-type-system/newtype-pattern/is-it-encapsulated.md
+++ b/src/idiomatic/leveraging-the-type-system/newtype-pattern/is-it-encapsulated.md
@@ -1,0 +1,54 @@
+---
+minutes: 5
+---
+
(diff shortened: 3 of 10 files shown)


src/idiomatic/leveraging-the-type-system.md
```diff
@@ -0,0 +1,47 @@
+---
+minutes: 5
+---
+
+# Leveraging the Type System
+
+Rust's type system is _expressive_.\
```
mgeisler: This is a style thing: I don't think we forced line breaks anywhere else in the course. I don't think most people know about the Markdown syntax either.

Personally, I don't like such line breaks because I don't want to dictate the formatting — CSS will do that and I only want to deal with the semantics (paragraphs, lists, tables).

Also, I generally like a very terse style for slides:

```markdown

One line introduction to the topic

- Some key point
- Another key point
```

I know we don't follow this very closely, but that [Variables](https://google.github.io/comprehensive-rust/types-and-values/variables.html) still show this style.

Basically, I would try to avoid paragraphs, especially with multiple sentences per paragraph.

But I'm fine to merge this now and then we'll copy-edit it much more going forward.

src/idiomatic/leveraging-the-type-system.md
```diff
@@ -0,0 +1,47 @@
...
+We can use types and traits to build abstractions that make our code harder to
+misuse. In some cases, we can even go as far as enforcing correctness at
+_compile-time_. Quite often, these abstractions have no runtime
+overhead[^zero-cost].
```
mgeisler: Foot notes are difficult to use on slides: they break the flow if the reader has to move from the text here to the bottom and back up.

I would avoid the footnote and instead explain the concept in the speaker notes.

src/idiomatic/leveraging-the-type-system.md
```diff
@@ -0,0 +1,47 @@
...
+and maintainability of the entire codebase.
+
+<details>
+
+Additional items speaker may mention:
+
+- Rust's type system borrows a lot of ideas from functional programming
+  languages.\
+  For example, Rust's enums are known as "algebraic data types" in languages
+  like Haskell and OCaml. You can take inspiration from learning material geared
```
mgeisler, line 25: I love this extra background, very useful!

If you happen to have a list of such helpful resources, then feel free to expand https://google.github.io/comprehensive-rust/other-resources.html -- perhaps with a new section on the type system!

Also, https://google.github.io/comprehensive-rust/glossary.html is an important page to know about.

src/idiomatic/leveraging-the-type-system/newtype-pattern.md
```diff
@@ -0,0 +1,49 @@
+---
+minutes: 5
+---
+
+# Newtype Pattern
+
+A _newtype_ is a wrapper around an existing type, often a primitive:
```
mgeisler, line 7: I would like a throw-back to the other page that introduces the newtype pattern (perhaps in the speaker notes, depending on how this looks on screen).

src/idiomatic/leveraging-the-type-system/newtype-pattern.md
```diff
@@ -0,0 +1,49 @@
...
+
+```rust
+/// A unique user identifier, implemented as a newtype around `u64`.
+pub struct UserId(u64);
+```
+
+Unlike type aliases, newtypes aren't interchangeable with the wrapped type:
+
+```rust,compile_fail
+# pub struct UserId(u64);
```
mgeisler, line 17: We have so far avoided the `#` construct: it make the Playground read-only and so the instructor cannot do any live coding (or engage with questions).

Also, the slide very quickly becomes too big. This is about as much space that you have:

image

The compiler error immediately ends up outside of the page:

image

I would suggest taking away the "This doesn't compile ❌" comments — let the instructor show, not tell.

mgeisler: The screenshots come from the artifact bundle uploaded for this PR test run: https://github.com/google/comprehensive-rust/actions/runs/16028068797 -- all the way at the bottom the page, you find zip files for each language (useful for me if I don't want to grab the branch and do `mdbook build` myself).

src/idiomatic/welcome.md
```diff
@@ -0,0 +1,26 @@
+---
+course: Idiomatic Rust
+session: Morning
+target_minutes: 180
+---
+
+# Welcome to Idiomatic Rust
```
djmitche, line 7: This new course should be mentioned in `src/running-the-course/course-structure.md` as well.

src/idiomatic/leveraging-the-type-system.md
LukeMathWalker: Thanks for the reference!
Do you think edfe9e83 is closer to the style we're looking for?

src/idiomatic/leveraging-the-type-system.md
LukeMathWalker: Converted into a speaker note in edfe9e83

src/idiomatic/leveraging-the-type-system.md
LukeMathWalker: What's the threshold for inclusion in "Other resources"? Right now, all resources mentioned are Rust-focused and relatively "bulky" in size. 
I expect we'll have a lot of references to articles—would those be suitable inclusions?

src/idiomatic/leveraging-the-type-system/newtype-pattern.md
LukeMathWalker: I've added bidirectional links in 51d050eb.

src/idiomatic/welcome.md
LukeMathWalker: Thanks for the pointer!
I've added it in 447904a9, although I had to change the course plugin slightly to handle multi-word course names.

djmitche: Oh, good fix, thanks!

src/idiomatic/leveraging-the-type-system/newtype-pattern.md
LukeMathWalker: Still adjusting to the slide format 😁 But "Show, don't tell" is a great framing for thinking about the medium, thanks.

Re: size: Even if we remove unnecessary comments, we'll struggle to fit compilation errors in the viewport. 
The code example in "Parse, Don't Validate" is relatively minimal, but it'll fill the screen if everything is shown (no `#`).

Is the issue with the Playground flow related to the "Copy" button not including the hidden code?

src/idiomatic/leveraging-the-type-system.md
gribozavr: I think as long as the "other resources" page is well-organized by topic, it is not a problem to grow it significantly. If a resource is directly related to what we are teaching, there should be no hesitation to link to it.

On the other hand maybe we don't want to fully duplicate [awesome-rust > Resources](https://github.com/rust-unofficial/awesome-rust?tab=readme-ov-file#resources) or [RustBooks](https://github.com/sger/RustBooks?tab=readme-ov-file) but can simply link to them.

src/idiomatic/leveraging-the-type-system.md
```diff
@@ -0,0 +1,47 @@
...
+  towards functional languages when looking for guidance on how to design with
+  types. ["Domain Modeling Made Functional"][1] is a great resource on the
+  topic, with examples written in F#.
+
+- Despite its functional roots, functional design patterns don't translate as-is
```
gribozavr:
```suggestion
- Despite Rust's functional roots, functional design patterns don't translate as-is
```

I had a hard time reading this because the word order suggests that "it" refers to "design patterns".

src/idiomatic/leveraging-the-type-system.md
```diff
@@ -0,0 +1,47 @@
...
+  to Rust. For instance, extensive use of higher-kinded functions and types can
```
gribozavr:
```suggestion
  to Rust. For instance, extensive use of higher-order functions and higher-kinded types can
```

I wouldn't necessarily mention specific features though - this is a reference in passing with some criticism of well-established ideas, and that can easily turn off some audience members. I think it would be better to use a vague phrase like "type-level programming" or "metaprogramming".

src/idiomatic/leveraging-the-type-system.md
```diff
@@ -0,0 +1,47 @@
...
+  result in code that is harder to read and maintain. Design patterns in Rust
+  must take into account (and leverage!) the granular control over mutability
+  that comes with its borrow-checker.
```
gribozavr:
```suggestion
  that comes with its borrow checker.
```

src/idiomatic/leveraging-the-type-system.md
```diff
@@ -0,0 +1,47 @@
...
+
+- The same caution should be applied to object-oriented design patterns. Rust
+  doesn't support inheritance, and object boundaries must be mindful of the
+  constraints introduced by the borrow-checker.
```
gribozavr:
```suggestion
  constraints introduced by the borrow checker.
```

src/idiomatic/leveraging-the-type-system.md
```diff
@@ -0,0 +1,47 @@
...
+  topic, with examples written in F#.
+
+- Despite its functional roots, functional design patterns don't translate as-is
+  to Rust. For instance, extensive use of higher-kinded functions and types can
+  result in code that is harder to read and maintain. Design patterns in Rust
+  must take into account (and leverage!) the granular control over mutability
+  that comes with its borrow-checker.
+
+- The same caution should be applied to object-oriented design patterns. Rust
+  doesn't support inheritance, and object boundaries must be mindful of the
```
gribozavr:
```suggestion
  doesn't support inheritance, and object decomposition should take into account the
```
Rewriting to avoid describing an inanimate thing as "mindful of". Also, "object decomposition" is a term of art in OOD.

src/idiomatic/leveraging-the-type-system.md
```diff
@@ -0,0 +1,47 @@
...
+  constraints introduced by the borrow-checker.
+
+- Mention that type-level abstractions are often referred to as "zero-cost
```
gribozavr:
```suggestion
- Mention that type-level programming can be often used to create "zero-cost
```
Pedantic comment, these are not the same thing.

src/idiomatic/leveraging-the-type-system/newtype-pattern.md
```diff
@@ -0,0 +1,53 @@
...
+fn double(n: u64) -> u64 {
+    n * 2
+}
+
+// This doesn't compile ❌
+double(UserId(1));
+```
+
+The Rust compiler won't implicitly convert to (or from) the underlying type.\
```
gribozavr:
```suggestion
The Rust compiler won't implicitly convert to (or from) the underlying type.
```

gribozavr: This line repeats "newtypes aren't interchangeable with the wrapped type" above, to save screen space maybe it is best left in speaker notes.

src/idiomatic/leveraging-the-type-system/newtype-pattern.md
gribozavr: As a middle-ground for "This doesn't compile ❌" - we could adopt a consistent way to mark such lines throughout the class. For example, end-of-line comments on the lines that have build errors:`// ❌🛠️` or `// ❌🔨` (less visually noisy, better for slides). This marker can be short (and appear on the same line - to save space), but it would be clear because people would see it often and they will quickly learn what it means.

src/idiomatic/leveraging-the-type-system/newtype-pattern/parse-don-t-validate.md
```diff
@@ -0,0 +1,62 @@
...
+#     TooLong { len: usize },
+# }
+```
+
+<details>
+
+- The newtype pattern, combined with Rust's module and visibility system, can be
+  used to _guarantee_ that instances of a given type satisfy a set of
+  invariants.\
+  In the example above, the raw `String` stored inside the `Username` struct
```
gribozavr:
```suggestion
  invariants.

  In the example above, the raw `String` stored inside the `Username` struct
```

src/idiomatic/leveraging-the-type-system/newtype-pattern/parse-don-t-validate.md
```diff
@@ -0,0 +1,62 @@
...
+  the `new` method to create instances. In turn, `new` performs validation, thus
+  ensuring that all instances of `Username` satisfy those checks.
+
+- The `as_str` method allows consumers to access the raw string representation
+  (e.g. to store it in a database) but, thanks to Rust's borrow checker, they
+  can't modify it.
+
+- Stress the importance of evaluating _the entire API surface_ exposed by a
+  newtype to determine if invariants are indeed bullet-proof.\
+  It is crucial to consider all possible interactions, including trait
```
gribozavr:
```suggestion
  newtype to determine if invariants are indeed bullet-proof.

  It is crucial to consider all possible interactions, including trait
```

src/idiomatic/leveraging-the-type-system/newtype-pattern/parse-don-t-validate.md
```diff
@@ -0,0 +1,62 @@
...
+  implementations, that may allow users to bypass the invariants. For example,
+  if the `Username` type implements the `DerefMut` trait, users can modify the
+  underlying string directly, bypassing the validation checks in `new`.
+
+- Type-level invariants have second-order benefits.\
+  The input is validated once, at the boundary, and the rest of the program can
```
gribozavr:
```suggestion
- Type-level invariants have second-order benefits.

  The input is validated once, at the boundary, and the rest of the program can
```

src/idiomatic/leveraging-the-type-system/newtype-pattern/parse-don-t-validate.md
```diff
@@ -0,0 +1,62 @@
...
+  used to _guarantee_ that instances of a given type satisfy a set of
+  invariants.\
+  In the example above, the raw `String` stored inside the `Username` struct
+  can't be accessed directly from other modules or crates, since it's not marked
+  as `pub` or `pub(in ...)`. Consumers of the `Username` type are forced to use
+  the `new` method to create instances. In turn, `new` performs validation, thus
+  ensuring that all instances of `Username` satisfy those checks.
+
+- The `as_str` method allows consumers to access the raw string representation
+  (e.g. to store it in a database) but, thanks to Rust's borrow checker, they
```
gribozavr:
```suggestion
  (e.g., to store it in a database) but, thanks to Rust's borrow checker, they
```

gribozavr: > thanks to Rust's borrow checker

Not really - this is fairly standard type system stuff. C++ can also do this.

src/idiomatic/leveraging-the-type-system/newtype-pattern/parse-don-t-validate.md
```diff
@@ -0,0 +1,62 @@
...
+  can't modify it.
+
+- Stress the importance of evaluating _the entire API surface_ exposed by a
+  newtype to determine if invariants are indeed bullet-proof.\
+  It is crucial to consider all possible interactions, including trait
+  implementations, that may allow users to bypass the invariants. For example,
+  if the `Username` type implements the `DerefMut` trait, users can modify the
```
gribozavr: It would be nice to have a slide with the `DerefMut` example ready.

src/idiomatic/leveraging-the-type-system/newtype-pattern/semantic-confusion.md
```diff
@@ -0,0 +1,72 @@
+---
+minutes: 5
+---
+
+# Semantic Confusion
+
+There is room for confusion whenever a function takes multiple arguments of the
+same type:
```
gribozavr:
```suggestion
When a function takes multiple arguments of the same type, call sites are unclear:
```

src/idiomatic/leveraging-the-type-system/newtype-pattern/semantic-confusion.md
```diff
@@ -0,0 +1,72 @@
...
+}
+
+# let password = "password";
+# let username = "username";
+// In another part of the codebase, we swap arguments by mistake.
+// Bug (best case), security vulnerability (worst case)
+login(password, username);
+```
+
+The newtype pattern can be used to prevent this class of errors at compile time:
```
gribozavr:
```suggestion
The newtype pattern can prevent this class of errors at compile time:
```

src/idiomatic/leveraging-the-type-system/newtype-pattern/semantic-confusion.md
```diff
@@ -0,0 +1,72 @@
...
+- Run both examples to show students the successful compilation for the original
+  example, and the compiler error returned by the modified example.
+
+- Stress the _semantic_ angle. The newtype pattern should be leveraged to use
+  distinct types for distinct concepts, thus ruling out this class of errors
+  entirely.
+
+- Nonetheless, note that there are legitimate scenarios where a function may
+  take multiple arguments of the same type. In those scenarios, if correctness
+  is of paramount important, consider using a struct with named fields as input:
```
gribozavr:
```suggestion
  is of paramount importance, consider using a struct with named fields as input:
```

src/idiomatic/leveraging-the-type-system/newtype-pattern/semantic-confusion.md
```diff
@@ -0,0 +1,72 @@
...
+  # let username = "username";
+
+  // No need to check the definition of the `login` function to spot the issue.
+  login(LoginArguments {
+      username: password,
+      password: username,
+  })
+  ```
+  Users are forced, at the callsite, to assign values to each field, thus
+  increasing the likelihood of spotting bugs.
```
gribozavr, line 70: This point is worthy of a slide. The fact that structs are so convenient to construct inline, and that they require field labels, makes structs like this a very effective substitute for the named arguments feature found in other languages.

src/idiomatic/leveraging-the-type-system.md
LukeMathWalker: The reference to specific features was intentional here—it's a pitfall I've seen many learners fall into when coming to Rust from a functional background. 
It is not meant to criticise the pattern in the functional context, and I'm happy to rephrase if it reads as such.

src/idiomatic/leveraging-the-type-system/newtype-pattern/parse-don-t-validate.md
LukeMathWalker: It is not meant to imply that Rust is unique in this regard. But, in the context of Rust, it is the borrow-checker that's going to stop you from modifying the string value via a shared reference.

src/idiomatic/leveraging-the-type-system/newtype-pattern/semantic-confusion.md
LukeMathWalker: I intend to add one in the "Foundations of API design" section. When that's done, I'll convert this note into a link.

src/idiomatic/leveraging-the-type-system/newtype-pattern.md
LukeMathWalker: That sounds like a good balance. I've updated to this more concise style in [93f1980](https://github.com/google/comprehensive-rust/pull/2800/commits/93f198051e5198ccf97b7b98ba76d1ef6ce7e8fc)

src/idiomatic/leveraging-the-type-system/newtype-pattern/parse-don-t-validate.md
gribozavr: > it is the borrow-checker that's going to stop you from modifying the string value via a shared reference

I don't think so. We don't need a borrow checker to resolve the list of methods on `&str`. This is essentially the same thing as `const` in C++ - overload resolution eliminates the methods that can modify the object.

src/idiomatic/leveraging-the-type-system/newtype-pattern/semantic-confusion.md
gribozavr: Please add a TODO here then.

src/idiomatic/leveraging-the-type-system.md
gribozavr: What rubs me the wrong way is the assertion that they "can result in code that is harder to read and maintain" without any proof. I have seen such debates in the Swift community, and the arguments need to become quite subtle to get good signal from the conversation. So this is the kind of thing that can trigger and distract certain types of people.

I believe there are non-trivial debates going on to this day about using open-coded loops vs. `map` and `filter` in many language communities where the language does not position itself as clearly of a functional descent. Different people have different preferences, so saying that these patterns can make code harder to maintain could trigger them.

If we're talking about "clearly bad" code, like expressions that chain 10 iterator combinators, the problem isn't higher-order functions themselves, it is the exact way that the code is written in. If we translated that into 10 nested loops, the code would be similarly bad! In other words, **overuse and misuse of anything is bad**; however, that is a trivial thing to say, and it is weird to single out specific features that way.

src/idiomatic/leveraging-the-type-system.md
```diff
@@ -0,0 +1,47 @@
...
+With careful design, you can improve the clarity and maintainability of the
+entire codebase.
+
+<details>
+
+Additional items speaker may mention:
+
+- Rust's type system borrows a lot of ideas from functional programming
+  languages.\
+  For example, Rust's enums are known as "algebraic data types" in languages
```
gribozavr:
```suggestion
  languages.

  For example, Rust's enums are known as "algebraic data types" in languages
```

src/idiomatic/leveraging-the-type-system/newtype-pattern.md
LukeMathWalker: I discussed this further with @gribozavr.
We will do our best to keep slides concise and to make them fit in a single viewport. For `#`, we will allow them under the assumption that #2811 is implemented before the "Idiomatic Rust" course is marked as complete.

src/idiomatic/leveraging-the-type-system.md
LukeMathWalker: Thanks for elaborating. I think there may be a misunderstanding at play here, which reflects on that note.

There is no value judgement: I don't want to imply that code using higher-order functions or higher-kinded types is better or worse than imperative-looking code.
I'm coming from a different angle—writing higher-order functions in Rust requires a deeper knowledge of the language (e.g., generic programming, `Fn*` bounds, closures, boxing, `fn` pointers), and there are limitations to what can be achieved (e.g., when it comes to GATs) which can be difficult to predict unless you've quite a bit of expertise on the subject. That's what I meant by "more difficult to read and maintain".

To be specific, I'm referring to writing higher-order construct (e.g. implementing the `Map` type in `std`), not using them (i.e. invoking `map` on an iterator).

My intent with this note was to steer the reader towards local mutation, where possible, since it'll usually put them on a path with lower friction.

Do you agree with this viewpoint?

src/idiomatic/leveraging-the-type-system/newtype-pattern/parse-don-t-validate.md
LukeMathWalker: OK, I see where you're coming from now—yes, it is enough to have a projection type with a restricted subset of methods attached, as `&str` in this case.
Fixed in bbd0631f

src/idiomatic/leveraging-the-type-system/newtype-pattern/semantic-confusion.md
LukeMathWalker: Done in 7c32ec27

LukeMathWalker:
> It would be helpful to see an overall outline for the new course.
> 
> * This would facilitate high-level review of the course content before you spend the time to write the slides
> * This would help us think about how long this course might last (is it just a morning course?)
> * This would provide an opportunity for others to contribute sections or slides
> * It would give us all more time to think about good exercises - I've found this to be the most time-consuming part of a section

You can find the outline for the course [here](https://hackmd.io/@6lzc4-pLRIiARouWpB8X8A/B1c_rLKBlx).
Consider it more of a syllabus of the topics we plan to cover rather than an exact breakdown into slides—as you can see in this PR, we need more than one slide to explore the new type pattern.

The outline has gone through a few iterations with @mgeisler's  and @gribozavr's feedback, but we can definitely iterate further.

src/idiomatic/leveraging-the-type-system/newtype-pattern/parse-don-t-validate.md
LukeMathWalker: Added in 3fe14761

src/idiomatic/leveraging-the-type-system/newtype-pattern/is-it-encapsulated.md
```diff
@@ -0,0 +1,54 @@
...
+
+impl Username {
+    pub fn new(username: String) -> Result<Self, InvalidUsername> {
+        // Validation checks...
+        Ok(Self(username))
+    }
+}
+
+impl DerefMut for Username {
+    fn deref_mut(&mut self) -> &mut Self::Target {
```
gribozavr, line 23: Flag this line with an exclamation mark emoji?

src/idiomatic/leveraging-the-type-system/newtype-pattern/is-it-encapsulated.md
```diff
@@ -0,0 +1,54 @@
...
+#     fn deref(&self) -> &Self::Target {
+#         &self.0
+#     }
+# }
+# pub struct InvalidUsername;
+```
+
+<details>
+
+- `DerefMut` allows users to get a mutable reference to the wrapped type.
```
gribozavr:
```suggestion
- `DerefMut` allows users to get a mutable reference to the wrapped value.
```

gribozavr:
> You can find the outline for the course [here](https://hackmd.io/@6lzc4-pLRIiARouWpB8X8A/B1c_rLKBlx).

WDYT about adding this outline under a `detail` tag in `src/idiomatic/welcome.md`?

src/idiomatic/leveraging-the-type-system.md
gribozavr: Yes, I agree with this argument. However, it confirms exactly what I was saying in my previous comment:

> I have seen such debates in the Swift community, and the arguments need to become quite subtle to get good signal from the conversation.

Originally the text was "extensive use of higher-order functions [...] more difficult to read and maintain" - which looks to me like a typical typical a black-and-white argument: X is harder to read than the alternative Y, don't do X, do the equivalent Y that's easier to read. But after we talked about it, it turns out that the actual argument you're making is more subtle: it takes at least a paragraph to explain, it is more mild guidance than a strict "don't", it is not about use, but instead about implementing something that looks and feels like a standard library API.

I do very much recommend replacing "more difficult to read and maintain" with something like "designing APIs with advanced language features like higher-order functions and higher-kinded types requires a deep knowledge of the language to do well", but since this text is inside instructor notes, I would recommend using even more words to explain what you mean to the instructors who will be teaching this section.

src/idiomatic/leveraging-the-type-system/newtype-pattern/is-it-encapsulated.md
LukeMathWalker: Good idea, done in 8befc72c

LukeMathWalker:
> > You can find the outline for the course [here](https://hackmd.io/@6lzc4-pLRIiARouWpB8X8A/B1c_rLKBlx).
> 
> WDYT about adding this outline under a `detail` tag in `src/idiomatic/welcome.md`?

Why not, added in 06865d4b.

src/idiomatic/leveraging-the-type-system.md
LukeMathWalker: I've tried a different phrasing in 71204b72. What do you think?

src/idiomatic/leveraging-the-type-system.md
```diff
@@ -0,0 +1,53 @@
...
+- Despite Rust's functional roots, not all functional design patterns can be
+  easily translated to Rust.
+
+  For example, you must have a solid grasp on a broad selection of advanced
+  topics to design APIs that leverage higher-order functions and higher-kinded
+  types in Rust.
+
+  Evaluate, on a case-by-case basis, whether a more imperative approach may be
+  easier to implement, relying on Rust's borrow-checker and type system to
+  control what can be mutated, and where.
```
gribozavr:
```suggestion
  Evaluate, on a case-by-case basis, whether a more imperative approach may be
  easier to implement. Consider using in-place mutation, relying on Rust's borrow-checker and type system to
  control what can be mutated, and where.
```

src/idiomatic/leveraging-the-type-system.md
gribozavr: Looks good, thanks!

src/idiomatic/leveraging-the-type-system/newtype-pattern/semantic-confusion.md
```diff
@@ -0,0 +1,72 @@
...
+      pub password: &str,
+  }
+  # fn login(i: LoginArguments) {}
+  # let password = "password";
+  # let username = "username";
+
+  // No need to check the definition of the `login` function to spot the issue.
+  login(LoginArguments {
+      username: password,
+      password: username,
```
djmitche, line 66: nit: these do not need to be repeated (and in fact the shorthand syntax makes a stronger case for this approach!)

djmitche:
> It would be helpful to see an overall outline for the new course.
> 
>     * This would facilitate high-level review of the course content before you spend the time to write the slides
> 
>     * This would help us think about how long this course might last (is it just a morning course?)
> 
>     * This would provide an opportunity for others to contribute sections or slides
> 
>     * It would give us all more time to think about good exercises - I've found this to be the most time-consuming part of a section

Reiterating this -- https://github.com/google/comprehensive-rust/pull/2806#issue-3202722424 has a nice summary of the planned content.

gribozavr:
@djmitche Is this ToC missing information that you're looking for?

> Why not, added in https://github.com/google/comprehensive-rust/commit/06865d4b4b2e51407235bdfd665305398b701a99.

Here's some more information.

The expected duration of this deep dive is 2 days (each day is morning + afternoon as usual).

The audience is people who have completed the fundamentals module (or have equivalent knowledge through other means). So we can assume broad knowledge of Rust basics, but little experience applying them effectively.

The goal is to help experienced software engineers who currently write C++ or Java transition to Rust. As @LukeMathWalker and I brainstormed ideas, we have gravitated towards two broad categories of topics: API design and implementation design. The newtype pattern belongs to the API design bucket, and so does the majority of "Leveraging the type system" and "Polymorphism in Rust" sections in the above commit. "Don't fight the borrow checker" is going to talk more about implementation design, but there's certainly some amount of API design there as well.

src/idiomatic/leveraging-the-type-system/newtype-pattern/semantic-confusion.md
LukeMathWalker: They don't qualify for the short-hand syntax in this specific example, since we have intentionally swapped them to mirror the mistake made in the slide example.

djmitche: Oops, I totally missed that! Sorry :)

src/idiomatic/welcome.md
```diff
@@ -0,0 +1,111 @@
...
+  - How do I pick the most appropriate pattern?
+- Working with generics
+  - Generic type parameter in a function or trait object as an argument?
+  - Trait bounds don't have to refer to the generic parameter
+  - Type parameters in traits: should it be a generic parameter or an associated
+    type?
+- Macros: a valuable tool to DRY up code when traits are not enough (or too
+  complex)
+
+### Error Handling
```
djmitche, line 97: This session will probably have substantial overlap with the fundamentals section on error handling. It might make sense to broaden it a little, to include panicking and panic safety, and maybe even async cancellation safety?

djmitche (commented):
> @djmitche Is this ToC missing information that you're looking for?

Nope, that's great -- sorry I missed it!

With good use of case studies and/or exercises, this seems like a good 2-day additional course.

src/idiomatic/welcome.md
LukeMathWalker: I don't think they will end up overlapping too much.
We may briefly recap some key concepts from Fundamentals, but we'll focus on how to _use_ those constructs effectively.

I'm open to broadening the scope, but I'd rather cross that bridge once we have a good sense of how long each of the sections currently in the ToC will take. Then we can "pad" with more content if needed.

djmitche: Sounds good

djmitche: (In general, the course has always tended to get longer and not shorter, so this is a good plan)

src/idiomatic/leveraging-the-type-system.md
mgeisler: Thanks!

src/idiomatic/leveraging-the-type-system.md
mgeisler: I agree with the sentiment of @gribozavr — we don't have an upper limit per se.

However, it's a good question where to put things and we haven't had a clear boundary yet. I have added lots of inline links to Wikipedia articles, for example.

When I originally introduced the "Other resources" section, my thinking was roughly that it was for Rust material that I think most developers would find useful. So a specialized link to type system theory is probably best for the details section on the particular page that talk about this material.

For more general things that don't belong to any particular part of the course, I would put them in the Other resources section.

src/idiomatic/leveraging-the-type-system/newtype-pattern.md
mgeisler: Thanks!

src/idiomatic/leveraging-the-type-system/newtype-pattern.md
mgeisler: Sounds good!

> Still adjusting to the slide format 😁

Yes, it's a constant struggle! @michael-kerscher has worked on a tool which can run through every slide and flag the ones that are too tall. It's not enabled by default yet in CI, but I believe it could be.

Simply rejecting slides that are taller than _NNN_ px is of course very crude and that's why we haven't insisted on it yet. Some scrolling is bound to happen, but it's worth avoiding. I've seen first hand the look of confusing on the faces when 20 people sit in a room and look at a gigantic imaging moving rapidly up and down. People lose track and cannot remember the part that is off screen.

> But "Show, don't tell" is a great framing for thinking about the medium, thanks.

Yeah, I found it to be quite a shift in thinking: instead of multi-step code blocks (as one would do in a book or doc comment!), I try to aim for a simple _initial state_ which the instructor can run. Something the error will be in the initial state, sometimes the instructor will introduce it.

mgeisler:
It's super exciting to see this take shape! :tada:

src/idiomatic/leveraging-the-type-system.md
mgeisler: Yes, thanks for updating!

PR #2805, opened 2025-07-03
A few fixes for the unsafe chapter

### LINKED ISSUE #2734: Raw pointers example using `String` is confusing
Our example for dereferencing raw pointers uses a raw pointer to a String:
```rust
fn main() {
    let mut s = String::from("careful!");

    let r1 = &raw mut s;
    let r2 = r1 as *const String;

    // SAFETY: r1 and r2 were obtained from references and so are guaranteed to
    // be non-null and properly aligned, the objects underlying the references
    // from which they were obtained are live throughout the whole unsafe
    // block, and they are not accessed either through the references or
    // concurrently through any other pointers.
    unsafe {
        println!("r1 is: {}", *r1);
        *r1 = String::from("uhoh");
        println!("r2 is: {}", *r2);
    }

    // NOT SAFE. DO NOT DO THIS.
    /*
    let r3: &String = unsafe { &*r1 };
    drop(s);
    println!("r3 is: {}", *r3);
    */
}
```

It then creates a `*const String` and a `*mut String`, mutates the String through the `*mut String`, and observes that mutation through the `*const String`. This suffices to show how raw pointers do not obey the same rules as references: mutable and constant ones can coexist, and we can observe mutation through them.

But when thinking about this in terms of references, it's strange to see a `String` accessed indirectly in this way. Rust programs don't often construct `&String` except as a temporary, and `&mut String` is similarly uncommon except as receiver of methods. When we think of `*const String` and `*mut String`, these are essentially pointers to a 3-field struct, and the assignment to `*r1` overwrites this struct. In doing so, it destroys the previous value, an object with a heap allocation and associated `Drop` impl. The `Drop` impl is still used when we drop the old value, but it is another thing to be aware of here, and it might be better to introduce raw pointers in a situation where ownership is not involved.

Something like this might be clearer:
```rust
fn main() {
    let mut s = Box::new(5);

    let r1: *mut i32 = &raw mut *s;
    let r2: *const i32 = r1 as *const i32;

    // SAFETY: r1 and r2 were obtained from a box and so are guaranteed to
    // be non-null and properly aligned, the objects underlying the references
    // from which they were obtained are live throughout the whole unsafe
    // block, and they are not accessed either through the box or
    // concurrently through any other pointers.
    unsafe {
        println!("r1 is: {}", *r1);
        *r1 = 100;
        println!("r2 is: {}", *r2);
    }

    // NOT SAFE. DO NOT DO THIS.
    /*
    let r3: &i32 = unsafe { &*r1 };
    drop(s);
    println!("r3 is: {:?}", *r3);
    */
}
```

fw-immunant:
See individual commits. Fixes #2734.

### DIFF
--- a/src/unsafe-rust/dereferencing.md
+++ b/src/unsafe-rust/dereferencing.md
@@ -8,27 +8,32 @@ Creating pointers is safe, but dereferencing them requires `unsafe`:
 
 ```rust,editable
 fn main() {
-    let mut s = String::from("careful!");
+    let mut x = 10;
 
-    let r1 = &raw mut s;
-    let r2 = r1 as *const String;
+    let p1: *mut i32 = &raw mut x;
+    let p2 = p1 as *const i32;
 
-    // SAFETY: r1 and r2 were obtained from references and so are guaranteed to
-    // be non-null and properly aligned, the objects underlying the references
-    // from which they were obtained are live throughout the whole unsafe
-    // block, and they are not accessed either through the references or
-    // concurrently through any other pointers.
+    // SAFETY: p1 and p2 were created by taking raw pointers to a local, so they
+    // are guaranteed to be non-null, aligned, and point into a single (stack-)
+    // allocated object.
+    //
+    // The object underlying the raw pointers lives for the entire function, so
+    // it is not deallocated while the raw pointers still exist. It is not
+    // accessed through references while the raw pointers exist, nor is it
+    // accessed from other threads concurrently.
     unsafe {
-        println!("r1 is: {}", *r1);
-        *r1 = String::from("uhoh");
-        println!("r2 is: {}", *r2);
+        dbg!(*p1);
+        *p1 = 6;
+        // Mutation may soundly be observed through a raw pointer, like in C.
+        dbg!(*p2);
     }
 
-    // NOT SAFE. DO NOT DO THIS.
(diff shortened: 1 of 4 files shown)


src/unsafe-rust/dereferencing.md
```diff
@@ -52,8 +57,8 @@ In the case of pointer dereferences, this means that the pointers must be
 
 In most cases the pointer must also be properly aligned.
 
-The "NOT SAFE" section gives an example of a common kind of UB bug: `*r1` has
-the `'static` lifetime, so `r3` has type `&'static String`, and thus outlives
-`s`. Creating a reference from a pointer requires _great care_.
+The "UNSOUND" section gives an example of a common kind of UB bug: `*p1` has the
+`'static` lifetime, so `r` has type `&'static i32`, and thus outlives `x`.
+Creating a reference from a pointer requires _great care_.
```
djmitche: `r` isn't outliving `x` anymore -- I think this paragraph needs to be updated

djmitche (commented):
Looks good, just one change. Thanks as always!

src/unsafe-rust/dereferencing.md
fw-immunant: Fixed.

djmitche (approved):

Issue #2590, opened 2025-01-28
Add a day 5 class on unsafe Rust


randomPoison:
Currently we briefly introduce unsafe Rust at the end of day 4, but there's a lot more to go into with unsafe Rust. I think we could have an entire day 5 course going into more detail with unsafe Rust. There's enough to cover with unsafe that we couldn't fit it all in one day, so I think we should aim for the goal of getting students ready to review unsafe code, rather than trying to get students to the point where they could write unsafe themselves.

andriyDev:
When you say "day 5 class" do you mean an additional optional day like the Concurrency class? Or do you mean extending the "main" course to be 5 days long?

randomPoison:
I meant an optional extra day like Android or concurrency, I don't wanna add a day 5 to the main course lol

PR #2806, opened 2025-07-04
Add Unsafe Rust Deep Dive

timClicks:
This PR adds the start of an unsafe deep dive to Comprehensive Rust, which will hopefully address #2590. This is a project that I intend to work on over the next several months.

Here is a tentative outline of a 10h (2 day) treatment:

Day 1: Using and Reviewing Unsafe

- Welcome
- Motivations: explain why the `unsafe` keyword exists
- Foundations: provide background knowledge; what is soundness? what is undefined behavior? what is validity in respect to pointers?
- Mechanics: what a safe `unsafe` block should look like
- Representations and Interoperability: explore how data is laid out in memory and how that can be sent across the wire and/or stored on disk.
- Reviewing unsafe
- Patterns for safer unsafe: Encapsulating unsafe code in safe-to-use abstractions, such as marking a type's constructor as `unsafe` so that invariants only need to be enforced once by the programmer. 

Day 2: Deploying Unsafe to Build Abstractions

- Welcome
- Validity in detail: A refresher. Emphasis on the details of the invariants that are being upheld by a “typical” unsafe block, such as aliasing, alignment, data validity, padding.
- Concurrency and thread safety: understanding `Send` and `Sync`, knowing how to implement them on a user-defined type
- Case study: Small string optimization
- Case study: Zero-copy parsing
- Review


src/running-the-course/course-structure.md
```diff
@@ -72,6 +72,13 @@ cargo run
 
 {{%course outline Concurrency}}
 
+### Unsafe (Work in Progress)
```
djmitche, line 85: I wonder if there are more positive titles? Perhaps "Safety and Soundness"?

The audience for this paragraph are people considering taking the class, or paying someone to teach the class to a group of engineers, or just people wondering how hard it is to learn Rust. I worry at the perception that it will take 4 days to learn Rust, then half that again to learn how to use it safely. That's not the case, but it's easy to misinterpret.

src/unsafe-deep-dive/foundations.md
```diff
@@ -0,0 +1 @@
+# Foundations
```
djmitche, line 1: You'll want a segment table-of-contents macro here

src/unsafe-deep-dive/motivations.md
```diff
@@ -0,0 +1,18 @@
+# Motivations
+
+We know that writing code without the guarantees that Rust provides ...
+
+> “Use-after-free (UAF), integer overflows, and out of bounds (OOB) reads/writes
+> comprise 90% of vulnerabilities with OOB being the most common.”
+>
+> --— **Jeff Vander Stoep and Chong Zang**, Google. "Queue the Hardening
+> Enhancements"
```
djmitche: Link to https://security.googleblog.com/2019/05/queue-hardening-enhancements.html

src/unsafe-deep-dive/motivations/data-structures.md
```diff
@@ -0,0 +1,20 @@
+# Data Structures
+
+Some families of data structures, are impossible to create in safe Rust.
+
+- graphs
+- bit stuffing
+- self-referential types
```
djmitche, line 11: intrusive data structures?

src/unsafe-deep-dive/foundations/data-structures-are-safe.md
```diff
@@ -0,0 +1,21 @@
+# Data structures are safe ...
+
+Data structures are inert. They cannot do any harm by themselves.
+
+It's possible to create a valid program with raw pointer entirely in safe code:
+
+```rust
+fn main() {
+    let n: i64 = 12345;
+    let safe = &n as *const _;
```
djmitche: AIUI the modern syntax for this is `&raw const n`.

src/running-the-course/course-structure.md
gribozavr: WDYT about "Unsafe Rust Fundamentals"?

djmitche: IMHO that's worse -- it suggests that students will spend two days learning about how unsafe Rust is, and can only expect to have a handle on the "fundamentals" after that time.

gribozavr: Sure, that is a possible interpretation, but it is a general problem with "Unsafe Rust" as a brand - not something we can change.

gribozavr: To help pick a better name, let me explain the expected content and the goals for this deep dive.

Internally at Google we have introduced a mandatory code review process for all unsafe Rust code. This is proving to be too onerous for engineers: even if unsafe code reviews complete quickly, people feel like we took away some of their agency.

Solution:

* ensure that engineers understand foundations of unsafe code, "the rules of the game" so to speak:

  - what the dangers of unsafe Rust are,
  - where to find information about specific unsafe operations,
  - how to read unsafe contracts,
  - how to satisfy unsafe contracts,
  - how to document all of the above in code comments for future readers,

* ensure engineers can distinguish simple cases (e.g., cookie-cutter FFI) from complex cases (e.g., something like the zerocopy crate),

* teach enough specifics about basic unsafe operations and do enough exercises that upon completion of the deep dive we can trust students to self-certify correctness of the simple unsafe code, and pull in experienced unsafe reviewers for complex cases.

src/running-the-course/course-structure.md
```diff
@@ -72,6 +72,13 @@ cargo run
...
+
+The [Unsafe](../unsafe-deep-dive/welcome.md) deep dive is a two day class on
+using the `unsafe` keyword correctly. It covers the fundamentals of what Rust's
```
gribozavr:
```suggestion
The [Unsafe](../unsafe-deep-dive/welcome.md) deep dive is a two-day class on
the *unsafe* Rust language. It covers the fundamentals of what Rust's
```

I feel like putting the unsafe keyword front and center is a bit reductionistic.

src/running-the-course/course-structure.md
```diff
@@ -72,6 +72,13 @@ cargo run
...
+safety guarantees are and why `unsafe` is needed, reviewing `unsafe` code, using
+FFI and building data structures.
```
gribozavr:
```suggestion
FFI, and building data structures that the borrow checker would normally reject.
```

src/unsafe-deep-dive/welcome.md
```diff
@@ -0,0 +1,39 @@
...
+> Please do not consider this module of Comprehensive Rust to be complete. With
+> that in mind, your feedback, comments, and especially your concerns, are very
+> welcome.
+>
+> To comment on this module's development, please use the
+> [GitHub issue tracker].
+
+[GitHub issue tracker]: https://github.com/google/comprehensive-rust/issues
+
+The `unsafe` keyword is a misunderstood, but at times essential, part of the
```
gribozavr: Consider something like "easy to type but hard to master" instead of "misunderstood".

src/unsafe-deep-dive/welcome.md
```diff
@@ -0,0 +1,39 @@
...
+What you'll learn:
+
+- What the terms undefined behavior, soundness, and safety mean
+- Why the `unsafe` keyword exists in the Rust language
+- How to write your own code using `unsafe` safely
+- How to review `unsafe` code
+
+## Links to other sections of the course
+
+The `unsafe` keyword has treatment
```
gribozavr:
```suggestion
The `unsafe` keyword has treatment in
```

src/unsafe-deep-dive/motivations/data-structures.md
```diff
@@ -0,0 +1,20 @@
+# Data Structures
+
+Some families of data structures, are impossible to create in safe Rust.
```
gribozavr:
```suggestion
Some families of data structures are impossible to create in safe Rust.
```

src/unsafe-deep-dive/motivations/data-structures.md
```diff
@@ -0,0 +1,20 @@
...
+
+- graphs
+- bit stuffing
```
gribozavr:
```suggestion
- bit twiddling
```

https://en.wikipedia.org/wiki/Bit_stuffing is a very specific technique in communication protocols, I think you're referring to "bit twiddling" https://en.wikipedia.org/wiki/Bit_manipulation

src/unsafe-deep-dive/motivations/perfomance.md
```diff
@@ -0,0 +1,6 @@
+# Performance
+
+> Stub for now
```
gribozavr:
```suggestion
> TODO: Stub for now
```

To make it easy to find these in the future.

src/unsafe-deep-dive/foundations/what-is-unsafe.md
```diff
@@ -0,0 +1,89 @@
...
+
+Let's create a working definition together.
+
+<details>
+
+## Clarifying "intentionally vague"
+
+It's likely that you'll have an audience member object to the phrase
+"intentionally vague". Perhaps refer to this statement from the standard library
+when discussing the [safety requirements of raw pointers].
```
gribozavr: > intentionally vague

I'd rather not take this angle at the very beginning. "Oh great, I'm about to sit in a 2-day class to hear about a thing where even the language authors are not sure about the rules", "Since the rules are intentionally vague, maybe I can be vague too"

There are two different things:

* what operations need to be unsafe: this is well defined (for example, dereferencing a pointer is an unsafe operation, there is no debate about that),
* the precise preconditions for some of the built-in unsafe operations: some parts of the preconditions are nailed down, some are intentionally left unspecified for now. What do I mean? Some scenarios where dereferencing a pointer is definitely OK have been delineated, and some scenarios where it is definitely UB are described, but there's a third group where the language lawyers have not yet made their decision.

The first one people probably have heard about, but the second one is very subtle. I wouldn't mention it in the first pass?

---

Instead, WDYT about the following angle to introduce the concept of unsafe: Rust has a formal definition of what needs to be "unsafe". It is a nice crisp definition - but unfortunately a useless to newcomers: one needs to understand many other things before reading the crisp formal definition. So let's start by reading quotes from the official documentation that tries to explain it in a more accessible way.

src/unsafe-deep-dive/foundations/what-is-unsafe.md
```diff
@@ -0,0 +1,89 @@
...
+> ...
+>
+> The precise rules for validity are not determined yet.
+
+[this module]: https://doc.rust-lang.org/std/ptr/index.html
+[safety requirements of raw pointers]: https://doc.rust-lang.org/std/ptr/index.html#safety
+
+## Definitions from authoritative docs:
+
+stdlib's keyword:
```
gribozavr: Sorry, I don't understand what it means because the keyword is a part of the language, not of the stdlib (this looks like a categorical error to me).

src/unsafe-deep-dive/foundations/what-is-unsafe.md
```diff
@@ -0,0 +1,89 @@
...
+Ask the class: **By raising your hand, indicate if you would feel comfortable
+defining unsafe?**
+
+If anyone's feeling confident, allow them to try to explain.
+
+### Part 2: Evidence gathering
+
+Ask the class to spend 3-5 minutes.
+
+- Find a a use of the unsafe keyword. What contract/invariant/pre-condition is
```
gribozavr:
```suggestion
- Find a use of the unsafe keyword. What contract/invariant/pre-condition is
```

src/unsafe-deep-dive/foundations/when-is-unsafe-used.md
```diff
@@ -0,0 +1,42 @@
+# When is unsafe used?
+
+The unsafe keyword indicates that the programmer is responsible for upholding
+Rust's safety guarantees.
+
+The keyword has two roles:
+
+- define pre-conditions must be satisfied
```
gribozavr:
```suggestion
- define pre-conditions that must be satisfied
```

src/unsafe-deep-dive/foundations/when-is-unsafe-used.md
```diff
@@ -0,0 +1,42 @@
...
+- verify that those defined pre-conditions are satisfied
```
gribozavr:
```suggestion
- assert to the compiler (= promise) that those defined pre-conditions are satisfied
```

src/unsafe-deep-dive/foundations/when-is-unsafe-used.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Further references
+
+- [The unsafe keyword chapter of the Rust Reference](https://doc.rust-lang.org/reference/unsafe-keyword.html)
+
+<details>
+
+Places where pre-conditions can be defined (Role 1)
+
+- [unsafe functions] (`unsafe fn { ... }`). Example: `get_unchecked` method on
```
gribozavr:
```suggestion
- [unsafe functions] (`unsafe fn foo() { ... }`). Example: `get_unchecked` method on
```

src/unsafe-deep-dive/foundations/data-structures-are-safe.md
```diff
@@ -0,0 +1,21 @@
+# Data structures are safe ...
+
+Data structures are inert. They cannot do any harm by themselves.
+
+It's possible to create a valid program with raw pointer entirely in safe code:
```
gribozavr:
```suggestion
It's possible to create a valid program with a raw pointer entirely in safe code:
```

src/unsafe-deep-dive/foundations/data-structures-are-safe.md
```diff
@@ -0,0 +1,21 @@
...
+fn main() {
+    let n: i64 = 12345;
+    let safe = &n as *const _;
+    println!("{safe:p}");
+}
+```
+
+<details>
+
+Consider a raw pointer to an integer, i.e. the value `safe` is the raw pointer
```
gribozavr:
```suggestion
Consider a raw pointer to an integer, i.e., the value `safe` is the raw pointer
```

src/unsafe-deep-dive/foundations/data-structures-are-safe.md
```diff
@@ -0,0 +1,21 @@
...
+type `*const i64`. Raw pointers can be out-of-bounds, misaligned, or refer to
```
gribozavr:
```suggestion
type `*const i64`. Raw pointers can be out-of-bounds, misaligned, or be
```

src/unsafe-deep-dive/foundations/data-structures-are-safe.md
gribozavr:
```suggestion
Safe Rust code can create raw pointers:
```

Suggesting shorter phrasing for slides.

src/unsafe-deep-dive/foundations/less-powerful.md
```diff
@@ -0,0 +1,18 @@
...
+
+let orig = b"RUST";
+let n: i32 = unsafe { transmute(orig) };
+
+println!("{}")
+```
+
+<details>
+
+Try to c different
```
gribozavr:
```suggestion
Try to c different TODO
```

src/unsafe-deep-dive/motivations/data-structures.md
timClicks: Fixed in 7fdbb45eaf55a2c864741c0fe973b4e9e46f55d1

src/unsafe-deep-dive/foundations/data-structures-are-safe.md
timClicks: Oh yes of course!

src/unsafe-deep-dive/motivations/data-structures.md
timClicks: Thank you for the close read. Fixed in a30d51838e40ec4fa964d30d5cbb95c9433096dd

src/unsafe-deep-dive/motivations/data-structures.md
timClicks: Very good point. I wasn't entirely confident and it seems that lack of confidence has been proven valid. Fixed 65d60494863936af3a04b16f78052350e410f131

src/unsafe-deep-dive/foundations/what-is-unsafe.md
timClicks: I like this adjustment. Saying it's "intentionally vague" isn't very fair to a newcomer at all.

Another place to put technicalities like this would be at the very end of the course. We have some buffer space in the "Review" section and some groups may enjoy a discussion about the precise semantics.

src/unsafe-deep-dive/foundations/what-is-unsafe.md
timClicks: You're correct, although keywords are documented in the standard library's documentation. https://doc.rust-lang.org/std/keyword.unsafe.html. I should have had that link in the original :)

src/running-the-course/course-structure.md
timClicks: Perhaps we should have the data structures item on its own line?

timClicks:
Many thanks for your close reads @djmitche @gribozavr. Many good suggestions in here. I'll have them addressed shortly with some further content.

As a matter of etiquette, I tend not to click the <kbd>Resolve conserveration</kbd> button. Let me know it would be better to 'clean as I cook'.

src/running-the-course/course-structure.md
djmitche: I was doing some of those reviews while I was there, and indeed it was a sticking point! That plan sounds fantastic, and will be a benefit to the Rust community as a whole.

My takeaway from that explanation is that the course is about `unsafe`, as that appears or is implied in all three bullet points. Safety and soundness are implied, but secondary. I'd consider "Working with Unsafe" but it seems long. So perhaps "Unsafe" is OK after all.

timClicks: Really encouraging to hear that you're supportive @djmitche. 

A pithy title seems like a fun thing to add, but I was wary about introducing too much into an initial review. That said, What about "Safe Unsafe" or "Unsafe Safely"?

djmitche: I like both of those, and they are nice and short! If I had to pick I'd prefer the first ("Safe Unsafe").

gribozavr:
> As a matter of etiquette, I tend not to click the Resolve conserveration button. Let me know it would be better to 'clean as I cook'.

Please click "resolve" if you believe that the reviewer would be happy with the way you resolved the comment (e.g., you accepted the suggested edit, you did roughly what they asked etc.) If you think the reviewer should take another look, leave a comment, but keep the conversation open.

src/running-the-course/course-structure.md
gribozavr: > Safe Unsafe

I'd be a pedant and say that this misuses the word "safe", the correct term would be "Sound Unsafe"

djmitche: Would "Safe and Sound" be too pithy, or is that a culturally-specific phrase? Anyway, I feel like we've spent too many words on this already.

gribozavr: While "Safe and Sound" is cute, it limits discoverability in a culturally-independent way: if someone is looking for information about unsafe, why would they click on "Safe and Sound"?

timClicks: > Sound Unsafe

I wonder if this verges on "Sounds Unsafe", which has a significantly more negative connotation.

src/unsafe-deep-dive/setup.md
```diff
@@ -21,3 +23,26 @@ We recommend that you install the [Bazel build system](https://bazel.build/insta
 This will allow you to easily compile project that combine multiple languages.
 
 -->
+
+## (Optional) Create a local instance of the course
+
+Having a local version of the course material is useful in case of any interruptions with the network and makes it easy to access it later.
```
gribozavr: A local version does not fully help: the playground widget runs code by performing an HTTP request to the standard Rust playground server :(

src/unsafe-deep-dive/motivations/interop.md
```diff
@@ -2,6 +2,29 @@
...
+> TODO: Add example
+
 # Interop
 
 Rust has no understanding of code generated from other languages.
+
+<details>
+
+Most libraries with a C ABI (which is most software libraries) provide an API
+that makes heavy use of `*void` pointers. The caller is provided no type
```
gribozavr: `void*` or `*c_void`

src/unsafe-deep-dive/motivations/interop.md
```diff
@@ -2,6 +2,29 @@
...
+information. Instead, users of the API are expected to read the documentation,
+read the correct number of bytes from that pointer and cast those bytes to the
+intended type.
```
gribozavr: Could you provide an example of that?

Where C libraries use opaque pointers, it is often to create opaque handle types that users should not dereference. For example:

https://sourcegraph.com/github.com/GNOME/libxml2/-/blob/include/libxml/xmlschemas.h?L89-90

There is of course a lot to say about dangers of pointers, but lack of type safety due to `void*` isn't the highest concern in most C APIs in my experience.

You can list issues such as lack of clarity about ownership or nullability, or expectations about lifetimes (for how long can I use the returned pointer value? for how long should I keep alive a pointer that I passed into a C library?)

timClicks: No, I don't have good examples of that. I have updated the content to focus primarily on lifetimes/ownership, away from type safety.

src/unsafe-deep-dive/motivations/interop.md
timClicks: Have removed this paragraph. I've replaced it with an expanded discussion later on with a code example. See `## Consideration: Ownership and lifetime management`.

src/unsafe-deep-dive/motivations/interop.md
timClicks: Marking as resolved as the updated content should have addressed this. Do feel free to re-open if you would like to discuss further.

src/unsafe-deep-dive/setup.md
timClicks: I've removed the commentary, but left the console instructions. Over time, I think it would make sense to move instructions like this to the prelude of the overall course.

timClicks:
@gribozavr You may wish to review additional content that is in the interop.md chapter, however I believe that I have addressed the points that you've raised. Thank you for these extensive comments.

src/unsafe-deep-dive/motivations/interop.md
```diff
@@ -0,0 +1,207 @@
...
+> }
+>
+> fn main() {
+> -    let a = random() as i64;
+> +    let a = random();
+>     println!("{a:?}");
+> }
+> ```
+
+It's also possible to completely erase the type. Stress that the Rust compiler
```
gribozavr: I'm not sure what you mean by "erase" here. Type erasure is somewhat of a different technique. Changing a type from an integer to a fixed-width array can significantly change the calling convention even if the bitwidth matches. (In this case, on x86-64, I believe, the actual returned random bytes are in a different register than where the caller is placing the array.)

src/unsafe-deep-dive/motivations/interop.md
```diff
@@ -0,0 +1,207 @@
...
+will trust that the wrapper is telling the truth.
+
+```rust
+unsafe extern "C" {
+    safe fn random() -> [u8; 64];
```
gribozavr: If you meant to match the bitwidth, it should have 8 elements.

src/unsafe-deep-dive/motivations/interop.md
```diff
@@ -0,0 +1,207 @@
...
+> -    let a = random() as i64;
+> -    println!("{a}");
+> +    let a = random();
+> +    println!("{a:?}");
+> }
+> ```
+
+Mention that type safety is generally not a large concern in practice.
+Auto-generated wrappers, i.e. those produced by bindgen and related tools, are
+excellent at reading header files and producing values of the correct type.
```
gribozavr: > Wrappers ... are excellent at reading

The wrappers are not doing the reading.

src/unsafe-deep-dive/motivations/interop.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Consideration: Ownership and lifetime management
+
+While libc's `random` function doesn't use pointers, may do. This creates the
```
gribozavr: "may do"?

src/unsafe-deep-dive/motivations/interop.md
```diff
@@ -0,0 +1,207 @@
...
+possibility that interacting with another programming language introduce
```
gribozavr: sp: "introduces"

This makes it sound like nothing that we talked about before could introduce unsoundness. That's not really true. Specifying an incorrect signature for an extern C function is unsound.

What you could say is that pointers on the FFI boundary create a unique type of unsoundness danger, and people spend a lot of time thinking about it when writing unsafe Rust. Other types of unsoundness are often easy to prevent because there are more or less algorithmic ways to check that we're using the correct extern C function signature, for example.

src/unsafe-deep-dive/motivations/interop.md
```diff
@@ -0,0 +1,207 @@
...
+
+For example, some C libraries expose functions that write to static buffers that
+are re-used between calls.
+
+```rust
+use std::ffi::{CStr, c_char};
+use std::time::{SystemTime, UNIX_EPOCH};
+
+unsafe extern "C" {
+    /// Create a formatted time based on time `t`, including trailing newline.
```
gribozavr, line 177: Add a few words about the ownership of the returned pointer? Maybe with a note that this is a short summary of the man page.

src/unsafe-deep-dive/motivations/interop.md
```diff
@@ -0,0 +1,207 @@
...
+    let future = now + 60;
+    let future_fmt = unsafe { format_timestamp(future) };
+    print!("future:  {}", future_fmt);
+
+    print!("now (2): {}", now_fmt);
+}
+```
+
+Bonus points: can anyone spot the lifetime bug? `format_timestamp()` should
+return a `&'static str`.
```
gribozavr, line 213: I think neither lifetime (`'a` or `'static`) is free of unsafe obligations, as subsequent calls to `ctime` overwrite the static buffer, breaking the AXM contract of immutable references.

That is, neither `&'a str` nor `&'static str` type fully describes the validity contract of the data, and thus all of the code that handles these values is burdened with an unsafe precondition (the code can't call `ctime` again).

src/unsafe-deep-dive/motivations/interop.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Consideration: Representation mismatch
+
+Different programming languages have made design decisions and this can create
```
gribozavr:
```suggestion
Different programming languages have made different design decisions and this can create
```

gribozavr (approved):

gribozavr:
LGTM for a first draft.

src/unsafe-deep-dive/motivations/interop.md
timClicks: No, it was intentionally the wrong size. I was tempted to add a note that it _shouldn't_ segfault but may.

timClicks: I've removed it in case it causes confusion in the future.

src/unsafe-deep-dive/motivations/interop.md
timClicks: I've rephrased the sentence to avoid the term of art.

src/unsafe-deep-dive/motivations/interop.md
timClicks: Indeed! Have fixed up.

src/unsafe-deep-dive/motivations/interop.md
timClicks: s/may/many/

src/unsafe-deep-dive/motivations/interop.md
timClicks: I've shortened this and I think that the result is more effective as a tighter sentence:

> While libc's `random` function doesn't use pointers, many do. This creates many
more possibilities for unsoundness.

src/unsafe-deep-dive/motivations/interop.md
timClicks: I was wary of introducing a safety section in the docstring because I didn't want to add a new concept out of order. Will leave a TODO in the source so that we can revisit.

src/unsafe-deep-dive/motivations/interop.md
timClicks: I spent a little bit of time thinking about this, which is where the "bonus points" comment came from. I decided that `'static` was probably better than `'a`, although I wasn't entirely satisfied. I've replaced the wording with a more neutral paragraph.

timClicks:
.. too soon.

src/unsafe-deep-dive/motivations/interop.md
```diff
@@ -92,12 +92,19 @@ fn main() {
 > }
 > ```
 
-It's also possible to completely erase the type. Stress that the Rust compiler
-will trust that the wrapper is telling the truth.
+It's also possible to completely ignore the intended type and create undefined
+behavior in multiple ways. The code below produces output most of the time, but
+generally results in a stack overflow. It may also produce illegal `char`
```
gribozavr, line 97: WDYM by a "stack overflow" here? I don't think that's what it does assuming the usual meaning of the term.

src/unsafe-deep-dive/motivations/interop.md
gribozavr: "read the man page" here perfectly communicates "I felt like I needed to write a comment, but I didn't want to spend time writing a thorough one". We can come back to this example in a future slide and show a proper version that's expected in Rust.

(no action needed in this PR)

src/running-the-course/course-structure.md
```diff
@@ -82,6 +82,15 @@ You should be familiar with the material in
 
 {{%course outline Idiomatic Rust}}
 
+### Unsafe (Work in Progress)
+
+The [Unsafe](../unsafe-deep-dive/welcome.md) deep dive is a two-day class on the
+_unsafe_ Rust language. It covers the fundamentals of what Rust's safety
+guarantees are and why `unsafe` is needed, reviewing `unsafe` code, using FFI,
+and building data structures that the borrow checker would normally reject.
```
gribozavr:
```suggestion
_unsafe_ Rust language. It covers the fundamentals of Rust's safety
guarantees, the motivation for `unsafe`, review process for `unsafe` code, FFI basics,
and building data structures that the borrow checker would normally reject.
```

An attempt to make list items parallel each other better.

gribozavr:
@timClicks Please squash-merge when you're ready.

src/unsafe-deep-dive/motivations/interop.md
timClicks: I was surprised too. Running [the code](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=afaf54388ab0681176cb75b11cd45dcf) a few times triggered an abort after the following message was sent to stderr:

> thread 'main' has overflowed its stack
> fatal runtime error: stack overflow, aborting

PR #2807, opened 2025-07-05
Update destructuring-structs.md

Enes1313:
In the #2749 , the match arm order has changed. Therefore, the word "second" should be changed to "first" in the notes.


djmitche (approved):
Good catch, thank you!

Issue #2811, opened 2025-07-07
Examples with hidden lines should be editable


LukeMathWalker:
# Problem

For this issue, there are two types of code blocks:

- Read-only ones, defined via ` ```rust `
- Editable ones, defined via ` ```rust,editable `

Read-only code blocks support the usage of `#` to hide code from the view that's rendered in the browser:

<img width="766" alt="Image" src="https://github.com/user-attachments/assets/f35329c2-cc90-4ddb-a4b0-90429583476d" />

You can click on the "Eye" toggle to show the hidden lines:

<img width="766" alt="Image" src="https://github.com/user-attachments/assets/7d932938-af53-43b5-911b-aae2960c260b" />

This makes it convenient to break an example across multiple code blocks without repeating content from the previous block (e.g., the definition of `UserId` in the example above).

Instructors often need to be able to edit code directly on the slide, but we lose the hiding feature if we mark the block as `editable`:

<img width="766" alt="Image" src="https://github.com/user-attachments/assets/6f5ec34d-13cb-4d6b-bb1b-8af20676e825" />

The `UserId` definition is shown, even if it was prefixed with `#` in the source.

# Solution

Modify the theme to introduce a new button, next to the "Copy" one, which allows instructors to switch from "Preview" to "Edit" mode.
In "Preview" mode, code is rendered as if the `editable` attribute had not been included.  
In "Edit" mode, all hidden code is shown and the instructor is free to edit.

We won't support going back from "Edit" to "Preview" mode.

gribozavr:
> Modify the theme to introduce a new button, next to the "Copy" one, which allows instructors to switch between "Preview" and "Edit" mode.

Could the "eye" button enable editing as well as revealing hidden code?

> When going back from "Edit" to "Preview", should the "Preview" show the original code or the code after the instructor edits?

I'd say it should use the code with the current edits.

LukeMathWalker:
> > Modify the theme to introduce a new button, next to the "Copy" one, which allows instructors to switch between "Preview" and "Edit" mode.
> 
> Could the "eye" button enable editing as well as revealing hidden code?

It could, if we want to.

> > When going back from "Edit" to "Preview", should the "Preview" show the original code or the code after the instructor edits?
> 
> I'd say it should use the code with the current edits.

[...]

PR #2812, opened 2025-07-09
Extension traits

LukeMathWalker:
Stacked on top of #2800.


src/idiomatic/leveraging-the-type-system/extension-traits.md
```diff
@@ -0,0 +1,37 @@
+---
+minutes: 5
+---
+
+# Extension Traits
+
+In Rust, you can't define new inherent methods for foreign types.
```
gribozavr: Is "foreign type" a standard term to refer to types from other crates? I feel like it can be easily misinterpreted as "types defined in C++" (similar to "foreign functions"). Is there an alternative?

src/idiomatic/leveraging-the-type-system/extension-traits/extending-foreign-types.md
```diff
@@ -0,0 +1,80 @@
...
+    impl StrExt for &str {
+        fn is_palindrome(&self) -> bool {
+            self.chars().eq(self.chars().rev())
+        }
+    }
+}
+
+// Bring the extension trait into scope..
+pub use ext::StrExt as _;
+// ..then invoke its methods as if they were inherent methods
```
gribozavr:
```suggestion
// Bring the extension trait into scope...
pub use ext::StrExt as _;
// ...then invoke its methods as if they were inherent methods
```

src/idiomatic/leveraging-the-type-system/extension-traits/method-resolution-conflicts.md
```diff
@@ -0,0 +1,79 @@
...
+
+  Change the signature of `StrExt::trim_ascii` to
+  `fn trim_ascii(&mut self) -> &str` and modify the invocation accordingly:
+
+  ```rust
+  assert_eq!((&mut " dad ").trim_ascii(), "dad");
+  ```
+
+  Now `StrExt::trim_ascii` is invoked, rather than the inherent method, since
+  `&mut self` is a more specific receiver than `&self`, the one used by the
```
djmitche: Is "more specific" defined somewhere?

djmitche (commented):
looking good!

src/idiomatic/leveraging-the-type-system/extension-traits/method-resolution-conflicts.md
LukeMathWalker: No, it's better to keep using the priority terminology. Reworded in [1670b64](https://github.com/google/comprehensive-rust/pull/2812/commits/1670b6468c86be8bfb96b53452ed14be5457f598)

src/idiomatic/leveraging-the-type-system/extension-traits/extending-foreign-types.md
```diff
@@ -0,0 +1,80 @@
...
+  require an additional import, and the familiarity of the method calling syntax
+  may not be enough to justify the boilerplate of a trait definition.
+
+  Nonetheless, extension methods can be **easier to discover** than free
+  functions. In particular, language servers (e.g. `rust-analyzer`) will suggest
+  extension methods if you type `.` after an instance of the foreign type.
+
+</details>
+
+[1]: https://rust-lang.github.io/rfcs/0445-extension-trait-rfc.html
```
randomPoison: This link is incorrect and gives a 404, the correct link is https://rust-lang.github.io/rfcs/0445-extension-trait-conventions.html

src/idiomatic/leveraging-the-type-system/extension-traits.md
```diff
@@ -0,0 +1,37 @@
...
+
+```rust,compile_fail
```
randomPoison, line 13: Can you make all the code snippets editable? It's often useful in class to be able to make modifications inline for demonstration purposes. I'm not going to leave a comment on all of the slides for this, but it looks like all of the code snippets in the slides are non-editable and need to be updated.

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
```diff
@@ -0,0 +1,86 @@
+---
+minutes: 15
+---
+
+# Extending Other Traits
```
randomPoison, line 5: It might be better to call this slide "Blanket Impls" since that's what this slide is demonstrating. At least personally, I don't think of blanket impls as "extending other traits", though that does seem like a reasonable way to look at it. But blanket impls are also generally useful outside of extension traits, so I think it's worth talking about blanket impls generally beyond their use in extension traits.

randomPoison (changes requested):
This looks good! I only have a couple of minor comments, but on the whole I like it :)

src/idiomatic/leveraging-the-type-system/extension-traits.md
LukeMathWalker: I will do a pass to mark them all as editable once #2811 is implemented!

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
LukeMathWalker: I see where you're coming from, but I disagree—blanket impls are a building block for this pattern, but they are a side character in this slide.
We could change the ToC to reserve some space for blanket impls, but then I would want to touch a variety of topics that we don't even mention here (e.g., overlap risk, common use cases to ease ergonomics, such as implementing traits for functions, etc.).

src/idiomatic/leveraging-the-type-system/extension-traits/extending-foreign-types.md
LukeMathWalker: Ops, thanks! Fixed in [c36f7fe](https://github.com/google/comprehensive-rust/pull/2812/commits/c36f7fe24d6b4758d259700d554ea8ce27d1b009)

src/idiomatic/leveraging-the-type-system/extension-traits.md
gribozavr: This one has no hidden lines though, why not make it editable right away?

src/idiomatic/leveraging-the-type-system/extension-traits.md
```diff
@@ -0,0 +1,37 @@
...
+        self.chars().eq(self.chars().rev())
+    }
+}
+```
+
+You can use the **extension trait pattern** to work around this limitation.
+
+<details>
+
+- Compile the example to show the compiler error that's emitted.
```
gribozavr, line 36: I think we should start by explaining what we want to achieve first: we want the user to be able to write something like `mystr.is_palindrome()`. Then transition to the obvious solution that does not work (the code snippet above). And then say that this is why we are using a more complex solution that does work.

Otherwise it might be confusing to some audience members: we are starting a new chapter by looking at a piece of code that does not compile (the impl block above), we want it to compile (why?..), but we actually wouldn't, instead we should do something else.

src/idiomatic/leveraging-the-type-system/extension-traits.md
```diff
@@ -0,0 +1,37 @@
...
+
+  Highlight how the compiler error message nudges you towards the extension
+  trait pattern.
+
+- Explain how many type-system restrictions in Rust aim to prevent _ambiguity_.
```
gribozavr, line 41: You can raise the level of abstraction even higher.

One effective way to approach evaluating features in programming design is to ask "what if everybody did this?"

"I want to be able to add methods to someone else's type! I want to add an is_palindrome method to a string" - "Yes, but what if two people did this?"

src/idiomatic/leveraging-the-type-system/extension-traits.md
```diff
@@ -0,0 +1,37 @@
...
+
+  If you were allowed to define new inherent methods on foreign types, there
+  would need to be a mechanism to disambiguate between distinct inherent methods
+  with the same name.
+
+  In particular, adding a new inherent method to a library type could cause
+  errors in downstream code if the name of the new method conflicts with an
+  inherent method that's been defined in the consuming crate.
```
gribozavr: Once you've said that "there would need to be a mechanism to disambiguate", the objection in this paragraph is rather weak. Yes, the methods could conflict, but we've already resigned to adding a disambiguation rule. So whenever we identify a case where methods conflict, we already know a fix - add a disambiguation rule, what's the big deal about it.

I think a better way to approach this explanation with fewer hypotheticals is to say that Rust doesn't even want to be in the business of designing these disambiguation rules, which might need to be quite complex in order to be effective at disambiguating. (You could draw a parallel with overload resolution in C++, which is extremely complex and subtle.)

src/idiomatic/leveraging-the-type-system/extension-traits/extending-foreign-types.md
```diff
@@ -0,0 +1,80 @@
...
+  crate where the trait is defined, otherwise you'll be blocked by Rust's
+  [_orphan rule_][2].
+
+- The extension trait must be in scope when its methods are invoked.
+
+  Comment out the `use` statement in the example to show the compiler error
+  that's emitted if you try to invoke an extension method without having the
+  corresponding extension trait in scope.
+
+- The `as _` syntax reduces the likelihood of naming conflicts when multiple
```
gribozavr: What does it do though? (I know what it does, this is a request to improve the text.)

src/idiomatic/leveraging-the-type-system/extension-traits/extending-foreign-types.md
```diff
@@ -0,0 +1,80 @@
...
+  traits are imported. It is conventionally used when importing extension
+  traits.
+
+- Some students may be wondering: does the extension trait pattern provide
```
gribozavr: I think this should be a separate slide: when to use an extension trait

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
gribozavr: > I don't think of blanket impls as "extending other traits"

I think this is a matter of perspective and background. "Protocol extensions" in Swift are a big deal (ex: [1](https://medium.com/@leandromperez/protocol-extensions-gotcha-9ef1a42c83b6)) and they are presented as extensions on the protocol, like "extending other traits" here.

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
```diff
@@ -0,0 +1,86 @@
...
+
+Extension traits can attach new methods to _all_ implementors of a given trait:
```
gribozavr: Like on the other slide, I'd suggest to start by explaining the desired result first.

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
```diff
@@ -0,0 +1,86 @@
...
+            format!("'{}'", self)
+        }
+    }
+}
+
+pub use ext::DisplayExt as _;
+
+assert_eq!("dad".quoted(), "'dad'");
+assert_eq!(4.quoted(), "'4'");
+assert_eq!(true.quoted(), "'true'");
```
gribozavr, line 29: This isn't runnable, is it?

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
```diff
@@ -0,0 +1,86 @@
...
+  experimental methods.
+
+  Stable methods are part of the trait definition.
+
+  Experimental methods are provided via an extension trait defined in a
+  different library, with a less restrictive stability policy. Some utility
+  methods are then "promoted" to the core trait definition once they have been
+  proven useful and their design has been refined.
+
+- Extension traits can be used to split a [dyn-incompatible trait][2] in two:
```
gribozavr, line 82: This is an excellent use case, but without a code example it is, most likely, impossible to learn from a mere description. The only type of student who would be satisfied with a terse description is someone who is already familiar with the idea.

src/idiomatic/leveraging-the-type-system/extension-traits/method-resolution-conflicts.md
```diff
@@ -0,0 +1,79 @@
...
+
+pub use ext::StrExt;
+// Which `trim_ascii` method is invoked?
+// The one from `StrExt`? Or the inherent one from `str`?
+assert_eq!(" dad ".trim_ascii(), "dad");
+```
+
+<details>
+
+- The foreign type may, in a newer version, add a new inherent method with the
```
gribozavr:
```suggestion
- The type may, in a newer version, add a new inherent method with the
```

src/idiomatic/leveraging-the-type-system/extension-traits/method-resolution-conflicts.md
```diff
@@ -0,0 +1,79 @@
...
+
+  Survey the class: what do the students think will happen in the example above?
+  Will there be a compiler error? Will one of the two methods be given higher
+  priority? Which one?
+
+  Add a `panic!("Extension trait")` in the body of `StrExt::trim_ascii` to
+  clarify which method is being invoked.
+
+- [Inherent methods have higher priority than trait methods][1], _if_ they have
+  the same name and the **same receiver**, e.g. they both expect `&self` as
```
gribozavr:
```suggestion
  the same name and the **same receiver**, e.g., they both expect `&self` as
```

src/idiomatic/leveraging-the-type-system/extension-traits/method-resolution-conflicts.md
```diff
@@ -0,0 +1,79 @@
...
+  input. The situation becomes more nuanced if the use a **different receiver**,
+  e.g. `&mut self` vs `&self`.
```
gribozavr:
```suggestion
  e.g., `&mut self` vs `&self`.
```

src/idiomatic/leveraging-the-type-system/extension-traits/method-resolution-conflicts.md
```diff
@@ -0,0 +1,79 @@
...
+  found in [an open PR to the Rust reference][3].
+
+- Avoid naming conflicts between extension trait methods and inherent methods.
+  Rust's method resolution algorithm is complex and may surprise users of your
+  code.
+
+## More to explore
+
+- The interaction between the priority search used by Rust's method resolution
+  algorithm and automatic `Deref`ering can be used to emulate
```
gribozavr:
```suggestion
  algorithm and automatic `Deref`ing can be used to emulate
```

src/idiomatic/leveraging-the-type-system/extension-traits/method-resolution-conflicts.md
```diff
@@ -0,0 +1,79 @@
+---
+minutes: 15
+---
+
+# Method Resolution Conflicts
+
+What happens when you have a name conflict between an inherent method and an
+extension method?
```
gribozavr, line 8: I think this content is missing transition. Creating naming conflicts is (obviously?) a bad idea, so how did we end up with code like this? Instead of figuring out what happens, shouldn't we remember simpler advice "don't do this"?

One way to introduce this slide is to set up a situation where we have added an extension method without conflicts, but then the type (owned by another crate) adds an inherent method.

src/idiomatic/leveraging-the-type-system/extension-traits/trait-method-conflicts.md
```diff
@@ -0,0 +1,62 @@
+---
+minutes: 5
+---
+
+# Trait Method Conflicts
+
+What happens when you have a name conflict between two different trait methods
```
gribozavr, line 7: We also need a setup here. This one is simple: imagine that Ext1 and Ext2 come from two different crates, and they chose the name "is_palindrome" independently, without coordination.

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
LukeMathWalker: It is, why not?

src/idiomatic/leveraging-the-type-system/extension-traits/method-resolution-conflicts.md
LukeMathWalker: That's the transition I envisioned—hence the first speaker note.
I would expect the instructor to set the stage accordingly, without pulling that paragraph into the slide itself.

Or do you prefer to be more explicit and pull that paragraph up?

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
LukeMathWalker: I placed this under "More to explore" to signal that the instructor can decide whether to mention/explain/dive into this topic.
I can expand it with code examples and more details, but then it'd make more sense to extract it into its own slide.
Alternatively, we can keep the terse explanation and provide a link to a more in-depth resource for those who want to dig deeper.

src/idiomatic/leveraging-the-type-system/extension-traits.md
LukeMathWalker: "Foreign type" and "foreign trait" are the terms used in the Rust reference when discussing orphan rules (see [this section](https://doc.rust-lang.org/stable/reference/items/implementations.html)). So I'd say they are standard terms in this specific context.

Alternatively, we could use "new inherent methods for a type defined in another crate" as an alternative phrasing. It could get fairly verbose in the speaker notes.

gribozavr: ack. Maybe then briefly mention in the speaker notes what these terms mean, hinting that the instructor should maybe explain them. I don't think it is a given that the audience understands coherence rules and understands why it matters where the trait is defined.

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
gribozavr: Because the `assert_eq!` should be in a function.

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
gribozavr: >  it'd make more sense to extract it into its own slide

Yes, that's what I'm hinting at :)

src/idiomatic/leveraging-the-type-system/extension-traits/method-resolution-conflicts.md
gribozavr: Ah sorry I missed that while writing this comment!

src/idiomatic/leveraging-the-type-system/extension-traits/trait-method-conflicts.md
LukeMathWalker: Same as above, is the first speaker note enough setup for the transition?

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
LukeMathWalker: The Rust playground will implicitly wrap them in a `main` function!

src/idiomatic/leveraging-the-type-system/extension-traits/trait-method-conflicts.md
```diff
@@ -0,0 +1,62 @@
...
+
+// Which method is invoked?
+// The one from `Ext1`? Or the one from `Ext2`?
+assert!("dad".is_palindrome());
+```
+
+<details>
+
+- The extended trait may, in a newer version, add a new trait method with the
+  same name of our extension method.
```
gribozavr:
```suggestion
- Another extension trait may, in a newer version, add a new trait method with the
  same name as our extension method.
```

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
randomPoison: Totally fair! I'm fine to leave things as-is.

src/idiomatic/leveraging-the-type-system/extension-traits.md
randomPoison: Personally I think it'd be fine to mark the blocks editable now, the hidden lines showing up doesn't seem like a big deal and they'll get hidden when #2811 is fixed. But since this isn't being actively taught it's not a big deal either way, as long as they're editable by the time this is being taught.

src/idiomatic/leveraging-the-type-system/extension-traits/extending-foreign-types.md
djmitche: Sub-slides can be good for this -- the main content is done, and there's a few more details to think about.

src/idiomatic/leveraging-the-type-system/extension-traits.md
LukeMathWalker: Added in [17ba065](https://github.com/google/comprehensive-rust/pull/2812/commits/17ba065295a8e0007a29ac1f41a599bc8eed7e65)

src/idiomatic/leveraging-the-type-system/extension-traits.md
LukeMathWalker: I restructured the flow a bit in [17ba065](https://github.com/google/comprehensive-rust/pull/2812/commits/17ba065295a8e0007a29ac1f41a599bc8eed7e65). What do you think?

src/idiomatic/leveraging-the-type-system/extension-traits.md
LukeMathWalker: Done in [17ba065](https://github.com/google/comprehensive-rust/pull/2812/commits/17ba065295a8e0007a29ac1f41a599bc8eed7e65)

src/idiomatic/leveraging-the-type-system/extension-traits.md
LukeMathWalker: Tried in [17ba065](https://github.com/google/comprehensive-rust/pull/2812/commits/17ba065295a8e0007a29ac1f41a599bc8eed7e65).

src/idiomatic/leveraging-the-type-system/extension-traits/extending-foreign-types.md
LukeMathWalker: I have expanded on this syntax in [17ba065](https://github.com/google/comprehensive-rust/pull/2812/commits/17ba065295a8e0007a29ac1f41a599bc8eed7e65), with a link to the reference.

src/idiomatic/leveraging-the-type-system/extension-traits/trait-method-conflicts.md
LukeMathWalker: Both cases are actually possible, so I reworded the paragraph to account for both in [17ba065](https://github.com/google/comprehensive-rust/pull/2812/commits/17ba065295a8e0007a29ac1f41a599bc8eed7e65)

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
LukeMathWalker: Done in 771a37a8

src/idiomatic/leveraging-the-type-system/extension-traits/extending-foreign-types.md
LukeMathWalker: Extracted in b58a2a52

src/idiomatic/leveraging-the-type-system/extension-traits.md
```diff
@@ -0,0 +1,63 @@
+---
+minutes: 15
+---
+
+# Extension Traits
+
+It may desirable to **extend** foreign types with new inherent methods. For
```
randomPoison, line 7:
```suggestion
It may desirable to **extend** foreign types with new methods. For
```

Since extension methods come from a trait they're not really "inherent" methods.

src/idiomatic/leveraging-the-type-system/extension-traits/method-resolution-conflicts.md
```diff
@@ -0,0 +1,78 @@
...
+  Change the signature of `StrExt::trim_ascii` to
+  `fn trim_ascii(&mut self) -> &str` and modify the invocation accordingly:
+
+  ```rust
+  assert_eq!((&mut " dad ").trim_ascii(), "dad");
+  ```
+
+  Now `StrExt::trim_ascii` is invoked, rather than the inherent method, since
+  `&mut self` has a higher priority than `&self`, the one used by the inherent
+  method.
```
randomPoison: I don't think this is an accurate explanation of what's happening here. The reference [explicitly states](https://doc.rust-lang.org/stable/reference/expressions/method-call-expr.html#r-expr.method.candidate-search) (in the info box in that section) that `&self` methods have higher priority than `&mut self` methods.

I think the reason why the `&mut self` version gets higher priority here is that the receiver expression is `&mut &str`. If I'm understanding the reference's explanation of method resolution correctly, this means that when it builds the list of candidate receiver types, `&mut &str` is the first candidate type in the list. It's then choosing between the inherent `&str` method and the `&mut &str` method coming from the trait, and the latter wins because it's the actual type of the expression `&mut " dad "`.

I think the confusion here is because we're implementing the trait on on `&str`, which is already a reference type. If I change the trait to be implemented on `str` directly (i.e. `impl StrExt for str`), when when I change the method to take `&mut self` the inherent method still gets called ([exmple in the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=40c1883946460bd6498dc7eab185a3d4)). Part of the reason for this is because when we do `(&mut " dad ")` we're not getting a `&mut str`, we're getting a `&mut &str`.


I think things would be a lot less ambiguous if we were demonstrating this on a regular, non-reference type such as `i32` or `struct Foo`.

src/idiomatic/leveraging-the-type-system/extension-traits/method-resolution-conflicts.md
```diff
@@ -0,0 +1,78 @@
...
+
+  Point the students to the Rust reference for more information on
+  [method resolution][2]. An explanation with more extensive examples can be
+  found in [an open PR to the Rust reference][3].
```
randomPoison:
```suggestion
  Point the students to the Rust reference for more information on
  [method resolution][2].
```

I think we can just link to the reference, I don't think linking to an open PR is necessary. Eventually the things in that PR will (hopefully) land, so just linking to the reference is enough imo.

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
```diff
@@ -0,0 +1,95 @@
...
+- Draw the students attention to the implementation of `DisplayExt::quoted`: we
+  can't make any assumptions about the type of `T` other than that it implements
+  `Display`. All our logic must either use methods from `Display` or
+  functions/macros that doesn't require `T` to implement any other trait.
+
+  We could introduce additional trait bounds on `T`, but it would restrict the
+  set of types that can leverage the extension trait.
+
+- Conventionally, the extension trait is named after the trait it extends,
+  following by the `Ext` suffix. In the example above, `DisplayExt`.
```
randomPoison: I'm not sure I agree with this, I think personally I'd name the trait something like `Quote` or `Quoted`. Calling it `DisplayExt` is reasonable if this is the only trait extending `Display`, but what if you had multiple traits extending `Display` in different ways? For example, I might have another trait that has a blanket impl on `T: Display + Clone` or something. Calling that `DisplayCloneExt` would be cumbersome and doesn't describe what behavior it's extending those traits with.

src/idiomatic/leveraging-the-type-system/extension-traits/trait-method-conflicts.md
```diff
@@ -0,0 +1,64 @@
...
+
+- The compiler rejects the code because it cannot determine which method to
+  invoke. Neither `Ext1` nor `Ext2` has a higher priority than the other.
+
+  To resolve this conflict, you must specify which trait you want to use. For
+  example, you can call `Ext1::is_palindrome("dad")` or
+  `Ext2::is_palindrome("dad")`.
+
+  For methods with more complex signatures, you may need to use a more explicit
+  [fully-qualified syntax][1].
```
randomPoison, line 63: It's probably worth showing the full syntax to students, since sometimes `Trait::method(foo)` isn't enough. Specifically, if the compiler can't infer the type of `foo` then it won't be able to resolve which type's trait implementation to use. In those cases you'd have to write `<Type as Trait>::method(foo)`. That syntax can be surprising for people new to the language (I know I was confused the first time I saw it), so I think showing it here (and explaining when it's necessary) would be good.

randomPoison (changes requested):
I think we might want to change the examples in this section to not be on `&str`, since implementing a trait on a reference leads to confusing behavior around method resolution (which I talk about in more detail in one of the comments here). I think things would be less confusing if we used a non-reference type like `i32` or `struct Foo`.

src/idiomatic/leveraging-the-type-system/extension-traits.md
LukeMathWalker: Extension trait methods are not inherent methods, I concur.
But in this paragraph, we're still talking about the _ideal_ case, which is to add an inherent implementation, as we try to do a few lines below before finding out it's not possible and falling back to the extension trait pattern.
So I'd say it's correct as it is.

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
LukeMathWalker: It's a matter of intent—if you have a trait with a distinct usecase that happens to need `Display` for a blanket implementation, then yes, you should name it according to its function.
But if the trait's purpose is to be an _extension trait_, then it should use `Ext` as its name suffix and this goes back to the naming recommendation/convention laid out in the relevant RFC (https://rust-lang.github.io/rfcs/0445-extension-trait-conventions.html#the-convention).

src/idiomatic/leveraging-the-type-system/extension-traits/method-resolution-conflicts.md
LukeMathWalker: Thanks for looking at it closely!
Re-reading through it, and cross-referencing with the RFC, I agree with your interpretation as to why things play out as they do in terms of precedence. I'll rework the example to something simpler.

src/idiomatic/leveraging-the-type-system/extension-traits/trait-method-conflicts.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+<details>
+
+- The trait you are extending may, in a newer version, add a new trait method
+  with the same name of your extension method. Or another extension trait for
+  the same type may define a method with a name that conflicts with your own
+  extension method.
+
+  Survey the class: what do the students think will happen in the example above?
```
djmitche: I think the approach of surveying the class is something specific to the instructor and situation. I would do this without prompting for small-group, in-person instruction, but for a large or shy group, or when using Zoom or Meet, it's not a very effective strategy (it results in lots of empty air).

So, it's a very minor point and not worth changing, but I think the question in the comment is sufficient here, and does not need repeating in the notes. Instructors will prompt from the comment if appropriate, or address it in a manner appropriate to the context and their teaching style.

djmitche (approved):
This looks great to me. The note below is stylistic and relates only to the speaker notes, so feel free to treat it as a mild preference at most.

src/idiomatic/leveraging-the-type-system/extension-traits.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+The Rust compiler won't allow it, though. But you can use the **extension trait
+pattern** to work around this limitation.
+
+<details>
+
+- Start by explaining the terminology.
+
+  A Rust item (be it a trait or a type) is referred to as:
```
gribozavr:
```suggestion
- A Rust item (be it a trait or a type) is referred to as:
```

Let's keep the speaker notes concise.

src/idiomatic/leveraging-the-type-system/extension-traits.md
gribozavr: LG, thanks!

src/idiomatic/leveraging-the-type-system/extension-traits.md
gribozavr: LG, thanks!

src/idiomatic/leveraging-the-type-system/extension-traits.md
```diff
@@ -0,0 +1,63 @@
...
+  If disambiguation happens implicitly, it can lead to surprising or otherwise
+  unexpected behavior. If disambiguation happens explicitly, it can increase the
+  cognitive load on developers who are reading your code.
+
+  Furthermore, every time a crate defines a new inherent method on a foreign
+  type, it may cause compilation errors in _your_ code, as you may be forced to
+  introduce explicit disambiguation.
+
+  Rust has decided to avoid the issue altogether by forbidding the definition of
+  new inherent methods on foreign types.
```
gribozavr, line 57:
```suggestion
  new inherent methods on foreign types.
  
- Other languages (e.g, Kotlin, C#, Swift) allow adding methods to existing types, often called "extension methods." This leads to different trade-offs in terms of potential ambiguities and the need for global reasoning.
```

src/idiomatic/leveraging-the-type-system/extension-traits.md
gribozavr: LG, thanks.

src/idiomatic/leveraging-the-type-system/extension-traits/extending-foreign-types.md
```diff
@@ -0,0 +1,66 @@
...
+  It communicates that the trait is primarily used for extension purposes, and
+  it is therefore not intended to be implemented outside the crate that defines
+  it.
+
+  Refer to the ["Extension Trait" RFC][1] as the authoritative source for naming
+  conventions.
+
+- The trait implementation for the chosen foreign type must belong to the same
+  crate where the trait is defined, otherwise you'll be blocked by Rust's
+  [_orphan rule_][2].
```
gribozavr:
```suggestion
- The extension trait implementation for a foreign type must be in the same crate as the trait itself, otherwise you'll be blocked by Rust's
  [_orphan rule_][2].
```

Consider this more succinct rewrite.

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
```diff
@@ -0,0 +1,95 @@
...
+pub use ext::DisplayExt as _;
+
+assert_eq!("dad".quoted(), "'dad'");
+assert_eq!(4.quoted(), "'4'");
+assert_eq!(true.quoted(), "'true'");
+```
+
+<details>
+
+- Highlight how we added new behaviour to _multiple_ distinct types at once.
```
gribozavr:
```suggestion
- Highlight how we added new behavior to _multiple_ types at once.
```

US spelling + removing unnecessary "distinct".

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
```diff
@@ -0,0 +1,95 @@
...
+  `.quoted()` can be called on string slices, numbers and booleans since they
+  all implement the `Display` trait.
+
+  This flavour of the extension trait pattern is built on top of
```
gribozavr:
```suggestion
  This flavour of the extension trait pattern uses
```

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
```diff
@@ -0,0 +1,95 @@
...
+  [_blanket implementations_][1].
+
+  Blanket implementations allow us to implement a trait for a generic type `T`,
+  as long as it satisfies the trait bounds specified in the `impl` block. In
+  this case, the only requirement is that `T` implements the `Display` trait.
```
gribozavr:
```suggestion
  A blanket implementation implements a trait for all types `T`
  that satisfy the trait bounds specified in the `impl` block. In
  this case, the only requirement is that `T` implements the `Display` trait.
```

Edits for clarity.

src/idiomatic/leveraging-the-type-system/extension-traits/method-resolution-conflicts.md
```diff
@@ -0,0 +1,78 @@
...
+pub use ext::StrExt;
+// Which `trim_ascii` method is invoked?
+// The one from `StrExt`? Or the inherent one from `str`?
+assert_eq!(" dad ".trim_ascii(), "dad");
+```
+
+<details>
+
+- The foreign type may, in a newer version, add a new inherent method with the
+  same name of our extension method.
```
gribozavr:
```suggestion
  same name as our extension method.
```

src/idiomatic/leveraging-the-type-system/extension-traits/method-resolution-conflicts.md
```diff
@@ -0,0 +1,78 @@
...
+  Survey the class: what do the students think will happen in the example above?
+  Will there be a compiler error? Will one of the two methods be given higher
+  priority? Which one?
+
+  Add a `panic!("Extension trait")` in the body of `StrExt::trim_ascii` to
+  clarify which method is being invoked.
+
+- [Inherent methods have higher priority than trait methods][1], _if_ they have
+  the same name and the **same receiver**, e.g., they both expect `&self` as
+  input. The situation becomes more nuanced if the use a **different receiver**,
```
gribozavr:
```suggestion
  input. The situation becomes more nuanced if they use a **different receiver**,
```

src/idiomatic/leveraging-the-type-system/extension-traits/method-resolution-conflicts.md
gribozavr: +1, I'd be happy if we explained a basic rule that people can remember, and simply mention that while the Rust language has rules to disambiguate in other cases, the rules are quite complex to remember and apply, and we'd rather not write code that depends on them.

src/idiomatic/leveraging-the-type-system/extension-traits/should-i-define-an-extension-trait.md
```diff
@@ -0,0 +1,40 @@
...
+<details>
+
+- A bespoke extension trait might be an overkill if you want to add a single
+  method to a foreign type. Both a free function and an extension trait will
+  require an additional import, and the familiarity of the method calling syntax
+  may not be enough to justify the boilerplate of a trait definition.
+
+  Nonetheless, extension methods can be **easier to discover** than free
+  functions. In particular, language servers (e.g. `rust-analyzer`) will suggest
+  extension methods if you type `.` after an instance of the foreign type.
```
gribozavr:
```suggestion
- Extension methods can be easier to discover than free functions. Language servers (e.g., `rust-analyzer`) will suggest them if you type `.` after an instance of the foreign type.

- However, a bespoke extension trait might be overkill for a single method. Both approaches require an additional import, and the familiar method syntax may not justify the boilerplate of a full trait definition.
```

Reordering for a stronger argument, focusing first on when to do it (answering the question on the slide), then commenting about the downsides.

gribozavr:
```suggestion
- **Discoverability:** Extension methods are easier to discover than free
  functions. Language servers (e.g., `rust-analyzer`) will suggest them if you
  type `.` after an instance of the foreign type.

- **Method Chaining:** A major ergonomic win for extension traits is method
  chaining. This is the foundation of the `Iterator` trait, allowing for fluent
  calls like `data.iter().filter(...).map(...)`. Achieving this with free
  functions would be far more cumbersome (`map(filter(iter(data), ...), ...)`).

- **API Cohesion:** Extension traits help create a cohesive API. If you have
  several related functions for a foreign type (e.g., `is_palindrome`,
  `word_count`, `to_kebab_case`), grouping them in a single `StrExt` trait is
  often cleaner than having multiple free functions for a user to import.

- **Tradeoffs:** Despite these advantages, a bespoke extension trait might be
  overkill for a single, simple function. Both approaches require an additional
  import, and the familiar method syntax may not justify the boilerplate of a
  full trait definition.
```

Adding a discussion of other advantages, a reference to the standard library, and points to think about.

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
```diff
@@ -0,0 +1,95 @@
...
+
+- Draw the students attention to the implementation of `DisplayExt::quoted`: we
+  can't make any assumptions about the type of `T` other than that it implements
```
gribozavr:
```suggestion
  can't make any assumptions about `T` other than that it implements
```

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
```diff
@@ -0,0 +1,95 @@
...
+  all implement the `Display` trait.
+
+  This flavour of the extension trait pattern is built on top of
+  [_blanket implementations_][1].
+
+  Blanket implementations allow us to implement a trait for a generic type `T`,
+  as long as it satisfies the trait bounds specified in the `impl` block. In
+  this case, the only requirement is that `T` implements the `Display` trait.
+
+- Draw the students attention to the implementation of `DisplayExt::quoted`: we
```
gribozavr:
```suggestion
- Draw the students' attention to the implementation of `DisplayExt::quoted`: we
```
possessive

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
```diff
@@ -0,0 +1,95 @@
...
+  can't make any assumptions about the type of `T` other than that it implements
+  `Display`. All our logic must either use methods from `Display` or
+  functions/macros that doesn't require `T` to implement any other trait.
```
gribozavr:
```suggestion
  functions/macros that don't require other traits..
```
"functions/macros" is plural + a concise rewrite.

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
gribozavr:
```suggestion
  followed by the `Ext` suffix. In the example above, `DisplayExt`.
```

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
```diff
@@ -0,0 +1,95 @@
...
+  different library, with a less restrictive stability policy. Some utility
+  methods are then "promoted" to the core trait definition once they have been
+  proven useful and their design has been refined.
+
+- Extension traits can be used to split a [dyn-incompatible trait][2] in two:
+
+  - A **dyn-compatible core**, restricted to the methods that satisfy
+    dyn-compatibility requirements.
+  - An **extension trait**, containing the remaining methods that are not
+    dyn-compatible. (e.g., methods with a generic parameter).
```
gribozavr:
```suggestion
    dyn-compatible (e.g., methods with a generic parameter).
```
The parenthetical continues the sentence.

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
```diff
@@ -0,0 +1,95 @@
...
+
+  Blanket implementations allow us to implement a trait for a generic type `T`,
+  as long as it satisfies the trait bounds specified in the `impl` block. In
+  this case, the only requirement is that `T` implements the `Display` trait.
+
+- Draw the students attention to the implementation of `DisplayExt::quoted`: we
+  can't make any assumptions about the type of `T` other than that it implements
+  `Display`. All our logic must either use methods from `Display` or
+  functions/macros that doesn't require `T` to implement any other trait.
+
```
gribozavr, line 52:
```suggestion

  For example, we can call `format!` with `T`, but can't call `.to_uppercase()` because it is not necessarily a `String`.
  
```

src/idiomatic/leveraging-the-type-system/extension-traits/extending-other-traits.md
```diff
@@ -0,0 +1,95 @@
...
+- Conventionally, the extension trait is named after the trait it extends,
+  following by the `Ext` suffix. In the example above, `DisplayExt`.
+
+- There are entire libraries aimed at extending foundational traits with new
+  functionality.
+
+  [`itertools`] provides a wide range of iterator adapters and utilities via the
+  [`Itertools`] trait. [`futures`] provides [`FutureExt`] to extend the
+  [`Future`] trait.
```
gribozavr:
```suggestion
- There are entire crates that extend standard library traits with new functionality.

  - `itertools` crate provides the `Itertools` trait that extends `Iterator`. It adds many iterator adapters, such as `interleave` and `unique`. It provides new algorithmic building blocks for iterator pipelines built with method chaining.
  
  - `futures` crate provides the `FutureExt` trait, which extends the `Future` trait with new combinators and helper methods.
```

More structure with a nested list + a bit more content.

src/idiomatic/leveraging-the-type-system/extension-traits/trait-method-conflicts.md
```diff
@@ -0,0 +1,64 @@
...
+
+// Which method is invoked?
+// The one from `Ext1`? Or the one from `Ext2`?
+assert!("dad".is_palindrome());
+```
+
+<details>
+
+- The trait you are extending may, in a newer version, add a new trait method
+  with the same name of your extension method. Or another extension trait for
```
gribozavr:
```suggestion
  with the same name as your extension method. Or another extension trait for
```

src/idiomatic/leveraging-the-type-system/extension-traits/trait-method-conflicts.md
```diff
@@ -0,0 +1,64 @@
...
+  Survey the class: what do the students think will happen in the example above?
+  Will there be a compiler error? Will one of the two methods be given higher
+  priority? Which one?
+
+- The compiler rejects the code because it cannot determine which method to
+  invoke. Neither `Ext1` nor `Ext2` has a higher priority than the other.
+
+  To resolve this conflict, you must specify which trait you want to use. For
+  example, you can call `Ext1::is_palindrome("dad")` or
+  `Ext2::is_palindrome("dad")`.
```
gribozavr:
```suggestion
  `Ext2::is_palindrome("dad")`. Demonstrate this syntax and that the updated code compiles.
```

src/idiomatic/leveraging-the-type-system/extension-traits/trait-method-conflicts.md
gribozavr: +1, that would be a good thing for the instructor to demonstrate. Please add a bullet point with the prompt for the instructor.

src/SUMMARY.md
```diff
@@ -437,6 +437,12 @@
     - [Semantic Confusion](idiomatic/leveraging-the-type-system/newtype-pattern/semantic-confusion.md)
     - [Parse, Don't Validate](idiomatic/leveraging-the-type-system/newtype-pattern/parse-don-t-validate.md)
     - [Is It Encapsulated?](idiomatic/leveraging-the-type-system/newtype-pattern/is-it-encapsulated.md)
+  - [Extension Traits](idiomatic/leveraging-the-type-system/extension-traits.md)
+    - [Extending Foreign Types](idiomatic/leveraging-the-type-system/extension-traits/extending-foreign-types.md)
+    - [Method Resolution Conflicts](idiomatic/leveraging-the-type-system/extension-traits/method-resolution-conflicts.md)
+    - [Should I Define An Extension Trait?](idiomatic/leveraging-the-type-system/extension-traits/should-i-define-an-extension-trait.md)
```
gribozavr: I think we could improve the flow if we moved the "should I" section after the "extending other traits" section.

The "should I" section discusses ideas that apply to all usages of this pattern. (It would also flow even better if you apply my suggestions about API coherence when adding multiple methods, because the "extending other traits" section would introduce that idea first)

gribozavr:
@LukeMathWalker I left another round of comments, but please don't be discouraged - this section looks very good and the comments are mostly minor. Please work through resolving the comments, rebase the PR to resolve conflicts with main, and merge. You have my LGTM.

src/idiomatic/leveraging-the-type-system/extension-traits/method-resolution-conflicts.md
tall-vase: I've reworked this slide to avoid the reference confusion and be more explicit about the rule system (as far as I understand it).

src/idiomatic/leveraging-the-type-system/extension-traits/should-i-define-an-extension-trait.md
```diff
@@ -0,0 +1,55 @@
...
+  calls like `data.iter().filter(...).map(...)`. Achieving this with free
+  functions would be far more cumbersome (`map(filter(iter(data), ...), ...)`).
+- **API Cohesion:** Extension traits help create a cohesive API. If you have
+  several related functions for a foreign type (e.g., `is_palindrome`,
+  `word_count`, `to_kebab_case`), grouping them in a single `StrExt` trait is
+  often cleaner than having multiple free functions for a user to import.
+- **Trade-offs:** Despite these advantages, a bespoke extension trait might be
+  overkill for a single, simple function. Both approaches require an additional
+  import, and the familiar method syntax may not justify the boilerplate of a
+  full trait definition.
```
gribozavr, line 56:
```suggestion
- **Discoverability:** Extension methods are easier to discover than free
  functions. Language servers (e.g., `rust-analyzer`) will suggest them if you
  type `.` after an instance of the foreign type.

- **Method Chaining:** A major ergonomic win for extension traits is method
  chaining. This is the foundation of the `Iterator` trait, allowing for fluent
  calls like `data.iter().filter(...).map(...)`. Achieving this with free
  functions would be far more cumbersome (`map(filter(iter(data), ...), ...)`).

- **API Cohesion:** Extension traits help create a cohesive API. If you have
  several related functions for a foreign type (e.g., `is_palindrome`,
  `word_count`, `to_kebab_case`), grouping them in a single `StrExt` trait is
  often cleaner than having multiple free functions for a user to import.

- **Trade-offs:** Despite these advantages, a bespoke extension trait might be
  overkill for a single, simple function. Both approaches require an additional
  import, and the familiar method syntax may not justify the boilerplate of a
  full trait definition.
```

src/idiomatic/leveraging-the-type-system/extension-traits/trait-method-conflicts.md
```diff
@@ -0,0 +1,64 @@
...
+- The compiler rejects the code because it cannot determine which method to
+  invoke. Neither `Ext1` nor `Ext2` has a higher priority than the other.
+
+  To resolve this conflict, you must specify which trait you want to use. For
+  example, you can call `Ext1::is_palindrome("dad")` or
+  `Ext2::is_palindrome("dad")`. Demonstrate this syntax and that the updated
+  code compiles.
+
+  For methods with more complex signatures, you may need to use a more explicit
+  [fully-qualified syntax][1].
```
gribozavr, line 63: I have previously left a comment here that wasn't acted upon:

> +1, that would be a good thing for the instructor to demonstrate. Please add a bullet point with the prompt for the instructor.

src/idiomatic/leveraging-the-type-system/extension-traits/trait-method-conflicts.md
```diff
@@ -0,0 +1,71 @@
...
+  To resolve this conflict, you must specify which trait you want to use.
+
+  Demonstrate: call `Ext1::is_palindrome(&"dad")` or
+  `Ext2::is_palindrome(&"dad")` instead of `"dad".is_palindrome()`.
+
+  For methods with more complex signatures, you may need to use a more explicit
+  [fully-qualified syntax][1].
+
+- Demonstrate: replace `"dad".is_palindrome()` with
+  `<&str as Ext1>::is_palindrome(&"dad")` or<&str as
```
gribozavr:
```suggestion
  `<&str as Ext1>::is_palindrome(&"dad")` or `<&str as
```

PR #2815, opened 2025-07-13
Fix typo: "appenix" → "appendix"

Enes1313:
If I'm not mistaken...


djmitche (approved):
"Appendix-itis" isn't the correct spelling either, but yes, I think this was a typo :)

PR #2818, opened 2025-07-15
Add a note about using rust-bindgen

JasonXJ:


src/unsafe-rust/exercise.md
```diff
@@ -74,3 +74,12 @@ functions and methods:
 
 {{#include exercise.rs:main}}
 ```
+
+<details>
+
+The `ffi` module here is for demonstration purposes. Normally, you should be
+using crates such as [rust-bindgen] to generate the binding code.
```
timClicks: Sorry to be pedantic, but while the repository is named `rust-bindgen`, `bindgen` is the name of the CLI. Strictly speaking, there are multiple crates in that repo, so I've suggested the term tool.

```suggestion
using tools such as [bindgen] to generate bindings.
```

src/unsafe-rust/exercise.md
```diff
@@ -74,3 +74,12 @@ functions and methods:
 
 {{#include exercise.rs:main}}
 ```
+
+<details>
+
+The `ffi` module here is for demonstration purposes. Normally, you should be
```
timClicks:
```suggestion
The `ffi` module here is for learning purposes. Normally, you should be
```

timClicks (commented):
Seems like a useful additional comment, especially for learners using the material to teach themselves. However, I do wonder whether it's necessary to make it clear that this is teaching code only, given that it is part of a course.

JasonXJ:
Hi Tim, thanks for the review!

> However, I do wonder whether it's necessary to make it clear that this is teaching code only, given that it is part of a course.

Are you suggesting removing the first sentence? Or are you suggesting that maybe we don't need the comment at all?

src/unsafe-rust/exercise.md
```diff
@@ -74,3 +74,12 @@ functions and methods:
 
 {{#include exercise.rs:main}}
 ```
+
+<details>
+
+The `ffi` module here is for learning purposes. Normally, you should be using
```
gribozavr:
```suggestion
The example code manually spells out `extern "C"` declarations in the `ffi` module only to make the source file self-contained, which is necessary to enable the exercise to be completed in the online playground. Normally, in production code you should be using
```

Yeah, I don't think "learning" / "demonstration" conveys the right nuance here. We are not trying to teach students to write `extern "C"` declarations in this exercise. The students get these as a part of the code template. Instead, we are providing these declarations in order to work around limitations of the Rust playground not being able to build complex projects (e.g., invoke a build script that runs bindgen).

JasonXJ: This makes sense, thanks! And sorry for bikesheding here, but the `ffi` module also contains struct definitions in addition to `extern "C"` stuff. What do you think about just keeping the note short without explaining too much? Please see the updated PR.

src/unsafe-rust/exercise.md
```diff
@@ -74,3 +74,12 @@ functions and methods:
 
 {{#include exercise.rs:main}}
 ```
+
+<details>
+
+FFI binding code is typically generated by tools like [bindgen], rather than
+being written manually as we are doing here.
```
gribozavr:
```suggestion
being written manually as we are doing here. However, bindgen can't run in an online playground.
```

src/unsafe-rust/exercise.md
gribozavr: See my new suggested edit - I found a way to make the explanation short.

src/unsafe-rust/exercise.md
JasonXJ: Thanks! Suggestion applied.

gribozavr (approved):

PR #2819, opened 2025-07-15
Unsafe deep dive - Mechanics segment

timClicks:
A course segment to describe the characteristics of well-written unsafe blocks.


timClicks:
Deleting for now as it is branched from another PR and those commits should probably be squashed.

PR #2821, opened 2025-07-16
add typestate pattern chapter for idiomatic rust

GlenDC:


src/idiomatic/leveraging-the-type-system/typestate-pattern.md
```diff
@@ -0,0 +1,86 @@
...
+## Typestate Pattern
+
+The typestate pattern uses Rust’s type system to make **invalid states
+unrepresentable**.
+
+```rust
+# use std::fmt::Write;
+#[derive(Default)]
+struct Serializer { output: String }
+struct SerializeStruct { serializer: Serializer }
```
gribozavr: I'd like to suggest to first show a slide that implements the most straightforward solution possible: all methods are on a single type, and there is no fluent interface (no `-> Self`). Use that example to explain the domain-specific problem that the code is trying to solve, and demonstrate the coding bug that you'd like to prevent (the main function does not call finish_struct).

src/idiomatic/leveraging-the-type-system/typestate-pattern.md
```diff
@@ -0,0 +1,86 @@
...
+- This example is inspired by
+  [Serde's `Serializer` trait](https://docs.rs/serde/latest/serde/ser/trait.Serializer.html).
+  For a deeper explanation of how Serde models serialization as a state machine,
+  see <https://serde.rs/impl-serializer.html>.
+
+- The typestate pattern allows us to model state machines using Rust’s type
+  system. In this case, the state machine is a simple serializer.
+
+- The key idea is that each state in the process, starting a struct, writing
+  fields, and finishing, is represented by a different type. Transitions between
```
gribozavr: This suggests that there should be 3 states?

Maybe you could work the String type into it, as the final state.

src/idiomatic/leveraging-the-type-system/typestate-pattern.md
```diff
@@ -0,0 +1,86 @@
...
+  states happen by consuming one value and producing another.
```
gribozavr: Consider adding an illustration? See https://google.github.io/comprehensive-rust/memory-management/review.html for an example of how to use our tooling. Show the state machine: states are types, transitions are methods.

src/idiomatic/leveraging-the-type-system/typestate-pattern.md
```diff
@@ -0,0 +1,86 @@
+---
+minutes: 15
+---
+
+## Typestate Pattern
+
+The typestate pattern uses Rust’s type system to make **invalid states
+unrepresentable**.
```
gribozavr: I'm a bit of a stickler for precise language. Typestate is the practice of encoding a part of the state of the value in its type, preventing incorrect/inapplicable operations from being called on the value.

So in this case, we didn't make invalid states unrepresentable - they were no invalid states to begin with. What we achieved is that we made the `Serializer` to `String` conversion not compile if the serializer is not ready for it.

"Make invalid states not representable" is better illustrated by an example like this, which is not using typestate:

```rust
enum Operation {
  Negate,
  Multiply,
}
struct Expr {
  op: Operation,
  lhs: Box<Expr>,
  rhs: Option<Box<Expr>>
}
// `Expr { op: Operation::Negate, lhs: Some(...), rhs: None }` is OK
// `Expr { op: Operation::Multiply, lhs: Some(...), rhs: Some(...) }` is OK
// `Expr { op: Operation::Multiply, lhs: Some(...), rhs: None }` is invalid, but possible to construct
```

```rust
// better:
enum Expr {
  Negate(lhs: Box<Expr>),
  Multiply(lhs: Box<Expr>, rhs: Box<Expr>)
}
```

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-generics.md
```diff
@@ -0,0 +1,141 @@
+## Typestate Pattern with Generics
+
+Generics can be used with the typestate pattern to reduce duplication and allow
+shared logic across state variants, while still encoding state transitions in
+the type system.
+
+```rust
```
gribozavr, line 8: Same feedback as for the serializer example - let's start this section with a slide that has a problem statement, a straightforward solution, and a demonstration of a coding bug that typestate could prevent here. Then show an illustration with the state machine that we're trying to implement with typestate and say that if we implement a separate type for each state like we did before, we would have a lot of types and a lot of code. Then you can reveal this example.

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-generics.md
```diff
@@ -0,0 +1,141 @@
...
+    like `Ready<T>`.
+  - Error messages from the compiler may become more opaque, especially if a
+    trait bound is not satisfied or a state transition is incomplete.
+
+    The error messages might also be hard to follow due to the complexity as a
+    result of the nested generics types.
+
+- Still, in return for this complexity, you get compile-time enforcement of
+  valid configuration, clear builder sequencing, and no possibility of
+  forgetting a required step or misusing the API at runtime.
```
gribozavr: I'd suggest finishing this section by showing the state graph again, and asking the audience to analyze how many user journeys there are really.

I think we have essentially two paths in the graph with an optional timeout in front (listing the edges):

* [optional timeout()], insecure(), connect()
* [optional timeout()], secure(), client_certificate(), connect()

Thus, while this is an excellent example of code sharing when using typestate, if we were to design this API for production, and we truly had only two user stories, using typestate would create a lot of API complexity. We would be much better off with two simple factory functions:

```
impl Connection {
  fn new_insecure(timeout: Option<u64>)
  fn new_secure(client_certificate: Vec<u8>, timeout: Option<u64>)
}
```

The moral is, you don't need to use all these tools only because you now know how.

src/idiomatic/leveraging-the-type-system/typestate-pattern.md
GlenDC: Great suggestion, I'll move this content to the 2nd slide.

src/idiomatic/leveraging-the-type-system/typestate-pattern.md
GlenDC: Hmm I think I would rather remove the ambiguity in my wording as adding the String type into it feels a bit far-fetched in this example. In a more realistic scenario you neither would want that String after each individual unit serialized.

Good catch though.

src/idiomatic/leveraging-the-type-system/typestate-pattern.md
GlenDC: I will do that for sure. It was in my head the entire time but I wasn't sure if it was desired or not. Good to know you think it is. Those illustrations look beautiful, I'll add one for sure. Do you see that as part of the speaker notes, or moved up to the main content?

src/idiomatic/leveraging-the-type-system/typestate-pattern.md
GlenDC: That's about Sum vs Product types, where you prevent combinations that are not valid. E.g. an error + success.

It's subtle but I think I get your point. Will take it into account.

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-generics.md
GlenDC: That means splitting it up in 2 slides, where the revealing happens because we move form slide A to slide B? Or if not, do you have an example in the book where I can see your desired approach in action?

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-generics.md
GlenDC: Ok, so you wanna make sure, like with most tools, that people need to be aware that you do not want to overuse it. But at the same time I think you agree that we cannot expand the example further then it already is? It's mostly just to teach that moral, right?

src/idiomatic/leveraging-the-type-system/typestate-pattern.md
gribozavr: One way to think about it is to look at the state diagram: invalid states vs. invalid transitions.

src/idiomatic/leveraging-the-type-system/typestate-pattern.md
gribozavr: I think it should be shown on the slide, it could be a good visualization.

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-generics.md
gribozavr: Yes, exactly. Mention this for students who might otherwise want to over-engineer their next API, and also acknowledge the feelings of those who are sitting there thinking "why did we need all this for a relatively simple use case". This is merely an example to illustrate code sharing, a real world API example where this pattern would be better than simple factory functions would be much more complex.

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-generics.md
gribozavr: Right, I'm suggesting to split into two slides: first slide to convey to students what it is that we're trying to achieve with this API on the domain level using straightforward but bug-prone implementation of a builder, and second slide with a typestate version of the builder (what you have right now).

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-generics.md
```diff
@@ -0,0 +1,141 @@
...
+
+fn main() -> std::io::Result<()> {
+    let _conn = Connection::new("db.local")
+        .secure()
+        .client_certificate(vec![1, 2, 3])
+        .timeout(10)
+        .connect()?;
+    Ok(())
+}
+```
```
randomPoison: This example is way too big. We generally want code examples like this to fit on the screen, or at least require minimal scrolling. This example would be very difficult to show to students in class as it would require a lot of scrolling up and down, and we can only fit a little bit of it on screen at a time.

randomPoison: One potential idea here is to continue to expand on the serializer example used in the previous slide, adding some shared state that can be de-duplicated with generic typestate generics.

randomPoison (changes requested):
As @gribozavr suggested, I think both of these slides would benefit from being split in two, where we start with a suboptimal API that allows incorrect usage, then demonstrate how typestate can make incorrect usage impossible.

GlenDC:
The four slides in the typestate pattern chapter are ready for review again.

Note that the generics slide currently exceeds the height limit and fails the corresponding test. I’ve left it as-is for now, since I wasn’t sure how you’d prefer to address that. Let me know how you’d like to proceed.

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-generics.md
```diff
@@ -0,0 +1,268 @@
...
+        # self.buffer
+    }
+}
+
+impl<S> Serializer<S> {
+    fn buffer_size(&self) -> usize {
+        // [...]
+        # self.buffer.len()
+    }
+}
```
randomPoison: I don't think we need `buffer_size`, I don't think printing out the buffer size as part of the demo is necessary and having it here adds a bit of noise to an already large example.

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-generics.md
randomPoison, line 1: I like the example here, I think it does a good job demonstrating a more complex case using typestate, but it is definitely a lot to fit into one slide. I think the move here is to break this up into smaller slides, where each slide focuses on just a portion of the structure, e.g. one for `impl<S> Serializer<Struct<S>>` and one for `impl<S> Serializer<List<S>>`, etc. Then we'd have a final file that shows how the API is used, along with demonstrating how it rejects incorrect usage. We may not even need to have a slide for every piece of this, maybe we just demonstrate a couple of the cases (e.g. structs and lists), and leave the rest of the setup in the playground.

What you may want to do is put this example in its own `.rs` file and then use anchors and imports to pull snippets of it into slides. This is the approach we use in a lot of the exercises.

randomPoison: Something else worth mentioning about this example is that all of the state structs are zero-sized, meaning there's no memory or runtime overhead from constructing complex state types.

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-generics.md
```diff
@@ -0,0 +1,268 @@
...
+    V           |   V         |            V     |
+                +                                |
+serializer --> structure --> property --> list +-+
+
+    |           |   ^           |          ^
+    V           |   |           |          |
+                |   +-----------+          |
+  String        |                          |
+                +--------------------------+
+```
```
randomPoison, line 35: Why is there an arrow from `structure` to `list` in this diagram? `Serializer<Struct>` only allows starting a property or finishing the struct.

src/idiomatic/leveraging-the-type-system/typestate-pattern.md
```diff
@@ -0,0 +1,89 @@
...
+-+---------------------
+ |
+ +--> serialize struct field
+      -+---------------------
+       |
+       +--> serialize struct field
+            -+---------------------
+             |
+             +--> serialize struct end
+```
```
randomPoison: I don't think we need this diagram. The intended flow here is pretty straightforward since the methods are intended to be used in the order they're declared. If you think it's worth explicitly calling out the intended flow then I think just writing it in a note would be enough.

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-example.md
```diff
@@ -0,0 +1,99 @@
...
++------------+  serialize struct   +-----------------+
+| Serializer | ------------------> | SerializeStruct | <------+
++------------+                     +-----------------+        |
+                                                              |
+   |   ^                             |     |                  |
+   |   |     finish struct           |     | serialize field  |
+   |   +-----------------------------+     +------------------+
+   |
+   +---> finish
+```
```
randomPoison, line 64: I think this diagram would be useful to show to students, so I think we should move it into the slide.

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-advanced.md
randomPoison, line 1: I think for this slide, rather than showing a bunch of stubbed out code, it might be better to show an example of the more complex serialized output that we want to support. That would then lead nicely into a few slides showing how we can build the more complex serializer.

randomPoison (commented):
This is coming together nicely! I think we still need to break up the generics example since it's too much for a single slide, but the example itself is a good demonstration of typestate.

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-generics.md
GlenDC: The goal was mostly to have at least 1 method that is use-able at any stage of the `Serializer`. Not something you deem useful? Or if you do but would like it to be shown differently, how?

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-generics.md
GlenDC: > What you may want to do is put this example in its own .rs file and then use anchors and imports to pull snippets of it into slides. This is the approach we use in a lot of the exercises.

Will investigate

> Something else worth mentioning about this example is that all of the state structs are zero-sized, meaning there's no memory or runtime overhead from constructing complex state types.

Fair, will see where that fits. Just need to make clear that it no longer would be a `ZST` as soon as you do add state in one of them (e.g. for avoiding duplicates in a `Struct`)

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-generics.md
GlenDC: Correct. It's because of your last option. Finishing a struct goes to its parent which can be:

- the root serializer
- another struct
- a list

As such it can also go back to a `list`.

src/idiomatic/leveraging-the-type-system/typestate-pattern.md
GlenDC: Sure, will handle that. Thank you for pointing it out.

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-example.md
GlenDC: In that case it will however no longer fit in the screen I think, when combined with the code. That is not an issue?

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-advanced.md
GlenDC: Sure thing, your other comments seem to allude to how to do that.
Gonna think about it a bit more and get back to you with a draft on it next time I get back to this PR. Thank you for all the feedback!

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-generics.md
randomPoison: I don't think that needs to be demonstrated directly. The example already shows how we can keep common state in `Serializer`, it's not a large leap to figure out this applies to methods as well. I think all we need here is a speaker note calling out that common state/behavior can go in `Serializer`.

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-generics.md
randomPoison: > Just need to make clear that it no longer would be a `ZST` as soon as you do add state in one of them (e.g. for avoiding duplicates in a `Struct`)

Ah, true. Maybe just mention that there's no additional overhead from composing these state marker types, and that ones that are ZSTs have no overhead at all. This can just be a speaker note, to be clear.

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-example.md
randomPoison: I think it's fine in this case. We do have a few slides that are a bit longer because they have some diagrams in them, e.g. the [Move Semantics slide](https://google.github.io/comprehensive-rust/memory-management/move.html), and I've found those reasonable to cover in class.

src/idiomatic/leveraging-the-type-system/typestate-pattern/typestate-advanced.md
randomPoison: Sounds good! Also, feel free to reach out if you want to further discuss any of my suggestions, I'm happy to discuss and brainstorm further.

djmitche:
I've removed myself as reviewer as a form of load-shedding, but I looked through this a week or so ago and had no objections, so consider this a soft approval :)

PR #2827, opened 2025-07-21
Add Mechanics segment to the Unsafe Deep Dive

timClicks:
This segment provides teaching material for what a well-written unsafe code block or function/trait should look like.


randomPoison:
Do you want a review of this as it is now? Or would it be better to wait until it's no longer WIP?

timClicks:
Closing; needs rethink.

PR #2832, opened 2025-07-27
Fix typo in Matching Values section

daivinhtran:
Given 

```
#[rustfmt::skip]
fn main() {
    let input = '1';
    match input {
        key if key.is_lowercase() => println!("Lowercase: {key}"),
        'q'                       => println!("Quitting"),
        'a' | 's' | 'w' | 'd'     => println!("Moving around"),
        '0'..='9'                 => println!("Number input"),
        _                         => println!("Something else"),
    }
}
```

the output will be

```
Number input
```

So in practice, failing the condition does result to the other arms if the other arms are after that (if I'm reading this correctly).

### DIFF
--- a/src/pattern-matching/match.md
+++ b/src/pattern-matching/match.md
@@ -43,7 +43,7 @@ Key Points:
   wish to concisely express more complex ideas than patterns alone would allow.
 - They are not the same as separate `if` expression inside of the match arm. An
   `if` expression inside of the branch block (after `=>`) happens after the
-  match arm is selected. Failing the `if` condition inside of that block won't
+  match arm is selected. Failing the `if` condition inside of that block will
   result in other arms of the original `match` expression being considered.
 - The condition defined in the guard applies to every expression in a pattern
   with an `|`.


gribozavr:
reference: https://doc.rust-lang.org/reference/expressions/match-expr.html#r-expr.match.guard.behavior

> When the pattern matches successfully, the pattern guard expression is executed. If the expression evaluates to true, the pattern is successfully matched against.
>
> Otherwise, the next pattern, including other matches with the `|` operator in the same arm, is tested.

src/pattern-matching/match.md
```diff
@@ -43,7 +43,7 @@ Key Points:
   wish to concisely express more complex ideas than patterns alone would allow.
 - They are not the same as separate `if` expression inside of the match arm. An
   `if` expression inside of the branch block (after `=>`) happens after the
-  match arm is selected. Failing the `if` condition inside of that block won't
+  match arm is selected. Failing the `if` condition inside of that block will
```
djmitche, line 46: This was correct. This is referring to a hypothetical `if` condition _inside_ the block
```
   'a'..'z' => { if c.is_lowercase() .. }
```

tidefield: Oh thanks for pointing out. Now that I read it again, it makes sense. I'll revert it.

gribozavr: @djmitche Thank you for catching this!

@daivinhtran Consider not simply reverting, but also adding a short code snippet that shows what this paragraph is talking about. You and I both misinterpreted the writing.

tidefield: PTAL https://github.com/google/comprehensive-rust/pull/2833

PR #2833, opened 2025-07-29
Clarify Matching Values section

daivinhtran:
Thanks @djmitche for https://github.com/google/comprehensive-rust/pull/2832#discussion_r2237935537. I reverted my previous change and added a code snippet to emphasize the original writing.

### DIFF
--- a/src/pattern-matching/match.md
+++ b/src/pattern-matching/match.md
@@ -41,8 +41,8 @@ Key Points:
 
 - Match guards as a separate syntax feature are important and necessary when we
   wish to concisely express more complex ideas than patterns alone would allow.
-- They are not the same as separate `if` expression inside of the match arm. An
-  `if` expression inside of the branch block (after `=>`) happens after the
+- Match guards are different from `if` expressions after the `=>`. An
+  `if` expression is evaluated after the
   match arm is selected. Failing the `if` condition inside of that block won't
   result in other arms of the original `match` expression being considered. In
   the following example, the wild card pattern is not matched after the `if`
--- a/src/pattern-matching/match.md
+++ b/src/pattern-matching/match.md
@@ -41,11 +41,11 @@ Key Points:
 
 - Match guards as a separate syntax feature are important and necessary when we
   wish to concisely express more complex ideas than patterns alone would allow.
-- Match guards are different from `if` expressions after the `=>`. An
-  `if` expression is evaluated after the
-  match arm is selected. Failing the `if` condition inside of that block won't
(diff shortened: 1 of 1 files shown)


src/pattern-matching/match.md
```diff
@@ -43,8 +43,29 @@ Key Points:
   wish to concisely express more complex ideas than patterns alone would allow.
 - They are not the same as separate `if` expression inside of the match arm. An
   `if` expression inside of the branch block (after `=>`) happens after the
```
gribozavr:
```suggestion
- Match guards are different from `if` expressions after the `=>`. An
  `if` expression is evaluated after the
```

src/pattern-matching/match.md
```diff
@@ -43,8 +43,29 @@ Key Points:
...
+        'a' | 's' | 'w' | 'd'     => println!("Moving around"),
+        '0'..='9'                 => println!("Number input"),
+        key if key.is_lowercase() => {
+            if input == 'y' {
+                println!("Lowercase: {key}")
+            }
+        },
+        _                         => println!("Something else"),
+    }
+}
```
gribozavr:
```suggestion
#[rustfmt::skip]
fn main() {
    let input = 'a';
    match input {
        key if key.is_uppercase() => println!("Uppercase"),
        key => if input == 'q' { println!("Quitting") },
        _   => println!("Bug: this is never printed"),
    }
}
```

src/pattern-matching/match.md
```diff
@@ -43,8 +43,29 @@ Key Points:
   wish to concisely express more complex ideas than patterns alone would allow.
 - They are not the same as separate `if` expression inside of the match arm. An
   `if` expression inside of the branch block (after `=>`) happens after the
-  match arm is selected. Failing the `if` condition inside of that block will
-  result in other arms of the original `match` expression being considered.
+  match arm is selected. Failing the `if` condition inside of that block won't
+  result in other arms of the original `match` expression being considered. In
+  the following example, the wild card pattern is not matched after the `if`
+  condition for `input == 'y'` fails.
```
gribozavr:
```suggestion
  result in other arms of the original `match` expression being considered. In
  the following example, the wildcard pattern `_ =>` is never even attempted.
```

tidefield:
LGTM @gribozavr

gribozavr (approved):

gribozavr:
@djmitche Please review at your convenience.

djmitche (approved):
That is a good illustration -- thanks!

PR #2846, opened 2025-08-12
Uninstall old mdbook-linkcheck while installing project tools.

### LINKED ISSUE #2773: link-check causing some build errors after 2024 update
After #2658, `cargo xtask build` (or`mdbook build`) is failing with the error mentioned here: https://github.com/Michael-F-Bryan/mdbook-linkcheck/issues/92. This can be fixed by removing the `[output.linkcheck]` section from book.toml (since I believe we're now using linkcheck2?).

However after that, the exclusions in the `[output.linkcheck2]` section don't seem to work, and building results in errors like `File not found: comprehensive-rust.pdf`.

egithinji:
Fixes #2773 .

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -141,7 +141,7 @@ fn uninstall_mdbook_linkcheck() {
         .arg("uninstall")
         .arg("mdbook-linkcheck")
         .output()
-        .expect("Failed to execute cargo uninstall mdbook-exerciser");
+        .expect("Failed to execute cargo uninstall mdbook-linkcheck");
 
     if !output.status.success() {
         if String::from_utf8_lossy(&output.stderr)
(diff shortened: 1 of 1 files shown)


xtask/src/main.rs
```diff
@@ -128,9 +129,35 @@ fn install_tools() -> Result<()> {
...
     Ok(())
 }
 
+fn uninstall_mdbook_linkcheck() {
+    println!("Uninstalling old mdbook-linkcheck if installed...");
+    let output = Command::new(env!("CARGO"))
+        .arg("uninstall")
+        .arg("mdbook-linkcheck")
+        .output()
+        .expect("Failed to execute cargo uninstall mdbook-exerciser");
```
qwandor: Typo, this should be mdbook-linkcheck.

qwandor (approved):

xtask/src/main.rs
egithinji: Ah good catch thanks.

Issue #2848, opened 2025-08-13
Solution for exercise 9.6.1: Use references to a Dynamically Sized Slice instead of references to a Fixed-Size Array.


iurii-kyrylenko:
The following functions get references to a **fixed-size** array:

https://github.com/google/comprehensive-rust/blob/301170f17eadec98bb1b5d1dc2e2a886c2ff3be8/src/references/exercise.rs#L17

https://github.com/google/comprehensive-rust/blob/301170f17eadec98bb1b5d1dc2e2a886c2ff3be8/src/references/exercise.rs#L26

 Don't think this is the best solution, especially after slices have just been discussed. The better option IMO would be to pass slices instead:
```rust
fn magnitude(vector: &[f64]) -> f64 {
    // ...
}

fn normalize(vector: &mut [f64]) {
    // ...
}
```

That way the functions accept any-dimensional vectors, e.g.

```rust
fn main() {
    println!(
        "Magnitude of a unit vector: {}",
        magnitude(&[0.0, 1.0, 0.0])
    );

    let mut v = [1.0, 2.0, 9.0];
    println!("Magnitude of {v:?}: {}", magnitude(&v));
    normalize(&mut v);
    println!("Magnitude of {v:?} after normalization: {}", magnitude(&v));

    let mut v4 = [1.0, 2.0, 3.0, 4.0];
    println!("4d vector: {v4:?}");
    normalize(&mut v4);
    println!("Normalized 4d vector: {v4:?}");
}

```

djmitche:
That's a good point -- I think this exercise used to be on the arrays segment, before slices were introduced. Do you want to make a PR with the fix?

iurii-kyrylenko:
@djmitche: Made a PR #2850

gribozavr:
Closing this issue because I believe it was fixed via [#2851](https://github.com/google/comprehensive-rust/pull/2851). Please reopen if not.

PR #2850, opened 2025-08-16
Pass slices to functions in exercise 9.6.1 (#2848)

iurii-kyrylenko:
Fixes issue #2848:
Use the references to a Dynamically Sized Slice instead of references to a Fixed-Size Array in function parameters.

### DIFF
--- a/src/references/exercise.rs
+++ b/src/references/exercise.rs
@@ -14,7 +14,7 @@
 
 // ANCHOR: solution
 /// Calculate the magnitude of the given vector.
-fn magnitude(vector: &[f64; 3]) -> f64 {
+fn magnitude(vector: &[f64]) -> f64 {
     let mut mag_squared = 0.0;
     for coord in vector {
         mag_squared += coord * coord;
@@ -23,7 +23,7 @@ fn magnitude(vector: &[f64; 3]) -> f64 {
 }
 
 /// Change the magnitude of the vector to 1.0 without changing its direction.
-fn normalize(vector: &mut [f64; 3]) {
+fn normalize(vector: &mut [f64]) {
     let mag = magnitude(vector);
     for item in vector {
         *item /= mag;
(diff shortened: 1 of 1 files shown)


djmitche (dismissed):
Thank you!

mgeisler:
Have people learned about slices at this point?

mgeisler:
> Have people learned about slices at this point?

Ah, yes, I see they have :smile: Slices are introduced in [9.3 Slices](https://google.github.io/comprehensive-rust/references/slices.html).

However, the exercise text says:

> utility functions for 3-dimensional geometry, representing a point as `[f64; 3]`.

So that is the reason why the functions use the `[f64; 3]` type — it's a simple concrete type which people should have an easy time understanding.

When I've been teaching the class (both inside and outside of Google), I'm sometimes met with questions about the math-related exercises. That is, while `magnitude` works for higher dimensions, the math is not the point of this exercise. The use of shared and unique references is the point.

I would suggest keeping the solution as-is, but add a speaker note showing that one could use slices instead. The speaker note should then discuss the tradeoffs: with slices, you have a _runtime_ length check, but with borrowed arrays, you have compile-time knowledge of the length (and the compiler can unroll the loops accordingly).

Basically, I don't think it's 100% clear that always using slices is the right call. A vector-library would want to use const generics instead, I believe.

djmitche:
That's a good point, @mgeisler -- thank you! I had missed that the instructions specifically mention arrays.

This solution already has a speaker note, so adding another regarding slices seems like a good idea (and this is the segment to do it in, since we have discussed both references and slices).

iurii-kyrylenko:
> I would suggest keeping the solution as-is, but add a speaker note showing that one could use slices instead. The speaker note should then discuss the tradeoffs.

@djmitche, Can we ask @mgeisler to add a speaker note?

We could mention that declaring slice `&[T]` as function params has a flexibility, based on [deref coercions](https://doc.rust-lang.org/book/ch15-02-deref.html#implicit-deref-coercions-with-functions-and-methods):
- if we have a slice `&[T]`, we can pass that directly;
- if we have an array `[T; N]`, we can pass a slice &[T] or a reference to the array &[T;N].
 
The same holds for [string slices as parameters](https://doc.rust-lang.org/book/ch04-03-slices.html#string-slices-as-parameters).

djmitche:
Fair point - I made #2851 for that purpose.

The Deref trait is covered (indirectly) later in the course, so I think it's best to omit deref coercion here.

iurii-kyrylenko:
Thank you, @djmitche !

mgeisler:
> @djmitche, Can we ask @mgeisler to add a speaker note?

Yes, certainly! I've been meaning to get back into the habit of contributing PRs! 🙂

Thanks for the PR, both of you!

PR #2851, opened 2025-08-17
Add speaker note regarding passing slices instead

djmitche:
This section covers both references and slices, but the exercise focuses on references. This speaker note briefly discusses the option to use slices instead, and a cost of doing so.


djmitche:
As suggested in #2850. Fixes #2848.

mgeisler (approved):
Awesome, thanks!

gribozavr (approved):

PR #2858, opened 2025-08-30
Fix link in README.md

mgeisler:
I don't know how, but the existing link points to an article that doesn't actually discuss the course. I updated it to the link I originally had in mind now.


djmitche (approved):
That makes more sense!

PR #2870, opened 2025-09-04
Update mdbook to version 0.4.52

mgeisler:
The most significant change for users is that the search index is downloaded on demand.

For us as a project, it’s good to see the server used by `mdbook serve` change from `warp` to `asum`: the former hadn’t been updated in a while, leading us to have old dependencies in our `Cargo.lock`.

I also ran `cargo update` to further unify the versions of our dependencies. This shrunk `cargo tree --duplicates` a little more.


qwandor (approved):

michael-kerscher:
The web tests have been failing. I fixed this by disabling /dev/shm usage in CI environments as this is too small there and also disabling the sandbox. This makes the test run successfully again without crashing.

mgeisler:
> The web tests have been failing. I fixed this by disabling /dev/shm usage in CI environments as this is too small there and also disabling the sandbox. This makes the test run successfully again without crashing.

Oh wow, thanks @michael-kerscher! I would never have found that myself :smile:

PR #2872, opened 2025-09-06
docs: improve language in std-traits section

mgeisler:


src/std-traits/operators.md
```diff
@@ -46,8 +46,8 @@ Discussion points:
 
 The `Not` trait (`!` operator) is notable because it does not "boolify" like the
```
gribozavr:
```suggestion
The `Not` trait (`!` operator) is notable because it does not convert the argument to bool like the
```

src/std-traits/operators.md
```diff
@@ -46,8 +46,8 @@ Discussion points:
...
 same operator in C-family languages; instead, for integer types it negates each
```
gribozavr:
```suggestion
same operator in C-family languages; instead, for integer types it flips each
```

src/std-traits/operators.md
```diff
@@ -46,8 +46,8 @@ Discussion points:
...
-bit of the number, which arithmetically is equivalent to subtracting it from -1:
-`!5 == -6`.
+bit of the number, which, arithmetically, is equivalent to subtracting it from
```
gribozavr:
```suggestion
bit of the number, which, arithmetically, is equivalent to subtracting the argument from
```

gribozavr (approved):

src/std-traits/operators.md
```diff
@@ -44,10 +44,10 @@ Discussion points:
 - You could implement `Add` for two different types, e.g.
   `impl Add<(i32, i32)> for Point` would add a tuple to a `Point`.
 
-The `Not` trait (`!` operator) is notable because it does not "boolify" like the
-same operator in C-family languages; instead, for integer types it negates each
-bit of the number, which arithmetically is equivalent to subtracting it from -1:
-`!5 == -6`.
+The `Not` trait (`!` operator) is notable because it does not convert the argument to bool like the
```
mgeisler:
```suggestion
The `Not` trait (`!` operator) is notable because it does not convert the argument to `bool` like the
```

src/std-traits.md
```diff
@@ -4,7 +4,7 @@
 
 <details>
 
-As with the standard-library types, spend time reviewing the documentation for
+As with the standard library types, spend time reviewing the documentation for
```
qwandor, line 7: I think this change makes it less clear. "standard-library types" makes it clear that these are types from the standard library, whereas without the hyphen it sounds they are "library types" which are standard somehow.

mgeisler: Okay, this is hard for me to decide on. Could @fw-immunant, @randomPoison, @djmitche chime in?

If this is a style choice, then documenting it in `GEMINI.md` would be super helpful going forward.

randomPoison: "standard library types" is how I would write it, personally. I don't think I've ever seen "standard-library" in the wild. I also think it doesn't much matter, the meaning is equally comprehensible either way.

djmitche: I agree with @randomPoison

mgeisler: Thanks all for chiming in. I don't think I can be the final arbiter on this, so I'll let you send PRs to improve the guidelines/text as you like.

PR #2876, opened 2025-09-06
docs: improve language in std-types section

mgeisler:

### DIFF
--- a/src/std-types/exercise.md
+++ b/src/std-types/exercise.md
@@ -10,7 +10,7 @@ It uses a
 to keep track of what values have been seen and how many times each one has
 appeared.
 
-The initial version of `Counter` is hard-coded to only work for `u32` values.
+The initial version of `Counter` is hardcoded to only work for `u32` values.
 Make the struct and its methods generic over the type of value being tracked,
 that way `Counter` can track any type of value.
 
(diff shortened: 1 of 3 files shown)


src/std-types/exercise.md
```diff
@@ -7,10 +7,10 @@ minutes: 20
 In this exercise you will take a very simple data structure and make it generic.
 It uses a
 [`std::collections::HashMap`](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html)
-to keep track of which values have been seen and how many times each one has
+to keep track of what values have been seen and how many times each one has
 appeared.
 
-The initial version of `Counter` is hard coded to only work for `u32` values.
+The initial version of `Counter` is hard-coded to only work for `u32` values.
```
gribozavr:
```suggestion
The initial version of `Counter` is hardcoded to only work for `u32` values.
```

gribozavr (approved):

fw-immunant:
The "which" -> "what" change in the exercise is not *incorrect* but certainly an oddly informal tone and less clear than the original phrasing in terms of removing emphasis from the fact that values are categorized by equality as members of the set of values in their type.

The first "which"->"that" change in `std.md` is incorrect but not in a way that's particularly confusing.

PR #2878, opened 2025-09-06
docs: improve language in references section

mgeisler:

### DIFF
--- a/src/references/exclusive.md
+++ b/src/references/exclusive.md
@@ -28,8 +28,8 @@ Key points:
   making an `&point.0` or changing `point.0` while `x_coord` is alive.
 
 - Be sure to note the difference between `let mut x_coord: &i32` and
-  `let x_coord: &mut i32`. The first one is a shared reference that can
-  be bound to different values, while the second is an exclusive
-  reference to a mutable value.
+  `let x_coord: &mut i32`. The first one is a shared reference that can be bound
+  to different values, while the second is an exclusive reference to a mutable
+  value.
 
 </details>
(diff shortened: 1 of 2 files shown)


src/references/exclusive.md
```diff
@@ -28,7 +28,7 @@ Key points:
   making an `&point.0` or changing `point.0` while `x_coord` is alive.
 
 - Be sure to note the difference between `let mut x_coord: &i32` and
-  `let x_coord: &mut i32`. The first one represents a shared reference which can
+  `let x_coord: &mut i32`. The first one represents a shared reference that can
   be bound to different values, while the second represents an exclusive
   reference to a mutable value.
```
gribozavr:
```suggestion
  `let x_coord: &mut i32`. The first one is a shared reference that can
  be bound to different values, while the second is an exclusive
  reference to a mutable value.
```

src/references/slices.md
```diff
@@ -47,7 +47,7 @@ fn main() {
     buffer.
   - You can't grow a slice to point to a larger section of the backing buffer.
     The slice loses information about the underlying buffer and so you can't
```
gribozavr: Either "creating a slice drops information" or "a slice does not have information".

"the slice loses" literally means that the slice had that information but lost it.

gribozavr (approved):

src/references/slices.md
mgeisler: Oh, nice catch! Let me reformulate it a bit:

```suggestion
    A slice does not have information about the length of the underlying buffer and so you can't
```

gribozavr (approved):

PR #2879, opened 2025-09-06
docs: improve language in pattern-matching section

mgeisler:

### DIFF
--- a/src/pattern-matching/infallible.md
+++ b/src/pattern-matching/infallible.md
@@ -37,8 +37,8 @@ fn main() {
 - Patterns are type-specific, including irrefutable patterns. Try adding or
   removing an element to the tuple and look at the resulting compiler errors.
 
-- Variable names are patterns that always match and bind the matched
-  value into a new variable with that name.
+- Variable names are patterns that always match and bind the matched value into
+  a new variable with that name.
 
 - `_` is a pattern that always matches any value, discarding the matched value.
 
--- a/src/pattern-matching/infallible.md
+++ b/src/pattern-matching/infallible.md
@@ -37,8 +37,8 @@ fn main() {
 - Patterns are type-specific, including irrefutable patterns. Try adding or
   removing an element to the tuple and look at the resulting compiler errors.
 
-- Variable names are patterns that always match and which bind the matched value
-  into a new variable with that name.
+- Variable names are patterns that always match, and which bind the matched
+  value into a new variable with that name.
 
 - `_` is a pattern that always matches any value, discarding the matched value.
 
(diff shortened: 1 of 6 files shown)


src/pattern-matching/infallible.md
```diff
@@ -37,8 +37,8 @@ fn main() {
 - Patterns are type-specific, including irrefutable patterns. Try adding or
   removing an element to the tuple and look at the resulting compiler errors.
 
-- Variable names are patterns that always match and which bind the matched value
-  into a new variable with that name.
+- Variable names are patterns that always match, and which bind the matched
```
gribozavr:
```suggestion
- Variable names are patterns that always match and bind the matched
```

src/pattern-matching/let-control-flow.md
```diff
@@ -4,7 +4,7 @@ minutes: 10
 
 # Let Control Flow
 
-Rust has a few control flow constructs which differ from other languages. They
+Rust has a few control flow constructs, which differ from other languages. They
```
gribozavr:
```suggestion
Rust has a few control flow constructs that differ from other languages. They
```

"which" should be used to provide non-essential information. This paragraph, on the other hand, is trying to convey the idea that that Rust has some unique control flow constructs.

src/pattern-matching/let-control-flow/while-let.md
```diff
@@ -2,7 +2,7 @@
 
 Like with `if let`, there is a
 [`while let`](https://doc.rust-lang.org/reference/expressions/loop-expr.html#predicate-pattern-loops)
-variant which repeatedly tests a value against a pattern:
+variant, which repeatedly tests a value against a pattern:
```
gribozavr:
```suggestion
variant that repeatedly tests a value against a pattern:
```

gribozavr (approved):

mgeisler:
I'll run `dprint fmt` here and on the other edited PRs to fix the formatting of the Markdown files.

gribozavr (approved):

src/pattern-matching/infallible.md
```diff
@@ -37,8 +37,8 @@ fn main() {
 - Patterns are type-specific, including irrefutable patterns. Try adding or
   removing an element to the tuple and look at the resulting compiler errors.
 
-- Variable names are patterns that always match and which bind the matched value
-  into a new variable with that name.
+- Variable names are patterns that always match and bind the matched value into
+  a new variable with that name.
```
qwandor, line 41: I think the previous text was clearer.

mgeisler: To me, the "which" seems redundant, but I guess this is a matter of taste? We need some more native speakers here (@djmitche, @randomPoison, @fw-immunant, etc...).

randomPoison: Both are equally fine imo. The presence of "which" doesn't change the meaning of the sentence, and I think it's down to personal taste whether you like the "which" or not.

PR #2881, opened 2025-09-06
docs: improve language in methods-and-traits section

mgeisler:

### DIFF
--- a/src/methods-and-traits/deriving.md
+++ b/src/methods-and-traits/deriving.md
@@ -30,7 +30,7 @@ fn main() {
   macros to add useful functionality. For example, `serde` can derive
   serialization support for a struct using `#[derive(Serialize)]`.
 
-  Derivation is usually provided for traits that have a common boilerplate
+- Derivation is usually provided for traits that have a common boilerplate
   implementation that is correct for most cases. For example, demonstrate how a
   manual `Clone` impl can be repetitive compared to deriving the trait:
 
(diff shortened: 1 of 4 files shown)


src/methods-and-traits/deriving.md
```diff
@@ -30,7 +30,7 @@ fn main() {
   macros to add useful functionality. For example, `serde` can derive
   serialization support for a struct using `#[derive(Serialize)]`.
 
-- Derivation is usually provided for traits that have a common boilerplate-y
+  Derivation is usually provided for traits that have a common boilerplate
```
gribozavr:
```suggestion
- Derivation is usually provided for traits that have a common boilerplate
```

gribozavr (approved):

fw-immunant:
The changes from "which" -> "that" in `methods.md` and `associated-types.md` are incorrect. See <https://www.grammarly.com/blog/commonly-confused-words/using-that-and-which-is-all-about-restrictive-and-non-restrictive-clauses/> for an explanation.

In `methods.md`, the changed text now implies that only constructors defined as static functions are called `new` by convention (but implies that perhaps other constructors might exist, defined other ways, for which this convention does not appyl). Instead, we want to say that static methods are typically used to define static functions, which by convention are often called "new". That's exactly what the old text conveyed.

In `associated-types.md`, again the change has confused the sense of the text by implying that being supplied by the trait implementation is a restriction that only applies to some of these types. Instead, the property of being supplied by the trait implementation is merely more (not grammatically required) information about associated types.

qwandor:
Agreed, this PR makes the text less clear.

PR #2886, opened 2025-09-06
docs: improve language in error-handling section

mgeisler:

### DIFF
--- a/src/error-handling/panics.md
+++ b/src/error-handling/panics.md
@@ -4,9 +4,7 @@ minutes: 3
 
 # Panics
 
-Rust handles fatal errors with a "panic".
-
-Rust will trigger a panic if a fatal error happens at runtime:
+Rust handles fatal errors with a "panic", which happens at runtime:
 
 ```rust,editable,should_panic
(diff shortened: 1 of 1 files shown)


src/error-handling/panics.md
```diff
@@ -4,9 +4,7 @@ minutes: 3
 
 # Panics
 
-Rust handles fatal errors with a "panic".
-
-Rust will trigger a panic if a fatal error happens at runtime:
+Rust handles fatal errors with a "panic", which happens at runtime:
```
gribozavr: I feel like it just smooshed together two sentences. All the information is there, but it is clumsy.

Consider: In case of a fatal runtime error, Rust triggers a "panic".

gribozavr (approved):

src/error-handling/panics.md
mgeisler: That was actually me who edited it a bit further :smile: I like your suggestion, it's very to the point:

```suggestion
In case of a fatal runtime error, Rust triggers a "panic":
```

PR #2887, opened 2025-09-06
docs: improve language in control-flow-basics section

mgeisler:

### DIFF
--- a/src/control-flow-basics/blocks-and-scopes.md
+++ b/src/control-flow-basics/blocks-and-scopes.md
@@ -4,9 +4,8 @@ minutes: 5
 
 # Blocks and Scopes
 
-A block in Rust contains a sequence of expressions, enclosed by braces `{}`.
-Each block has a value and a type, which are those of the last expression of the
-block:
+A block in Rust contains a sequence of expressions, enclosed by braces {}. The
+final expression of a block determines its value and type:
 
 ```rust,editable
 fn main() {
--- a/src/control-flow-basics/blocks-and-scopes.md
+++ b/src/control-flow-basics/blocks-and-scopes.md
@@ -5,7 +5,7 @@ minutes: 5
 # Blocks and Scopes
 
 A block in Rust contains a sequence of expressions, enclosed by braces {}. The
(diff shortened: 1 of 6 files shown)


src/control-flow-basics/blocks-and-scopes.md
```diff
@@ -4,9 +4,8 @@ minutes: 5
 
 # Blocks and Scopes
 
-A block in Rust contains a sequence of expressions, enclosed by braces `{}`.
-Each block has a value and a type, which are those of the last expression of the
-block:
+A block in Rust contains a sequence of expressions, enclosed by braces {}. The
+final expression of a block determines its value and type:
```
gribozavr:
```suggestion
final expression of a block determines the value and type of the whole block:
```

gribozavr (approved):

fw-immunant:
I think the change to `blocks-and-scopes.md` regresses clarity. Using code formatting for the literal braces seems intentional, and it is novel that blocks have values and types, so noting that they have these properties is worth doing upfront and explicitly rather than in passing.

src/control-flow-basics/blocks-and-scopes.md
```diff
@@ -4,9 +4,8 @@ minutes: 5
 
 # Blocks and Scopes
 
-A block in Rust contains a sequence of expressions, enclosed by braces `{}`.
-Each block has a value and a type, which are those of the last expression of the
-block:
+A block in Rust contains a sequence of expressions, enclosed by braces {}. The
```
qwandor, line 7: Don't remove the backticks.

mgeisler: Yeah, sorry about that!

mgeisler: Will be fixed in #2901.

mgeisler:
> Using code formatting for the literal braces seems intentional,

Yes, I agree!

> and it is novel that blocks have values and types, so noting that they have these properties is worth doing upfront and explicitly rather than in passing.

I'll try reformulating this in #2901.

PR #2888, opened 2025-09-06
docs: improve language in closures section

mgeisler:

### DIFF
--- a/src/closures/exercise.md
+++ b/src/closures/exercise.md
@@ -4,8 +4,8 @@ minutes: 10
 
 # Exercise: Log Filter
 
-Building on the generic logger from this morning, implement a `Filter` which
-uses a closure to filter log messages, sending those which pass the filtering
+Building on the generic logger from this morning, implement a `Filter` that uses
+a closure to filter log messages, sending those that pass the filtering
 predicate to an inner logger.
 
 ```rust,compile_fail,editable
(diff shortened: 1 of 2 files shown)


gribozavr (approved):

src/closures/exercise.md
```diff
@@ -4,8 +4,8 @@ minutes: 10
 
 # Exercise: Log Filter
 
-Building on the generic logger from this morning, implement a `Filter` which
-uses a closure to filter log messages, sending those which pass the filtering
+Building on the generic logger from this morning, implement a `Filter` that uses
+a closure to filter log messages, sending those that pass the filtering
```
qwandor, line 8: I think that the second 'which' makes more sense.

mgeisler: I discussed this a bit with Gemini and learned that "which" should only be used for non-restrictive clauses, and "that" is for restrictive clauses. I'm obviously not an expert here... we could add further guidance for this in the `GEMINI.md` file.

Cc @djmitche, @randomPoison, @fw-immunant and other native speakers so you can discuss this a bit if you have input on this.

djmitche: Either is technically fine in this case (Gemini is leading you down the garden path: both can be used for restrictive cases, while only which is only suitable for non-restrictive cases).  So it's a matter of taste and common usage, and I agree with @qwandor there.

Also, this is a matter of activation energy in a PR -- neither of these uses bothered much of anyone, but since it came up in a PR we've all given it more thought than it was really due. I'd suggest we just let this merged PR be.

PR #2893, opened 2025-09-06
docs: improve language in borrowing section

mgeisler:


src/borrowing/borrowck.md
```diff
@@ -67,7 +67,7 @@ fn main() {
 
 - Technically multiple mutable references to a piece of data can exist at the
```
gribozavr:
```suggestion
- Technically, multiple mutable references to a piece of data can exist at the
```

PR #2896, opened 2025-09-07
ci: add rustfmt check to formatting workflow

### LINKED ISSUE #2794: Apply rustfmt to exercises
The exercises do not conform to the current rustfmt standard, e.g. [error-handling exercise](https://google.github.io/comprehensive-rust/error-handling/exercise.html). Maybe you can discuss if adding a ci job that checks for rustfmt changes makes sense here?

lucienfer:
## What does this PR do?

This PR extends the existing `format` job in `.github/workflows/build.yml` to
**check Rust formatting with nightly `rustfmt`**:

```yaml
- name: Check Rust formatting (rustfmt --check)
  run: cargo fmt --all -- --check
  ```

## Why?

•	The repository already defines a rustfmt.toml with nightly-only settings (imports_granularity, use_small_heuristics = "Max", etc.).
•	The CI installs nightly rustfmt but wasn’t actually using it.
•	Currently, dprint/check verifies formatting for Markdown/PO/JS/etc., but not Rust code.
This PR makes sure PRs fail if Rust code is not formatted according to rustfmt.toml.

## How was it tested?

•	Ran locally with:
```bash
rustup default nightly
rustup component add rustfmt
cargo fmt --all -- --check
dprint fmt --check
```
Both checks pass on a clean branch.

•	CI workflow updated accordingly.
If Rust code is unformatted, the format job will now fail.

**Closes #2794**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,6 @@ on:
   push:
     branches:
       - main
-      - fix-2794
   workflow_dispatch:
 
 env:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - main
+      - fix-2794
   workflow_dispatch:
 
 env:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,6 +27,9 @@ jobs:
           rustup default nightly
           rustup component add rustfmt
 
+      - name: Check Rust formatting (rustfmt --check)
+        run: cargo fmt --all -- --check
+
       - name: Check formatting
         uses: dprint/check@v2.3
 
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -21,9 +21,9 @@
 
 use anyhow::{Ok, Result, anyhow};
 use clap::{Parser, Subcommand};
+use std::env;
 use std::path::{Path, PathBuf};
-use std::process::Stdio;
-use std::{env, process::Command};
+use std::process::{Command, Stdio};
 
 fn main() -> Result<()> {
     if let Err(e) = execute_task() {
@@ -49,7 +49,8 @@ enum Task {
     InstallTools,
     /// Runs the web driver tests in the tests directory.
     WebTests {
-        /// Optional 'book html' directory - if set, will also refresh the list of slides used by slide size test.
+        /// Optional 'book html' directory - if set, will also refresh the list
+        /// of slides used by slide size test.
         #[arg(short, long)]
         dir: Option<PathBuf>,
     },
@@ -61,7 +62,8 @@ enum Task {
         #[arg(short, long)]
         language: Option<String>,
 
-        /// Directory to place the build. If not provided, defaults to the book/ directory (or the book/xx directory if a language is provided).
+        /// Directory to place the build. If not provided, defaults to the book/
+        /// directory (or the book/xx directory if a language is provided).
         #[arg(short, long)]
         output: Option<PathBuf>,
     },
@@ -71,7 +73,8 @@ enum Task {
         #[arg(short, long)]
         language: Option<String>,
 
-        /// Directory to place the build. If not provided, defaults to the book/ directory (or the book/xx directory if a language is provided).
+        /// Directory to place the build. If not provided, defaults to the book/
+        /// directory (or the book/xx directory if a language is provided).
         #[arg(short, long)]
         output: Option<PathBuf>,
     },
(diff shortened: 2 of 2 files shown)


mgeisler:
Hey @lucienfer, thanks for the PR! I'm surprised this is necessary: `dprint` is supposed to run `rustfmt` as well, but I guess it's somehow not doing this (since you include formatting changes in this PR)?

lucienfer:
You’re right, after digging a bit I realized that dprint wasn’t actually checking the Rust formatting properly because it was configured to run rustfmt on the stable toolchain. Since the repo’s `rustfmt.toml` uses unstable options like `imports_granularity = "module"`, those settings were simply being ignored.

To fix this I made two changes:
	1.	I added the `unstable_features = true` flag in `rustfmt.toml` to make sure nightly-specific settings are explicitly enabled.
	2.	Instead of relying on `cargo fmt` in CI, I updated the `dprint.json` config so that it directly runs `rustup run nightly rustfmt`. This way dprint is aligned with the nightly toolchain already used in CI, and it will fail if the code isn’t correctly formatted.

This should ensure that formatting is consistently checked both locally (via dprint) and in CI.

mgeisler:
Awesome that you investigated, super helpful! I remember we had #2168 a while ago, which made us go from nightly to stable for this. I think we should revisit my comment from then and simply pin the nightly version for this. Something like `nightly-2025-09-01` or simila.

When I did this recently in my new company, I ran into `dprint` being slow or basically erroring out: I had not primed the cache on CI, so I think `dprint` basically tried to auto-install the nightly version on the fly for every file at the same time.

I filled https://github.com/dprint/dprint/issues/1023 about this in `dprint` just now, perhaps there is an elegant solution. In the meantime, I suggest we pin the version and add a `rustup install --profile minimal nightly-2025-09-01` step before calling `dprint` in CI.

For people at home calling `dprint fmt`, we should also be helpful: we could install the right nightly toolchain with `cargo xtask install-tools` or similar? Cc @egithinji and @michael-kerscher who have both looked at our build infrastructure.

lucienfer:
I updated the PR to pin rustfmt to `nightly-2025-09-01`, pre-install it in CI, make dprint call that exact toolchain, and wire the same install into `cargo xtask install-tools`. This should keep dprint fast on CI and ensure identical formatting locally and in CI.

xtask/src/main.rs
```diff
@@ -91,6 +94,30 @@ fn execute_task() -> Result<()> {
 
 fn install_tools() -> Result<()> {
     println!("Installing project tools...");
+
+    const PINNED_NIGHTLY: &str = "nightly-2025-09-01";
+
+    let rustup_steps = [
+        // Installe la toolchain sans les composants inutiles
```
mgeisler: Heh, let's write this in English :smile:

.github/workflows/build.yml
```diff
@@ -22,10 +22,10 @@ jobs:
           sudo apt update
           sudo apt install gettext yapf3
 
-      - name: Install nightly rustfmt
+      - name: Install pinned nightly for rustfmt
         run: |
-          rustup default nightly
-          rustup component add rustfmt
+          rustup toolchain install --profile minimal nightly-2025-09-01
+          rustup component add rustfmt --toolchain nightly-2025-09-01
```
mgeisler: Thank you for this! One suggestion: could you use a variable for this so it's easier to update in a few months?

Cc @egithinji, could we use `cargo xtask install-tools` here? Not now, but in a later PR since this just refactors the existing install commands.

CONTRIBUTING.md
```diff
@@ -53,6 +53,16 @@ this:
...
 Run `dprint fmt` to automatically format all files.
 
+**Note:** To make sure you have the correct version of `rustfmt` installed,
+please run:
+
+```bash
+cargo xtask install-tools
+```
+
+This will install the pinned nightly toolchain (`nightly-2025-09-01`) and add
```
mgeisler: Let's avoid mentioning the precise version here since we'll surely forget to update this later :smile:

xtask/src/main.rs
```diff
@@ -91,6 +94,30 @@ fn execute_task() -> Result<()> {
...
+        ["toolchain", "install", "--profile", "minimal", PINNED_NIGHTLY],
+        // Ajoute rustfmt pour cette toolchain
+        ["component", "add", "rustfmt", "--toolchain", PINNED_NIGHTLY],
+    ];
+
+    for args in rustup_steps {
+        let status = std::process::Command::new("rustup")
+            .args(args)
+            .status()
+            .expect("Failed to execute rustup");
```
mgeisler: Please don't use `expect` (or `unwrap`), instead return the `Result` back with `?`. @egithinji, can you review this part? We should (eventually) have some useful patterns in the `xtask` code to easily and quickly execute shell commands.

xtask/src/main.rs
lucienfer: Sorry, I didn't see that I left it.

CONTRIBUTING.md
lucienfer: I had even forgotten that it needed to be updated 😂

xtask/src/main.rs
lucienfer: I noticed that expect was used further down, so I thought it would be fine to use it here as well. Would you like me to remove it too?

mgeisler: Ah, I see! Let's update the existing code in a different PR to keep this focused on your changes.

PR #2897, opened 2025-09-07
fix(docs): remove stray/trailing s>

daveworth:
It looks like changes from the last commit (`085b534`) to the `<details>...</details>` block had a stray `s>` included.


djmitche (approved):
Good thing we humans still have a role in cleaning up behind the robots!

mgeisler:
Yes, thanks for finding and fixing this! :see_no_evil:

PR #2901, opened 2025-09-10
Fix blocks-and-scopes.md after code review

mgeisler:
Thanks to @fw-immunant and @qwandor for the comments in #2887.

### DIFF
--- a/src/control-flow-basics/blocks-and-scopes.md
+++ b/src/control-flow-basics/blocks-and-scopes.md
@@ -4,8 +4,9 @@ minutes: 5
 
 # Blocks and Scopes
 
-A block in Rust contains a sequence of expressions, enclosed by braces {}. The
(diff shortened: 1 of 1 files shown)


src/control-flow-basics/blocks-and-scopes.md
```diff
@@ -4,8 +4,9 @@ minutes: 5
 
 # Blocks and Scopes
 
-A block in Rust contains a sequence of expressions, enclosed by braces {}. The
-final expression of a block determines the value and type of the whole block:
+- A block in Rust contains a sequence of expressions, enclosed by braces `{}`.
+- The final expression of a block determines the value and type of the whole
+  block:
```
mgeisler: Ups, this should not end with `:` now that it's a separate list item.

```suggestion
  block.
```

djmitche (approved):

Issue #2803, opened 2025-07-03
Running cargo xtask triggers upstream deprecation warning


timClicks:
When someone has their own installation of `cargo-xtask` installed, building the project with the vendored version triggers a deprecation warning.

```console
$ cargo xtask serve
warning: user-defined alias `xtask` is shadowing an external subcommand found at: `/path/to/.cargo/bin/cargo-xtask`
This was previously accepted but is being phased out; it will become a hard error in a future release.
For more information, see issue #10049 <https://github.com/rust-lang/cargo/issues/10049>.
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/xtask serve`
```

michael-kerscher:
@timClicks (also @djmitche) what crate offers cargo-xtask? This is a common mechanism to define external commands that can be run as a cargo subcommand - see also https://github.com/google/comprehensive-rust/pull/2688.

I'm not convinced to remove the xtask alias in the future as this provides command discoverability by using `cargo xtask`. As I don't see what crate provides cargo-xtask I wonder where this is from

mgeisler:
> I'm not convinced to remove the xtask alias in the future as this provides command discoverability by using `cargo xtask`.

Funny, I had exactly the same instinct in my comment on the PR: https://github.com/google/comprehensive-rust/pull/2804#issuecomment-3057247204

In some sense, the error is a user error: you should not have a `cargo-xtask` binary installed on your system when you work on Comprehensive Rust. That is of course a bit of a big ask, and a global one to boot.

An alternative would be for Comprehensive Rust to pick a different name for its top-level tooling. Perhaps `cargo run-tool` or something else along those lines?

djmitche:
Interesting, https://github.com/matklad/cargo-xtask still suggests using `cargo xtask foo`.

egithinji:
Is it the case that the only way of running into this warning is by running `cargo install cargo-xtask` before attempting any of the `cargo xtask` commands? If so, I think that can be mitigated by having a brief explanation in the README of how we use xtask (similar to [rust-analyzer](https://github.com/rust-lang/rust-analyzer/blob/master/.cargo/config.toml) in their contributing guide) and that it need not be installed by the user.

I think having a discoverable `cargo xtask` command that shows the user all possible tasks they can run in the project is better from a usability perspective, especially for first time open source contributors.

michael-kerscher:
[...]

PR #2904, opened 2025-09-15
Add more prominent references to `cargo xtask`

mgeisler:
I found that Gemini didn’t know how to install the necessary tools
make `dprint fmt` work, so I’m adding more references to `cargo xtask`
in our documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,15 +65,15 @@ git clone https://github.com/google/comprehensive-rust/
 cd comprehensive-rust
 ```
 
-Then install these tools with:
+Then run the following command to install the correct versions of all tools
+mentioned above:
 
 ```shell
 cargo xtask install-tools
 ```
 
-> **Note** We use `xtask` for task automation within the project (e.g.
-> installing required tools). Xtask is not a package that you should install.
-> Visit https://github.com/matklad/cargo-xtask for more information.
+This uses `cargo install` to install the tools, so you will find them in your
+`~/.cargo/bin/` directory afterwards.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -65,15 +65,15 @@ git clone https://github.com/google/comprehensive-rust/
...
-Then install these tools with:
+Then run the following command to install the correct versions of all tools
+mentioned above:
 
 ```shell
 cargo xtask install-tools
 ```
 
-> **Note** We use `xtask` for task automation within the project (e.g.
-> installing required tools). Xtask is not a package that you should install.
```
egithinji, line 75: We added this note about not needing to install xtask because of #2803. But it was quite an edge case so might not really be necessary. Just wanted to highlight that so you can decide whether you think its worth retaining.

egithinji:
Nice to have you back @mgeisler!

README.md
mgeisler: Thanks for the reminder! When reading the README, I felt it was drawing too much attention to itself somehow. That is, people can read up on this themselves if they want (or need to), but I don't think we should link to it from here.

I would also like to remove the many links to the tools we use and thus make the README more focused (the tools are a bit of an implementation detail now, thanks to your work on `cargo xtask install-tools`).

mgeisler:
> Nice to have you back @mgeisler!

Thanks a lot! Ironically, I have more mental capacity to work on this now that I no longer work at Google :smile:

PR #2907, opened 2025-09-16
refactor cargo xtasks

michael-kerscher:
* improve error handling
* show executed command lines
* reduce code duplication (run_command, run_mdbook_command)

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -19,18 +19,14 @@
 //! `cargo xtask install-tools` and the logic defined here will install
 //! the tools.
 
-use anyhow::{Ok, Result, anyhow};
+use anyhow::{Context, Result, anyhow};
 use clap::{Parser, Subcommand};
 use std::env;
 use std::path::{Path, PathBuf};
-use std::process::{Command, Stdio};
+use std::process::Command;
 
 fn main() -> Result<()> {
-    if let Err(e) = execute_task() {
-        eprintln!("{e}");
-        std::process::exit(-1);
-    }
-    Ok(())
+    execute_task()
 }
 
 #[derive(Parser)]
@@ -83,11 +79,28 @@ enum Task {
 fn execute_task() -> Result<()> {
     let cli = Cli::parse();
     match cli.task {
-        Task::InstallTools => install_tools()?,
-        Task::WebTests { dir } => run_web_tests(dir)?,
-        Task::RustTests => run_rust_tests()?,
-        Task::Serve { language, output } => start_web_server(language, output)?,
-        Task::Build { language, output } => build(language, output)?,
+        Task::InstallTools => install_tools(),
+        Task::WebTests { dir } => run_web_tests(dir),
+        Task::RustTests => run_rust_tests(),
+        Task::Serve { language, output } => start_web_server(language, output),
+        Task::Build { language, output } => build(language, output),
+    }
+}
+
+/// Executes a command and returns an error if it fails.
+fn run_command(cmd: &mut Command) -> Result<()> {
+    let command_display = format!("{cmd:?}");
+    println!("> {command_display}");
+    let status = cmd
+        .status()
+        .with_context(|| format!("Failed to execute command: {command_display}"))?;
+    if !status.success() {
+        let exit_description = if let Some(code) = status.code() {
+            format!("exited with status code: {}", code)
+        } else {
+            "was terminated by a signal".to_string()
+        };
+        return Err(anyhow!("Command `{command_display}` {exit_description}"));
     }
     Ok(())
 }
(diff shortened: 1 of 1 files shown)


xtask/src/main.rs
```diff
@@ -19,18 +19,14 @@
...
-use std::process::{Command, Stdio};
+use std::process::Command;
 
 fn main() -> Result<()> {
-    if let Err(e) = execute_task() {
-        eprintln!("{e}");
-        std::process::exit(-1);
-    }
-    Ok(())
+    execute_task()
```
michael-kerscher, line 29: this uses the Termination trait which returns a FAILURE code and attempts to print to stderr

xtask/src/main.rs
```diff
@@ -97,214 +110,139 @@ fn install_tools() -> Result<()> {
...
+        ("mdbook-svgbob", "0.2.2"),
+        ("mdbook-pandoc", "0.10.4"),
+        ("mdbook-i18n-helpers", "0.3.6"),
+        ("i18n-report", "0.2.0"),
+        ("mdbook-linkcheck2", "0.9.1"),
     ];
 
-    for args in &install_args {
-        let status = Command::new(env!("CARGO"))
-            .arg("install")
```
michael-kerscher, line 139: I have split this into external tools and internal tools because I want to use `binstall` (with a flag) in the next commit I have in my pipeline. This prepares the stage for this.

xtask/src/main.rs
mgeisler: Yep, sounds good!

xtask/src/main.rs
```diff
@@ -97,214 +110,139 @@ fn install_tools() -> Result<()> {
...
+        cmd.args(["install", tool, "--version", version, "--locked"]);
+        run_command(&mut cmd)?;
+    }
+
+    // Install local tools from the workspace.
+    let workspace_dir = Path::new(env!("CARGO_WORKSPACE_DIR"));
+    let local_tools = ["mdbook-exerciser", "mdbook-course"];
+    for tool in local_tools {
+        let mut cmd = Command::new(cargo);
+        cmd.args(["install", "--path"])
```
mgeisler, line 146: I think the `mut cmd` is unnecessary here:

```suggestion
        let cmd = Command::new(cargo);
            .args(["install", "--path"])
```

xtask/src/main.rs
```diff
@@ -97,214 +110,139 @@ fn install_tools() -> Result<()> {
...
-            println!("mdbook-linkcheck not installed. Continuing...");
-        } else {
-            eprintln!(
-                "An error occurred during uninstallation of mdbook-linkcheck:\n{:#?}",
-                String::from_utf8_lossy(&output.stderr)
-            );
+        let stderr = String::from_utf8_lossy(&output.stderr);
+        // This specific error is OK, it just means the package wasn't installed.
+        if !stderr.contains("did not match any packages") {
+            return Err(anyhow!(
```
mgeisler, line 169: I think you could use [`bail!`](https://docs.rs/anyhow/latest/anyhow/macro.bail.html) here:

```suggestion
            bail!(
```

xtask/src/main.rs
```diff
@@ -97,214 +110,139 @@ fn install_tools() -> Result<()> {
...
+        run_command(&mut cmd)?;
     }
 
-    let path_to_tests_dir = Path::new(env!("CARGO_WORKSPACE_DIR")).join("tests");
-    let mut command = Command::new("npm");
-    command.current_dir(path_to_tests_dir.to_str().unwrap());
-    command.arg("test");
+    let tests_dir = workspace_dir.join("tests");
+    let mut cmd = Command::new("npm");
+    cmd.current_dir(tests_dir).arg("test");
```
mgeisler, line 197: Something like this should work, especially if `run_command` would take ownership of the `Command` passed in:

```suggestion
    let cmd = Command::new("npm");
        .current_dir(tests_dir).arg("test");
```

mgeisler: Ah, eh... I see now that you conditionally modify `cmd` below. So the `mut` is still needed.

xtask/src/main.rs
```diff
@@ -83,11 +79,28 @@ enum Task {
...
+        Task::InstallTools => install_tools(),
+        Task::WebTests { dir } => run_web_tests(dir),
+        Task::RustTests => run_rust_tests(),
+        Task::Serve { language, output } => start_web_server(language, output),
+        Task::Build { language, output } => build(language, output),
+    }
+}
+
+/// Executes a command and returns an error if it fails.
+fn run_command(cmd: &mut Command) -> Result<()> {
```
mgeisler, line 91: I have a hunch that `run_command` should take `Command` instead of a reference: I don't think you have a case where you want to run the command twice. But you can also keep it like this if you find it more flexible.

mgeisler (approved):
This is really nice, thank you!

mgeisler:
Ups, I had not see that it would auto-merge :smile: Luckily, my comments were only stylistic.

egithinji (commented):
Nice refactor!

Issue #2394, opened 2024-10-02
Speaker notes headers have bad contrast in Rust theme


InSyncWithFoo:
Basically black-on-brown:

![](https://github.com/user-attachments/assets/c9792b0f-cee3-4427-8196-e8f659187066)

randomPoison:
A thing to note here: This only happens when the speaker notes are inline. If you pop out the speaker notes the lighter background color from the main slide is used (instead of the darker brown color): 

<img width="1346" height="701" alt="Image" src="https://github.com/user-attachments/assets/46b6dd02-257d-4ddf-8420-d78b813bb465" />

PR #2914, opened 2025-09-18
Fix speaker notes header and link contrast in Rust theme

juliatuttle:
Despite the !important in the .content details style, links were showing up in the link color and headers were showing up in the (main) foreground color, which in the Rust theme has poor and no contrast, respectively.

Fix this by explicitly forcing both to the sidebar foreground color instead.

### DIFF
--- a/theme/css/speaker-notes.css
+++ b/theme/css/speaker-notes.css
@@ -5,6 +5,10 @@
   padding: 0.25em;
 }
 
+.content details a:link, .content details a:visited {
+  color: var(--sidebar-fg) !important;
+}
+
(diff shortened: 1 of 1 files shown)


randomPoison:
Worth noting that this is attempting to fix https://github.com/google/comprehensive-rust/issues/2394

randomPoison:
Hm, I'm trying this on my machine and I'm still seeing headers in the speaker notes having low contrast with the background:

image

Unfortunately I don't know much webdev stuff, so I'm not sure what about your fix isn't working here.

PR #2915, opened 2025-09-19
Speed up rust crate installs in workflow by using cargo-binstall

michael-kerscher:
When cargo xtask install-tools --binstall is used, precompiled binaries are fetched if available.
Triggers a regular (compilation) install if binary releases do not exist.

This is now configured to be used by the CI to speed up the build and deploy workflows where possible.

Currently binary packages exist for:
- mdbook
- mdbook-pandoc


michael-kerscher:
Discussion around bininstall is in #2767

qwandor (approved):

PR #2916, opened 2025-09-20
Optimize `apt update` calls

mgeisler:
This introduces a new local GitHub action for installing Debian packages. I noticed that it often takes several minutes to install a few small packages, and a large part of that time is spent updating the man page database:

    2025-09-20T11:39:25.3001853Z Setting up yapf3 (0.33.0-1) ...
    2025-09-20T11:39:25.3045329Z Processing triggers for libc-bin (2.39-0ubuntu8.5) ...
    2025-09-20T11:39:25.4420933Z Processing triggers for man-db (2.12.0-4build2) ...
    2025-09-20T11:41:42.8522570Z Processing triggers for install-info (7.1-3build2) ...

Excluding the documentation files should make the man-db run much faster. Compare these two runs:

* [Before](https://github.com/google/comprehensive-rust/actions/runs/17879390708/job/50845084995): 2m 41s.
* [After](https://github.com/google/comprehensive-rust/actions/runs/17880418155/job/50847334471): 11s.

### DIFF
--- a/.github/workflows/apt-get-install/action.yml
+++ b/.github/workflows/apt-get-install/action.yml
@@ -1,0 +1,44 @@
+name: Setup Apt and Install Packages
+description: Configures apt, runs update once per job, and installs packages.
+
+inputs:
+  packages:
+    description: A space-separated list of packages to install.
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Configure dpkg and apt for CI
+      shell: bash
+      run: |
+        # Avoid time-consuming man-db updates.
+        sudo tee /etc/dpkg/dpkg.cfg.d/99-no-doc > /dev/null <<EOF
+        path-exclude /usr/share/doc/*
+        path-exclude /usr/share/man/*
+        path-exclude /usr/share/info/*
+        EOF
+
+        # Exclude translations.
+        sudo tee /etc/apt/apt.conf.d/99-no-translations > /dev/null <<EOF
+        Acquire::Languages "none";
+        EOF
+
+        # Exclude command-not-found utility.
+        sudo rm -f /etc/apt/apt.conf.d/50command-not-found
+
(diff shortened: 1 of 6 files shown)


.github/workflows/install-mdbook/action.yml
```diff
@@ -12,10 +12,13 @@ runs:
       run: cargo xtask install-tools --binstall
       shell: bash
 
-    - name: Install dependencies for mdbook-pandoc
+    - name: Install mdbook-pandoc dependencies
+      uses: ./.github/workflows/apt-get-install
+      with:
+        packages: texlive texlive-luatex texlive-lang-cjk texlive-lang-arabic librsvg2-bin fonts-noto
+
+    - name: Install mdbook-pandoc
```
michael-kerscher, line 20: This is not installing mdbook-pandoc itself but the dependencies AFAIU, the step name could be missleading

.github/workflows/apt-get-install/action.yml
```diff
@@ -0,0 +1,44 @@
+name: Setup Apt and Install Packages
```
michael-kerscher, line 1: nice action, I like the idea of updating once, excluding documentation and being the go-to point for further tweaks on this topic!

michael-kerscher (approved):
Nice improvement. Thank you!

.github/workflows/install-mdbook/action.yml
mgeisler: The installation of the dependencies has been split off to the new `apt-get-install` action above. So I think this step only installs `mdbook-pandoc` now.

mgeisler:
As a further optimization, I was playing with adding `--no-install-recommends` to the `apt-get` invocation. However, that would need some fiddling since we do depend on some of the recommended packages (the build fails when I add it).

PR #2917, opened 2025-09-21
docs: Clarify speaker note style for instructors

gribozavr:
The previous STYLE.md advised against using speaker notes as a 'script' but didn't provide clear examples of what that meant. This could be interpreted by LLMs as disallowing any kind of instructor prompt.

This change clarifies the guideline by:

1.  Distinguishing between long-form, verbatim 'scripts' (bad) and short, actionable 'teaching prompts' (good).

2.  Providing concrete 'good' and 'bad' examples to make the distinction unambiguous for future contributors.

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -220,9 +220,23 @@ collapsed or removed entirely from the slide.
 
   - Where to pause and engage the class with questions.
 
-- Speaker notes are not a script for the instructor. When teaching the course,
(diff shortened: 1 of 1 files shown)


mgeisler (approved):
Nice, this is great! It was exactly the earlier style that made me add the warnings against making this a script. Turning it into tips about good things to do is super useful.

PR #2921, opened 2025-09-22
"Token Types" chapter of Idiomatic Rust

tall-vase:
Materials on "token types."

The "Branded tokens" section of this one requires a little extra scrutiny. It may be the case it is too complex for this stage, or that the explanation isn't suitable for the audience/instructors. Either way, I'm eager for input.


src/idiomatic/leveraging-the-type-system/token-types.md
```diff
@@ -0,0 +1,57 @@
...
+    if let Some(token) = token::get_token() {
+        // We have a token, so we can do this work.
+        work_that_requires_a_token(token);
+    } else {
+        // We could not get a token, so we couldn't call the function that
+        // needs a value of this type.
+        println!("Could not get token.");
+    }
+}
+```
```
mgeisler, line 32: Sorry to be annoying, but the code block here looks way too long. We typically don't have room for 25 lines of code in a classroom setting.

mgeisler: @djmitche and @gribozavr, I would like to see the red box CSS trick be resurrected so that it's easier to know when a slide becomes too big to fit.

@michael-kerscher, would it be feasible to flag new slides automatically with the CI code you wrote?

tall-vase: Not annoying at all! I'm still picking up practical instruction details in feedback cycles, though right now structure and content feedback is greatly appreciated 🙂

STYLE.md
```diff
@@ -226,9 +226,23 @@ collapsed or removed entirely from the slide.
...
+  > a boolean value indicating if the string is the same forwards and
+  > backwards..."_
+
+  **Instead, prefer** bullet points with background information or actionable
+  **teaching prompts**:
+  > **Good:**
+  >
+  > - Note: The `Ext` suffix is a common convention.
+  > - Ask: What happens if the `use` statement is removed?
+  > - Demo: Comment out the `use` statement to show the compiler error.
```
gribozavr, line 245: I think this content already exists in the main branch? Please rebase to hide the spurious diff.

src/idiomatic/leveraging-the-type-system/token-types.md
```diff
@@ -0,0 +1,57 @@
...
+
+<details>
+
+- Token types let us use the privacy tools of types and modules to control when
```
gribozavr: This sentence is a bit too long. Please think about the reader, in this case the instructor. They only have a brief moment to glance at the speaker notes, so shorter and less complex sentences are preferred, when possible.

gribozavr: I'd start the narrative by describing we want to achieve, and explaining the how later. "use the privacy tools", "have it be a requirement that a consumer" go into implementation details from the get-go, and "A user of the API cannot construct a value of a certain type on their own", "because in rust you cannot pass "null" values for any type" explain why this technique works. However the audience is still unsure what we're trying to do.

Consider something like "A token represents a permission to perform certain actions. We represent tokens as types, and actions as functions that require a value of a token type. This way, the function can only be called by someone who has a token - in other words, by someone who has permission."

While the instructor explains the idea we could have slide up with a couple of bullet points:

* A token represents a permission to perform an action
* A token is a distinct type
* Token creation is tightly controlled
* The action is a function, requires a token

and a piece of code to illustrate the idea behind the pattern (it does not compile!)

```rust
struct Token { ... }

fn check_permissions(args: ...) -> Option<Token> { ... }

fn protected_work(_: Token) { ... }
```

gribozavr: This explanation is a bit too abstract and sounds very similar to how we introduce the typestate pattern, "How can we ensure that only valid operations are allowed on a value based on its current state?", except that we're not talking about a specific state. Could we make it more concrete? Should we merge the two sections? Should this section be ordered before typestate but after newtype?

My thinking is that newtype wraps a concrete piece of data, and represents a proof that the data satisfies a predicate.

What I see in the token types section suggests that a token type represents a proof that some data elsewhere, or the program control flow, satisfies a certain predicate.

The current progression seems to be newtype -> typestate -> token types.

This does not seem ideal. First we introduce a foundational concept (Newtype), then jump to a complex application (the generic serializer in Typestate), and then present a simpler concept (token types). It seems to me that token types are a more straightforward extension of the initial newtype pattern.

src/idiomatic/leveraging-the-type-system/token-types.md
```diff
@@ -0,0 +1,57 @@
+---
+minutes: 0
+---
+
+# Token Types
```
gribozavr, line 5: Please add a short definition of what token types are.

src/idiomatic/leveraging-the-type-system/token-types.md
```diff
@@ -0,0 +1,57 @@
...
+
+```rust,editable
+pub mod token {
+    pub struct Token {
+        proof: (),
+    }
```
gribozavr: This can be made shorter, given the current narrative in the speaker notes we don't really need the `proof` field.

src/idiomatic/leveraging-the-type-system/token-types.md
```diff
@@ -0,0 +1,57 @@
...
+
+    pub fn get_token() -> Option<Token> {
+        Some(Token { proof: () })
+    }
+}
+
+pub fn work_that_requires_a_token(token: token::Token) {
```
gribozavr: Consider a shorter name, like "protected_work".

src/idiomatic/leveraging-the-type-system/token-types.md
```diff
@@ -0,0 +1,57 @@
...
+}
+
+fn main() {
+    if let Some(token) = token::get_token() {
+        // We have a token, so we can do this work.
+        work_that_requires_a_token(token);
+    } else {
+        // We could not get a token, so we couldn't call the function that
+        // needs a value of this type.
+        println!("Could not get token.");
```
gribozavr:
```suggestion
        // We could not get a token, so we can't call `protected_work`.
        println!("Could not get token.");
```

src/idiomatic/leveraging-the-type-system/token-types/permission-tokens.md
```diff
@@ -0,0 +1,56 @@
+---
+minutes: 0
+---
+
+# Permission Tokens
+
+```rust,editable
+mod sudo {
+    // A public type with private fields kept behind a module
+    // to enforce privacy boundaries.
```
gribozavr: I think this comment belongs better in token-types.md where we explain the pattern and show an implementation skeleton.

src/idiomatic/leveraging-the-type-system/token-types.md
```diff
@@ -0,0 +1,57 @@
...
+<details>
+
+- Token types let us use the privacy tools of types and modules to control when
+  an API consumer has access to a value of a certain type, and have it be a
+  requirement that a consumer has a value of that type in order to perform
+  certain actions.
+
+  A user of the API cannot construct a value of a certain type on their own, so
+  they need to go through "proper" channels to do so. The designer of the API
+  gets to choose those proper channels and how they behave.
```
gribozavr: It would be nice to prompt the instructor to try to construct the token in main and show that the compiler does not allow doing that.

src/idiomatic/leveraging-the-type-system/token-types.md
```diff
@@ -0,0 +1,57 @@
...
+
+- There's some overlap between this and general API design, because in rust you
```
gribozavr: I don't quite understand what "general API design" refers to here.

src/idiomatic/leveraging-the-type-system/token-types.md
```diff
@@ -0,0 +1,57 @@
...
+  cannot pass "null" values for any type. If a function asks for a value, you
+  have to construct that value! If you can't construct that value yourself, you
+  need to call functions that can!
```
gribozavr: It would be nice to refer back to the same idea that the newtype pattern uses.

src/idiomatic/leveraging-the-type-system/token-types.md
```diff
@@ -0,0 +1,57 @@
...
+
+- By an API user showing they have access to a value of a certain type, we can
```
gribozavr: Please simplify the writing, break it into multiple sentences etc.

src/idiomatic/leveraging-the-type-system/token-types.md
```diff
@@ -0,0 +1,57 @@
...
+  assume that whatever invariants we put around the construction of that type
+  likely apply.
```
gribozavr: This is also a repetition of the idea from parse-don-t-validate.md, so it would be good to ground everything in newtype. You can of course say that token types are more specialized, abstract etc.

src/idiomatic/leveraging-the-type-system/token-types/permission-tokens.md
```diff
@@ -0,0 +1,56 @@
...
+        );
+    } else {
+        eprintln!("Incorrect password! Could not prove privileges.")
+    }
+}
+```
+
+<details>
+
+- This example shows modelling gaining root privileges with a password and
```
gribozavr: Sorry, but this example feels contorted and disconnected from the reality because Unix privileges are enforced by the kernel, not by userspace. This would be confusing or distracting for some people, who would be thinking about this discrepancy while the example is presented, instead of thinking about token types.

---

A more realistic example could be a fragment of a hypothetical bulletin board implementation.

First, show the implementation without token types. The `login()` function returns an `Option<User>`. There's also a function `User::is_admin()`. The function `delete_message()` takes the message ID. It has a doc comment that says that the caller must check that the user is an admin. The instructor explains that this is error-prone.

Second slide: the `login()` function returns a pair `(User, Option<AdminToken>)`, and the function to delete a message requires an `AdminToken` in addition to the message ID.

---

Another alternative would be a document approval workflow:

- `Document::approve()` returns an `ApprovalToken`.
- A `publish()` function requires the document and the `ApprovalToken` as proof that the document was indeed approved.

---

Please pick one of these examples, or think of another one that feels more realistic, and replace the `SuToken` example with the new one.

src/idiomatic/leveraging-the-type-system/token-types.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+<details>
+
+- Token types let us use the privacy tools of types and modules to control when
+  an API consumer has access to a value of a certain type, and have it be a
+  requirement that a consumer has a value of that type in order to perform
+  certain actions.
+
+  A user of the API cannot construct a value of a certain type on their own, so
```
gribozavr: > A user of the API cannot construct a value of a certain type on their own

The instructor should emphasize that the private field `proof` (that is otherwise unused!) is critical to prevent users from constructing tokens on their own. Removing the "unused" field to simplify the code would be an easy mistake to make.

This could be also a good opportunity for the instructor to first ask the class why the `proof` field is there, and what would happen if it was removed to simplify the code.

src/idiomatic/leveraging-the-type-system/token-types.md
gribozavr: +1 about the code being a bit too long. Please see my other comments, I think they will lead to some changes in the flow for this slide, which will also affect the code. Once done, please evaluate the code length again.

src/idiomatic/leveraging-the-type-system/token-types/mutex-guard.md
```diff
@@ -0,0 +1,39 @@
+---
+minutes: 5
+---
+
+# Mutex Guards – Token Types with Data
+
+A token type might only be useful when there's other data, a mutex guard is an
+example of permission + data.
```
gribozavr: I'm not sure what this sentence is trying to say ("might only be useful when" throws me off). If I were to summarize MutexGuard myself, I'd first recall that a newtype wraps the data by value. In contrast, a token type can hold a reference to data that it does not own. A plain Rust reference is a permission to access data that it points to, and so can be a token type. Most of this would go into the speaker notes, and a condensed version would be on the slide.

src/idiomatic/leveraging-the-type-system/token-types/mutex-guard.md
```diff
@@ -0,0 +1,39 @@
...
+
+```rust,editable
+use std::sync::{Arc, Mutex, MutexGuard};
+
+fn main() {
+    let mutex = Arc::new(Mutex::new(42));
+    let try_mutex_guard: Result<MutexGuard<'_, _>, _> = mutex.lock();
+    if let Ok(mut guarded) = try_mutex_guard {
+        // We've proven we have safe access to the data through the successful
+        // construction of
```
gribozavr: Incomplete sentence.

src/idiomatic/leveraging-the-type-system/token-types/mutex-guard.md
```diff
@@ -0,0 +1,39 @@
...
+example of permission + data.
+
+```rust,editable
+use std::sync::{Arc, Mutex, MutexGuard};
+
+fn main() {
+    let mutex = Arc::new(Mutex::new(42));
+    let try_mutex_guard: Result<MutexGuard<'_, _>, _> = mutex.lock();
+    if let Ok(mut guarded) = try_mutex_guard {
+        // We've proven we have safe access to the data through the successful
```
gribozavr:
```suggestion
        // We've proven we have exclusive access to the data through the successful
```

src/idiomatic/leveraging-the-type-system/token-types/mutex-guard.md
gribozavr: Consider rewriting the comment in a more concise way:

"Existence of a MutexGuard instance proves that we have exclusive access to the data."

src/idiomatic/leveraging-the-type-system/token-types/mutex-guard.md
```diff
@@ -0,0 +1,39 @@
+---
+minutes: 5
+---
+
+# Mutex Guards – Token Types with Data
```
gribozavr: I think the opposite order would be better, "Token types with data: Mutex guards"

src/idiomatic/leveraging-the-type-system/token-types/mutex-guard.md
```diff
@@ -0,0 +1,39 @@
...
+        // We've proven we have safe access to the data through the successful
+        // construction of
+        *guarded = 451;
+    }
+}
+```
+
+<details>
+
+<!-- TODO: Reference the Mutex section of the RAII chapter once that is merged. -->
```
gribozavr:
```suggestion
<!-- TODO: Reference the Mutex section of the RAII chapter once that is merged.
Remind the students that the RAII section specifically covered automatic mutex unlocking and did not talk about access to the data.
-->
```

src/idiomatic/leveraging-the-type-system/token-types/mutex-guard.md
```diff
@@ -0,0 +1,39 @@
...
+
+- Mutexes need to enforce mutual exclusion of read/write access to a value.
+  We've covered Mutexes earlier in this course already (See: RAII/Mutex), but
+  here we're looking at `MutexGuard` specifically.
+
+- `MutexGuard` is a value generated by a `Mutex` that "proves" you have
```
gribozavr: Why quotes around "proves"?

src/idiomatic/leveraging-the-type-system/token-types/mutex-guard.md
```diff
@@ -0,0 +1,39 @@
...
+  read/write access at that point in time, as well as holding onto a reference
```
gribozavr: Please break up into two sentences.

src/idiomatic/leveraging-the-type-system/token-types/mutex-guard.md
```diff
@@ -0,0 +1,39 @@
...
+  to the data it wraps for reading and writing of that value.
+
+  If a `MutexGuard` isn't returned by `mutex.lock()`, you don't have permission
```
gribozavr: This could be misunderstood as talking about a hypothetical situation where we have a `MutexGuard` but its provenance is incorrect.

I think what you mean is that if `mutex.lock()` didn't return a `MutexGuard`, the user does not have permission to access the data. What's important to emphasize here is that not only the user does not have permission (which is also true, e.g., for C++ mutexes), the user **does not have a way** to access the data (which sets Rust mutexes apart).

src/idiomatic/leveraging-the-type-system/token-types/branded-tokens.md
```diff
@@ -0,0 +1,84 @@
+---
+minutes: 0
+---
+
+# Lifetime-branded tokens
+
+We can tie a token to a specific value by using lifetimes.
```
gribozavr:
```suggestion
Branded types are token types tied to a specific value through Rust's lifetimes.
```

src/idiomatic/leveraging-the-type-system/token-types.md
```diff
@@ -0,0 +1,57 @@
+---
+minutes: 0
```
gribozavr: Here and everywhere: please try to act out the presentation and estimate the timing.

src/idiomatic/leveraging-the-type-system/token-types/branded-tokens.md
```diff
@@ -0,0 +1,84 @@
...
+- This kind of token is **highly restrictive**, but the things that it makes
+  possible to prove as safe within the rust type system are meaningful. The data
+  structures we end up passing to the closures cannot be returned on their own.
+
+  [GhostCell](https://plv.mpi-sws.org/rustbelt/ghostcell/paper.pdf), a structure
+  that allows for safe cyclic data structures in rust, uses this principle to
+  make sure cells can't "escape" a context where we know where cyclic operations
+  are safe.
+
+</details>
```
gribozavr: This is all technically correct, but too dense for a slide. I think we need to have a more gentle flow, broken down into multiple slides, starting with motivation (in this example it is not clear what the token is achieving). When we move on to explain the implementation, we should first start with a sketch that has the right structure but omits some of the details that make the code correct for the sake of brevity (e.g., the variance mechanics).

---

Here's an idea, but you can of course come up with your own.

Motivation: if you decide to use the `AdminToken` example in the previous slide, you can build a segue to this idea. Remind the students that the web server processes multiple requests, but only one at a time. We could use one user's credentials to log in and get an `AdminToken`, stash it in a global variable somewhere, and then process other user's message deletion request using that token.

=> Idea for a solution: the `Board::delete_message()` function should only accept a token that was returned by `Board::login()` by **the same object**.

gribozavr (changes requested):

src/idiomatic/leveraging-the-type-system/token-types.md
tall-vase: We can't quite do that if we're covering branded types, as that depends on the Borrowck invariants work (introduction of PhantomData, use of Lifetimes to enforce invariants). If we cut the Branded types content, then this re-ordering would be feasible.

gribozavr: Sorry, I tried referencing those sections, but I don't understand. The token types slide does not depend on PhantomData or complex use of lifetimes. Could you explain this issue in more detail?

gribozavr: To clarify, I don't mean moving the entirety of this PR between newtype and typestate. Only this slide.

src/idiomatic/leveraging-the-type-system/token-types/permission-tokens.md
```diff
@@ -0,0 +1,47 @@
+---
+minutes: 5
+---
+
+# Permission Tokens
+
+Token types work well as a trivial "proof of checked permission."
```
gribozavr:
```suggestion
Token types work well as a proof of checked permission.
```

src/idiomatic/leveraging-the-type-system/token-types.md
```diff
@@ -0,0 +1,64 @@
...
+- Motivation: We want to be able to restrict user's access to functionality
+  until they've performed a specific task.
+
+  We can do this by defining a type the API consumer cannot construct on their
+  own, through privacy tools of structs and modules.
+
+- Ask: What is the purpose of the `proof: ()` field here?
+
+  Without `proof: ()`, `Token` would have no private fields and users would be
+  able to construct values of `Token` arbitrarily.
```
gribozavr, line 50: It would be nice to prompt the instructor to try to construct the token in main and show that the compiler does not allow doing that.

src/idiomatic/leveraging-the-type-system/token-types.md
```diff
@@ -0,0 +1,64 @@
...
+```
+<!-- dprint-ignore-end -->
+
+<details>
+
+- Motivation: We want to be able to restrict user's access to functionality
+  until they've performed a specific task.
+
+  We can do this by defining a type the API consumer cannot construct on their
+  own, through privacy tools of structs and modules.
```
gribozavr: It would be nice remind students that the newtype pattern is based on the same idea.

src/idiomatic/leveraging-the-type-system/token-types/mutex-guard.md
```diff
@@ -0,0 +1,43 @@
+---
+minutes: 10
+---
+
+# Token Types with Data: Mutex Guards
+
+Sometimes, a token type needs additional data. A mutex guard is an example of
+permission + data.
```
gribozavr:
```suggestion
a permission + data.
```

src/idiomatic/leveraging-the-type-system/token-types/mutex-guard.md
```diff
@@ -0,0 +1,43 @@
...
+
+- `MutexGuard` is a value generated by a `Mutex` that proves you have read/write
+  access at that point in time.
+
+  `MutexGuard` also holds onto a reference to the `Mutex` that generated it,
+  with `Deref` and `DerefMut` implementations that give access to the data of
+  `Mutex` while the underlying `Mutex` keeps that data private from the user.
+
+- If `mutex.lock()` does not return a `MutexGuard`, you don't have permission to
+  change the value within the mutex.
```
gribozavr, line 41: What's important to emphasize here is that not only the user does not have permission (which is also true, e.g., for C++ mutexes), the user does not have a way to access the data (which sets Rust mutexes apart).

src/idiomatic/leveraging-the-type-system/token-types/branded-01-motivation.md
```diff
@@ -0,0 +1,55 @@
+---
+minutes: 10
+---
+
+# Value Specific Tokens (Branding 1/4)
```
gribozavr:
```suggestion
# Value-Specific Tokens (Branding 1/4)
```

or "value-bound"

src/idiomatic/leveraging-the-type-system/token-types/branded-01-motivation.md
```diff
@@ -0,0 +1,55 @@
...
+
+What if we want there to be a token tied to a specific value?
```
gribozavr:
```suggestion
What if we want to tie a token to a specific value?
```

src/idiomatic/leveraging-the-type-system/token-types/branded-01-motivation.md
```diff
@@ -0,0 +1,55 @@
...
+
+impl Data {
+    fn get_token(&self) -> ValueSpecificToken {
+        ValueSpecificToken {}
+    }
+    fn use_token(&self, token: &ValueSpecificToken) {}
+}
+
+fn main() {
+    let branded_1 = Data {};
```
gribozavr: Consider s/branded/value/ in variable names to better match the terms used in the introductory sentence above: "token" bound to a "value".

src/idiomatic/leveraging-the-type-system/token-types/branded-01-motivation.md
```diff
@@ -0,0 +1,55 @@
...
+
+    // But how do we prevent this?
+    branded_1.use_token(&token_2);
+    branded_2.use_token(&token_1);
+}
+```
+
+<details>
+
+- What if we want to tie a token to a _specific variable_ in our code? Can we do
```
gribozavr, line 37: The slide uses the term "value". I think the term "variable" is indeed more precise, because "value" can be understood in a more abstract sense (like the one that Stepanov uses, the one that relates to the current state of the object, rather than the identity).

Here we are binding the token to the identity of a specific object, hence "variable" is better. If you agree, please update the slide.

src/idiomatic/leveraging-the-type-system/token-types/branded-01-motivation.md
```diff
@@ -0,0 +1,55 @@
...
+  this in Rust's type system?
+
+- Motivation: Say we want "proof of index" for a value of some type, but we
```
gribozavr: I'm not sure what "index" refers to here.

Consider: "Say we want a token to prove that some check has been done to a given variable. However that token should be incompatible with other variables of the same type."

src/idiomatic/leveraging-the-type-system/token-types/branded-01-motivation.md
```diff
@@ -0,0 +1,55 @@
...
+# Value Specific Tokens (Branding 1/4)
+
+What if we want there to be a token tied to a specific value?
+
+```rust,editable
+struct Data {}
+struct ValueSpecificToken {}
+
+impl Data {
+    fn get_token(&self) -> ValueSpecificToken {
```
gribozavr: s/get_token/check_permission/ to follow the narrative from before, and emphasize that there is some checking happening. Consider even returning an Option. You can switch it up to a different use case later, but I think continuity about using "permissions" as a motivation would help understand this example easier.

src/idiomatic/leveraging-the-type-system/token-types/branded-01-motivation.md
```diff
@@ -0,0 +1,55 @@
...
+fn main() {
+    let branded_1 = Data {};
+    let token_1 = branded_1.get_token();
+    branded_1.use_token(&token_1); // Works fine!
+    let branded_2 = Data {};
+    let token_2 = branded_2.get_token();
+
+    // But how do we prevent this?
+    branded_1.use_token(&token_2);
+    branded_2.use_token(&token_1);
```
gribozavr:
```suggestion

    // But how do we prevent this?
    let branded_2 = Data {};
    branded_2.use_token(&token_1);
```

A shorter way to illustrate the same point.

First show that var_1 works with the token minted for it specifically.

Then show that the same token also works for a new object.

That's all we need to demonstrate to show the problem.

src/idiomatic/leveraging-the-type-system/token-types/branded-01-motivation.md
```diff
@@ -0,0 +1,55 @@
...
+}
+```
```
gribozavr, line 33: Excellent motivational example! 👍  Simple, just shows the problem, builds on top of what readers already know, and omits the complexity of the solution.

src/idiomatic/leveraging-the-type-system/token-types/branded-01-motivation.md
```diff
@@ -0,0 +1,55 @@
...
+<details>
+
+- What if we want to tie a token to a _specific variable_ in our code? Can we do
+  this in Rust's type system?
+
+- Motivation: Say we want "proof of index" for a value of some type, but we
+  don't want that index to "cross over" to values of the same or different
+  types.
+
+  Or what if we want to model "arena allocation" somehow?
```
gribozavr: This needs more words for the instructor to be able to expand on it. (I don't get the connection yet.)

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,31 @@
...
+struct InvariantLifetime<'id>(PhantomData<*mut &'id ()>);
+```
+
+<details>
+
+<!-- TODO: Link back to PhantomData in the borrowck invariants chapter.
+- We saw `PhantomData` back in the Borrow Checker Invariants chapter.
+-->
+
+- Note: There are two parts to this, the first part is `PhantomData`, and how
```
gribozavr: I'm not sure the instructor would be able to expand this into specific words to say.

I'd expect the instructor notes to start by guiding the audience through

(a) what they see on the screen, mechanically speaking, because `*mut &'id ()` is just token soup for most people,

(b) what this achieves at a high level,

(c) how it achieves that.

If you find it difficult to write speaker notes, try writing a complete script first, then condense it into the speaker notes.

I wouldn't even mind you pasting a suggested complete script in a separate `<details>`, tag, this is one of the most complex slides in the whole repository. Just keep it separate from the speaker notes, which should be shorter and bullet-point-based.

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,31 @@
...
+
+  How type parameter of `InvariantLifetime`'s internal `PhantomData` relates to
+  [Subtyping of lifetimes](https://doc.rust-lang.org/stable/reference/subtyping.html)
+  is what forces this "branding" between lifetimes to apply.
+
+  Without it, the compiler would see the lifetimes on the types we're handling
+  as "similar enough" (able to be subtyped) and users would be able to use the
+  token for one value with a different value.
+
+</details>
```
gribozavr, line 175: Alternative flow for the "branded types" section could be to introduce the solution on the next slide but avoiding the `InvariantLifetime` complexity. Then once the general structure of the solution is clear, point out this hole, introduce `InvariantLifetime`, and show how to fix the hole.

The advantage of this suggested flow is that we're transitioning directly from the problem statement to the solution, without a detour through `InvariantLifetime`. We introduce `InvariantLifetime` later, once the students clearly see an issue to fix.

src/idiomatic/leveraging-the-type-system/token-types/branded-03-impl.md
```diff
@@ -0,0 +1,64 @@
...
+struct BrandedToken<'id>(InvariantLifetime<'id>);
+
+struct MyStructure<'id>(Vec<u8>, InvariantLifetime<'id>);
+
+impl<'id> MyStructure<'id> {
+    fn new<T>(
+        // The data we want to modify in this context.
+        data: Vec<u8>,
+        // We want a function whose lifetime is specific to each time
+        // `new` is called, not tied to any one data structure other than
```
gribozavr:
```suggestion
        // We want a function whose lifetime is specific to each call to
        // `new`, not tied to any one data structure other than
```

src/idiomatic/leveraging-the-type-system/token-types/branded-tokens.md
gribozavr: You made great improvements, but the example is still abstract.

src/idiomatic/leveraging-the-type-system/token-types/branded-01-motivation.md
```diff
@@ -0,0 +1,55 @@
...
+  Expect students to not reach a good implementation from this, but be willing
+  to experiment and follow through on things.
+
+- This kind of token-association is called Branding. Doing this lets us expand
+  the "proof of work from elsewhere" to more general aspects of rust.
+
+- [`GhostCell`](https://plv.mpi-sws.org/rustbelt/ghostcell/paper.pdf) is a
+  prominent user of this, later slides will touch on it.
+
+</details>
```
gribozavr, line 78: Next, once we have shown the general idea, I'd suggest transitioning to a concrete API design problem (not an abstract one like here).

src/idiomatic/leveraging-the-type-system/token-types/branded-03-impl.md
```diff
@@ -0,0 +1,64 @@
...
+    fn use_token(&mut self, token: &BrandedToken<'id>) {}
+}
+```
+
+<details>
+
+- The underlying Branded Data Structure we're going to use here is just a
+  `Vec<u8>` (the data) and an `InvariantLifetime`.
+
+- The constructor for this type will take **data for the `Vec<u8`** plus a
```
gribozavr:
```suggestion
- The constructor for this type will take **data for the `Vec<u8>`** plus a
```

src/idiomatic/leveraging-the-type-system/token-types/branded-03-impl.md
```diff
@@ -0,0 +1,64 @@
...
+  function to manipulate the data constructed by `MyStructure::new`.
+
+- Ask: Does anyone know what the `for <'a>` is for?
+
+  Expect not much, it's "for" in the sense of "forall" from mathematics.
+
+- The `for<'a> [trait bound that uses 'a]` binding of `'a` means the lifetime is
+  "self contained."
```
gribozavr: I think we can explain this in more depth, rather than through an analogy.

src/idiomatic/leveraging-the-type-system/token-types/permission-tokens.md
```diff
@@ -0,0 +1,47 @@
...
+- This example shows modelling gaining administrator privileges for a chat
+  client with a password and giving a user a moderator rank once those
+  privileges are gained. The `AdminToken` type acts as "proof of correct user
+  privileges."
+
+  The user asked for a password in-code and if we get the password correct, we
+  get a `AdminToken` to perform administrator actions within a specific
+  environment (here, a chat client).
+
+  Once the permissions are gained, we can call a `add_moderator` function.
```
gribozavr:
```suggestion
  Once the permissions are gained, we can call the `add_moderator` function.
```

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
tall-vase: I think the narrative can be bent towards "this is a tangent, but a relevant one" but I agree it's a bit clunky. I'll take a stab at making the non-stub iteration on this slide not be too reliant on if it's before or after the implementation.

src/idiomatic/leveraging-the-type-system/token-types/branded-tokens.md
tall-vase: The topic is now based off the GhostCell paper's `BrandedVec` implementation, if this isn't concrete enough then I'd be open to some suggestions. I am quite used to abstract concepts, the grounding in what is more useful to show off in the classroom is greatly appreciated.

gribozavr: > The topic is now based off the GhostCell paper's BrandedVec implementation

Yep, I see it now, this is good, let's keep this as a running example.

src/idiomatic/leveraging-the-type-system/token-types/branded-01-motivation.md
```diff
@@ -0,0 +1,70 @@
...
+    bytes: Vec<u8>,
+}
+struct ProvenIndex(usize);
+
+impl Bytes {
+    fn get_index(&self, ix: usize) -> Option<ProvenIndex> {
+        if ix < self.bytes.len() { Some(ProvenIndex(ix)) } else { None }
+    }
+    fn get_proven(&self, token: &ProvenIndex) -> u8 {
+        self.bytes[token.0]
```
gribozavr: Use get_unchecked to hammer the point that we don't need range checking once we have a proven index? But add a debug assert to trigger the panic necessary for the narrative.

src/idiomatic/leveraging-the-type-system/token-types/branded-01-motivation.md
```diff
@@ -0,0 +1,70 @@
...
+    }
+}
+
+fn main() {
+    let data_1 = Bytes { bytes: vec![0, 1, 2] };
+    if let Some(token_1) = data_1.get_index(2) {
+        data_1.get_proven(&token_1); // Works fine!
+        let data_2 = Bytes { bytes: vec![0, 1] };
+        // data_2.get_proven(&token_1); // Panics! How do we prevent this at compile time?
```
gribozavr:
```suggestion
        data_1.get_proven(&token_1); // Works fine!

        // let data_2 = Bytes { bytes: vec![0, 1] };
        // data_2.get_proven(&token_1); // Panics! How do we prevent this at compile time?
```

Trying to make it visually clear that the last two lines belong together

src/idiomatic/leveraging-the-type-system/token-types/branded-01-motivation.md
```diff
@@ -0,0 +1,70 @@
...
+}
+```
+
+<details>
+
+- What if we want to tie a token to a _specific variable_ in our code? Can we do
+  this in Rust's type system?
+
+- Motivation: We want to have a Token Type that represents a known, valid index
+  for a a byte array that we know exists.
```
gribozavr:
```suggestion
  into a byte array.
```

src/idiomatic/leveraging-the-type-system/token-types/branded-01-motivation.md
```diff
@@ -0,0 +1,70 @@
...
+  indexes at compile time.
+
+- Ask: How might we try to do this?
+
+  Expect students to not reach a good implementation from this, but be willing
+  to experiment and follow through on suggestions.
+
+- Ask: What are the alternatives, why are they not good enough?
+
+  Expect runtime checking of index bounds, especially as `get_index` already
```
gribozavr:
```suggestion
  Expect runtime checking of index bounds, especially as `Vec::get()` already
```

src/idiomatic/leveraging-the-type-system/token-types/branded-01-motivation.md
```diff
@@ -0,0 +1,70 @@
...
+  uses runtime checking.
+
+  Runtime checking of bounds does not prevent the erroneous crossover in the
```
gribozavr:
```suggestion
  Runtime bounds checking does not prevent the erroneous crossover in the
```

src/idiomatic/leveraging-the-type-system/token-types/branded-01-motivation.md
```diff
@@ -0,0 +1,70 @@
...
+  first place, only the `panic` result. That erroneous checking
```
gribozavr:
```suggestion
  first place, it only guarantees a panic. That erroneous checking
```

gribozavr: "That erroneous checking" incomplete sentence

src/idiomatic/leveraging-the-type-system/token-types/branded-01-motivation.md
```diff
@@ -0,0 +1,70 @@
...
+
+- The kind of token-association we will be doing here is called Branding. Doing
+  this lets us expand the "proof of work from elsewhere" to more general aspects
+  of rust.
```
gribozavr:
```suggestion
- The kind of token-association we will be doing here is called Branding.
  This is an advanced technique that expands applicability of token types to more API designs.
```

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,124 @@
+---
+minutes: 20
+---
+
+# `PhantomData` and Lifetime Subtyping (Branding 2/4)
+
+We need a mechanism to make lifetimes distinct enough from each other to not
+compile when we try to compare them.
```
gribozavr:
```suggestion
Idea:
- Use a lifetime as a unique brand for each token.
- Make lifetimes sufficiently distinct so that they don't implicitly convert into each other.
```

(1) It would be nice to clearly state the intent behind what we're about to do ("Use a lifetime as a unique brand for each token")

(2) It was not clear to me what "compare" was referring to

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,124 @@
...
+```
+<!-- dprint-ignore-end -->
+
+<details>
+
+<!-- TODO: Link back to PhantomData in the borrowck invariants chapter.
+- We saw `PhantomData` back in the Borrow Checker Invariants chapter.
+-->
+
+- **Goal**: We want two lifetimes that the rust compiler cannot determine if one
```
gribozavr, line 60: Maybe step back a little and say that Rust defines a subtyping relationship between references, and references with longer lifetimes implicitly convert into shorter lifetimes. This also happens when a user-defined type has a lifetime parameter, Rust defines a subtyping relationship.

I know you're talking more about subtyping later, but maybe a brief mention upfront would be good. I would like you to consider how you'd introduce the flow from the idea ("Use a lifetime as a unique brand for each token") to the suggested solution (disable subtyping) in the shortest way possible when first introducing the slide.

... thus our goal: eliminate that subtying relationship, in order for each Token's lifetime to be unique and incompatible with every other Token.

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,124 @@
...
+#[derive(Default)]
+struct InvariantLifetime<'id>(PhantomData<&'id ()>); // The main focus
+
+struct Wrapper<'a> { value: u8, invariant: InvariantLifetime<'a> }
+
+fn lifetime_separator<T>(value: u8, f: impl for<'a> FnOnce(Wrapper<'a>) -> T) -> T {
+    f(Wrapper { value, invariant: InvariantLifetime::default() })
+}
+
+fn compare_lifetimes<'a>(left: Wrapper<'a>, right: Wrapper<'a>) {}
```
gribozavr: Would "try_coerce_lifetimes", "are_coercible_lifetimes", or "check_common_type" be better names?

I would like to avoid using the word "compare" because that's just not a thing that the type system in Rust does.

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,124 @@
...
+
+<!-- TODO: Link back to PhantomData in the borrowck invariants chapter.
+- We saw `PhantomData` back in the Borrow Checker Invariants chapter.
+-->
+
+- **Goal**: We want two lifetimes that the rust compiler cannot determine if one
+  outlives the other.
+
+  We are using `compare_lifetimes` as a compile-time check to see if the
+  lifetimes are being subtyped.
```
gribozavr:
```suggestion
  lifetimes are in a subtyping relationship.
```

or "are subtypes", or "have a common shorter lifetime"

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,124 @@
...
+
+- Note: This slide compiles, by the end of this slide it should only compile
+  when `subtyped_lifetimes` is commented out.
+
+- There are two important parts of this code:
+  - The `impl for<'a>` bound on the closure passed to `lifetime_separator`.
+  - The way lifetimes are used in the parameter for `PhantomData`.
+
+- `for<'a> [trait bound]` is a way of introducing a new lifetime variable to a
```
gribozavr: I know I asked for a direct explanation, but I think we can customize it a bit here. I'd rather refer to "function types" rather than "trait bounds" here (even though that's technically true, but a bit too abstract).

Once you start talking about the types of functions, you can customize the rest of the explanation. What's the type of sqrt? `Fn(f32)->f32`. Similarly, `for<'a> Fn(Wrapper<'a>)->X` is the type of a function (closure) that is generic over a lifetime. What does it mean to be generic over a lifetime? The function body (closure body) must be type-safe and memory-safe regardless of which lifetime it is called with. Since it is not in control of which lifetime it is called with (the caller is!), it must work for _any_ lifetime. That bridges us to the forall quantifier analogy.

Maybe this is worth putting on a separate slide.

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,124 @@
...
+  lifetime parameters to make them "used."
+
+- Ask: What can we do with `PhantomData`?
+
+  Expect mentions of the Typestate pattern, tying together the lifetimes of
+  owned values.
+
+- Ask: In other languages, what is subtyping?
+
+  Expect mentions of Inheritance, being able to use a value of type `B` when a
```
gribozavr:
```suggestion
  Expect mentions of inheritance, being able to use a value of type `B` when a
```

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,124 @@
...
+  of the lifetime captured by `InvariantLifetime` is too lenient.
+
+  <!-- Note: We've been using "invariants" in this module in a specific way, but subtyping introduces _invariant_, _covariant_, and _contravariant_ as specific terms. -->
+
+- Ask: How can we make it more restrictive?
+
+  Expect or demonstrate: Making it `&'id mut ()` instead. This will not be
+  enough!
+
+  We need to use a
```
gribozavr, line 137: This is a very complex topic. What's your intent? Mention it for those who are in the know, or ensure that everyone in the audience understands it?

If it is the former (given that this slide is not introducing this concept, I think it must be the former), I think we need to mention this in the notes for the instructor that they should not try to make the whole class understand variance.

But is it really an option, given what follows? WDYT?

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,124 @@
...
+  lifetime but not type).
+
+  Those last two should not compile, which means we've finally found candidates
+  for how to bind lifetimes to `PhantomData` so they can't be compared to one
+  another in this context.
+
+- Wrap up: We've introduced ways to stop the compiler from deciding that
+  lifetimes are "similar enough" by choosing a Variance for a lifetime captured
+  in `PhantomData` that is restrictive enough to prevent this slide from
+  compiling.
```
gribozavr: Maybe add a final summary sentence in even simpler English, something like "thus, we now can create token values with unique lifetimes, and one token does not implicitly convert to any other"

src/idiomatic/leveraging-the-type-system/token-types.md
```diff
@@ -0,0 +1,72 @@
...
+  runtime.
+
+- Ask: What is the purpose of the `proof: ()` field here?
+
+  Without `proof: ()`, `Token` would have no private fields and users would be
+  able to construct values of `Token` arbitrarily.
+
+  Demonstrate: Try to construct the token manually in `main`, then remove the
+  `proof` field from `Token` to show how users would be able to construct
+  `Token` without restriction if it had no private fields.
```
gribozavr:
```suggestion
  Demonstrate: Try to construct the token manually in `main` and show the compilation error.
  Demonstrate: Remove the
  `proof` field from `Token` to show how users would be able to construct
  `Token` if it had no private fields.
```

Trying to make it clear that there two demos here, and also using shorter sentences.

src/idiomatic/leveraging-the-type-system/token-types/branded-03-impl.md
```diff
@@ -0,0 +1,70 @@
...
+
+<details>
+
+- Motivation: We want to have "proven indexes" for a type, and we don't want
+  those indexes to be usable by different variables of the same type. We also
+  don't want those indexes to escape a scope.
+
+  Our Branded Type will be `Bytes`: a byte array.
+
+  Our Branded Token will be `ProvenIndex`: an index known to be occupied.
```
gribozavr:
```suggestion
  Our Branded Token will be `ProvenIndex`: an index known to be in range.
```

src/idiomatic/leveraging-the-type-system/token-types/branded-03-impl.md
```diff
@@ -0,0 +1,70 @@
...
+- There are several notable parts to this implementation:
+  - `new` does not return a `Bytes`, instead asking for "starting data" and a
+    use-once Closure that is passed a `Bytes` when it is called.
+  - That `new` function has a `for<'a>` on its trait bound.
+  - We have both a getter for an index and a getter for a values with a proven
+    index.
+
+- Ask: Why does `new` not return a `Bytes`?
+
+  Answer: At least partially because we need to introduce a lifetime to `Bytes`.
```
gribozavr:
```suggestion
  Answer: Because we need `Bytes` to have a unique lifetime.
```

If there are more reasons, feel free to expand on them.

gribozavr: Consider adding a hypothetical:

Question: So what if `new()` returned `Bytes`, what is the specific harm that it would cause?

Answer: Think about the signature of that hypothetical `new()` method:

`fn new<'a>() -> Bytes<'a> { ... }`

this signature means that the caller chooses what `'a` is. We can't allow that, because we must ensure that every lifetime arg in every `Bytes` object is unique. If caller could choose `'a`, they could repeatedly call `new` with the same lifetime.

src/idiomatic/leveraging-the-type-system/token-types/branded-03-impl.md
```diff
@@ -0,0 +1,70 @@
...
+
+- Ask: Why do we need both a `get_index` and a `get_proven`?
+
+  Expect "Because we can't know if an index is occupied at compile time"
+
+  Ask: Then what's the point of the proven indexes?
+
+  Answer: The throughline of preventing proven indexes "crossing over" to arrays
+  of the same type, causing panics.
```
gribozavr: I'd rather motivate them by avoiding repeated bounds checks. Maybe this should be explained on the branded-01 slide?

src/idiomatic/leveraging-the-type-system/token-types/branded-03-impl.md
tall-vase: Sure thing, I'll just copy this in more or less verbatim for the sake of getting this done. Can be iterated on later.

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,176 @@
...
+
+## `for<'a>` bound on a Closure
+
+- We are using `for<'a>` as a way of introducing a new lifetime variable to a a
+  function type and asking that the type of that function be true for all
+  possible lifetimes.
+
+  What this also does is remove some ability of the compiler to make assumptions
+  about that specific lifetime for the function argument, as it must meet rust's
+  borrow checking rules regardless of the "real" lifetime its arguments capture.
```
gribozavr:
```suggestion
  borrow checking rules regardless of the "real" lifetime its arguments are going to have.
```

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,176 @@
...
+  Only the callsite is able to determine the "real" lifetime, the function
```
gribozavr:
```suggestion
  The caller is substituting in actual lifetime, the function
```

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,176 @@
...
+- Note: This slide compiles, by the end of this slide it should only compile
+  when `subtyped_lifetimes` is commented out.
+
+- There are two important parts of this code:
+  - The `impl for<'a>` bound on the closure passed to `lifetime_separator`.
+  - The way lifetimes are used in the parameter for `PhantomData`.
+
+## `for<'a>` bound on a Closure
+
+- We are using `for<'a>` as a way of introducing a new lifetime variable to a a
```
gribozavr:
```suggestion
- We are using `for<'a>` as a way of introducing a lifetime generic parameter to a
```

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,176 @@
...
+  function type and asking that the type of that function be true for all
```
gribozavr:
```suggestion
  function type and asking that the body of the function to work for all
```

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,176 @@
...
+  about that specific lifetime for the function argument, as it must meet rust's
+  borrow checking rules regardless of the "real" lifetime its arguments capture.
+  Only the callsite is able to determine the "real" lifetime, the function
+  itself cannot.
+
+  This is analogous to a forall (Ɐ) quantifier in mathematics, or the way we
+  introduce `<T>` as type variables, but only for lifetimes in trait bounds.
+
+  When we write a function generic over a type `T`, we can't determine that type
+  from within the function itself. Even if we call a function `Fn(A, B) -> C`
```
gribozavr:
```suggestion
  from within the function itself. Even if we call a function `fn foo<T, U>(first: T, second: U)`
```

I understand what you mean, but it was written as a non-generic function.

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,176 @@
...
+  with two arguments of the same type, the body of this function cannot
+  determine if `A` and `B` are the same type until it's called.
```
gribozavr:
```suggestion
  determine if `T` and `U` are the same type.
```

It can't do that even when it is called, can it?

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,176 @@
...
+  asked for a value of type `A` because `B` is a "subtype" of `A`.
+
+- Rust does have Subtyping! But only for lifetimes.
+
+  Ask: If one lifetime is a subtype of another lifetime, what might that mean?
+
+  A lifetime is a "subtype" of another lifetime when it _outlives_ that other
+  lifetime.
+
+- The way that lifetimes captured by `PhantomData` behave depends not only on
```
gribozavr:
```suggestion
- The way that lifetimes used by `PhantomData` behave depends not only on
```

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,176 @@
...
+  where the lifetime "comes from" but on how the reference is defined too.
+
+  The reason this compiles is that the
+  [**Variance**](https://doc.rust-lang.org/stable/reference/subtyping.html#r-subtyping.variance)
+  of the lifetime captured by `InvariantLifetime` is too lenient.
```
gribozavr:
```suggestion
  of the lifetime inside of `InvariantLifetime` is too lenient.
```

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,176 @@
...
+
+  Reason: `*mut` means
+  [mutable raw pointer](https://doc.rust-lang.org/reference/types/pointer.html#r-type.pointer.raw).
+  Rust has mutable pointers! But you cannot reason about them in safe rust.
+  Making this a mutable raw pointer to a reference that has a lifetime
+  complicates the compiler's ability subtype because it cannot reason about
+  mutable raw pointers within the borrow checker.
+
+- Wrap up: We've introduced ways to stop the compiler from deciding that
+  lifetimes are "similar enough" by choosing a Variance for a lifetime captured
```
gribozavr:
```suggestion
  lifetimes are "similar enough" by choosing a Variance for a lifetime
```

src/idiomatic/leveraging-the-type-system/token-types/branded-02-phantomdata.md
```diff
@@ -0,0 +1,176 @@
...
+  with two arguments of the same type, the body of this function cannot
+  determine if `A` and `B` are the same type until it's called.
+
+  This also prevents _the API consumer_ from defining a lifetime themselves,
+  which would allow them to circumvent the restrictions we want to impose.
+
+## PhantomData and Lifetime Variance
+
+- We already know `PhantomData`, which we can use to capture unused type or
+  lifetime parameters to make them "used."
```
gribozavr:
```suggestion
- We already know `PhantomData`, which can introduce a formal no-op usage of an otherwise unused type or a lifetime parameter.
```

src/idiomatic/leveraging-the-type-system/token-types/branded-03-impl.md
```diff
@@ -0,0 +1,84 @@
...
+  our ability to guarantee that the lifetimes between different instances of
+  `Bytes` are unique and unable to be subtyped to one another.
+
+- Ask: Why do we need both a `get_index` and a `get_proven`?
+
+  Expect "Because we can't know if an index is occupied at compile time"
+
+  Ask: Then what's the point of the proven indexes?
+
+  Answer: Avoiding bounds checking and maintaining
```
gribozavr: Incomplete sentence

src/idiomatic/leveraging-the-type-system/token-types/branded-01-motivation.md
```diff
@@ -0,0 +1,75 @@
...
+<details>
+
+- What if we want to tie a token to a _specific variable_ in our code? Can we do
+  this in Rust's type system?
+
+- Motivation: We want to have a Token Type that represents a known, valid index
+  into a byte array.
+
+  Once we have these proven indexes we would be able to avoid bounds checks
+  entirely, as the tokens would act as the _proof of an existing index_.
```
gribozavr, line 44:
```suggestion
  entirely, as the tokens would act as the _proof of an existing index_.
  
  Since the index is known to be valid, `get_proven()` can skip the bounds check.
```

PR #2922, opened 2025-09-22
feat(docs): Centralize code block style guide

gribozavr:
- Moves the canonical documentation for code block annotations and language identifiers into STYLE.md.
- Removes the redundant, partial documentation from GEMINI.md.
- Updates GEMINI.md to instruct the AI assistant to always refer to STYLE.md for guidance.

This change makes STYLE.md the single source of truth for all markdown and code style conventions, ensuring consistency for both human and AI contributors.


STYLE.md
```diff
@@ -161,11 +161,34 @@ Contributors should design their slides with this interactivity in mind. The
...
+
+You can add annotations to Rust code blocks to control how they are tested and
+displayed:
+
+- **`editable`**: Makes the code block an interactive playground where users can
+  edit and run the code. This should be used for most Rust examples.
+- **`compile_fail`**: Indicates that the code is expected to fail compilation.
+  This is used to demonstrate specific compiler errors.
+- **`should_panic`**: Indicates that the code is expected to panic when run.
+- **`warnunused`**: Indicates that the code is expected to produce a compiler
```
mgeisler: I don't find any trace of this in `mdbook`: [search](https://github.com/search?q=repo%3Arust-lang%2FmdBook%20warnunused&type=code).

mgeisler: Nor for `rustdoc` in general... is it perhaps an invention by an AI? :smile:

gribozavr: I thought so too but then I checked we use it in `src/types-and-values/variables.md` and `src/testing/lints.md`. There's processing in `theme/book.js`. It was added in 5f7e0c3f6.

mgeisler: Aha! That's a tad confusing, even if the idea is great in itself :smile: Cc @djmitche.

I created #2924 since we need to better track how we modify `book.js`.

PR #2923, opened 2025-09-23
Don’t automatically install recommended packages

mgeisler:
PR to show what happens if we don't install the recommended Debian packages (things fail right now).

### DIFF
--- a/.github/workflows/apt-get-install/action.yml
+++ b/.github/workflows/apt-get-install/action.yml
@@ -41,4 +41,4 @@ runs:
     - name: Installing ${{ inputs.packages }}
       shell: bash
       run: |
-        sudo apt-get install --quiet --yes ${{ inputs.packages }}
+        sudo apt-get install --quiet --yes --no-install-recommends ${{ inputs.packages }}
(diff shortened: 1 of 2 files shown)


michael-kerscher:
I think that PR has is worth pursuing because it makes the packages we depend on more explicit and clear - and has the benefit of removing unnecessary bloat

michael-kerscher (changes requested):
looks good - once the installed packages are manually extended by what is actually in use and the tests are green again.

mgeisler:
> I think that PR has is worth pursuing because it makes the packages we depend on more explicit and clear - and has the benefit of removing unnecessary bloat

Yes, I also think this is the right long-term approach! I just need to find the time to fiddle with the lists to determine the right minimum :smile:

I'm not sure when I'll do this, so anyone is welcome to jump in and build on this PR.

michael-kerscher:
That might be enough: 0aa44f626735a2564b79869d478273b6d8f756d1

michael-kerscher:
To review this, the generated PDFs should be reviewed. I had a look at the en, de, zh-CN pdfs and did not spot anything different

michael-kerscher (commented):
To me this looks complete, but I would like to have additional review to not miss something critical

qwandor (approved):

PR #2932, opened 2025-09-27
Re-enable the red box to show available space

mgeisler:
Way back in #187, I introduced a hacky tool to show the available space on a slide: it was a `mdbook` plugin which injected the necessary CSS on each slide. Crude, but it got the job done.

The logic was moved from Python to a real CSS file with associated JavaScript in #1842. In #1917, the functionality was moved to a dedicated “instructor menu”, together with functionality for saving the state of the Playground on each slide.

Unfortunately, the whole thing was disabled in #1935 since I found that the Playgrounds lost their code with the saving logic. I was also not 100% happy with dedicating space on each slide for a menu only used by instructors.

However, I really think we need a tool to let slide editors know about the available space, so I would like to re-introduce the red box. This time via a keyboard shortcut to make it easy to toggle as needed.

I’m suggesting enabling this for everyone, with the expectation that most people won’t find the shortcut and will quickly disable the box if they do (there is a dedicated button to hide it again).

End-to-end tests have been added for the new functionality.


mgeisler:
Cc @gribozavr, @LukeMathWalker, @timClicks since I hope you folks will find it useful for the new course material you're working on :smile:

mgeisler:
Thanks @qwandor! It seems I need to rework the new tests since they somehow fail in CI 😄 I'll try to look at this soon!

mgeisler:
@michael-kerscher, do you have an idea why the test failed with

> [0-3] 2025-09-28T08:25:15.544Z ERROR webdriver: WebDriverError: tab crashed
> [0-3]   (Session info: chrome=140.0.7339.185) when running "element" with method "POST" and args "{"using":"css selector","value":"#content > main"}"
> [0-3] tab crashed in "Slide welcome-day-1.html.should not be higher than 1333 pixels or wider than 750 pixels"
> tab crashed: WebDriverError: tab crashed
>  (Session info: chrome=140.0.7339.185) when running "element" with method "POST" and args "{"using":"css selector","value":"#content > main"}"
>    at async Slide.violates_max_size (/home/runner/work/comprehensive-rust/comprehensive-rust/tests/src/objects/slide.ts:39:31)
>    at async Context.<anonymous> (/home/runner/work/comprehensive-rust/comprehensive-rust/tests/src/slide-style-guide.test.ts:52:13)
> [0-3] FAILED in chrome - file:///src/slide-style-guide.test.ts

I think this might be unrelated to my changes, but I'm mystified why we see this here and not on every other PR.

qwandor:
Isn't this #2941, which is affecting all PRs?

mgeisler:
> Isn't this #2941, which is affecting all PRs?

Oh, thanks! I hope that is the case... I rebased on top of `main` and hopefully the tests will pass this time. Cc @michael-kerscher

Issue #2941, opened 2025-10-02
Web tests fail due to crashing tabs while checking for style guide conformance check


michael-kerscher:
The WebDriver has issues due to tabs crashing while checking slides for conformance with the style guide.

It seems like this is failing mostly in (because of `with method "POST" and args "{"using":"css selector","value":"#content > main"}"` https://github.com/google/comprehensive-rust/blob/17a04070ed4dc896f60d008d2e61b4c247bb64bd/tests/src/objects/slide.ts#L23-L25

I'm not sure on *why* this happens. A quick mitigation could be to disable the test entirely. Another mitigation option is to only check files that have been changed in the PR at all. The last one should not be too difficult and has a better efficiency and speeds up the checks massively!

This is visible in runs like https://github.com/google/comprehensive-rust/actions/runs/18159303020/job/51777458019

```
Run ./src/slides/create-slide.list.sh
~/work/comprehensive-rust/comprehensive-rust/book/comprehensive-rust-en/html ~/work/comprehensive-rust/comprehensive-rust/tests
~/work/comprehensive-rust/comprehensive-rust/tests
> test
> wdio run ./wdio.conf.ts
Execution of 4 workers started at 2025-10-02T09:02:33.271Z
[0-0] RUNNING in chrome - file:///src/generic-page.test.ts
[0-1] RUNNING in chrome - file:///src/playground.test.ts
[0-3] RUNNING in chrome - file:///src/speaker-notes.test.ts
[0-2] RUNNING in chrome - file:///src/slide-style-guide.test.ts
[0-3] 2025-10-02T09:02:48.800Z WARN webdriver: WebDriverError: element not interactable: [object HTMLButtonElement] has no size and location
[0-3]   (Session info: chrome=140.0.7339.185) when running "actions" with method "POST" and args "{"actions":[{"id":"action1","type":"wheel","parameters":{},"actions":[{"type":"scroll","x":-2531,"y":-15,"deltaX":0,"deltaY":0,"duration":0,"origin":{"element-6066-11e4-a52e-4f735466cecf":"f.AF0A63E12952C6A7284DBF28DCFE6552.d.CEBF99EA81288D278FF1E7AC260C9F05.e.34"}}]}]}"
[0-3] PASSED in chrome - file:///src/speaker-notes.test.ts
[0-0] PASSED in chrome - file:///src/generic-page.test.ts
[0-2] slide concurrency/sync-exercises/link-checker.html is on the exemption list
[...]

PR #2942, opened 2025-10-02
web tests: run style guide compliance check only for modified markdown files

michael-kerscher:
In the CI environment, only changes in the current PR are taken into account.
This speeds up the test process and reduces failure potential if no Markdown files have been changed.

Mitigates #2941 but does not fix the root cause

### DIFF
--- a/tests/src/slides/create-slide.list.sh
+++ b/tests/src/slides/create-slide.list.sh
@@ -24,19 +24,52 @@ if [[ ! -f "${TEST_BOOK_DIR}/index.html" ]]; then
   exit 1
 fi
 
-pushd "${TEST_BOOK_DIR}"
-# exclude special pages that should never be tested
-SLIDES=$(grep -L "Redirecting to..."  -R --include "*.html" \
-    --exclude "exercise.html" \
-    --exclude "solution.html" \
-    --exclude "toc.html" \
-    --exclude "print.html" \
-    --exclude "404.html" \
-    --exclude "glossary.html" \
-    --exclude "index.html" \
-    --exclude "course-structure.html"
-    )
-popd
+# These special pages should never be tested for slide size.
+EXCLUDE_FILES=(
+  "exercise.html"
+  "solution.html"
+  "toc.html"
+  "print.html"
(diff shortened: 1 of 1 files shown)


tests/src/slides/create-slide.list.sh
```diff
@@ -24,19 +24,52 @@ if [[ ! -f "${TEST_BOOK_DIR}/index.html" ]]; then
...
+SLIDES=""
+if [[ -n "${CANDIDATE_SLIDES}" ]]; then
+  # From the candidate slides, filter out:
+  # - Files that are just redirects.
+  # - Files that are in the EXCLUDE_FILES list.
+  # - Files that do not exist in the TEST_BOOK_DIR (by using ls)
+  pushd "${TEST_BOOK_DIR}" > /dev/null
+  EXCLUDE_PATTERN=$(IFS="|" ; echo "${EXCLUDE_FILES[*]}")
+  SLIDES=$(echo "${CANDIDATE_SLIDES}" | grep -v -E "${EXCLUDE_PATTERN}" \
+    | xargs ls 2>/dev/null \
```
mgeisler, line 64: I'm not sure what a call to `ls` can do here? Is it normalizing a path or so? Then something like `readlink` is probably better?

michael-kerscher: This is checking for files that exist, non-existant files are reported to /dev/null. Not the most beautiful solution probably, I give you that.

michael-kerscher:
Thank you for the honest review. Yes it increases complexity that I actually would like to remove. This needs some bigger refactor. For now it would mitigate some of these web test issues.

qwandor (approved):

tests/src/slides/create-slide.list.sh
mgeisler: Mkay, I see! 😄

mgeisler:
> Thank you for the honest review.

Always! I know I'm not telling you anything you don't know with my comments about complexity. Thanks for putting up the PR in the first place and helping fight against flaky tests.

PR #2946, opened 2025-10-10
Update "Solutions" in Rust-in-Chromium guide.

flowerhack:
Previously the guide only listed a series of abandoned patches for the solutions for this part of the guide.

However, we've moved some of these exercises into the Chromium tree as standalone build targets to make it more convenient for Rust learners.

So, let's make sure they're linked in this guide.


gribozavr:
@flowerhack LGTM, but please fix the formatting linter to unblock merging.

flowerhack:
Updated; should be fixed now

PR #2953, opened 2025-10-19
Add glossary entries for Day 1

mgeisler:
These Rust-specific entries are for concepts taught on Day 1 and could thus be useful to have in the glossary.

### DIFF
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -30,6 +30,9 @@ h1#glossary ~ ul > li:first-line {
 
 - allocate:\
   Dynamic memory allocation on [the heap](memory-management/review.md).
+- array:\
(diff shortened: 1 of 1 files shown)


gribozavr (approved):

michael-kerscher:
Regarding the error in the en build... It appears that the shell script did something wrong which I do not understand fully but also is not relevant anymore with porting the shell script in https://github.com/google/comprehensive-rust/pull/2957 to more safe rust code.

I think this behavior originates in the branch being located in this repository and not in a forked repository. I'm not sure if GITHUB_BASE_REF contains what I expect it to contain...

michael-kerscher:
Unfortunately to make this check pass, you would need to sync past the fix once it is submitted

PR #2957, opened 2025-10-28
tests: Migrate create_slide.list.sh into cargo xtask function.

michael-kerscher:
The new xtask function makes the helper code
- more readable
- more reliable due to better error checking
- be in the same place as other helper functions
- and more aligned to the skillset relevant for contributing in this repository.

The shell script grew and was not readable for everyone anymore without deeper knowledge.

mitigates #2941 in a more reliable way but does still not fully fix the root cause

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -258,7 +258,8 @@ fn create_slide_list(html_directory: PathBuf) -> Result<()> {
             .arg("--")
             // Retrieve all modified files in the src directory.
             // Pathspec syntax: https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-pathspec
-            // * can match path separators, thus matches also files in subdirectories
+            // `*` can match path separators, thus matches also files in
+            // subdirectories
             .arg("src/*.md");
         println!("> {cmd:?}");
         let output = cmd.output().context("Failed to run git diff")?;
@@ -267,7 +268,7 @@ fn create_slide_list(html_directory: PathBuf) -> Result<()> {
             .map(|line| {
                 let path = Path::new(line);
                 // We know the path starts with "src/" because of the pathspec in the
-                // git diff command. and we need it relative to the html base
+                // `git diff` command, and we need it relative to the html base
                 // directory
                 let stripped_path = path.strip_prefix("src").unwrap();
                 let mut html_path = stripped_path.to_path_buf();


michael-kerscher:
In the linked broken PR the previous shell script did something unexpected and worked with .js files and directories, which is wrong.
This is one of many reasons to abandon the shell script in favor of rust code. The new xtask code would check if files actually exist before listing them as file to check.

xtask/src/main.rs
```diff
@@ -210,6 +213,124 @@ fn run_web_tests(dir: Option<PathBuf>) -> Result<()> {
...
+    // - all html files in the provided directory otherwise
+    let candidate_slides: Vec<PathBuf> = if env::var("CI").is_ok() {
+        println!("CI environment detected, checking only modified slides.");
+        // GITHUB_BASE_REF is available in PRs. Default to 'main' for other CI
+        // contexts.
+        let base_ref = env::var("GITHUB_BASE_REF").unwrap_or("main".to_string());
+        let mut cmd = Command::new("git");
+        cmd.arg("diff")
+            .arg("--name-only")
+            .arg(format!("{}...", base_ref))
```
michael-kerscher, line 257: this is one of the things I'm not 100% sure if this would create issues. This git diff should provide the list of filenames (`--name-only`) that was changed by this PR (over all commits). But I'm not sure about GITHUB_BASE_REF in various situations. I might need to use a more reliable way to do this

michael-kerscher: GITHUB_BASE_REF might only be set in forked repositories (but not sure if this is the case), which would explain why I did not see issues with the previous shell script

xtask/src/main.rs
```diff
@@ -210,6 +213,124 @@ fn run_web_tests(dir: Option<PathBuf>) -> Result<()> {
...
+            .arg("--")
+            // Retrieve all modified files in the src directory.
+            // Pathspec syntax: https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-pathspec
+            // * can match path separators, thus matches also files in subdirectories
```
gribozavr:
```suggestion
            // `*` can match path separators, thus matches also files in subdirectories
```

xtask/src/main.rs
```diff
@@ -210,6 +213,124 @@ fn run_web_tests(dir: Option<PathBuf>) -> Result<()> {
...
+            .arg("src/*.md");
+        println!("> {cmd:?}");
+        let output = cmd.output().context("Failed to run git diff")?;
+        String::from_utf8(output.stdout)?
+            .lines()
+            .map(|line| {
+                let path = Path::new(line);
+                // We know the path starts with "src/" because of the pathspec in the
+                // git diff command. and we need it relative to the html base
```
gribozavr:
```suggestion
                // `git diff` command, and we need it relative to the html base
```

gribozavr (approved):

PR #2962, opened 2025-10-29
Start a small playbook on how to deal with failing webtests.

michael-kerscher:
This is a starting point on how to deal with failures in the web-test checks. They are supposed to warn for legitimate issues but if they fail, they can be overridden.

The web-tests will be separated into its own job to better differentiate if issues in building the English translation or web-testing the English translation are happening

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -17,7 +17,32 @@ when `npm start` or `npm test` is executed.
 > **Tip:** Use `cargo xtask web-tests` to run the tests in this directory from
 > anywhere in the repository.
 
-For local testing and quick iterations it is possible to use `mdbook serve`
+For local testing and quick iterations it is possible to use `cargo xtask serve`
 which creates a small HTTP server on port 3000 by default. There is a special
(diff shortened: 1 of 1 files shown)


tests/README.md
```diff
@@ -17,7 +17,32 @@ when `npm start` or `npm test` is executed.
...
+
+When you see tests failing they should already indicate what checks broke for
+specific pages.
+
+### Legitimate warnings
+
+You might e.g. need to reduce the length of an overlong page (or get an
+exemption) or updating some mdbook infrastructure had a breaking change that
+breaks functionality. These issues need to be fixed before this change is
+merged.
```
gribozavr, line 35:
```suggestion
Determine if the warning is flagging a real problem and respond accordingly.

You might need, for example, to reduce the length of an overlong page (or get an
exemption). If you're updating mdbook infrastructure, a pre-merge check might be flagging a breaking change. These issues need to be fixed before this change is
merged.
```

PR #2964, opened 2025-10-31
Rework lifetimes section

randomPoison:
This PR attempts to rework the lifetimes section to better contextualize how lifetimes are used and to try to lead through a series of examples more naturally. I think the current version of the lifetimes section falls short in a handful of ways that make it hard to teach in class. For a long time I've been experimenting with different ways to discuss the topic, and in my recent classes I think I've finally hit on an approach that works. The changes in this PR capture the approach I've been using.

The core idea is to contextualize lifetime annotations by looking at what happens when we return a reference from a function. We start with a simple example that doesn't return a reference to show that the borrow ends when the function returns. Then we look at a function that returns a reference and show that it extends the borrow that was passed into the function. With that context, we then look at cases where we need lifetime annotations, specifically examples where there are two ref arguments and a returned ref. We look at both possible cases: A function where either of the argument refs may be returned, and a case where only one will be returned. In both cases we show how we use lifetime annotations to tell the compiler what the returned ref is borrowing.

I haven't had a chance to use these slides in class yet so I'm putting this up as a draft PR while I continue to iterate on it so that other folks can start giving feedback.

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -198,6 +198,7 @@ use-boolean-and = true
 "generics/trait-objects.html" = "../traits/trait-objects.html"
 "hello-world/basic-syntax/functions-interlude.html" = "../../control-flow-basics/functions.html"
 "hello-world/hello-world.html" = "../types-and-values/hello-world.html"
+"lifetimes/lifetime-annotations.html" = "../lifetimes.html"
 "memory-management/manual.html" = "../memory-management/approaches.html"
 "memory-management/rust.html" = "../memory-management/ownership.html"
 "memory-management/scope-based.html" = "../memory-management/approaches.html"


src/lifetimes/returning-borrows.md
```diff
@@ -0,0 +1,38 @@
...
+fn identity(x: &i32) -> &i32 {
+    x
+}
+
+fn main() {
+    let mut x = 123;
+
+    let out = identity(&x);
+
+    // x = 5; // `x` is still borrowed!
```
gribozavr:
```suggestion
    // x = 5; // 🛠️❌ `x` is still borrowed!
```

In new content we started using the emoji sequence 🛠️❌ to mark a line that breaks the build. WDYT about using it here?

https://google.github.io/comprehensive-rust/idiomatic/leveraging-the-type-system/newtype-pattern.html

src/lifetimes/multiple-borrows.md
```diff
@@ -0,0 +1,50 @@
...
+}
+
+fn main() {
+    let mut a = 5;
+    let mut b = 10;
+
+    let r = multiple(&a, &b);
+
+    // Which one is borrowed? Should either of these
+    // be allowed?
```
gribozavr:
```suggestion
    // Which one is still borrowed?
    // Should either mutation be allowed?
```

src/lifetimes/multiple-borrows.md
```diff
@@ -0,0 +1,50 @@
...
+  Which one should be extended? Do we need to see the body of `multiple` to
+  figure this out?
+
+- When borrow checking, the compiler doesn't look at the body of `multiple` to
+  reason about the borrows flowing out, instead it looks only at the signature
+  of the function for borrow analysis.
+
+- In this case there is not enough information to determine if `a` or `b` will
+  be borrowed by the returned reference. Show students the compiler errors and
+  introduce the lifetime syntax.
```
gribozavr: Could you add the code sample (here, within `<details>`) that the instructor should end up with at the end of the explanation and editing?

src/lifetimes/borrow-both.md
```diff
@@ -0,0 +1,46 @@
...
+}
+
+fn main() {
+    let mut a = 5;
+    let mut b = 10;
+
+    let r = pick(true, &a, &b);
+
+    // Which one is borrowed? Should either of these
+    // be allowed?
```
gribozavr: Please update the comment here to match the other slide if you accept the suggested edit there.

src/lifetimes/borrow-both.md
```diff
@@ -0,0 +1,46 @@
...
+  `c`, which means we can't know at compile time which one will be returned.
+
+- To express this to the compiler, we use the same lifetime for both `a` and
+  `b`, along with the return type. This means that the returned reference will
+  borrow BOTH `a` and `b`!
+
+- Uncomment both of the commented lines and show that `r` is borrowing both `a`
+  and `b`, even though at runtime it will only point to one of them.
+
+- Change the first argument to `pick` to show that the value you pass in doesn't
```
gribozavr:
```suggestion
- Change the first argument to `pick` to show that the flag value doesn't
```

"value you pass in" is a bit ambiguous because there are three of them.

src/lifetimes/borrow-one.md
```diff
@@ -0,0 +1,77 @@
+---
+minutes: 5
+---
+
+# Borrow One
+
+In this example `find_nearest` takes in multiple borrows but will only return
```
gribozavr:
```suggestion
In this example `find_nearest` takes in multiple borrows but returns only
```

Simplifying

src/lifetimes/borrow-both.md
randomPoison: I went with a slightly different wording that emphasizes that whether `a` or `b` is returned doesn't impact borrow analysis.

randomPoison:
@gribozavr I think I have addressed all of your feedback.

mgeisler:
Looks really good, I wish I would have had those examples when teaching the class two weeks ago :smile:

PR #3018, opened 2026-01-09
Fix typo in unsafe deep dive

ia0:
The implementation is safe.

### DIFF
--- a/src/unsafe-deep-dive/rules-of-the-game/copying-memory/safe.md
+++ b/src/unsafe-deep-dive/rules-of-the-game/copying-memory/safe.md
@@ -23,7 +23,7 @@ fn main() {
 
 <details>
 
-“The implementation only uses unsafe Rust
+“The implementation only uses safe Rust
 
 What can we learn from this?
 


src/unsafe-deep-dive/rules-of-the-game/copying-memory/safe.md
```diff
@@ -23,7 +23,7 @@ fn main() {
 
 <details>
 
-“The implementation only uses unsafe Rust
+“The implementation only uses safe Rust
```
mgeisler, line 26: @timClicks, what is up with all the `“` found in these new slides? I don't think we've used this style anywhere else in the course and as such, it looks inconsistent? I might have missed it.

If you intend the instructor to read these notes out aloud, then notice that this [conflicts with the style guide](https://github.com/google/comprehensive-rust/blob/main/STYLE.md#:~:text=Avoid%20long%2C%20narrative%20paragraphs%20meant%20to%20be%20read%20aloud).

mgeisler: Thanks @gribozavr for fixing a bunch of these. However, the style is still different than the rest of the course and I hope it can be aligned better.

At a high level, the notes serve as cue cards for the instructor: talk about A  B  and C. So they're not a script to be followed directly.

The notes are also the tool that enables self-learners to use the material.

PR #3019, opened 2026-01-09
Fix leftover in encapsulated-unsafe

ia0:
This seems to be a leftover of a copy/paste from the safe version. This is redundant with the unsafe implementation.


mgeisler (approved):
Thanks @ia0 , let's get this merged. @timClicks can let us know anytime if there was some special meaning behind these lines 😅

PR #3020, opened 2026-01-09
Fix missing text in exposed-unsafe

ia0:
I'm guessing the first characters of that sentence got somehow deleted. I filled them with what I believe is the most probable option.


mgeisler:
Thanks @ia0!

PR #3022, opened 2026-01-11
docs: fix typos and grammar in unsafe deep dive pages

gribozavr:
Fix grammar, spelling, and capitalization across the Unsafe Rust deep
dive chapters. Ran `dprint fmt`.

### DIFF
--- a/src/unsafe-deep-dive/ffi/abs.md
+++ b/src/unsafe-deep-dive/ffi/abs.md
@@ -32,7 +32,7 @@ unsafe extern "C" {
 }
 ```
 
-Explain that many POSIX functions are available in Rust because cargo links
+Explain that many POSIX functions are available in Rust because Cargo links
 against the C standard library (libc) by default, which brings its symbols into
 the program’s scope.
 
--- a/src/unsafe-deep-dive/pinning/self-referential-buffer/rust-pin.md
+++ b/src/unsafe-deep-dive/pinning/self-referential-buffer/rust-pin.md
@@ -2,7 +2,7 @@
 minutes: 10
 ---
 
-# With Pin&lt;Ptr&gt;
+# With `Pin<Ptr>`
 
 Pinning allows Rust programmers to create a type which is much more similar to
 C++ classes.
--- a/src/unsafe-deep-dive/safety-preconditions/determining.md
+++ b/src/unsafe-deep-dive/safety-preconditions/determining.md
@@ -38,7 +38,7 @@ Click the "convertible to a reference" hyperlink to the "Pointer to reference
 conversion"
 
 Track down the rules for converting a pointer to a reference, i.e., whether it
-is "_dereferenceable_".
+is "dereferenceable".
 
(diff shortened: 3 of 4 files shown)


src/unsafe-deep-dive/ffi/abs.md
```diff
@@ -32,9 +32,9 @@ unsafe extern "C" {
 }
 ```
 
-Explain that many POSIX functions are available Rust because cargo links against
-the C standard library (libc) by default, which makes its symbols into the
-program’s scope.
+Explain that many POSIX functions are available in Rust because cargo links
```
mgeisler:
```suggestion
Explain that many POSIX functions are available in Rust because Cargo links
```

src/unsafe-deep-dive/pinning/self-referential-buffer/rust-pin.md
```diff
@@ -5,7 +5,7 @@ minutes: 10
 # With Pin&lt;Ptr&gt;
```
mgeisler: I forget, is there some reason to use HTML here instead of just regular Markdown?

```suggestion
# With `Pin<Ptr>`
```

src/unsafe-deep-dive/safety-preconditions/determining.md
```diff
@@ -37,8 +37,8 @@ Highlight Safety section.
 Click the "convertible to a reference" hyperlink to the "Pointer to reference
 conversion"
 
-Track down the rules for converting a pointer to a reference, aka is
-"_deferencerable_".
+Track down the rules for converting a pointer to a reference, i.e., whether it
+is "_dereferenceable_".
```
mgeisler: Scare quotes and italic seems a bit much to me 😅 I would go with just the quotes:

```suggestion
is "dereferenceable".
```

src/unsafe-deep-dive/setup.md
```diff
@@ -39,7 +39,7 @@ $ cargo serve # then open http://127.0.0.1:3000/ in a browser
 <details>
 
 Ask everyone to confirm that everyone is able to execute `rustc` with a version
-older that 1.87.
+older than 1.87.
```
mgeisler: I'm guessing this should be _newer_ than 1.87? We normally don't require people to use older versions. Cc @timClicks.


```suggestion
newer than 1.87.
```

src/unsafe-deep-dive/setup.md
```diff
@@ -39,7 +39,7 @@ $ cargo serve # then open http://127.0.0.1:3000/ in a browser
...
 
 For those people who do not, tell them that we'll resolve that in the break.
```
mgeisler, line 44: These notes should be improved my making them actionable:

- please don't mske a script (the instructor won't have time to read it and should certainly not read from the slides)
- add the necessary information to _expand_ on the slides. Don't leave the actual help out, tell people who read this at home to run `rustup` in such-and-such a way.

mgeisler: This is really for @timClicks to fix, not you @gribozavr.

PR #3026, opened 2026-01-29
Avoid unnecessary borrow in exercise.rs

mgeisler:
We haven't talked about copy types yet, but the rows are infact `Copy`. So we can avoid the borrow syntax, which hasn't yet been introduced.

Also inline the variable in the format string like elsewhere in the course.


hurryabit (approved):
That does make things simpler.

PR #3028, opened 2026-01-31
Remove reqwest's rustls-tls feature in link checker exercise

hurryabit:
As of version 0.13 of `reqwest`, the `rustls-tls` feature does not exist anymore. The feature has been renamed to `rustls` and is turned on by default. See https://github.com/seanmonstar/reqwest/releases/tag/v0.13.0 for details.

Thus, we remove the feature from the exercise to avoid students running into issues.


mgeisler (approved):
Thanks!

PR #3029, opened 2026-01-31
style: update style guide for precise language

mgeisler:
Add a section on precision to the style guide, discouraging the use of
'weasel words' (e.g., 'likely', 'probably', 'maybe', 'usually') in
favor of authoritative and definite statements.

This establishes the standard for a series of follow-up PRs that
clean up the existing course material.

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -191,6 +191,16 @@ throughout the course. When editing course content, use the glossary to anchor
 concepts and ensure consistency in terminology. Terms should be defined and used
 consistently with their glossary entries.
 
+### Precision
+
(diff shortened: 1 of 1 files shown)


mgeisler:
> Makes sense. Thanks for taking the initiative.

Happy to! I had Gemini generate a description for the solution to the Fibonacci exercise. It wrote something like "in release mode, numbers will wrap around, likely leading to incorrect results for this algorithm" and this annoyed me 😄 It not "likely", it's given for a simple Fibonacci algorithm.

A few iterations later, and we have PRs which try to strengthen the language in the whole course.

PR #3033, opened 2026-01-31
concurrency: remove weasel words

mgeisler:
This PR removes imprecise language ('weasel words') to align with the new precision guidelines in #3029.

---
*Note: This PR was generated using Gemini. Please review the changes accordingly. If the new style is not desired for this section, feel free to close this PR.*

### DIFF
--- a/src/concurrency/welcome.md
+++ b/src/concurrency/welcome.md
@@ -10,8 +10,8 @@ Rust has full support for concurrency using OS threads with mutexes and
 channels.
 
 The Rust type system plays an important role in making many concurrency bugs
-compile time bugs. This is known as _fearless concurrency_ since you can rely on
-the compiler to ensure correctness at runtime.
+compile time errors. This idea is known as _fearless concurrency_ since you can
+rely on the compiler to ensure correctness at runtime.
 
 ## Schedule
 


src/concurrency/async/runtimes/tokio.md
```diff
@@ -37,9 +37,9 @@ async fn main() {
 
 **Further exploration:**
 
-- Why does `count_to` not (usually) get to 10? This is an example of async
-  cancellation. `tokio::spawn` returns a handle which can be awaited to wait
-  until it finishes.
+- Why does `count_to` not get to 10? This is an example of async cancellation.
```
hurryabit, line 40: Shouldn't there be only one L in "cancelation" since we use American English here?

src/concurrency/welcome.md
```diff
@@ -10,8 +10,8 @@ Rust has full support for concurrency using OS threads with mutexes and
 channels.
 
 The Rust type system plays an important role in making many concurrency bugs
-compile time bugs. This is often referred to as _fearless concurrency_ since you
-can rely on the compiler to ensure correctness at runtime.
+compile time bugs. This is known as _fearless concurrency_ since you can rely on
```
hurryabit: I feel like this this whole section isn't very precise. They main thing the compiler rules out are data races. And you can't use mutexes/locks the wrong way. That being said, let's improve in small steps:

This should probably be "compile time errors" and not "compile time bugs". I would also start the next sentence with "this idea" or "this principle" instead of just "this".

src/concurrency/welcome.md
```diff
@@ -10,8 +10,8 @@ Rust has full support for concurrency using OS threads with mutexes and
...
+the compiler to ensure correctness at runtime.
```
mgeisler:
```suggestion
compile time errors. This idea is known as _fearless concurrency_ since you can
rely on the compiler to ensure correctness at runtime.
```

src/concurrency/welcome.md
mgeisler: Thanks, happy to rephrase like you suggest!

src/concurrency/async/runtimes/tokio.md
mgeisler: I had no idea that this could be spelled with one L 😄 When I look it up, it seems that this is actually the recommended spelling: [Merriam-Webster](https://www.merriam-webster.com/dictionary/cancel#:~:text=Cancellation%20is%20the%20usual%20spelling%20everywhere%2C%20though%20cancelation%20is%20also%20sometimes%20used):

> The forms of cancel in American English are typically canceled and canceling; in British English they are cancelled and cancelling. Cancellation is the usual spelling everywhere, though cancelation is also sometimes used.

hurryabit: Wow, the (American) English language is so inconsistent. Let's leave it as is then.

mgeisler: Yeah, thanks for the review!

PR #3044, opened 2026-01-31
unsafe-deep-dive: remove weasel words

mgeisler:
This PR removes imprecise language ('weasel words') to align with the new precision guidelines in #3029.

---
*Note: This PR was generated using Gemini. Please review the changes accordingly. If the new style is not desired for this section, feel free to close this PR.*

### DIFF
--- a/src/unsafe-deep-dive/case-studies/intrusive-linked-list.md
+++ b/src/unsafe-deep-dive/case-studies/intrusive-linked-list.md
@@ -3,7 +3,7 @@
 > Current as of tokio v1.48.0
 
 The Tokio project maintains an [intrusive linked list implementation][ill] that
-demonstrates many use cases of `unsafe` and a number of types and traits from
+demonstrates several use cases of `unsafe` and a number of types and traits from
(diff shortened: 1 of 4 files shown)


gribozavr:
Sorry @mgeisler, I'm not sure this is an improvement. For example, "many" is replaced with "several", why is that better?

mgeisler:
> Sorry @mgeisler, I'm not sure this is an improvement. For example, "many" is replaced with "several", why is that better?

As you know, I'm not a native speaker... 😄 the reasoning seems to be that "several" is considered a little more precise than "many":

> [_many_](https://www.merriam-webster.com/dictionary/many): consisting of or amounting to a large but indefinite number
>
> [_several_](https://www.merriam-webster.com/dictionary/several#dictionary-entry-2): an indefinite number more than two and fewer than many

So I guess we should use "many" when the number of cases/examples/... is so many that nobody will want to count them? And use "several" when there are so few cases/examples/... shown that one can count them on roughly two hands?

Now, I have no problem with us not wanting to make such a subtle change. I would not have come up with this myself from looking at the English text!

mgeisler:
Let me close this again and the improvements are not obvious.

PR #3047, opened 2026-01-31
borrowing: add explanatory commentary to solution

mgeisler:


hurryabit (changes requested):
This seems out of date. The borrowing exercise is no longer about health reports but rather about wizards.

mgeisler:
Yes, you're quite right! Let me close this one and then I can make another one later.

PR #3063, opened 2026-01-31
tuples-and-arrays: add explanatory commentary to solution

mgeisler:

### DIFF
--- a/src/tuples-and-arrays/solution.md
+++ b/src/tuples-and-arrays/solution.md
@@ -3,3 +3,27 @@
 ```rust,editable
 {{#include exercise.rs:solution}}
 ```
+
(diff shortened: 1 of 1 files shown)


src/tuples-and-arrays/solution.md
```diff
@@ -3,3 +3,27 @@
 ```rust,editable
 {{#include exercise.rs:solution}}
 ```
+
+- **Array Types:** The type `[[i32; 3]; 3]` represents an array of size 3, where
+  each element is itself an array of 3 `i32`s. This is how multi-dimensional
+  arrays are typically represented in Rust.
```
gribozavr, line 9:
```suggestion
- **Array Types:** The type `[[i32; 3]; 3]` is a multi-dimensional array. It is an array of size 3, where
  each element is itself an array of 3 `i32`s.
```

PR #3068, opened 2026-01-31
content: add daily recaps

mgeisler:
This adds recap slides to the start of Day 2, 3, and 4 to reinforce learning from the previous days.


mgeisler:
> Nice! I think this is far more useful as a focusing moment than setting an agenda for the day.

My thoughts exactly – I wanted to give people a feeling of accomplishment by reminding them that we've covered large parts of the language already by Day 2.

PR #3077, opened 2026-02-01
pattern-matching: Replace `Foo` with more realistic struct

mgeisler:
The old example used a `Foo` struct with some nested integer fields. We now use a `Move` struct which could easily appear in a real computer game.

This was the last use of code from Rust by Example, so the whole third-party directory can now be removed.


djmitche (approved):
That's much more natural -- nice!

PR #3079, opened 2026-02-02
Fix the pseudo-definition reference to a MutexGuard

gribozavr:
This came up during a class. An audience member said that this snippet looks like a type definition, but actually MutexGuard also has a lifetime parameter.

### DIFF
--- a/src/concurrency/send-sync/examples.md
+++ b/src/concurrency/send-sync/examples.md
@@ -33,9 +33,10 @@ Typically because of interior mutability:
 These types are safe to access (via shared references) from multiple threads,
 but they cannot be moved to another thread:
 
-- `MutexGuard<T>`: Uses OS level primitives which must be deallocated on
-the thread which created them. However, an already-locked mutex can have its
-guarded variable read by any thread with which the guard is shared (unless `T` itself is `!Sync`).
+- `MutexGuard<T>`: Uses OS level primitives which must be deallocated on the
+  thread which created them. However, an already-locked mutex can have its
+  guarded variable read by any thread with which the guard is shared (unless `T`
+  itself is `!Sync`).
 
 ## `!Send + !Sync`
 


src/concurrency/send-sync/examples.md
```diff
@@ -33,7 +33,7 @@ Typically because of interior mutability:
 These types are safe to access (via shared references) from multiple threads,
 but they cannot be moved to another thread:
 
-- `MutexGuard<T: Sync>`: Uses OS level primitives which must be deallocated on
+- `MutexGuard<'a, T: Sync>`: Uses OS level primitives which must be deallocated on
   the thread which created them. However, an already-locked mutex can have its
   guarded variable read by any thread with which the guard is shared.
```
sam-mccall:
```suggestion
- `MutexGuard<T>`: Uses OS level primitives which must be deallocated on
the thread which created them. However, an already-locked mutex can have its
guarded variable read by any thread with which the guard is shared (unless `T` itself is `!Sync`).
```

sam-mccall: The lifetime seems like noise here. If we're assuming that `MutexGuard<T: Sync>` implies we're writing a definition, then `MutexGuard<'a, T: Sync>` is misleading as it implies you can only have mutexguards of sync types.

I like the current text best but if we don't want to use `: Sync` I think we should replace it with english text.

src/concurrency/send-sync/examples.md
```diff
@@ -33,9 +33,9 @@ Typically because of interior mutability:
 These types are safe to access (via shared references) from multiple threads,
 but they cannot be moved to another thread:
 
-- `MutexGuard<T: Sync>`: Uses OS level primitives which must be deallocated on
-  the thread which created them. However, an already-locked mutex can have its
-  guarded variable read by any thread with which the guard is shared.
+- `MutexGuard<T>`: Uses OS level primitives which must be deallocated on
+the thread which created them. However, an already-locked mutex can have its
+guarded variable read by any thread with which the guard is shared (unless `T` itself is `!Sync`).
```
gribozavr:
```suggestion
- `MutexGuard<T>`: Uses OS level primitives which must be deallocated on the
  thread which created them. However, an already-locked mutex can have its
  guarded variable read by any thread with which the guard is shared (unless `T`
  itself is `!Sync`).
```

PR #3097, opened 2026-02-14
Add license headers to Markdown files

gribozavr:
This commit adds the `SPDX-License-Identifier: CC-BY-4.0` header to all Markdown (`.md`) files in the `src/` directory.

It also adds `SPDX-License-Identifier: Apache-2.0` headers to Rust code blocks within these Markdown files. The headers in code blocks are prefixed with `#` to hide them in the rendered book.

The year of the copyright notice is determined from the first commit that added the file to the repository.


src/android/interoperability/cpp/shared-enums.md
```diff
@@ -1,12 +1,23 @@
...
+-->
+
 # Shared Enums
 
 ```rust,ignore
+# // Copyright 2023 Google LLC
+# // SPDX-License-Identifier: Apache-2.0
+#
 {{#include ../../../../third_party/cxx/book/snippets.rs:shared_enums_bridge}}
 ```
```
mgeisler, line 13: I'm sure a Google lawyer told you to add these 😄 but did the lawyer also consider that there can be a minimum size for a published work to be copyrightable?

Even the FSF does not consider it necessary to get a copyright assignment for changes that touch only a few lines (they say [less than 15 or so](https://www.gnu.org/prep/maintain/html_node/Legally-Significant.html)).

mgeisler (approved):
I think these changes are unnecessary as the license is stated clearly in the README and in larger files. However, I assume a friendly Google lawyer asked for this, so here we go.

mgeisler:
This should be reverted as it caused #3101, which degrades the experience of teaching the course. I'll be teaching a new class tomorrow, so I'm removing this from my own branch until we can fix it here.

PR #3108, opened 2026-02-24
Make string construction explicit in dyn slide

hurryabit:
Let's make it crystal clear that we're taking a reference to a `String` instead of hiding it behind a call to the overloaded `to_owned` method.


gribozavr (approved):

mgeisler:
Thank you, I much prefer the `String::from` constructor as well 🙂

PR #3125, opened 2026-03-09
Add 404 error page and update book configuration

### LINKED ISSUE #2022: 404 page breaks translations
When you call `mdbook build`, you [automatically get a `404.html` page](https://rust-lang.github.io/mdBook/continuous-integration.html#404-handling) in the root of the output directory.

Since we run `mdbook build` for each translation, we end up with a structure like

```
/404.html
/da/404.html
/de/404.html
...
```

The root `404.html` page is automatically used by [GitHub Pages](https://docs.github.com/en/pages/getting-started-with-github-pages/creating-a-custom-404-page-for-your-github-pages-site) when a page cannot be found. Note that the URL doesn't change: the content of `404.html` is simply returned to the browser.

Now, this doesn't work well with our translations: you end up with the English `404.html` regardless of which translation you're currently viewing. We should fix this so that people get the correct version. The [`output.html.input-404`](https://rust-lang.github.io/mdBook/format/configuration/renderers.html#html-renderer-options) setting is for that.

Simply setting this is not enough: GitHub Pages doesn't know anything about the translation sub-directories, so it will continue to show people the root `404.html` page. I think we can fix this by injecting a bit of JavaScript into the `404.md` file which `mdbook build` uses to generate the `404.html` page.

yl275:
This PR adds a small JavaScript snippet to `404.md`, so that when a user visits a broken link under a translated version of the book (for example `/es/does-not-exist.html`), they are redirected to the corresponding language-specific 404 page (`/es/404.html`, `/zh-CN/404.html`, etc.) instead of the English 404.

I tested this, but it seems that `404.md` is not going through the translation pipeline. 
Am I on the right track with this approach?

fixes #2022


mgeisler:
> I tested this, but it seems that `404.md` is not going through the translation pipeline. 

I haven't tested this myself, but what you write makes sense: we extract text from the book "chapters", which are the files mentioned in `SUMMARY.md`. So the 404 is not there, so I would not expect the text to be extracted.

To improve upon this, I think we would need to add code in [mdbook-i18n-helpers](https://github.com/google/mdbook-i18n-helpers/blob/main/i18n-helpers/src/xgettext.rs#L117) to detect the `input-404` config setting (I hope we have access to the `book.toml` data there?) and extract the data manually.

We would then also need to test if the page can be translated later...

> Am I on the right track with this approach?

Yes, I think this looks like a great approach!

I don't have a lot of time to really test this out right now, but it looks like a nice approach to me.

gribozavr:
@yl275 Please sign the CLA.

PR #3157, opened 2026-04-10
Idiomatic: Trim slides from naming conventions section

randomPoison:
I think the naming conventions section of Idiomatic is a bit longer than it needs to be. I think it's useful to point out common conventions, but the section currently is pretty long and is imo not where we want to be spending a lot of time. I'm suggesting trimming the following slides:

- Get and Push - These are specific to the collection types, and are not generally naming conventions that students would be using when writing their own code. Pointing these out can be good for helping students orient themselves when looking at the collection types in std, but I think we should focus more on naming conventions that students would use when writing their own code.
- With in normal use - I don't think we need a slide to point out when "with" isn't used as part of a convention, I think we can trust students to figure that out on their own.
- Into inner - This is a special case of "into", I don't think we need a separate slide for it.
- Unchecked and Raw parts - I don't think it's worth covering unsafe-specific conventions in Idiomatic, since we don't really focus on unsafe in this course.

I also moved the "by" slide to the end of the section since it was in the middle of a bunch of slides that are all related to type conversion. I think that'll help things flow a bit better.


mgeisler:
Yeah, I tend to agree. @gribozavr, I'll let you decide since you might have worked on an internal outline for this with other thoughts?

For me, it's more problematic that there aren't any references to [Rust Naming Conventions](https://rust-lang.github.io/api-guidelines/naming.html). We should ground our slides in references when at all possible, and that is still the most official reference I know of when it comes to naming in Rust.

mgeisler:
> * With in normal use - I don't think we need a slide to point out when "with" isn't used as part of a convention, I think we can trust students to figure that out on their own.

Yes, very much this! Less is sometimes more :smile:

randomPoison:
> For me, it's more problematic that there aren't any references to [Rust Naming Conventions](https://rust-lang.github.io/api-guidelines/naming.html).

Yeah that's a good point, it could be nice to have a slide that talks about the general naming style conventions (i.e. snake_case vs CamelCase). That doc doesn't really cover many other naming conventions, though.

randomPoison:
> We should ground our slides in references when at all possible

I do think the naming conventions section is grounded in references, it just references specific examples from std rather than there being a specific document that it can cite.

mgeisler (approved):
Let's merge this —— if people disagree later, we can revive the pages again.

Issue #3155, opened 2026-04-03
Code snippets without `main` pass tests but fail to compile in the slide


randomPoison:
The "PartialEq and Eq" slide in the idiomatic course [has a code snippet](https://github.com/google/comprehensive-rust/blob/b9e63dd1229dbdc847820ac21f12c65d43f0e1f9/src/idiomatic/foundations-api-design/predictable-api/common-traits/partialeq-eq.md?plain=1#L18-L41) that doesn't compile or run, but is not marked `ignore` or `compile_fail`, and `mdbook test` does not fail.

The issue is that the code snippet is missing a `main` function, and so fails to compile/run when trying to run it in the slide. But when running tests, it looks like a `main` function wrapper is implicitly added if the snippet doesn't have its own `main`. This means that the tests pass, even though the code doesn't work when trying to teach the class.

I think the implicit main behavior is useful and can allow us to write more concise code examples, but we either need to add that behavior when running the code snippets in the slide, or we need to remove this behavior from the tests so that tests can catch when code snippets don't work.

Issue #3101, opened 2026-02-19
Copyright notice in code snippets


hurryabit:
Most code snippets in the course show a copyright header of the form

```rust
// Copyright 20xx Google LLC
// SPDX-License-Identifier: Apache-2.0
```

This is a problem since it's visual noise that takes up valuable screen space.

Some code blocks don't show this header though. My preliminary investigation suggests that blocks opened by only

<code>```rust</code>

don't show the copyright header whereas those opened by more complex fences, such as

<code>```rust,editable</code>

do show it.

mgeisler:
Yes, I can reproduce this. The problem is that mdbook does _not_ support hiding code when the text areas are editable. It knows to strip away the leading `#`, but that is it.

Basically, hiding lines in the output is not compatible with also giving the user access to edit the code block.

mgeisler:
The problematic PR is #3097.

PR #3159, opened 2026-04-10
Idiomatic: Fix various broken code examples

randomPoison:
Many of the code snippets in Idiomatic are broken in various ways that make them either not run in the slides or not show the error the slide is intended to illustrate. This PR fixes a number of common issues:

- **Invalid syntax**, e.g. `let` inside a struct definition.
- **Missing `main` fn**. Not all examples need a `main`, but several code examples had code that only compiles if it's in a function, and so was failing to compile in the slides. Note that these were not caught in tests because [there seems to be a difference in behavior between how we test the code vs how the code works in the slides](https://github.com/google/comprehensive-rust/issues/3155).
- **Missing `fn` keyword** on method definitions.

In a few places I made more opinionated changes that I'm including here because splitting them out into a separate PR would just result in merge conflicts later:

- `hash.md` - Removed the `friends` field. It wasn't necessary for demonstrating the `Hash` trait, and removing it allows the slide to be more concise.
- `newtype_pattern.md` - Tweak the exercise to show that you can't pass an inner type where a newtype wrapper is expected (inverse of what it was previously demonstrating). I think this is a slightly clearer way to show what the slide is demonstrating.
- A few minor tweaks to formatting.
- Remove a few extraneous comments.
- Add empty lines to help space things out and make them easier to read.
- Remove `pub` keyword in a few places.
- Remove the `#` prefix on lines that I think should be kept visible in the code snippet (though note that none of the hidden lines are ever hidden because the code snippets are editable, see https://github.com/google/comprehensive-rust/issues/2811).

### DIFF
--- a/src/idiomatic/foundations-api-design/predictable-api/naming-conventions/new.md
+++ b/src/idiomatic/foundations-api-design/predictable-api/naming-conventions/new.md
@@ -16,12 +16,11 @@ method name.
 # // Copyright 2025 Google LLC
 # // SPDX-License-Identifier: Apache-2.0
 #
-impl <T> Vec<T> {
-    // Creates an empty vec.
+impl<T> Vec<T> {
     fn new() -> Vec<T>;
 }
 
-impl <T> Box<T> {
+impl<T> Box<T> {
     fn new(T) -> Box<T>;
 }
 ```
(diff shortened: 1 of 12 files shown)


src/idiomatic/foundations-api-design/predictable-api/naming-conventions/new.md
```diff
@@ -16,12 +16,11 @@ method name.
 # // Copyright 2025 Google LLC
 # // SPDX-License-Identifier: Apache-2.0
 #
-impl <T> Vec<T> {
-    // Creates an empty vec.
+impl<T> Vec<T> {
```
mgeisler, line 19: Thanks for fixing the formatting (and all the other things!)

It just occurred to me that the `# ` syntax very likely is what prevents `dprint` from formatting this code block. It knows it's Rust from the info string and it knows to split on `,`, but `rustfmt` will not be happy with the `# ` lines.

Cc @gribozavr as another side-effect of #3101.

mgeisler (approved):
Thank you for bringing this up to standard.

PR #3163, opened 2026-04-13
Idiomatic: Various tweaks and cleanup to common traits section

randomPoison:
Various tweaks and cleanup to the "Implementing Common Traits" section of Idiomatic based on my first time teaching the class.

- Remove the "When to implement" notes from the slides. I found these to be unhelpful, as the answer for "when do I implement this trait" is always "when you need to". I would generally not encourage people to implement traits, even derivable ones, just because they _can_ do so. Where necessary we have speaker notes that call out more specifically when there are extra caveats to implementing a trait, which I think better covers the nuances here.
- Remove the commented-out code showing the definition of the traits. This is cruft that adds visual noise, and I'd prefer to just pull up the standard library docs to show students what the trait API looks like.
- Add `main` function with example code to a few slides that were not demonstrating the behavior of the traits.
- Fix various issues with markdown formatting in slides and speaker notes.
- `copy.md`
  - Note that `Copy` is an implicit operation, whereas `Clone` is an explicit, user-defined operation.
  - Note that `Copy` and `Clone` should be derived together.
  - Note that `Copy` types cannot impl `Drop`, and reorganize notes so that they render more clearly in the speaker notes.
- `display.md` - Remove reference to the `Error` trait and shift focus to highlight the differences between `Debug` and `Display`. I don't think mentioning `Error` is bad per se, but I think it's the wrong thing to focus on here.\
- `from-into.md`
  - Rename `ObviousImplementation` to `Wrapper` to make things more concise and easier to read.
  - Change `From` impls to be for `&str` and `i32`, to better demonstrate how `From` can support conversions from multiple unrelated types.
  - Add example code showing how `Into` is a more natural trait bound than `From`.

### DIFF
--- a/src/idiomatic/foundations-api-design/predictable-api/common-traits/copy.md
+++ b/src/idiomatic/foundations-api-design/predictable-api/common-traits/copy.md
@@ -40,7 +40,7 @@ fn main() {
 - Has the same caveat as `Clone`: If duplicating the values would break an
   invariant, the type shouldn't implement `Copy`.
 
-- Always derive `Clone` and `Copy` together! Do NOT manually implement `Clone`
+- Always derive `Clone` and `Copy` together! Do _not_ manually implement `Clone`
   when implementing `Copy`.
 
   - Copy operations do not invoke the `clone` method, so a custom `Clone` impl
--- a/src/idiomatic/foundations-api-design/predictable-api/common-traits/copy.md
+++ b/src/idiomatic/foundations-api-design/predictable-api/common-traits/copy.md
@@ -50,21 +50,21 @@ fn main() {
 
 - Cannot be implemented on types with `Drop` or non-`Copy` fields.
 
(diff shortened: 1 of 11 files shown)


src/idiomatic/foundations-api-design/predictable-api/common-traits/copy.md
```diff
@@ -34,31 +29,42 @@ fn main() {
...
+  an implicit, bitwise copy.
+
+- Should generally only be implemented on "plain data" types that should act
+  like primitive values. For example, primitive numeric types for a linear
+  algebra library.
+
+- Has the same caveat as `Clone`: If duplicating the values would break an
+  invariant, the type shouldn't implement `Copy`.
+
+- Always derive `Clone` and `Copy` together! Do NOT manually implement `Clone`
```
mgeisler:
```suggestion
- Always derive `Clone` and `Copy` together! Do _not_ manually implement `Clone`
```

mgeisler (approved):

mgeisler (approved):

PR #3164, opened 2026-04-13
Idiomatic: Various tweaks and cleanup to naming conventions section

randomPoison:
Tweaks and feedback to the "Naming Conventions" section of Idiomatic based on my experience teaching the class, mostly focusing on making the slides more concise. Related to https://github.com/google/comprehensive-rust/pull/3157, which also attempts to trim the section by removing some slides that I don't think are necessary. As noted in that PR, I think the section is overly long and would benefit from being a quicker tour of naming conventions, without needing to spend a lot of time on any of the slides.


mgeisler (approved):
Thanks for condensing this!

mgeisler (approved):

PR #3165, opened 2026-04-13
Idiomatic: Misc. minor tweaks to the type system section

randomPoison:
A handful of minor fixes to the type systems section of Idiomatic. Mostly just fixing wording or markdown formatting.

Also rename the "Parse, Don't Validate" slide to "Enforce Invariants at Construction". The naming of this slide seems odd to me when the example code is validating inputs, not doing any parsing. I think what the slide demonstrates is fine, but makes more sense framed as "enforce invariants at construction time, then you can rely on those invariants later".


mgeisler (approved):
Thanks!

mgeisler (approved):

PR #3167, opened 2026-04-13
Simplify and flesh out extension trait speaker notes

randomPoison:
- Remove some redundant notes at the top that just repeat information in the other bullet points.
- Add a note pointing out that you need a trait to use with generics and `dyn`.

### DIFF
--- a/src/idiomatic/leveraging-the-type-system/extension-traits/should-i-define-an-extension-trait.md
+++ b/src/idiomatic/leveraging-the-type-system/extension-traits/should-i-define-an-extension-trait.md
@@ -36,14 +36,6 @@ The main advantage of extension traits is **ease of discovery**.
 
 <details>
 
-- Extension methods can be easier to discover than free functions. Language
-  servers (e.g., `rust-analyzer`) will suggest them if you type `.` after an
-  instance of the foreign type.
-
-- However, a bespoke extension trait might be overkill for a single method. Both
-  approaches require an additional import, and the familiar method syntax may
-  not justify the boilerplate of a full trait definition.
-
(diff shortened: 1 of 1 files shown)


src/idiomatic/leveraging-the-type-system/extension-traits/should-i-define-an-extension-trait.md
```diff
@@ -36,14 +36,6 @@ The main advantage of extension traits is **ease of discovery**.
 
 <details>
 
-- Extension methods can be easier to discover than free functions. Language
-  servers (e.g., `rust-analyzer`) will suggest them if you type `.` after an
-  instance of the foreign type.
```
mgeisler, line 41: Thanks for removing this! I super dislike this argument myself: I find it very hard to "discover" mysterious methods that appear on well-known types in a nearly implicit fashion.

Free functions are much clearer: they are imported by name (barring any glob imports) and used by name. It doesn't get more clear than this when reading the code (which is what we typically optimize for, writing the code while having a full IDE environment powered up is just a small part of the life cycle).

mgeisler (approved):

src/idiomatic/leveraging-the-type-system/extension-traits/should-i-define-an-extension-trait.md
randomPoison: Yeah agreed, I'd generally suggest starting with free functions and only reaching for an extension trait if you need some functionality that a trait provides (polymorphism or method call syntax). Discoverability is still listed as a motivation in this slide, but I don't think I'm going to emphasize that much while teaching.

PR #3168, opened 2026-04-13
Idiomatic: Rework Option-in-Drop example to emphasize move semantics

randomPoison:
Rework the example in the "Drop: Option" slide to emphasize how `Option` is used to achieve move operations in `Drop` code. The current version of the slide emphasizes the wrong thing in my opinion, and fails to actually demonstrate the pattern of needing to perform a move operation in `drop`. I've modified `Handle` to have a `fn close(self)` instead of a `Drop` impl, that way `File::drop` needs to move out of its inner field in order to call `close`.

### DIFF
--- a/src/idiomatic/leveraging-the-type-system/raii/drop_option.md
+++ b/src/idiomatic/leveraging-the-type-system/raii/drop_option.md
@@ -21,55 +21,47 @@ impl File {
     }
 
     fn write(&mut self, data: &str) -> std::io::Result<()> {
+        // We have to go through the `Option` to get the `Handle`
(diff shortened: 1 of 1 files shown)


mgeisler (approved):

randomPoison:
@mgeisler can you hit the merge button if this looks good? I no longer have permissions to do so myself for some reason.

mgeisler:
> @mgeisler can you hit the merge button if this looks good? I no longer have permissions to do so myself for some reason.

Yes! Sorry about the delay — @gribozavr, can you help with permissions here?

PR #3173, opened 2026-04-23
memory-management: rename `as_dependency` to `to_dependency`

m4tx:
Because this function takes a reference to `self` and returns an owned object, it's more idiomatic to call it `to_` rather than `as_`: https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

This was brought up when discussing the code with the students when running the course.


mgeisler (approved):
Thanks, that makes sense to me!

gribozavr (approved):

PR #3185, opened 2026-05-21
ru: translate types-and-values segment

yabanci:
Translates the remaining 19 untranslated entries in `src/types-and-values/`, closing this chapter to 100% in Russian.

## Scope

All untranslated entries in:
- `hello-world.md` (intro slide + multi-paradigm note)
- `variables.md` (let bindings, mut, i32 / type inference explanation)
- `values.md` (built-in types intro + underscore-in-numbers note)
- `arithmetic.md` (function-takes-three-ints explanation, integer overflow C/C++ vs Rust, change-to-i16 exercise, constant-expression overflow note)
- `exercise.md` (Fibonacci prompt + code-comment hints + `"Implement this"` literal)

No existing translations were modified. Fuzzy entries in this chapter are left as-is for the original translator (@baltuky) to revisit.

## Translation choices

- Followed the existing glossary in `ru.po` for established terms (Function → Функция, Type → Тип, etc.).
- Kept Rust API method names (`overflowing_*`, `saturating_*`, `carrying_*`, `wrapping_*`, `checked_*`) in English with backticks — these are identifiers, not concepts.
- For integer overflow vocabulary, used the canonical RU Rust translation: wrap-around → обёртывание (matching the Rust Book RU at doc.rust-lang.ru/book/ch03-02).
- Format-string and code-comment literals translated (`"Hello 🌍!"` → `"Привет, 🌍!"`, `// The base case.` → `// Базовый случай.`); technical identifier placeholders like `"x: {x}"` left alone.

## Verification

```
$ msgfmt --statistics -o /dev/null po/ru.po
571 translated messages, 72 fuzzy translations, 3117 untranslated messages.
```

Before: 552 translated, 72 fuzzy, 3136 untranslated. Δ = +19 translated, -19 untranslated. Fuzzy count unchanged.

### DIFF
--- a/po/ru.po
+++ b/po/ru.po
@@ -2925,7 +2925,7 @@ msgstr ""
 
 #: src/types-and-values/hello-world.md
 msgid "\"Hello 🌍!\""
-msgstr ""
+msgstr "\"Привет, 🌍!\""
 
 #: src/types-and-values/hello-world.md
 msgid "What you see:"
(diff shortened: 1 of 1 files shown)


gribozavr (approved):

gribozavr:
@yabanci Thanks! Please resolve the merge conflict and the pre-merge check failure.

yabanci:
Rebased onto current main (conflict was only the PO-Revision-Date header line). Also fixed the `check-msgid-changes` failure: the `Arithmetic is very similar...` entry had its msgid split into multi-line format (empty string + continuation) — reverted to the original single-line form; translation is preserved in msgstr.

mgeisler:
Hi @yabanci, thanks for the rebase!

The problem is this part of the diff:

image

The rather strict checks we run only allow changes to the `msgid` fields (even whitespace changes) if the `POT-Creation-Date` header is also updated. The logic behind this is that the English text for a translation (the `msgid` fields) cannot change out of the blue. Only when a new `.pot` file is incorporated, should the fields change.

In this case, I'm not sure how those whitespace changes snug into the file?

yabanci:
@mgeisler Good catch — thanks for the detailed explanation.

The stray `msgid` whitespace changes came from my PO editor re-wrapping long lines (moving the trailing space from end-of-line to start-of-next-line). To rule that out completely, I rebuilt this branch directly on top of current `main` and injected **only** the `msgstr` translations, leaving every `msgid` byte-identical to upstream. The diff is now strictly +43/-6 (the 6 deletions are `msgstr ""` placeholders).

I reproduced the CI check locally against `origin/main` and it now passes:

```
$ python3 .github/workflows/check-msgid-changes.py
$ echo $?
0
```

(Same check fails on the previous revision at `po/ru.po:3004`, confirming the wrapping was the only issue.) No `POT-Creation-Date` change. Ready for another look.

gribozavr (approved):

PR #3187, opened 2026-05-22
Fix redirects

### LINKED ISSUE #3186: Many redirects target a non existent slide
In [`book.toml`](https://github.com/google/comprehensive-rust/blob/main/book.toml#L91), many redirects target a non existent slide.

For example, https://google.github.io/comprehensive-rust/async/control-flow/select.html redirects to https://google.github.io/comprehensive-rust/async/control-flow/select.html, which is also incorrect.

I propose fixing the redirects and then adding a CI check to ensure all redirects remain valid.

bridadan:
Fixes #3186

NOTE: the fixes to the actual redirects were done with the assistance of an LLM. I've looked through most of them and they seem to be generally accurate, but it's hard to say for sure given the original sources have been deleted. Please feel free to suggest an alternative!


book.toml
```diff
@@ -274,17 +274,17 @@ use-boolean-and = true
 "unsafe/extern-functions.html" = "../unsafe-rust/unsafe-functions.html"
 "unsafe.html" = "unsafe-rust/unsafe.html"
 "unsafe/mutable-static-variables.html" = "../unsafe-rust/mutable-static.html"
-"unsafe/mutable-static-variables.md" = "mutable-static-variables.html"
+"unsafe/mutable-static-variables.md" = "../unsafe-rust/mutable-static.html"
```
djmitche: It looks like this was incorrect when it was added and is redundant to the previous line. I think this line can be removed, and if it's easy it might be good to flag such lines as invalid in the GH check.

djmitche (commented):
Note that this also needs `dprint fmt` run for the Python script.

qwandor (approved):

book.toml
bridadan: Good catch! I've removed this line but also added a check to ensure all redirects end in either `.html` or nothing (when redirecting back to a directory like [this existing one](https://github.com/google/comprehensive-rust/blob/main/book.toml#L286))

bridadan:
@djmitche Should be ready for re-review, thanks!

.github/workflows/check-redirects.py
```diff
@@ -33,6 +33,32 @@
 import subprocess
 import sys
 
+ALLOWED_REDIRECT_EXTENSIONS = [".html", ""]
+ALLOWED_REDIRECT_EXTENSIONS_STRING = ",".join(
+    f'"{ext}"' for ext in ALLOWED_REDIRECT_EXTENSIONS)
```
djmitche: This will be `".html,"` which isn't especially readable in an error message.

djmitche (approved):
I'm happy enough with this! I don't have merge perms so @qwandor is the ultimate judge anyway.

.github/workflows/check-redirects.py
bridadan: Good point, I've updated the error to add more clarity.

PR #3197, opened 2026-06-09
bazel: Build and cache local tools with Bazel

### LINKED ISSUE #1168: Rewrite CI to only test languages that have actually changed
We currently build and test all translations — resulting in some 32 GitHub workflows being executed per PR. This is often unnecessary when only the original English text changes.

I'm considering if it would be worth it to use a big build system such as Bazel. This would allow us to get very fine-grained dependency tracking and this would allow us to skip unnecessary steps. However, I believe it will be significantly more complex to maintain the build system files that way.

Thoughts or ideas?

mgeisler:
This PR is the first step of several aiming at introducing better caching and reproducibility with Bazel. It will eventually fix #1168.

I have been introducing Bazel for a large polyglot build at work, and this gave me hands on experience with building Rust with Bazel. In this project, we have so far been using shell scripts. YAML files, and a bit of `cargo xtask` to keep the build here going.

However, it's time to admit that we have a non-trivial build process:

- we build several `mdbook` plugins, some from our repository, some from external repositories (we install them globally with `cargo install`).

- we use `build.sh` to backdate the sources for translations (we leave behnd a dirty working copy).

- we have end-to-end tests which rely on NodeJS being installed on the system.

All this complexity can be rained in with Bazel: we can build our `mdbook` plugins hermetically and put them on the `PATH` only for the final `mdbook build` invocation. We can put our backdated sources into temporary directories and keep the working copy clean. This will let us skip building translations that haven't changed, thus speeding up local and CI builds.

I have not yet looked at the NodeJS parts, but I'm sure that can be integrated with Bazel as well.

The approach used lets `rules_rust` read the `Cargo.lock` file:

```python
crate = use_extension("@rules_rust//crate_universe:extensions.bzl", "crate")
crate.from_cargo(
    name = "crates",
    cargo_lockfile = "//:Cargo.lock",
    manifests = ["//:Cargo.toml"],
)
use_repo(crate, "crates")
```

This is the documented on
https://bazelbuild.github.io/rules_rust/crate_universe_bzlmod.html.

It build on the existing Cargo setup and is thus compatible with it. We can discuss later if we want to keep Cargo compatibility or move completely to Bazel.


mgeisler:
If people are okay with this, I have other PRs after this:

- move building all our Rust code into the Bazel sandbox
- move the backdating logic to Bazel repositories
- setting up `rustfmt` via Bazel (no more ad-hoc pinning of a nightly version)

gribozavr (approved):

mgeisler:
Thanks for the quick review, @gribozavr! I'll put up more PRs next and migrate the build to Bazel.

PR #3202, opened 2026-06-10
bazel: migrate external `mdbook` plugins to Bazel

mgeisler:
Configure mdbook-i18n-helpers, mdbook-linkcheck2, mdbook-pandoc, and mdbook-svgbob as Bazel external plugins using `rules_rust`.

Update `xtask/src/main.rs` to build all preprocessor tools via Bazel, programmatically resolve their paths, copy them to `~/.cargo/bin`.

This is a drop-in replacement for the old Cargo based approach. It will go away as we move the `mdbook build` call itself to Bazel, but it's useful in its own since it establishes that we can build the `mdbook` plugins with Bazel.


mgeisler:
Blocked on #3212.

MODULE.bazel
```diff
@@ -23,10 +23,67 @@ use_repo(llvm, "llvm_toolchain")
...
+rust_host_tools.host_tools(
+    name = "rust_host_tools_nightly",
+    sha256s = {
+        "2025-09-01/cargo-nightly-x86_64-unknown-linux-gnu.tar.xz": "9a701f2eb103703c018518066fa7deb476f7ebab548b1c4e2ea0df81ee42a20f",
+        "2025-09-01/clippy-nightly-x86_64-unknown-linux-gnu.tar.xz": "a55b532605896ed0a22f44d68c4eba8ba0ad66ea0f62f3c62469b0a3d01ab19c",
+        "2025-09-01/llvm-tools-nightly-x86_64-unknown-linux-gnu.tar.xz": "a0da791e41770b888c60ef74a0c63c5f6fe11bddb2a8fa88075341a9acb442a6",
+        "2025-09-01/rust-std-nightly-x86_64-unknown-linux-gnu.tar.xz": "b3c2d890d9405285e015ab4ceb78087e3ec55c64b0b3030d79102dfe5e622e09",
+        "2025-09-01/rustc-nightly-x86_64-unknown-linux-gnu.tar.xz": "7f5486ae6ece8a734149e5fce6dc8f1bcdcb36b5c5cb53a819569109b445c700",
+    },
+    version = "nightly/2025-09-01",
```
djmitche, line 36: That was a few nights ago...

djmitche (approved):

MODULE.bazel
```diff
@@ -23,10 +23,67 @@ use_repo(llvm, "llvm_toolchain")
...
+crate.spec(
+    artifact = "bin",
+    package = "mdbook-pandoc",
+    repositories = ["mdbook_plugins"],
+    version = "0.11.0",
+)
+crate.from_specs(
+    name = "mdbook_plugins",
+    generate_binaries = True,
+    host_tools = "@rust_host_tools_nightly",
```
qwandor, line 72: Why do we need to build these with the nightly toolchain? (Assuming that that is what this is saying.)

MODULE.bazel
```diff
@@ -23,10 +23,67 @@ use_repo(llvm, "llvm_toolchain")
...
+    artifact = "bin",
+    package = "mdbook-linkcheck2",
+    repositories = ["mdbook_plugins"],
+    version = "0.12.0",
+)
+crate.spec(
+    artifact = "bin",
+    package = "mdbook-pandoc",
+    repositories = ["mdbook_plugins"],
+    version = "0.11.0",
```
qwandor, line 67: Can we configure Dependabot to update these dependencies too?

MODULE.bazel
```diff
@@ -23,10 +23,67 @@ use_repo(llvm, "llvm_toolchain")
...
+)
+crate.from_specs(
+    name = "mdbook_plugins",
+    generate_binaries = True,
+    host_tools = "@rust_host_tools_nightly",
+)
+use_repo(crate, "mdbook_plugins")
+
+# mdbook-svgbob depends transitively on sauron-core, which has a
+# "=0.3.30" dependency on futures. This conflicts with other plugins.
```
qwandor, line 77: Does Bazel not allow multiple versions of a library crate in the tree? That seems like a big difference from cargo that could relate in other issues in future.

xtask/src/main.rs
```diff
@@ -155,33 +155,46 @@ fn install_tools(binstall: bool) -> Result<()> {
...
-    let tools = [
-        ("mdbook", "0.5.3"),
-        ("mdbook-svgbob", "0.3.0"),
-        ("mdbook-pandoc", "0.11.0"),
-        ("mdbook-i18n-helpers", "0.4.0"),
-        ("i18n-report", "0.2.0"),
-        ("mdbook-linkcheck2", "0.12.0"),
-    ];
+    // Tools not yet migrated to be installed with Bazel.
+    let tools = [("mdbook", "0.5.3"), ("i18n-report", "0.2.0")];
```
qwandor, line 159: Why can't these be moved to Bazel at the same time?

MODULE.bazel
mgeisler: Yeah, I know :smile: It's what we have hard-coded in `main.rs`. We should of course update it once in a while.

MODULE.bazel
mgeisler: Yeah, I was asking myself (and Gemini) the same thing! What is happening here is that we want Bazel to run the equivalen to of `cargo install`.

It turns out that this isn't exposed in a stable way: there is a `-Z bindeps` flag ([reference](https://doc.rust-lang.org/cargo/reference/unstable.html#artifact-dependencies)), but this requires a nightly compiler.

This is what `rules_rust` [uses internally](https://bazelbuild.github.io/rules_rust/crate_universe_bzlmod.html#binary-dependencies) when we specify that we want the binary artifacts.

MODULE.bazel
mgeisler: Yeah, it is different. I've been playing with this for a few weeks now and what I've learnt is that

- We had the same problem with `cargo install`, we just didn't know. When you `cargo install foo`, a small workspace is created on the fly and `foo` is compiled in that. Our different `cargo install` calls created different workspaces and could thus depend on whatever incompatible crate versions they like.
- With the declarations here, we create a bigger workspace called `mdbook_plugins` and compile most `mdbook` plugins in this workspace. We thus save a little time since transitive dependencies are only compiled once.
- We are creating multiple Bazel "repositories" with `rules_rust`: here `mdbook_plugins` and `svgbob_plugin`. We also have a repository called `crates` which is populated with the dependencies from our actual Cargo workspace via
    ```python
    crate.from_cargo(
        name = "crates",
        cargo_lockfile = "//:Cargo.lock",
        manifests = ["//:Cargo.toml"],
    )
    use_repo(crate, "crates")
    ```

    This uses our `Cargo.lock` file and I believe this means that the same crate can be available in multiple different versions. I'm not 100% sure, though.

I hope this helps clarify things a little?

xtask/src/main.rs
mgeisler: Actually, they could... I was just starting with the `mdbook` plugins to round off that part of the build.

I'll migrate everything eventually — and remove `cargo xtask install-tools` since everything will be installed/compiled on the fly with Bazel.

MODULE.bazel
mgeisler: Actually... I don't think it is supported. There is [Dependabot support for Bazel modules](https://blog.bazel.build/2026/02/20/dependabot.html), but the version number here is from crates.io and Bazel doesn't really know anything about it.

So I think we're in the same boat as before where this was hard-coded in `main.rs`.

xtask/src/main.rs
mgeisler: Being moved in #3215!

PR #3204, opened 2026-06-10
bazel: Configure LLVM toolchain

mgeisler:
I noticed warnings when compiling Rust code both locally and in GitHub CI:

> INFO: From Compiling Rust bin mdbook-exerciser (1 file):
> warning: the gold linker is deprecated and has known bugs with Rust
>  |
>  = help: consider using LLD or ld from GNU binutils instead

From what I understands, this is because we haven't configured any C/C++ toolchain for Bazel and it thus picks up what is on the CI runner image.

Configuring `toolchains_llvm` fixes the warnings and ensures a hermetic build.


mgeisler:
I'm curious how long it will take to download and configure this on the first build (and if subsequent builds are faster due to the cache). We're talking 10+ GB of unpacked data.

mgeisler:
Hmm... this was suspiciously fast... on my laptop, it took a small eternity to download and unpack the LLVM toolchain and then recompile everything.

mgeisler:
I tried clearing the caches and trigger a build again — it recompiled everything and so I believe this is working as intended (it's just faster than I had expected).

qwandor (approved):

PR #3205, opened 2026-06-10
bazel: Define tests for everyting tested with Cargo

mgeisler:


qwandor:
Does this mean all tests have to be added to these build files in future, rather than being picked up automatically by cargo? What happens if someone forgets to do that?

mgeisler:
> Does this mean all tests have to be added to these build files in future, rather than being picked up automatically by cargo?

Hey Andrew! Yeah, that's a stark difference... it's the usual Bazel argument that explicit is better than implicit. Whether that works out in practice, I cannot really say as we're only just migrating to Bazel now at Proton.

> What happens if someone forgets to do that?

The test will not be run and nothing will detect this.

I guess the mixed setup with both Cargo and Bazel compatibility is making this worse compared to a pure Bazel-based build. There, the developer writing a new test would run `bazel test` locally while working on it, and so they would quickly notice if the test isn't hooked up anywhere. If they now develop locally with `cargo test`, but CI runs `bazel test`, then they might never realize that their test is ignored. Tooling could of course help here — and I'm actually interested in this for my `$DAYJOB` as well :smile:

Cc @gribozavr as well who probably have thoughts on this.

djmitche:
Gazelle can do some of that automatically, or buildozer can be used to script it

mgeisler:
> Gazelle can do some of that automatically, or buildozer can be used to script it

Oh, weird... I knew of Gazelle for generating build files for Go, but it now also supports Rust: Calsign/gazelle_rust: Gazelle language plugin for Rust. https://github.com/Calsign/gazelle_rust

I call it weird since the `rules_rust` documentation doesn't mention it at all and I have also not come across it a single time in the last three months when looking at Rust and Bazel.

But thanks for the pointer, we could look into this to save us the trouble of updating this.

djmitche:
I'm mostly using it for Go and it's not _awesome_ per se. I've also done some scripting with buildozer which for basic things might be easy to either run by hand or automate.

djmitche (approved):
These look good!

BTW, there is a Bazel Slack instance with a Rust channel that seems fairly active.

mgeisler:
> BTW, there is a Bazel Slack instance with a Rust channel that seems fairly active.

Thanks, I'll have to check it out!

PR #3208, opened 2026-06-13
bazel: disable textual diffs for ``MODULE.bazel.lock`

mgeisler:
The file has extremely long lines, which makes my poor Emacs very confused. We can clean this up if/when
https://github.com/bazelbuild/rules_rust/issues/3946 is fixed.


mgeisler:
Blocked on #3212.

djmitche (approved):

PR #3209, opened 2026-06-13
bazel: automatically show test failures

mgeisler:
Without this, the errors are only available in a Bazel log file.


mgeisler:
Blocked on #3212.

djmitche (approved):

PR #3210, opened 2026-06-13
bazel: avoid caching huge repository cache

mgeisler:
The LLVM toolchain repository is around 5 GB in the cache, which means that we quickly fill up the 10 GB cache quota we have for GitHub Actions. Downloading it should be roughtly equally fast.


djmitche (approved):

mgeisler:
Blocked on #3212.

PR #3213, opened 2026-06-14
bazel: ensure Cargo and Bazel lock files are in sync

mgeisler:
We lacked a lint job which ensures that `Cargo.lock` is in sync with all the `Cargo.toml` files in the workspace. This is now added: from what I can tell[1], running `cargo update --workspace --locked` is the easiest way to do this. I tested it locally and it displays a nice little table of what would be updated on error:

    % cargo update --workspace --locked
        Updating crates.io index
         Locking 2 packages to latest compatible versions
        Updating zerocopy v0.8.50 -> v0.8.52
        Updating zerocopy-derive v0.8.50 -> v0.8.52
    note: pass `--verbose` to see 131 unchanged dependencies behind latest
    error: cannot update the lock file /home/mg/src/comprehensive-rust/Cargo.lock because --locked was passed to prevent this
    help: to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.

Similarly, we lacked a job to ensure that the new `MODULE.bazel.lock` file is in sync with the Cargo world. This is now also added.

[1]: https://users.rust-lang.org/t/check-if-the-cargo-lock-is-up-to-date-without-building-anything/91048/5


mgeisler:
Blocked on #3212.

djmitche (approved):

PR #3215, opened 2026-06-16
bazel: build all Rust tools

mgeisler:
This moves the remaining two tools into Bazel: `i18n-report` and `mdbook` itself. They are still installed with

    cargo xtask install-tools

to make this less disruptive. This will go away as we start using the `mdbook` rule in Bazel to build the many books.


xtask/src/main.rs
```diff
@@ -135,54 +135,23 @@ fn copy(src: &Path, dest: &Path) -> Result<u64> {
     Ok(len)
 }
 
-fn install_tools(binstall: bool) -> Result<()> {
+fn install_tools(_binstall: bool) -> Result<()> {
```
qwandor: Why not remove the binstall parameter, rather than just ignoring it?

mgeisler: I'm aiming towards removing the whole `install-tools`, so I figured it would only lead to more churn if I updated the call site.

But I can remove it, no problem.

mgeisler: You were right, I should have removed it completely. Done now.

qwandor (approved):

djmitche (approved):

PR #3220, opened 2026-06-20
bazel: build and test bare-metal examples

mgeisler:
This configures Bazel targets for the bare-metal examples. We now also have test targets for them:

    bazel test //:bare_metal_tests

will compile all the code and run a simple test in QEMU (boot the binaries, send `"q"` to them on stdin).

The Cargo-based workflow has been kept: I expect that students will keep using it in class, so we should keep testing it in CI.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -3,11 +3,25 @@ module(
     version = "0.1.0",
 )
 
+bazel_dep(name = "platforms", version = "1.1.0")
 bazel_dep(name = "rules_rust", version = "0.70.0")
+bazel_dep(name = "rules_shell", version = "0.8.0")
+
+# Remove when https://github.com/bazelbuild/rules_rust/pull/4019 is
+# released and we update past version 0.70.0.
+git_override(
+    module_name = "rules_rust",
+    commit = "d2d856d13f89f234298db1606ece7afd1c3eee1f",
+    remote = "https://github.com/bazelbuild/rules_rust.git",
+)
 
 rust = use_extension("@rules_rust//rust:extensions.bzl", "rust")
 rust.toolchain(
     edition = "2024",
+    extra_target_triples = [
+        "aarch64-unknown-none",
+        "thumbv7em-none-eabihf",
+    ],
 )
 use_repo(rust, "rust_toolchains")
 
@@ -77,3 +91,62 @@ crate.from_specs(
     host_tools = "@rust_host_tools_nightly",
 )
 use_repo(crate, "svgbob_plugin")
+
+# Bare-Metal and Microcontroller Examples
+#
+# Note: Host triple is required in supported_platform_triples below so
+# cargo-bazel can invoke cargo tree on the host during splicing.
+crate.from_cargo(
+    name = "bare_metal_alloc_example",
+    cargo_lockfile = "//src/bare-metal/alloc-example:Cargo.lock",
+    manifests = ["//src/bare-metal/alloc-example:Cargo.toml"],
+    supported_platform_triples = [
+        "x86_64-unknown-linux-gnu",
+        "aarch64-unknown-none",
+    ],
+)
+use_repo(crate, "bare_metal_alloc_example")
+
+crate.from_cargo(
+    name = "bare_metal_aps",
+    cargo_lockfile = "//src/bare-metal/aps/examples:Cargo.lock",
+    manifests = ["//src/bare-metal/aps/examples:Cargo.toml"],
+    supported_platform_triples = [
+        "x86_64-unknown-linux-gnu",
+        "aarch64-unknown-none",
+    ],
+)
+use_repo(crate, "bare_metal_aps")
+
+crate.from_cargo(
+    name = "bare_metal_microcontrollers",
+    cargo_lockfile = "//src/bare-metal/microcontrollers/examples:Cargo.lock",
+    manifests = ["//src/bare-metal/microcontrollers/examples:Cargo.toml"],
+    supported_platform_triples = [
+        "x86_64-unknown-linux-gnu",
+        "thumbv7em-none-eabihf",
+    ],
+)
+use_repo(crate, "bare_metal_microcontrollers")
+
+crate.from_cargo(
+    name = "bare_metal_compass",
+    cargo_lockfile = "//src/exercises/bare-metal/compass:Cargo.lock",
+    manifests = ["//src/exercises/bare-metal/compass:Cargo.toml"],
+    supported_platform_triples = [
+        "x86_64-unknown-linux-gnu",
+        "thumbv7em-none-eabihf",
+    ],
+)
+use_repo(crate, "bare_metal_compass")
+
+crate.from_cargo(
+    name = "bare_metal_rtc",
+    cargo_lockfile = "//src/exercises/bare-metal/rtc:Cargo.lock",
+    manifests = ["//src/exercises/bare-metal/rtc:Cargo.toml"],
+    supported_platform_triples = [
+        "x86_64-unknown-linux-gnu",
+        "aarch64-unknown-none",
+    ],
+)
+use_repo(crate, "bare_metal_rtc")
--- a/src/bare-metal/alloc-example/BUILD.bazel
+++ b/src/bare-metal/alloc-example/BUILD.bazel
@@ -1,0 +1,28 @@
+"""Bazel BUILD file for AArch64 Bare-Metal Allocator Example.
+
+This package defines the allocator example target. Because this code
+targets AArch64 bare-metal platforms and not the host OS:
+
+1. The raw `rust_binary` is constrained using `target_compatible_with`
+   to run only on `os:none` to prevent compilation errors during
+   wildcard host builds.
+"""
+
+load("@rules_rust//rust:defs.bzl", "rust_binary")
+
+package(default_visibility = ["//visibility:public"])
+
+rust_binary(
+    name = "alloc-example",
+    srcs = ["src/main.rs"],
+    edition = "2024",
+    rustc_flags = [
+        "-C",
+        "linker=rust-lld",
+    ],
+    target_compatible_with = ["@platforms//os:none"],
+    deps = [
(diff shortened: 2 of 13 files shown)


src/bare-metal/aps/examples/BUILD.bazel
```diff
@@ -0,0 +1,373 @@
...
+   `target_compatible_with` to run only on `os:none` to prevent
+   compilation errors during wildcard host builds.
+
+2. We extract flat binary images via `genrule` targets using the
+   hermetic host toolchain's `rust-objcopy`.
+
+3. We wrap each binary target in a platform transition
+   (`aarch64_binary`) so they are built for the AArch64 bare-metal
+   target platform during host-executed tests.
+"""
```
djmitche, line 18: This file is quite repetitive -- could it be improved with a rule or a macro?

djmitche (commented):
I admit I'm not quite sure why all of this is happening!

mgeisler:
> I admit I'm not quite sure why all of this is happening!

Hey @djmitche, yeah, fair point! #1168 is in some sense the starting point — it has bothered me for years that

1. we build too much in every CI run and
2. we don't have an easy way to locally replicate what `publish.yml` does

At Proton, I've been part of introducing Bazel for our monorepo: building Rust, Kotlin, Swift, Typescript with it... So I naturally started looking deeper into it and realized that it's not so complicated. I kind of knew this from Google where we also use Bazel (Blaze) with `rules_rust`.

What you, @qwandor and @gribozavr haven't seen yet is the Bazel build rules that actually do something interesting: namely configure Bazel "repositories" (those things that start with `@`...) to contain the back-dated sources for each language.

With that logic neatly wrapped in a repository rule in Bazel, it becomes easy to do (typed from memory):

```python
mdbook(
  name = "da_html",
  srcs = "@backdated_sources//:da_content",
  book = "@backdated_sources//:da_book",
)
```

and this gives you a `mdbook` generated HTML book for Danish. You then simply `bazel build //:da_html` to get this generated.

I have this on my laptop which is not infront of me right now. But that is the end goal: generate per-language repositories which in turn are used to generate HTML, PDF, exerciser, ... outputs.

Now, I must admit that the step towards that have been more bumpy that I expected! The LLVM toolchain sage (resolved for now in #3221) made things strangely unstable, unlike what I've seen on my own laptop. So I was expecting

image

but it has been a little more than 20 minutes by now :smile: 

I will try and put up the other bits and pieces next so you and the rest of the admins can look at the whole thing together.

src/bare-metal/alloc-example/BUILD.bazel
```diff
@@ -0,0 +1,28 @@
...
+
+package(default_visibility = ["//visibility:public"])
+
+rust_binary(
+    name = "alloc-example",
+    srcs = ["src/main.rs"],
+    edition = "2024",
+    rustc_flags = [
+        "-C",
+        "linker=rust-lld",
```
qwandor, line 21: We don't have this flag in the `Cargo.toml`, why is it needed here?

src/bare-metal/aps/examples/BUILD.bazel
```diff
@@ -0,0 +1,373 @@
...
+rust_binary(
+    name = "improved",
+    srcs = glob(["src/**/*.rs"]),
+    compile_data = glob(["src/**/*.S"]),
+    crate_root = "src/main_improved.rs",
+    edition = "2024",
+    linker_script = ":image.ld",
+    rustc_flags = [
+        "-C",
+        "linker=rust-lld",
```
qwandor, line 36: Again, why does this need to be specified here?

src/bare-metal/aps/examples/BUILD.bazel
```diff
@@ -0,0 +1,373 @@
...
+load("@rules_shell//shell:sh_binary.bzl", "sh_binary")
+load("@rules_shell//shell:sh_test.bzl", "sh_test")
+load("//platforms:transition.bzl", "aarch64_binary")
+
+package(default_visibility = ["//visibility:public"])
+
+rust_binary(
+    name = "improved",
+    srcs = glob(["src/**/*.rs"]),
+    compile_data = glob(["src/**/*.S"]),
```
qwandor, line 30: What's the difference between `srcs` and `compile_data`?

src/bare-metal/aps/examples/run_test.sh
```diff
@@ -0,0 +1,22 @@
...
+# Detect if running as an automated test under 'bazel test'
+NON_INTERACTIVE="${TEST_SRCDIR:-}"
+
+if [ ! -f "$BIN_PATH" ]; then
+    echo "Error: Binary not found at $BIN_PATH" >&2
+    exit 1
+fi
+
+echo "Running ${BIN_NAME}.bin under QEMU..."
+if [ -n "$NON_INTERACTIVE" ]; then
```
qwandor, line 18: If the point of this script is to test the binaries, why does it need to be different in interactive shells? If they intention is to use it to run examples interactive then I think the script should be called `run_example.sh` rather than `run_test.sh`, and maybe take a flag rather than detecting whether it is interactive. Though I'm still not really keen on the parallel build systems, who is going to actually use this?

src/bare-metal/microcontrollers/examples/BUILD.bazel
```diff
@@ -0,0 +1,146 @@
+"""Bazel BUILD file for ARM Cortex-M Microcontroller Examples.
+
+This package defines board support and hardware abstraction layer
+(HAL) binaries for microcontrollers. Because this code targets ARM
```
qwandor, line 4: Nit: this description is misleading, the binaries are examples of how to use PAC, HAL and BSP libraries, they aren't in themselves the HAL or anything.

src/bare-metal/microcontrollers/examples/BUILD.bazel
```diff
@@ -0,0 +1,146 @@
...
+    srcs = glob(["src/bin/**/*.rs"]),
+    crate_root = "src/bin/board_support.rs",
+    edition = "2024",
+    rustc_flags = [
+        "-C",
+        "linker=rust-lld",
+        "-C",
+        "link-arg=-Tlink.x",
+    ],
+    target_compatible_with = ["@platforms//os:none"],
```
qwandor, line 27: Where do you specify the architecture?

src/exercises/bare-metal/compass/BUILD.bazel
qwandor, line 1: Will this file be included in the generated exercise template that students download? We should make sure it isn't, as it presumably won't work in that context and will be confusing.

src/exercises/bare-metal/rtc/BUILD.bazel
qwandor, line 1: Ditto, we should make sure this isn't included in the exercise template to confuse students.

src/exercises/bare-metal/rtc/rtc_test.sh
```diff
@@ -0,0 +1,21 @@
...
+NON_INTERACTIVE="${TEST_SRCDIR:-}"
+
+if [ ! -f "$BIN_PATH" ]; then
+    echo "Error: rtc.bin not found at $BIN_PATH" >&2
+    exit 1
+fi
+
+echo "Running rtc.bin under QEMU..."
+if [ -n "$NON_INTERACTIVE" ]; then
+    echo "q" | qemu-system-aarch64 -machine virt,gic-version=3 -cpu max -serial mon:stdio -display none -kernel "$BIN_PATH"
```
qwandor, line 18: What is the point of this? The model solution doesn't read any input from the serial console.

MODULE.bazel
```diff
@@ -87,3 +101,62 @@ crate.from_specs(
...
+crate.from_cargo(
+    name = "bare_metal_alloc_example",
+    cargo_lockfile = "//src/bare-metal/alloc-example:Cargo.lock",
+    manifests = ["//src/bare-metal/alloc-example:Cargo.toml"],
+    supported_platform_triples = [
+        "x86_64-unknown-linux-gnu",
+        "aarch64-unknown-none",
+    ],
+)
+use_repo(crate, "bare_metal_alloc_example")
```
qwandor, line 108: Why do we need both these and the build files in the various directories? How do they interact?

djmitche:
Thanks for the explanation! We use bazel at $WORK too and roughly the same experience: conceptually really need, in practice lots of unexpected sharp edges!

src/bare-metal/aps/examples/run_test.sh
mgeisler: > If the point of this script is to test the binaries, why does it need to be different in interactive shells?

Oh, I just realized that the examples can be used for both: executed by Bazel for testing and by hand if someone wants to test things out. But I think you're right, the interactive mode could be removed.

The overall goal was to add test coverage to something that was never tested in CI in the past.

> Though I'm still not really keen on the parallel build systems, who is going to actually use this?

My thinking is that GitHub CI will use this -- ultimately, CI should just be `bazel test //...` in this repository. Everything will be cached, and only the minimum amount of work necessary will be done.

src/bare-metal/aps/examples/BUILD.bazel
mgeisler: The `srcs` are the Rust source files passed to `rustc`, the `compile_data` is other files which are made available in the Bazel sandbox during compilation (for example for `include_str!` macro calls).

src/bare-metal/aps/examples/BUILD.bazel
mgeisler: Bazel will default to use the C/C++ toolchain configured whereas [Cargo uses `rust-lld` by default from 1.90](https://blog.rust-lang.org/2025/09/01/rust-lld-on-1.90.0-stable/).

So it's a matter of being more explicit to get the existing behavior in Bazel.

src/exercises/bare-metal/rtc/rtc_test.sh
mgeisler: Ah, it was just copied from the other test script. Removed now.

Issue #2996, opened 2025-12-09
Cannot serve or build the Portuguese (pt-BR) translation locally


osseamdevel:
### What I'm trying to do
I want to read the course in **Portuguese (Brazilian - pt-BR)** on my local machine.

### What I tried
- `cargo xtask install-tools` → installs `mdbook-course` successfully
- `cargo xtask serve` → works for English, but I cannot figure out how to serve or build the pt-BR version

### Expected result
Open the full course in Portuguese in my browser (localhost or static files).

### System information
- OS: Ubuntu (or whatever you have)
- Rust version: `rustc 1.91.1` (run `rustc --version` if you want to be precise)
- `mdbook-course --version` → 0.4.52 (or whatever it shows)

### Question
What is the correct command in the current main branch to:
1. Serve the pt-BR version on localhost:3000, or  
2. Build the static pt-BR version so I can open `book/pt-BR/index.html`

Thank you very much! This course is amazing and I really want to study it in Portuguese.

michael-kerscher:
To answer your first questions: You can serve or build any translation locally using the  -l  ( --language ) flag with  cargo xtask :
  
  1. To serve the Portuguese (pt-BR) translation locally on localhost:3000:
```
cargo xtask serve -l pt-BR
```
  
  2. To build static HTML files for the pt-BR translation:
```
cargo xtask build -l pt-BR
```
The generated HTML entrypoint will be located at  `book/pt-BR/html/index.html`.

But tat unfortunately does not work at the moment for pt-BR (likely other languages as well depending on similar conditions, not all translations are linkcheck2ed yet...)

On the production site, clicking unsafe traits on the Portuguese Traits Marker page currently leads to a broken 404 link: https://google.github.io/comprehensive-rust/pt-BR/concurrency/send-sync/marker-traits.html -> (HTTP 404) https://google.github.io/comprehensive-rust/pt-BR/concurrency/unsafe/unsafe-traits.html

### Why local builds are failing
  
  If you run `cargo xtask build -l pt-BR` on `main`, the build failed with the following error from  linkcheck2 :
```
 INFO Invoking the "linkcheck2" renderer
error: File not found: ../unsafe/unsafe-traits.md
   ┌─ concurrency/send-sync/marker-traits.md:13:21
   │
[...]

PR #3223, opened 2026-06-25
Bazel backdating

mgeisler:
Hey @qwandor and @djmitche, this is a WIP of where I'm taking the project. On this branch,

```
% bazel build :da
# ...
Target //:da up-to-date:
  bazel-bin/da_html
  bazel-bin/da_pandoc
```

produces a Danish book!

I've been spinning off small bits and pieces of this WIP commit over the last few weeks in the PRs you've reviewed.


mgeisler:
From this infrastructure, it's a small step to make a `:website` target which builds all known languages, puts them into the correct sub-directory and voila, we can now move all logic out of shell scripts and YAML files (and into `.bzl` files).

The complexity is still there... but it's now suddenly in a form that allows us to reproduce things locally that we could not really do easily today.

backdated_sources.bzl
```diff
@@ -0,0 +1,230 @@
...
+    book = toml.decode(repository_ctx.read(base_book_path))
+
+    # Apply pristine Rust/settings if it is backdated.
+    if pristine_book and "rust" in pristine_book:
+        book["rust"] = pristine_book["rust"]
+
+    # Patch language-specific settings.
+    book["book"]["language"] = lang
+    book["output"]["html"]["site-url"] = "/comprehensive-rust/{}/".format(lang)
+    book["output"]["html"].pop("redirect", None)
```
michael-kerscher, line 64: is that removing the redirects completely? In my understanding this is necessary to be kept at the state at which it was backdated.

backdated_sources.bzl
```diff
@@ -0,0 +1,230 @@
...
+
+    book["output"].pop("linkcheck", None)
+    book["output"].pop("linkcheck2", None)
```
michael-kerscher, line 67: When we are modifying the entire file alrready, should we care about the linkcheck2 results? https://github.com/google/comprehensive-rust/issues/2996 - this would have been avoidable or at least we could have fixed the translations if linkcheck2 would have been enabled

backdated_sources.bzl
```diff
@@ -0,0 +1,230 @@
...
+
+We do this by creating isolated, read-only external repositories for
+each required point in time. This ensures the local workspace remains
+clean and the builds are hermetic.
+"""
+
+load("@toml.bzl", "toml")
+load("@yaml.bzl", "yaml")
+
+def _book_sources_repo_impl(repository_ctx):
```
michael-kerscher, line 16: this function has a high mental load. I would prefer to split that into multiple rules. Maybe we could even separate some build rules and depend on them here. e.g. the archive.tar.gz creation, the config (book.toml) modification and reassembly

backdated_sources.bzl
mgeisler: Hey @michael-kerscher, it removes them just like we do today in `build.sh`: 

https://github.com/google/comprehensive-rust/blob/565f4c0d5def68ed2d43290a1a171ba6a5d71bb2/.github/workflows/build.sh#L36-L39

But as the comment says, we might not need to do this?

My gut feeling would be that we should not expect translators to know about or maintain redirects. But I guess they get them from free via updates to `main`, so yeah, we could stop clearing them.

That would be a change in functionality and so far I've only aimed at re-creating our current workflow, but in Bazel.

backdated_sources.bzl
mgeisler: Yeah, I agree that there is a lot going on! I have oscillated between having things more split and more centralized... the version here is more centralized.

I found that the split versions were complicated because of another reason: they would have implicit dependencies between different files and rules. So one rule would register a `@backdated_sources//:da_srcs` and another rule in another file would use this... that's (also) not nice.

mgeisler:
Hey folks, just a note that this PR was only for demonstration so you could get a feeling for what I've been experimenting with over the last few months.

Issue #3021, opened 2026-01-09
Issues in Unsafe Deep Dive


ia0:
Here's a list of issues in the Unsafe Deep Dive chapter which I don't feel confident to fix myself:
- The `may_overflow` [section](https://github.com/google/comprehensive-rust/blob/main/src/unsafe-deep-dive/introduction/may_overflow.md) in the introduction is incorrect. Arithmetic overflow as in `1 + i32::MAX` is [not undefined behavior](https://doc.rust-lang.org/reference/behavior-not-considered-unsafe.html?highlight=overflow-checks#integer-overflow). It is well defined to either panic or wrap-around at the implementation's discretion.
- There seems to be a leftover sentence in the "Arrays of uninit" [section](https://google.github.io/comprehensive-rust/unsafe-deep-dive/initialization/maybeuninit/arrays.html), namely the one about `slice_assume_init_ref`. It says the example uses `&buf[..input.len()]` while it doesn't.
- The [pinning chapter](https://google.github.io/comprehensive-rust/unsafe-deep-dive/pinning.html) contains an empty "details" in the speaker notes.
- The "What pinning is" [section](https://google.github.io/comprehensive-rust/unsafe-deep-dive/pinning/what-pinning-is.html) contains an unfinished sentence before the speaker notes, namely "This is explained in".
- In the same section, the sentence "A “pointer type” wraps every type that implements Deref with a target that implements Unpin." seems wrong. It's too imprecise to actually refute, but Unpin should not be a concern in this sentence.
- The outline in the "Self-Referential Buffer Example" [section](https://google.github.io/comprehensive-rust/unsafe-deep-dive/pinning/self-referential-buffer.html) is wrong. It's the one of the pinning chapter.
- The "With a raw pointer" [section](https://google.github.io/comprehensive-rust/unsafe-deep-dive/pinning/self-referential-buffer/rust-raw-pointers.html) states "`unsafe` blocks lack of safety comments. Therefore, this code is unsound." which is wrong. This Unsafe Deep Dive chapter uses the terminology "not safe" for `unsafe` blocks without safety comments. The terminology "unsound" means it is possible to trigger UB while satisfying safety preconditions.
- In the same section, the answer is insufficient. The fact that the object can move should be addressed. It should be mentioned that in that case, either cursor will point outside the expected data range (for non-overlapping moves) or it will still point inside but at a possibly incorrect position. In both cases, the rest of the function is sound due to this assertion.
[...]

PR #3244, opened 2026-07-27
unsafe-deep-dive: replace may_overflow with a real safety precondition

### LINKED ISSUE #3122: Unsafe Rust: Integer overflow is not UB
Slide https://google.github.io/comprehensive-rust/unsafe-deep-dive/introduction/may_overflow.html in the Unsage Rust course claims that integer overflow is UB in release mode. This is simply not true. See https://doc.rust-lang.org/book/ch03-02-data-types.html#integer-overflow for details.

rmyndharis:
The "Example: may_overflow" slide taught that signed integer overflow is
undefined behavior in release mode and that the function therefore needs
the `unsafe` keyword. Both claims are false: integer overflow is well
defined in Rust (debug builds panic, release builds wrap), and
`a + i32::MAX` does not require `unsafe` at all.

Replace the example with an `unsafe fn element_at` built on
`slice::get_unchecked`, which has a genuine safety precondition: the
caller must keep the index in bounds. This preserves the slide's purpose
(a concrete example of responsibility shifting to the programmer) and
leads directly into the "Safety Preconditions" segment. The speaker notes
state plainly that integer overflow is defined behavior and never requires
`unsafe`.

Closes #3122.

### DIFF
--- a/src/unsafe-deep-dive/introduction/may_overflow.md
+++ b/src/unsafe-deep-dive/introduction/may_overflow.md
@@ -7,53 +7,51 @@ Copyright 2026 Google LLC
 SPDX-License-Identifier: CC-BY-4.0
 -->
 
-# Example: may_overflow function
+# Example: a safety precondition
 
-```rust,should_panic,editable
+```rust,editable
 # // Copyright 2026 Google LLC
 # // SPDX-License-Identifier: Apache-2.0
 #
-/// Adds 2^31 - 1 to negative numbers.
-unsafe fn may_overflow(a: i32) -> i32 {
-    a + i32::MAX
+/// Returns the element at `index` without bounds checking.
+///
+/// # Safety
+///
+/// The caller must guarantee that `index < slice.len()`.
+unsafe fn element_at(slice: &[i32], index: usize) -> i32 {
+    unsafe { *slice.get_unchecked(index) }
 }
 
 fn main() {
-    let x = unsafe { may_overflow(123) };
-    println!("{x}");
+    let numbers = [10, 20, 30];
+    // Safety: `1` is a valid index into `numbers`.
+    let value = unsafe { element_at(&numbers, 1) };
+    println!("{value}");
 }
 ```
 
 <details>
 
-“The `unsafe` keyword may have a subtly different meaning than what some people
-assume.”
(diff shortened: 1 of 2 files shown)


src/unsafe-deep-dive/introduction/may_overflow.md
```diff
@@ -7,53 +7,51 @@ Copyright 2026 Google LLC
...
-/// Adds 2^31 - 1 to negative numbers.
-unsafe fn may_overflow(a: i32) -> i32 {
-    a + i32::MAX
+/// Returns the element at `index` without bounds checking.
+///
+/// # Safety
+///
+/// The caller must guarantee that `index < slice.len()`.
+unsafe fn element_at(slice: &[i32], index: usize) -> i32 {
+    unsafe { *slice.get_unchecked(index) }
```
mgeisler, line 22: Could you add a `// Safety` comment here as well to show how this `unsafe` blocks needs a proof - and that we can only prove safety by relying on the `# Safety` block attached to the function itself.

We show this pattern of shifting the responsibility to the caller in the part about bare-metal Rust, but I think it's good to show it here again.

mgeisler (approved):
Thanks a lot for cleaning up this big misconception!

I'm happy to merge this as-is, my point about adding another `// Safety` (or `// SAFETY` as I normally see) comment can come later.

src/unsafe-deep-dive/introduction/may_overflow.md
mgeisler: Cc @gribozavr and @qwandor, may I suggest that you find someone from Google with strong experience in unsafe Rust to review these slides critically? Perhaps someone in the old 20% group would be up for this?

mgeisler: Ah, I see @ia0 alrady opened #3021 with a large list of things to fix.

I am unfortunately rather busy at work these days, but I would like to send out PRs for this — or let an LLM do it, rather, as I'm sure they're rather simple to fix now that the robot knows where to look. If someone gets to it before me, by all means!

src/unsafe-deep-dive/introduction/may_overflow.md
qwandor, line 1: The file should also be renamed, it is no longer about overflowing.

PR #3245, opened 2026-07-27
Fix clippy::useless_borrows_in_formatting warnings in modules exercise

dvrdnz:
Fixes `clippy::useless_borrows_in_formatting` warnings introduced by newer
clippy versions in the modules exercise. Removes redundant `&` references
in three `writeln!` calls — no behavior change, `String` and `&String`
both implement `Display`.

### DIFF
--- a/src/modules/exercise.rs
+++ b/src/modules/exercise.rs
@@ -88,7 +88,7 @@ impl Widget for Window {
         // TODO: Change draw_into to return Result<(), std::fmt::Error>. Then use the
         // ?-operator here instead of .unwrap().
         writeln!(buffer, "+-{:-<inner_width$}-+", "").unwrap();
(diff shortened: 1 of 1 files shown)


mgeisler (approved):
Thanks @dvrdnz!

mgeisler:
I think you need to merge all the fixes into one PR... sorry about the trouble, I would have liked to fix this in smaller chunks just like you did, but it won't pass CI.

What we really need to do is to pin the version of Clippy used here, likely by moving this part of the CI pipeline to Bazel as well. See https://bazelbuild.github.io/rules_rust/rust_clippy.html if you want to help with that!

dvrdnz:
Looks like this was already fixed upstream in 47b3a7ff0 (#3257) — closing in favor of that. Thanks for the review anyway!

mgeisler:
I see, thanks for the PR!

PR #3246, opened 2026-07-27
Fix clippy::useless_borrows_in_formatting warnings in dining philosophers exercise

dvrdnz:
Fixes `clippy::useless_borrows_in_formatting` warnings introduced by newer
clippy versions in the dining philosophers exercise. Removes redundant `&`
references in `format!`/`println!` calls — no behavior change, `String`
and `&String` both implement `Display`.

Same category of fix as #3245, kept as a separate PR since it's an
unrelated file/exercise.


dvrdnz:
already fixed upstream in 47b3a7ff0 (#3257)

PR #3257, opened 2026-08-01
Fix updated clippy warnings

djmitche:
I assume these are from a rust version update.


mgeisler (approved):
Weird that this snug in, but I guess we'll fix this with Bazel as well since then no longer do `cargo update` to a new random version.

PR #3263, opened 2026-08-06
fix: correct literal value in pattern matching exercise note

coreele:
Corrected the variable reference from '10' to '0' in the example.


coreele:
there is no `10` in the first arm. It's `0`

fw-immunant (approved):

PR #3265, opened 2026-08-24
Add initial version of macros course

fw-immunant:
This covers declarative and procedural macros, and should be a good complement to the fundamentals course


djmitche:
Wow, this is exciting! I am pretty far from the course now, and will not have the time to review. Sorry!

src/macros/kinds/derive.md
```diff
@@ -0,0 +1,50 @@
...
+- **Additive:** They cannot modify or delete the original type definition; they
+  can only output additional code (such as `impl` blocks).
+- **Implementation:** Always implemented as procedural macros.
+
+<details>
+
+- Explain that `#[derive(Debug)]` generates an implementation of
+  `std::fmt::Debug` for the struct.
+- Note that derive macros can also support "helper attributes" (attributes
+  placed on struct fields, like `#[serde(rename = "name")]`), which configure
```
randomPoison, line 43: A better example might be the `#[default]` attribute from the `Default` derive, which is used to specify the default variant for an enum. The example code could probably also be changed to demonstrate and enum with that attribute.

src/macros/streams.md
```diff
@@ -0,0 +1,54 @@
...
+}
+
+fn main() {
+    // The macro returns tokens: 1 + 2
+    // Since expansion inserts a structured AST fragment, precedence is preserved!
+    println!("3 * 10 = {}", add_rpn!(2 1 +) * 10); // Outputs "30"
+}
+```
+
+- Because macro output is inserted as an AST subtree, we don't need to wrap
```
randomPoison, line 40:
```suggestion
Because macro output is inserted as an AST subtree, we don't need to wrap
```

src/macros/macro_rules/pattern-matching.md
```diff
@@ -0,0 +1,49 @@
...
+    command!(restart);
+}
+```
+
+In this example, the tokens `start`, `stop`, and `restart` are matched exactly
+as literal identifiers. Literal identifiers and custom symbols (such as `=>`,
+`,`, `:`, etc.) can be sequenced to design readable and mnemonic input syntax
+for your macros.
+
+Note, this macro invokes itself recursively in its transcriber body.
```
randomPoison, line 38: This line would be better as a speaker note, mostly just to keep the slide a bit more compact.

src/macros/macro_rules/fragment-specifiers.md
```diff
@@ -0,0 +1,54 @@
...
+- `vis`: An optional visibility specifier in an item, e.g. `pub` or
+  `pub(crate)`, or nothing.
+- `meta`: The possible _contents_ of a `#[...]` attribute (not the full
+  attribute with hash and brackets).
+
+There is overlap between a number of these specifiers, e.g. we could write our
+own pattern for blocks using `stmt` instead of using `block` directly. In
+general, try to reach for the most semantically appropriate fragment specifier
+where possible, which will catch more syntactic edge cases and integrate more
+seamlessly into the language.
```
randomPoison, line 41: I think this bit could be a speaker note.

src/macros/macro_rules/repetition.md
randomPoison, line 1: Do you think it'd be useful here to talk about making the trailing separator optional? The way the example is written you can't have a trailing `,`. If the matcher were changed to `$( $val:expr, )+` then a trailing separator would be mandatory. You can make the trailing comma optional by making the matcher `$( $val:expr ),+ $(,)?`, which is a little cumbersome but makes the macro syntax more flexible.

If this isn't the right place to call that out, maybe we could have a slide for this in the "declarative macro techniques" section?

src/macros/macro_rules/techniques/accumulators.md
```diff
@@ -0,0 +1,61 @@
...
+<!--
+Copyright 2026 Google LLC
+SPDX-License-Identifier: CC-BY-4.0
+-->
+
+# Push-Down Accumulators
+
+Because all function-like macro calls must evaluate to valid Rust syntax (most
+frequently, expressions), we cannot directly generate and return loose token
+lists like `a, b, c` from helper macros to be embedded inside templates.
```
randomPoison, line 14: I feel like this could use a bit more explanation. It might be worth demonstrating a macro that attempts to do this and walking through the errors that arise. I tried to do this with the turtle program example before reading the accumulator slide and found that it was unclear why my macro wasn't working.

[Here's a playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=c3f25e56564723af62850fb096dc943f) with two approaches I tried after reading the tt muncher slide. I first tried to build a single array expr, which didn't work (as this slide notes). I then tried a different approach where I built a sequence of `push` statements, and that actually did work. Maybe it'd be worth covering at least the non-working expr-based approach to show students what happens when they try to build a list from a recursive macro invocation.

src/macros/macro_rules/techniques/accumulators/exercise.md
```diff
@@ -0,0 +1,94 @@
...
+            TurtleOp::SetColor(Color::Black),
+        ]
+    );
+}
+```
+
+<details>
+
+- Explain why push-down accumulation is necessary when building flat collections
+  like a `Vec<T>` or array: individual macro recursive calls cannot emit looose
```
randomPoison, line 90:
```suggestion
  like a `Vec<T>` or array: individual macro recursive calls cannot emit loose
```

src/macros/macro_rules/techniques/accumulators/exercise.md
```diff
@@ -0,0 +1,94 @@
...
+#[derive(Debug, PartialEq)]
+pub enum TurtleOp {
+    Forward(i32),
+    Reverse(i32),
+    Left(i32),
+    Right(i32),
+    PenUp,
+    PenDown,
+    SetColor(Color),
+}
```
randomPoison, line 55: A minor note about this and the other turtle program exercise: We probably don't need this many operations. Each operation means another arm of the macro, which is added boilerplate when the actual logic in the macro arms are basically the same. If we stripped out a few of these ops then the exercises would be a little bit less tedious to do.

My suggested simplification would be something like this:

```rust
pub enum TurtleOp {
    Forward(i32),
    Turn(i32),
    PenUp,
    PenDown,
}
```

src/macros/macro_rules/techniques/tt-bundling.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+The signature of the recursive arm does not change regardless of how complex the
+bundled parameters may become.
+
+<details>
+
+- In the code example, `(present: fail, absent: pass)` represents the bundle,
+  which is matched recursively as a single `$bundle:tt` and passed downstream.
+- Macros with substantial state are most likely to benefit from this technique.
```
randomPoison, line 56:
```suggestion
- Macros with substantial state are most likely to benefit from this technique.
- It might be helpful to debug print the `$_skip` variable in the recursive case to better visualize the way that the recursive case pulls tokens off the front until we hit either of the base cases.
```

At least, I found it helpful to add debug printing to make it easier to follow what the macro was doing.

src/macros/hygiene/what-is-hygiene.md
```diff
@@ -0,0 +1,56 @@
...
+}
+```
+
+In Rust, **neither of these examples compile**. Both produce the error:
+`error[E0425]: cannot find value 'local' in this scope`.
+
+Rust's macro system treats variables hygienically, protecting from silent
+namespace pollution.
+
+However, declarative macros in Rust are not fully hygienic!
```
randomPoison, line 56: These lines could be pulled into speaker notes to make the slide a bit more concise.

src/macros/hygiene/what-is-hygiene.md
randomPoison, line 1: The two examples from this slide could probably be broken into two separate slides. Maybe three: One for the "What Is Macro Hiegiene" section at the top, and then two more slides for the two examples.

src/macros/hygiene/rust-macro-hygiene.md
```diff
@@ -0,0 +1,85 @@
...
+which we could not decrease lexical boilerplate.
+
+On the other hand, hygiene helps us write reliable code, so it is desirable for
+any internal operations that a macro may want to perform. Luckily for us, Rust
+does provide a solution for hygienic references to items.
+
+### The `$crate` variable
+
+Item and crate paths are unhygienic, so item paths within a macro definition
+could will refer to a different item than intended if their leading module or
```
randomPoison, line 32:
```suggestion
will refer to a different item than intended if their leading module or
```

src/macros/hygiene/rust-macro-hygiene.md
```diff
@@ -0,0 +1,85 @@
...
+Frequently, rust macros are used as shorthand to refer to existing types and
+traits, e.g. when defining `impl`s. In this situation, hygienic macros would
+always need to accept all relevant items as arguments, imposing a floor beneath
+which we could not decrease lexical boilerplate.
+
+On the other hand, hygiene helps us write reliable code, so it is desirable for
+any internal operations that a macro may want to perform. Luckily for us, Rust
+does provide a solution for hygienic references to items.
+
+### The `$crate` variable
```
randomPoison, line 29: This section on `$crate` should be pulled out into its own slide.

src/macros/hygiene/rust-macro-hygiene.md
```diff
@@ -0,0 +1,85 @@
...
+- They are hygienic with respect to: local variables, parameters, loop labels,
+  and the special `$crate` variable.
+- They are **not** hygienic with respect to: items, types, methods, and traits.
+
+## Rationale
+
+Frequently, rust macros are used as shorthand to refer to existing types and
+traits, e.g. when defining `impl`s. In this situation, hygienic macros would
+always need to accept all relevant items as arguments, imposing a floor beneath
+which we could not decrease lexical boilerplate.
```
randomPoison, line 23: This could benefit from an example showing how macros that interact with with items get more verbose when you have to explicitly pass in everything. Maybe something that implements a trait for a bunch of types, and would have to have the trait and all of the method names passed in explicitly?

src/macros/hygiene/rust-macro-hygiene.md
```diff
@@ -0,0 +1,85 @@
...
+On the other hand, hygiene helps us write reliable code, so it is desirable for
+any internal operations that a macro may want to perform. Luckily for us, Rust
+does provide a solution for hygienic references to items.
+
+### The `$crate` variable
+
+Item and crate paths are unhygienic, so item paths within a macro definition
+could will refer to a different item than intended if their leading module or
+crate name is defined differently at the call site than the macro author
+expected.
```
randomPoison: This could use an example (maybe as its own slide) showing how a path in a macro can get interpreted differently depending on where the macro is invoked.

randomPoison, line 34: This could use a dedicated example showing how a path in a macro gets interpreted relative to where the macro is expanded, maybe also splitting this into its own slide.

src/macros/hygiene/rust-macro-hygiene.md
```diff
@@ -0,0 +1,85 @@
...
+// Macro-consuming crate that alters meaning of the `std` crate name
+#![no_std]
+
+// libcore exports many similar APIs to libstd, but not `io::print`
+extern crate core as std;
+
+fn main() {
+    my_macros::print_something!()
+}
+```
```
randomPoison, line 68: This example is very confusing to me:

- It references `std::io::print`, which doesn't appear to be a thing.
- Uses `#[no_std]`, which is something students may not understand but would be awkward to explain here.
- It tries to illustrate that using `#[no_std]` in a downstream crate would change what `std` means in the macro, except the macro only references `my_macro_helper`. The speaker note somewhat explains what's going on here, but I think the code example needs to illustrate that better.

I think the better example would be to have two crates that define the same function, and show that using `$crate` allows the macro to always reference the one from the crate where the macro is defined.

<details><summary>Alternative example code</summary>

Crate A, which defines a function and a macro that references the function:

```rust
pub fn print() {
    println!("Hello from crate A");
}

macro_rules! print_something {
    () => {
        // With $crate this will always reference `print` in this
        // same crate, without it the macro will call whatever
        // `print` is in scope where the macro is invoked.
        $crate::print()
    };
}
```

Crate B, which defines its own `print` and uses the macro from crate A:

```rust
use crate_a::print_something;

pub fn print() {
    println!("Hello from crate B");
}

fn main() {
    print_something!();
}
```

We'd then show that the program prints "Hello from crate A", and if `$crate` is removed then it'll print "Hello from crate B". Not sure if we can actually demonstrate that in the slides though, since it involves multiple crates interacting :/

</details>

src/SUMMARY.md
```diff
@@ -240,6 +240,72 @@ SPDX-License-Identifier: CC-BY-4.0
...
+  - [Attribute Macros](macros/proc-macros/attr.md)
+  - [Handling Errors](macros/proc-macros/errors.md)
+  - [Procedural Macros In The Wild](macros/proc-macros/in-the-wild.md)
+  - [Writing Procedural Macros](macros/proc-macros/writing.md)
+    - [Dependencies](macros/proc-macros/writing/deps.md)
+      - [The `proc_macro` Crate](macros/proc-macros/writing/deps/proc_macro.md)
+      - [The `proc_macro2` Crate](macros/proc-macros/writing/deps/proc_macro2.md)
+      - [The `syn` and `quote` Crates](macros/proc-macros/writing/deps/syn-quote.md)
+        - [The `syn` AST](macros/proc-macros/writing/deps/syn-ast.md)
+        - [The `quote!` macro](macros/proc-macros/writing/deps/quote-macro.md)
```
randomPoison, line 303:
```suggestion
  - [Writing Procedural Macros](macros/proc-macros/writing.md)
  - [Dependencies](macros/proc-macros/writing/deps.md)
  - [The `proc_macro` Crate](macros/proc-macros/writing/deps/proc_macro.md)
  - [The `proc_macro2` Crate](macros/proc-macros/writing/deps/proc_macro2.md)
  - [The `syn` and `quote` Crates](macros/proc-macros/writing/deps/syn-quote.md)
  - [The `syn` AST](macros/proc-macros/writing/deps/syn-ast.md)
  - [The `quote!` macro](macros/proc-macros/writing/deps/quote-macro.md)
```

I prefer to avoid deep nesting of slides like this. While teaching I often look at the table of contents to remind myself how many slides I have left, which helps with time management. Nesting slides like this somewhat interferes with my ability to do that quickly.

If you think it's worth de-nesting this a bit, then I think the morning slides could be structured a bit more flatly as well.

src/macros/proc-macros/exercise/exercise.rs
```diff
@@ -0,0 +1,45 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+// ANCHOR: solution
+// ANCHOR: Derive
+use proc_macro2::TokenStream;
+use quote::quote;
+use syn::{DeriveInput, parse2};
+
+fn derive_display_impl(input: TokenStream) -> TokenStream {
+    // ANCHOR-END: Derive
```
randomPoison, line 11:
```suggestion
    // ANCHOR_END: Derive
```

src/macros/proc-macros/exercise/exercise.rs
```diff
@@ -0,0 +1,45 @@
...
+    let name_str = name.to_string();
+
+    // Generate the Display implementation using quote!
+    quote! {
+        impl std::fmt::Display for #name {
+            fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                write!(f, "{}", #name_str)
+            }
+        }
+    }
```
randomPoison, line 26: This example solution doesn't do anything with the fields of the struct, is that intentional? I would think it would be better to have students actually handle the fields, since that'd also require them to handle repetitions in `quote!`.

src/macros/proc-macros/exercise.md
```diff
@@ -0,0 +1,53 @@
...
+# Exercise: `Display` Derive
+
+In this exercise, you will walk through the conceptual implementation of a
+custom **`Display` derive macro** using the `syn` and `quote` crates.
+
+Since we are running this in a single-file environment without a separate
+`proc-macro = true` crate setup, we will write a normal Rust function that takes
+simulated token streams (using `proc_macro2` and `quote!`), parses them with
+`syn`, generates the output with `quote!`, and asserts that the generated code
+is correct.
```
randomPoison, line 19: I don't love this setup :/ I think it'd be better to have students actually setup a local workspace with a separate crate for the proc macro, and then test the proc macro's implementation by using the derive in actual code.

That said, I'm not sure there's a good way give students that setup, since we're somewhat limited by the fact that we give students the starting code in the slides. Maybe we just give students the test code and have them setup the local crate structure? e.g. we give them something like:

```rust
#[derive(Display)]
struct MyAwesomeType {
    field: i32,
}

fn main() {
    let my_type = MyAwesomeType { field: 123 };
    let string = my_type.to_string(); // Goes through `Display`.
    assert_eq!(my_string, "MyAwesomeType: 123"); // Or whatever we want the output string to be.
}
```

And then we have them setup the crates and proc macro to get that to compile and run.

randomPoison (commented):
I think this is off to a good start! I have various suggestions, but most of them are pretty minor. I think there's room to further flesh out the proc-macro section, especially with more exercises, but I don't think we need to iterate on that in this PR.

For the new exercises, it would be good to split the code into their own files and then pull sections into the slides with `#include`, that way the exercise and solution code stay in sync. At least that's how we do it for the Fundamentals exercises (and you did this already for the proc macro exercise).

src/macros/macro_rules/repetition.md
fw-immunant: I think this is worth a speaker note on the repetition slide. Will add.

src/macros/hygiene/rust-macro-hygiene.md
fw-immunant: My motivation here was to produce a somewhat realistic situation where this would occur--the most likely way macros might rely on ambiently-available names is via the standard library, but some consumers may violate that expectation. For some reason I thought there was a non-macro `std::io::print` in libstd, but the closest thing actually exposed is probably `std::io::stdout().write_all(...)`.

But given that `#![no_std]` is not something folks are likely to be familiar with when they want to learn how to use macros, you're probably right that an example with two hypothetical crates is better.

And yeah, I don't think we can provide a working interactive example here because what's really happening is a cross-crate interaction. But I'll change the example to something closer to what you suggest.

fw-immunant: Having thought about this more, I think it might be best to split this slide in half. The first can demonstrate how the lack of hygiene for paths can make a macro refer to different items when invoked from different locations, and show how we might use absolute paths to try to avoid that situation. This can be done with multiple modules in a single crate. The second can show that even absolute paths do not solve this problem completely in the presence of multiple crates, and motivate/explain the $crate keyword. We kind of try to do this already, but the existing "what is hygiene" slide is really just introducing the concept without demonstrating how the semantics of Rust item paths interact with it.

src/SUMMARY.md
fw-immunant: The nesting is semantically helpful to understand the structure of the information, but I agree and have a similar workflow when teaching--I often want to see how many slides I need to cover and which are coming up next. What if we made it possible (or the default) to expand all children?

fw-immunant: Maybe a conservative compromise for now is to just unindent the individual syn and quote slides.

randomPoison: Sure, I think changing the summary to default to expanding all children would at least address my use case of wanting to see how many slides are left. I'm still inclined to avoid nesting regardless, I don't think it adds much value in practice and ultimately students experience the slides as a flat list, but I don't think it matters much either way as long as we have an easy way to visualize how many slides are left.

PR #3275, opened 2026-09-04
fix(unsafe): use unchecked_add in may_overflow to accurately model UB

### LINKED ISSUE #3122: Unsafe Rust: Integer overflow is not UB

Aditya-9-6:
## Summary

Updates the `may_overflow` example in the Unsafe Deep Dive chapter to use `i32::unchecked_add` instead of standard `+`.

### Background

In standard Rust, integer overflow (`a + i32::MAX`) is not undefined behavior: it panics with overflow checks enabled (debug mode) and performs two's-complement wrapping in release mode.

Using `unchecked_add`:
1. Genuinely justifies the `unsafe` function signature and block.
2. In debug builds, it triggers the standard library's precondition panic:
   ```text
   unsafe precondition(s) violated: i32::unchecked_add cannot overflow
   ```
3. In release builds, it produces actual Undefined Behavior (UB) as the slide intends to teach.

The speaker notes have been updated accordingly to clarify the distinction between safe Rust arithmetic (which wraps in release mode) and unchecked operations (which produce UB).

Fixes #3122
Relates #3021

### DIFF
--- a/src/unsafe-deep-dive/introduction/may_overflow.md
+++ b/src/unsafe-deep-dive/introduction/may_overflow.md
@@ -13,9 +13,14 @@ SPDX-License-Identifier: CC-BY-4.0
 # // Copyright 2026 Google LLC
 # // SPDX-License-Identifier: Apache-2.0
 #
-/// Adds 2^31 - 1 to negative numbers.
+/// Adds 2^31 - 1 to negative numbers without checking for overflow.
+///
+/// # Safety
+///
+/// Calling this with `a >= 1` causes signed integer overflow, which is
+/// undefined behavior with `unchecked_add`.
 unsafe fn may_overflow(a: i32) -> i32 {
-    a + i32::MAX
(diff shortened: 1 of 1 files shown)


fw-immunant (approved):
LGTM, this corrects the factual error and explains the situation clearly. Thanks!

I think we should merge this as a conservative correctness fix, but it's worth noting that there is another open PR (#3244) addressing the same issue in a different way (by changing to an unrelated example with a safety contract around indexing). But that PR is not yet ready to merge and is a more substantial alteration to the curriculum, so I'd rather merge this now and later decide whether #3244 should merge at all and if so in what form.

PR #3277, opened 2026-09-11
Improve naming in std types exercise

fw-immunant:
This makes this exercise a bit cleaner and easier to follow.

### DIFF
--- a/src/std-types/exercise.rs
+++ b/src/std-types/exercise.rs
@@ -20,38 +20,48 @@ use std::hash::Hash;
 
 /// Counter counts the number of times each value of type T has been seen.
 struct Counter<T> {
-    values: HashMap<T, u64>,
+    counts: HashMap<T, u64>,
 }
 
 impl<T: Eq + Hash> Counter<T> {
     /// Create a new Counter.
     fn new() -> Self {
-        Counter { values: HashMap::new() }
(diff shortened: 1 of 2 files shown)


src/std-types/exercise.rs
```diff
@@ -20,38 +20,48 @@ use std::hash::Hash;
...
+        charctr.counts.keys().max_by_key(|&c| charctr.times_seen(*c))
+    );
+
+    let mut intctr = Counter::new();
+    intctr.count(13);
+    intctr.count(14);
+    intctr.count(16);
+    intctr.count(14);
+    intctr.count(14);
+    intctr.count(11);
```
randomPoison, line 61: Maybe it'd better to not use ints as the thing we're counting? At least I think it's a little confusing since there's the number we're counting and then the number of times we've seen that number. Maybe have this one count strings instead?

randomPoison (approved):
Looks good! I had one small suggestion but otherwise I think this is a good improvement.

src/std-types/exercise.rs
fw-immunant: We count strings and characters in the other uses of `Counter` in `main`, so I think this is probably clear in context.

thedataking (approved):
Approving on behalf of @randomPoison to unblock merging. Edit: same problem :|

PR #3279, opened 2026-09-17
Fundamentals: Add more loggers so we can be more generic

randomPoison:
In a couple of the slides/exercises in Fundamentals day 2 we use the example of a generic logger that does some kind of filtering and wraps another underlying logger type. Except in both places where we use this example, we don't actually demonstrate swapping out the underlying logger 😅 

This PR updates the slide/exercise to add a `StdoutLogger` alongside the existing `StderrLogger`, and demonstrates that we can swap out these underlying loggers using generics.

In the closures exercise I also change the closures to actually capture a variable so that they can't decay into function pointers. I've seen students use a function pointer (`fn(u8, &str) -> bool`) instead of the actual closure trait, which worked because the closures didn't capture anything.


src/closures/exercise.rs
```diff
@@ -37,13 +45,13 @@ struct Filter<L, P> {
 
 impl<L, P> Filter<L, P>
 where
-    L: Logger,
     P: Fn(u8, &str) -> bool,
```
fw-immunant, line 41: Why drop the L bound but not the P bound?

fw-immunant (approved):
Seems like a reasonable change. The way I usually demonstrate the flexibility of the generics here is by nesting `Filter` loggers.

src/closures/exercise.rs
randomPoison: The `L` bound wasn't doing anything, but the `P` bound helps with type inference: If you remove it you start getting type inference errors for the closure arguments.

randomPoison:
> The way I usually demonstrate the flexibility of the generics here is by nesting Filter loggers.

That's also a cool idea too, though maybe a little harder to follow for students? idk

PR #3280, opened 2026-09-17
Add review of for loops to iterator section

randomPoison:
I think the iterators section would benefit from an additional slide at the beginning that reviews the fact that `for` loops can accept different things that represent a sequence of values. This is something that students have already seen, but it's good to go back over it as it sets the stage for the next slide that discusses how iteration works in general.

I also tweaked the speaker notes for the original `for` loops slide to better call out this fact that `for` loops work with different kinds of iterators.


fw-immunant (approved):
LGTM!

PR #3281, opened 2026-09-18
Split out slide for IntoIterator on references

randomPoison:
Currently the slide on `IntoIterator` demonstrates implementing it directly on a type, and then links to a playground in order to demonstrate implementing `IntoIterator` on references. This PR pulls that playground out into its own slide and adds some speaker notes to guide the discussion and call out how this pattern is used in the standard library.

This is how I've been teaching this section for a while, and I think it goes pretty smoothly, so it's probably worth making into an actual slide.


src/iterators/intoiterator-refs.md
```diff
@@ -0,0 +1,85 @@
...
+            self.j += 1;
+            if self.j >= self.grid.y_coords.len() {
+                return None;
+            }
+        }
+        let res = Some((self.grid.x_coords[self.i], self.grid.y_coords[self.j]));
+        self.i += 1;
+        res
+    }
+}
```
fw-immunant, line 45: It's idiomatic for `Ref` and `Mut` iterator versions to return references into the collection, like so:
```suggestion
impl<'a> IntoIterator for &'a Grid {
    type Item = (&'a u32, &'a u32);
    type IntoIter = GridRefIter<'a>;
    fn into_iter(self) -> GridRefIter<'a> {
        GridRefIter { grid: self, i: 0, j: 0 }
    }
}

struct GridRefIter<'a> {
    grid: &'a Grid,
    i: usize,
    j: usize,
}

impl<'a> Iterator for GridRefIter<'a> {
    type Item = (&'a u32, &'a u32);

    fn next(&mut self) -> Option<(&'a u32, &'a u32)> {
        if self.i >= self.grid.x_coords.len() {
            self.i = 0;
            self.j += 1;
            if self.j >= self.grid.y_coords.len() {
                return None;
            }
        }
        let res = Some((&self.grid.x_coords[self.i], &self.grid.y_coords[self.j]));
        self.i += 1;
        res
    }
}
```

This isn't terribly useful in this case, but important for consistency with other types, especially generic ones.

src/iterators/intoiterator.md
```diff
@@ -81,18 +84,8 @@ Note that `IntoIter` and `Item` are linked: the iterator must have the same
...
-`GridRefIter` that iterates by reference. A version with both `GridIter` and
-`GridRefIter` is available [in this playground][1].
-
-The same problem can occur for standard library types: `for e in some_vector`
-will take ownership of `some_vector` and iterate over owned elements from that
-vector. Use `for e in &some_vector` instead, to iterate over references to
-elements of `some_vector`.
+Uncomment the commented line in `main` and show the compiler error. Note that
+`IntoIterator::into_iter` takes ownership of `self`. The next slide shows how we
+can fix this by implementing `IntoIterator` on `&Grid`.
```
fw-immunant:
```suggestion
can avoid this by iterating over `&Grid` with an additional  `IntoIterator` implementation.
```

Merely implementing the trait for the other type doesn't change what happens for this code, but is necessary as a dependency of changing from iterating over `grid` to `&grid`.

fw-immunant (changes requested):
A few tweaks commented inline, but this is a good change; the flow of teaching this can be annoying with the current slides.

PR #3282, opened 2026-09-18
Add slide on enumerate

randomPoison:
`enumerate` is a super commonly used iterator helper, and addresses a use case that students often ask about, so I think it's worth having a quick slide to introduce it.


randomPoison:
It looks like CI failed on an unrelated build step, maybe a spurious failure? I'm not a member of this repo so I can't trigger a re-run.

